PR #13, opened 2015-04-29
Django

xming:
Django1.5 version bump


ncopa:
we have minetest-0.4.12 an py-djano-1.8.4 now

PR #15, opened 2015-05-06
Add F# - PR for redundancy

haf:
I have also sent an e-mail.


ncopa:
It fails to build here:
```
...
make -C FSharp.Build-proto Configuration=proto build-proto
make[3]: Entering directory '/home/ncopa/aports/testing/fsharp/src/fsharp-3.1.1.32/src/fsharp/FSharp.Build-proto'
MONO_ENV_OPTIONS= xbuild /p:Configuration=Proto
Segmentation fault
Makefile:9: recipe for target 'build-proto' failed
make[3]: *** [build-proto] Error 139
```

I suspect mono needs be pax marked (to disable the memory protections)

haf:
Isn't that what [this does](https://github.com/alpinelinux/aports/blob/master/testing/mono/APKBUILD#L48)?

clandmeter:
On 7 May 2015 at 09:10, Henrik Feldt <notifications@github.com> wrote:

> Isn't that what this does
> <https://github.com/alpinelinux/aports/blob/master/testing/mono/APKBUILD#L48>
> ?
>

only for the build. We probably need to set paxmarks for all bins.

clandmeter:
please check http://git.alpinelinux.org/cgit/aports/commit/?id=f9accc21

Should be able to build now with mono. If confirmed we can move it to main.

haf:
Doesn't seem to work properly:

```
Can't find custom attr constructor image: /app/bin/heka.exe mtoken: 0x0a00001b
* Assertion at class.c:5748, condition `!mono_loader_get_last_error ()' not met

Stacktrace:


=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================

Killed
```

haf:
I've tested mono and it seems to work for me.

ncopa:
i think this is fixed already. thanks!

haf:
You have.

Worth noting is that F# is still not quite working despite the efforts.

PR #18, opened 2015-11-09
community/openal-soft

xming:
Compile fix


ncopa:
patch applied. thanks

PR #21, opened 2015-11-15
Update rsyslog.conf

mterron:
Fix emergency broadcast. This prevented rsyslog from starting up.


Barthalion:
[Merged](http://git.alpinelinux.org/cgit/aports/commit/?id=c7efbf47debca380e2a8d7d50d74870f4ff59dee) with changed commit message. Thanks.

PR #22, opened 2015-11-15
Update rsyslog.confd

mterron:
Remove compatibility switch.


Barthalion:
Merged with [updated commit message](http://git.alpinelinux.org/cgit/aports/commit/?id=8d5e97348b47766a50e5ab8183da7b4141823b6d), thank you.

PR #24, opened 2015-11-25
freeswitch - add mod_json_cdr to modules.conf

robwdux:
provide mod_json_cdr with default freeswitch package


ncopa:
thi sseems to been merge. thanks

PR #25, opened 2015-11-27
testing/stress: new stress testing package

andyshinn:
Port of the http://people.seas.harvard.edu/~apw/stress/ stress test utility.


ncopa:
this seems to be added. thanks

PR #26, opened 2016-01-31
WIP: OCaml ports

avsm:
This adds support for camlp4, OPAM and the aspcud dependencies (clang, gringo, etc).

I'm sending this as a WIP port since it doesn't quite work yet -- some interaction (possibly with musl or a bash-ism) is causing the aspcud invocation to fail and the solver to error out with an inconsistent solution.  Still debugging...


avsm:
Rebased with a clean commit history and mergeable commit messages.

PR #28, opened 2016-02-13
add mozjpeg

wjordan:
Builds [mozjpeg](https://github.com/mozilla/mozjpeg) with `turbojpeg` and `jpeg` compatibility enabled, so it can be used in place of `libjpeg-turbo`.

This is my first APK submission and I'm not sure I set all of the fields correctly for optimal use of the library, so any feedback on best-practices would be welcome!


wjordan:
@ncopa thanks for the comments, I've made your suggested changes (left as a separate commit to preserve commit-specific comments, can rebase back into a single commit if you'd prefer).

fabled:
@wjordan rebased commit is preferred so the main aports would not contain the broken interim version.

ncopa:
@wjordan yes please rebase it to a single commit.Thanks!

ncopa:
it looks like mozjpeg is not supposed to be a general purpose library. it gives better compression at the expense of performance. I don't know if we want introduce a provides that may cause mozjpeg be pulled in unintentionally.

http://www.libjpeg-turbo.org/About/Mozjpeg

clandmeter:
Closing this as outlined by ncopa.

PR #29, opened 2016-02-13
Libvips

wjordan:
Builds [`libvips`](https://github.com/jcupitt/libvips) with various optional dependencies enabled. One of the optionally-enabled dependencies is `mozjpeg`, which I've provided on its own in #28.


AVVS:
Completely missed that PR and was struggling to implement an APK for the first time :) Well, I've succeeded, but it seems have too many redundant packages in my own (https://gist.github.com/AVVS/c56026dee71f7aff29167532d2b96960). Can we get this into community repo and also bump libvips to 8.3.1 as the current version?

@ncopa @wjordan ?

ncopa:
seems like mozjpeg conflicts with libjpeg-turbo. They both provide same library file name `usr/lib/libturbojpeg.so.0` which means that both cannot be installed at the same time. This means that we have the following options:
- rename the library to something else for mozjpeg (why is it not called `libmozjpeg.so.0` already?)
- replace libjpeg-turbo with mozjpeg system wide
- make libvips use libjpeg-turbo instead of mozjpeg

mozjpeg is designed for a special case while libjpeg-turbo is designed for more general purpose, so I don't think just replacing libjpeg-turbo with mozjpeg is a good idea.

ncopa:
i pushed the pakcage as `testing/vips` with some minor modifications. thanks!

PR #30, opened 2016-02-13
Node-sharp

wjordan:
Builds [`sharp`](https://github.com/lovell/sharp). Depends on both mozjpeg (#28) and libvips (#29).

I'm not sure if an npm package makes an appropriate candidate for an APK or the right approach for building it using npm, since no other npm-package APKs currently exist in the repository. I thought it would be suitable because the package has C extensions that must be compiled at install-time, and it's much simpler to just include the APK and have the dependency graph make sure the required dynamic libraries are all installed correctly. Feedback on the overall approach would be appreciated!


clandmeter:
Closing due to inactivity. Please reopen if needed.

PR #31, opened 2016-02-14
Libdir should point to /lib

fabn:
As in [linux-pam](https://github.com/alpinelinux/aports/blob/master/main/linux-pam/APKBUILD#L52) package libdir should be `/lib` otherwise pam module is installed in the wrong folder.


fabled:
This was applied in 24684747f588033d6a7ea41a366e0ff6871ce00e with modified commit message.

PR #33, opened 2016-02-21
main/alpine-mirrors: update desc and add HTTP to FIT CTU mirror

jirutka:


fabled:
Merged in commit a71af9ede1e5b62f8a658bbdc1a153d411d18866. Thanks.

PR #34, opened 2016-02-26
testing/ssh-getkey-ldap: new aport

jirutka:
https://github.com/jirutka/ssh-getkey-ldap
A simple script to be used as AuthorizedKeysCommand in OpenSSH server to look up user's public keys in LDAP.


ncopa:
i rebased the patch. thanks!

PR #35, opened 2016-03-06
testing/py-setproctitle: new aport

jirutka:
https://github.com/dvarrazzo/py-setproctitle
A Python module to customize the process title.


ncopa:
applied. thanks

PR #37, opened 2016-03-18
testing/maven: refactor abuild and upgrade to 3.3.9

jirutka:


jirutka:
@ncopa I’ve fixed error with creating directories; sorry for that, I had probably forgotten to clear pkgdir that it didn’t occur to me. Now it should be fully working.

ncopa:
applied. thanks!

PR #38, opened 2016-03-23
main/uwsgi: set default plugin_dir

jirutka:
Default value of plugin_dir is . (current working directory). This is
bad, because plugins can't be loaded using simple name  (e.g. uwsgi
--plugin python) and absolute path must be used instead. It also breaks
some applications that use uwsgi wrappers which expects this variable to
be set correctly.

This approach is adopted from Gentoo ebuild www-servers/uwsgi.
See also http://projects.unbit.it/uwsgi/wiki/Guide4Packagers

### DIFF
--- a/main/uwsgi/0005-uwsgiconfig-plugin_build_dir.patch
+++ b/main/uwsgi/0005-uwsgiconfig-plugin_build_dir.patch
@@ -1,0 +1,32 @@
+From 6dc5730a234f4bc8a0cbfb0bd3360e860d39e340 Mon Sep 17 00:00:00 2001
+From: Jakub Jirutka <jakub@jirutka.cz>
+Date: Sat, 26 Mar 2016 00:17:35 +0100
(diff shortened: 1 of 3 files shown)


ncopa:
i wonder why upstream does not do this?

jirutka:
What exactly do you mean? If setting `plugin_dir`, then they actually do this, see [buildconf/package.ini](https://github.com/unbit/uwsgi/blob/fe95a7334cc351634e69c671b2b32c15f516624b/buildconf/package.ini). But this profile doesn’t set `main_plugin` to empty and `embedded_plugins` to null as the core profile that’s currently used.

If you mean `plugin_build_dir`, then I really don’t know.

ncopa:
I mean `0005-use-plugin_build_dir.patch`. Do you think you could try submit that upstream? Even if they reject it, we would still know why.

jirutka:
I had submitted the patch to upstream and they already accepted it. I backported the exact patch I sent to upstream to this pull request.

ncopa:
good job!

PR #39, opened 2016-03-26
testing/py-pyldap: new abuild

jirutka:
https://github.com/pyldap/pyldap
Python modules for implementing LDAP clients


ncopa:
applied. thanks!

PR #41, opened 2016-03-28
testing/elasticsearch: new abuild

jirutka:
https://www.elastic.co/products/elasticsearch
Open Source, Distributed, RESTful Search Engine

### DIFF
--- a/testing/elasticsearch/APKBUILD
+++ b/testing/elasticsearch/APKBUILD
@@ -1,0 +1,73 @@
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+pkgname=elasticsearch
+pkgver=2.2.1
+pkgrel=0
+pkgdesc="Open Source, Distributed, RESTful Search Engine"
+url="https://www.elastic.co/products/elasticsearch"
+arch="noarch"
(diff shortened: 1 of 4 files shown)


testing/elasticsearch/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+	$pkgname.confd
+	"
+
+_modules="lang-expression lang-groovy"
+for _mod in $_modules; do
+	subpackages="$subpackages $pkgname-$_mod:_${_mod//-/_}"
+	eval "_${_mod//-/_}() { _builtin_module $_mod; }"
+done
+
+_builddir="$srcdir/$pkgname-$pkgver"
```
Barthalion: I think it's being set by default as "$builddir" now. @ncopa, could you confirm?

ncopa:
fixed a whitespace damage and pushed. thanks!

testing/elasticsearch/APKBUILD
ncopa: yes, abuild use "$builddir" now, due to the `default_prepare` func in abuild.

jirutka: Thanks for notice, I didn’t know about it. Could you please add it to wiki http://wiki.alpinelinux.org/wiki/APKBUILD_Reference#Variables?

PR #45, opened 2016-04-07
testing/java-jansi-native: new abuild

jirutka:
https://github.com/fusesource/jansi-native
Native JNI libraries used by the Jansi project.


jirutka:
Replaced by #51.

PR #49, opened 2016-04-08
testing/jruby: new abuild

jirutka:
http://jruby.org
An implementation of Ruby on the JVM

Depends on #45, #46


jirutka:
Replaced by #51.

PR #53, opened 2016-04-11
testing/munin: new aport

stwa:


ncopa:
do you think that you could merge in the c15902e commit? For example using `git rebase -i`

stwa:
Sorry, done

jirutka:
@stwa You should use tabs for indentation in APKBUILD (as specified in [.editorconfig](https://github.com/alpinelinux/aports/blob/master/.editorconfig) in the root of this repository), not spaces.

testing/munin/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+pkgname=munin
+pkgver=2.0.25
+pkgrel=0
+pkgdesc="A distributed monitoring/graphing tool"
+url="http://munin-monitoring.org/"
+arch="noarch"
+license="GPL"
+perl_modules="perl-rrd perl-net-snmp perl-log-log4perl perl-html-template
+perl-net-ssleay perl-net-server perl-date-manip perl-io-socket-inet6
+perl-file-copy-recursive perl-fcgi perl-uri"
```
jirutka: Please indent multiline strings (using one tab), to be easily recognizable.

```diff
- perl_modules="perl-rrd perl-net-snmp perl-log-log4perl perl-html-template
- perl-net-ssleay perl-net-server perl-date-manip perl-io-socket-inet6
- perl-file-copy-recursive perl-fcgi perl-uri"
+ perl_modules="perl-rrd perl-net-snmp perl-log-log4perl perl-html-template
+ 	perl-net-ssleay perl-net-server perl-date-manip perl-io-socket-inet6
+ 	perl-file-copy-recursive perl-fcgi perl-uri"
```

testing/munin/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+depends="$pkgname-node bash rrdtool perl $perl_modules"
+depends_dev="perl perl-module-build"
+makedepends="$depends_dev"
```
jirutka: Since you don’t provide `-dev` subpackage, it doesn’t make sense to declare `depends_dev`. `makedepends` is for build/compile dependencies, `depends_dev` is for dependencies of `-dev` subpackage.

testing/munin/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+install="$pkgname.pre-install $pkgname.post-install $pkgname.post-upgrade"
+subpackages="$pkgname-node"
+source="http://downloads.munin-monitoring.org/munin/stable/$pkgver/$pkgname-$pkgver.tar.gz
+Makefile.config munin.cron.d munin-node.initd munin.logrotate munin-node.logrotate"
```
jirutka: Same as above…

```diff
- source="http://downloads.munin-monitoring.org/munin/stable/$pkgver/$pkgname-$pkgver.tar.gz
- Makefile.config munin.cron.d munin-node.initd munin.logrotate munin-node.logrotate"
+ source="http://downloads.munin-monitoring.org/munin/stable/$pkgver/$pkgname-$pkgver.tar.gz
+     Makefile.config
+     munin.cron.d
+     munin-node.initd
+     munin.logrotate
+     munin-node.logrotate"
```

testing/munin/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+
+_builddir="$srcdir/$pkgname-$pkgver"
+build() {
+  cd "$_builddir"
+  cp ../Makefile.config .
+  make -j1 || return 1
```
jirutka: Return value of the last statement becomes return value of the function, so it’s unnecessary to add `|| return 1` on the last line of the function.

testing/munin/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+
+node() {
+  depends="bash rrdtool perl $perl_modules"
+  install="$subpkgname.pre-install $subpkgname.post-install $subpkgname.post-upgrade"
+
+  mkdir -p "$subpkgdir"
+
+  cd "$_builddir"
+  make DESTDIR="$subpkgdir" install-common-prime install-node-prime install-plugins-prime || return 1
+  rm -rf $subpkgdir/usr/share/man || return 1
```
jirutka: `$subpkgdir` should be always quoted, especially when it’s used as an argument for `rm -rf`…

testing/munin/munin-node.initd
```diff
@@ -0,0 +1,36 @@
+#!/sbin/openrc-run
+
+# munin-node init.d file for alpine linux.
+
+name=munin-node
+
+pidfile=/var/run/munin/${name}.pid
```
jirutka: `/var/run` is deprecated, it’s just a symlink to `/run` for backward compatibility.

testing/munin/munin-node.initd
```diff
@@ -0,0 +1,36 @@
...
+depend() {
+  need net
+  after firewall
+}
+
+start() {
+  if [ ! -d /var/run/munin ]; then
+    mkdir -p /var/run/munin
+    chown munin:munin /var/run/munin
+  fi
```
jirutka: 1. This should be in `start_pre()`, not `start()` ([man openrc-run](http://manpages.org/openrc-run/8)).
2. openrc-run provides handy function for this: `checkpath -d -o munin:munin -m755 /run/munin` ([man openrc-run](http://manpages.org/openrc-run/8)).

testing/munin/munin-node.initd
```diff
@@ -0,0 +1,36 @@
...
+
+  ebegin "Starting ${name}"
+    start-stop-daemon --start --quiet \
+      --pidfile ${pidfile} \
+      --exec ${command} -- ${command_args}
```
jirutka: This is just redundant. Remove the `start()` function and declare variable: `command_background="yes"` (you already have `command`, `command_args`, and `pidfile`).

testing/munin/munin-node.initd
```diff
@@ -0,0 +1,36 @@
...
+  eend $?
+}
+
+stop() {
+  ebegin "Stopping ${name}"
+    start-stop-daemon --stop --quiet \
+      --pidfile ${pidfile} \
+      --exec ${command}
+  eend $?
+}
```
jirutka: This is also redundant and unnecessary, just remove the `stop()` function.

testing/munin/Makefile.config
```diff
@@ -0,0 +1,172 @@
...
+JAR        := jar
+
+# Check if the java compiler works
+#    Note that we defer JCVALID evaluation to runtime, 
+#    since $(JC) can be redefined later in a specific Makefile.config
+#    The core Makefile.config is then used as a Makefile.default
+JCVALID    = no
+
+# Check whether setruid functionality can be used
+HASSETR    := $(shell perl -e 'use Config; my @vars=("d_setruid", "d_setreuid", "d_setresuid"); foreach my $$var (@vars) { if ($$Config{$$var} eq "define") { print "1\n"; exit 0; } } print "0\n"; exit 0;' )
```
jirutka: Where does this Makefile come from? Have you written it from scratch, is it copied from other distribution or is it a patched upstream’s Makefile?

testing/munin/munin.pre-install
```diff
@@ -0,0 +1 @@
+munin-node.pre-install
```
jirutka: `munin` depends on `munin-node`, so this is unnecessary.

testing/munin/munin.post-install
```diff
@@ -0,0 +1 @@
+munin-node.post-install
```
jirutka: `munin` depends on `munin-node`, so this is unnecessary.

testing/munin/munin-node.post-install
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# create directories
+mkdir -p /usr/share/munin/docs
```
jirutka: Why do you create `/usr/share/munin/docs` here? If this package installs some docs, then it must be in APKBUILD. If it doesn’t then it doesn’t make sense to create this directory.

testing/munin/munin-node.post-install
```diff
@@ -0,0 +1,14 @@
...
+mkdir -p /var/lib/munin/spool
+mkdir -p /var/log/munin
+mkdir -p /var/run/munin
```
jirutka: `/var/run/munin` is a symlink to `/run` that is a tmpfs, i.e. it’s wiped on every restart. Therefore it doesn’t make any sense to create it here, moreover you already creates it in runscript.

testing/munin/munin-node.post-install
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# create directories
+mkdir -p /usr/share/munin/docs
+mkdir -p /var/lib/munin/spool
+mkdir -p /var/log/munin
```
jirutka: `/var/lib/munin/spool` and `/var/log/munin` is IMO better to create in runscript using `checkpath`. Then you can completely omit these `post-{install,upgrade}` scripts.

testing/munin/Makefile.config
stwa: It is copied from the munin source package and modified for easier distribution.

jirutka: Then it’d be better to add your modifications as a patch.

testing/munin/APKBUILD
```diff
@@ -0,0 +1,84 @@
...
+install="$pkgname.post-install"
+subpackages="$pkgname-node"
+source="http://downloads.munin-monitoring.org/munin/stable/$pkgver/$pkgname-$pkgver.tar.gz
+	munin-config.patch
+	munin.cron.d
+	munin-node.initd
+	munin.logrotate
+	munin-node.logrotate"
+
+_builddir="$srcdir/$pkgname-$pkgver"
```
jirutka: `builddir` is now supported variable in abuild tool, so rename `_builddir` to `builddir` (I know, it’s not updated in template or wiki yet).

testing/munin/APKBUILD
```diff
@@ -0,0 +1,84 @@
...
+prepare() {
+	local i
+	cd "$_builddir"
+	for i in $source; do
+		case $i in
+			*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1
+		esac
+	done
+}
```
jirutka: This is now a default behaviour when the `prepare()` function is not declared in APKBUILD, so you can just omit it. It’s currently undocumented, I found it yesterday when reading source code of abuild [here](http://git.alpinelinux.org/cgit/abuild/tree/abuild.in#n543).

testing/munin/munin-node.initd
```diff
@@ -0,0 +1,22 @@
...
+
+# munin-node init.d file for alpine linux.
+
+name=munin-node
+
+pidfile=/run/munin/${name}.pid
+
+command=/usr/sbin/munin-node
+command_args="${EXTRA_OPTS}"
+command_background="no"
```
jirutka: Are you sure with this? Does it really work as you expect now?

stwa: Actually yes, it works as expected (tested on one of my servers). With command_background=yes the the status of the process is „crashed“ because the munin-node puts itself into background and there is no way to prevent it from doing so.
As munin-node is creating its own pid-file this should be ok, right?

ncopa:
I get this error on build server:
````
>>> munin: Checking sanity of /home/buildozer/aports/testing/munin/APKBUILD...
>>> WARNING: munin: munin.post-install: found chown
             Permissions should be fixed in APKBUILD package()
>>> munin: Analyzing dependencies...
(1/3) Installing perl (5.22.1-r0)
(2/3) Installing perl-module-build (0.4214-r0)
(3/3) Installing .makedepends-munin (0)
Executing busybox-1.24.2-r2.trigger
OK: 328 MiB in 100 packages
>>> munin: Cleaning temporary build dirs...
>>> munin: Checking sha512sums...
munin-2.0.25.tar.gz: OK
munin-config.patch: FAILED
munin.cron.d: OK
munin-node.initd: OK
munin.logrotate: OK
munin-node.logrotate: OK
>>> ERROR: munin: all failed
>>> munin: Uninstalling dependencies...
````

I have to revert it sorry.

stwa:
The patch applied wasn't the latest commit, I will close this pull request and send it to alpine-aports instead.

PR #55, opened 2016-04-13
testing/elasticsearch: fix runscript (#5390) and change arch to noarch

jirutka:
* You can't "die" in runscript, it's not defined in openrc-run.
* checkpath doesn't create intermediate directories.
* Variable "builddir" is now allowed to be declared in APKBUILD.

Fixes: http://bugs.alpinelinux.org/issues/5390


ncopa:
appiled. thanks

PR #56, opened 2016-04-13
testing/java-jna: new abuild

jirutka:
https://github.com/java-native-access/jna
JNA provides Java programs easy access to native shared libraries.


ncopa:
applied. thanks.

PR #57, opened 2016-04-14
main/freeswitch: upgrade freeswitch to 1.6.7, also add perl and perlesl subpackagages

mepholic:
Upgrade freeswitch to 1.6.7.1 and add subpackages for freeswitch-perl and freeswitch-perlesl

### DIFF
--- a/main/freeswitch/APKBUILD
+++ b/main/freeswitch/APKBUILD
@@ -2,8 +2,8 @@
 # Contributor: Michael Mason <ms13sp@gmail.com>
 # Contributor: Cameron Banta <cbanta@gmail.com>
 pkgname=freeswitch
-pkgver=1.6.6
-pkgrel=2
+pkgver=1.6.7
+pkgrel=0
 pkgdesc="A communications platform written in C from the ground up"
 url="http://www.freeswitch.org"
 arch="all"
@@ -14,6 +14,7 @@ pkggroups="freeswitch"
 makedepends="
 	bash
 	bsd-compat-headers
+	nasm
 	curl-dev
 	flac-dev
 	flite-dev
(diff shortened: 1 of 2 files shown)


main/freeswitch/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Contributor: Michael Mason <ms13sp@gmail.com>
 # Contributor: Cameron Banta <cbanta@gmail.com>
 pkgname=freeswitch
-pkgver=1.6.6
+pkgver=1.6.7
 pkgrel=2
```
jirutka: When bumping `pkgver`, reset `pkgrel` to zero.

main/freeswitch/APKBUILD
```diff
@@ -2,8 +2,8 @@
 # Contributor: Michael Mason <ms13sp@gmail.com>
 # Contributor: Cameron Banta <cbanta@gmail.com>
 pkgname=freeswitch
-pkgver=1.6.6
-pkgrel=2
+pkgver=1.6.7
+pkgrel=1
```
jirutka: When bumping pkgver, reset pkgrel to zero.

main/freeswitch/APKBUILD
```diff
@@ -115,6 +117,11 @@ build() {
 	make libfreeswitch.la || return 1
 	# finally we build the rest
 	make -j1 all || return 1
+	
+	# build perlesl module
+	cd "$_builddir"/libs/esl
+	make -j1 perlmod || return 1
```
jirutka: `|| return 1` is unnecessary behind the last statement of the function.

ncopa: but it does not hurt either.

jirutka: Yes, but it’s redundant, so it only adds a noise.

mepholic:
Okay, I squashed the commits down to one, and set the pkgrel to 0. I might add the explicit " || return 1" back, as it won't hurt anything.

main/freeswitch/APKBUILD
Barthalion: This a nitpick and actually doesn't add any noise, except for you.

Barthalion:
Merged in bf8fe206f1bfbfa9466f12466d0a7647340542a7.

PR #59, opened 2016-04-20
dhcp: add /sbin/dhclient-script

little-dude:
fix https://bugs.alpinelinux.org/issues/4855


little-dude:
@ncopa the `/sbin/dhclient-script` starts with `#! /bin/bash`. It works with `#!/bin/ash`, should I add a path for this, since `ash` is the default on alpine?

ncopa:
i'd prefer to use `#!/bin/sh` but then you should also check for bashisms. you can use `checkbashisms` for that

little-dude:
I added a patch to remove bashisms. Thanks for `checkbashisms`, it quite handy :)

ncopa:
very nice. I modified the commit message a bit and pushed it.

Thanks!

little-dude:
Thank you! I'm closing this then.

PR #62, opened 2016-04-25
Update composer to current stable release

skwashd:
On 21 April composer 1.0.2 was released. beta1 is now ~8 weeks old.


ncopa:
i fixed the commit message and pushed it. thanks!

skwashd:
@ncopa thanks for that. What is the correct format for commit messages? I'd like to get it right next time.

ncopa:
````
<repo>/<package>: short desc

longer desc which answers the question "why was this change made" if that is not obvious.
if it has fix for specific issue in bugtracker then also add:
fixes #<issuenumber>
````

skwashd:
OK, I'll try to do that next time.

PR #63, opened 2016-04-30
unmaintained/ruby-bundler: move back into main and upgrade to 1.12.1

jirutka:
Bundler is important, it’s needed to install Ruby _applications_.


ncopa:
applied thanks!

PR #65, opened 2016-05-01
Upgrade and move apache-ant and jetty-runner to the community repo

jirutka:


ncopa:
You cannot move those to `main` because they depend on `openjdk8` which is in `community`. We can move them to `community` though.

jirutka:
@ncopa Ah, you’re right, I’ve updated the PR and moved them to the community repo instead of main.

ncopa:
maybe also fix commit message which still say "move to the main repo" while it actually moves to `community`

jirutka:
Sorry, fixed.

ncopa:
applied. Thanks!

PR #66, opened 2016-05-01
Introduce a metapackage for Java JDK/JRE (WIP)

jirutka:
⚠️ **Do not merge it yet, we have to discuss this change!**
- Add metapackages java-jdk, java-jre and java-jre-base (headless).
- Change openjdk7 and openjdk8 packages to _provide_ the above metapackages.
- Clean-up openjdk7 and openjdk8 packages.

This patch will allow packages to depend on _any_ JDK or JRE, without specifying exact version (7, or 8) or implementation (OpenJDK). In this case, the latest OpenJDK is installed. If any JRE/JDK is already installed, then it satisfies the requirement (thanks to the directive `provides`).

OpenJDK 7 is not publicly supported by upstream anymore (but we need it to build OpenJDK 8), so it may seem that this change is unnecessary since we have just one truly supported version. Well, for **now**; OpenJDK 9 comes in a year and nightly builds are already available.

Moreover, the sad reality is, that some ~~sh*tty~~ enterprise software still doesn’t work on JRE/JDK 8, so openjdk7 is not used just for building openjdk8 in the wild. And also some people insist on installing Oracle JDK on Alpine, for whatever absurd reason (like some “security certificates”). I’m not for adding Oracle JDK to Alpine (for many reasons), but why not to allow people using their own JDK abuild with any Alpine packages depending on any JDK/JRE? Lastly, there are also [many other Java VM implementations](https://en.wikipedia.org/wiki/List_of_Java_virtual_machines); that’s why I named it java-\* and not openjdk-*.

EDIT: I don’t understand why openjdk7 fails on Travis (`tar (child): icedtea-2.6.3.tar.gz: Cannot open: No such file or directory`), it works on my machine.

### DIFF
--- a/community/java/APKBUILD
+++ b/community/java/APKBUILD
@@ -1,0 +1,39 @@
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+pkgname=java
+pkgver=8
+pkgrel=0
+pkgdesc="Metapackage for Java runtime (JRE) and development (JDK)"
+url="http://git.alpinelinux.org/aports.git"
+arch="noarch"
+license="custom"
+depends="$pkgname-jdk"
+makedepends=""
+subpackages="$pkgname-jdk $pkgname-jre $pkgname-jre-base:jre_base"
+source=""
+
+build() {
+	return 0
+}
+
+package() {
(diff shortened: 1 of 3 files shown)


Andrei-Pozolotin:
https://bugs.alpinelinux.org/issues/5760

PR #67, opened 2016-05-02
community/openjdk[78]: fix messed indentation, global vars and modernize it

jirutka:
These patches are extracted from #66, because #66 will take more time.


ncopa:
applied. thanks!

PR #68, opened 2016-05-02
Fixes of various packages in the main that fails to build

jirutka:
@ncopa All the packages from [this list](http://tpaste.us/2z91) are fixed, except nagiosql that is IMHO not worth fixing.

* audacious  –  already works
* awstats  –  fixed incorrect checksums
* debootstrap  –  fixed by upgrade to the latest version
* ghostscript  –  fixed incorrect checksums
* gimp  –  didn't try yet
* ldapvi  – fixed missing dependency on ncurses-dev
* libeventlog  –  already works
* liblogging  –  source tarball is not available anymore, so I modified abuild to build from sources
* nagiosql  –  very old and the project is already dead, IMHO it should be moved to unmaintained
* perl-carp  –  fixed by upgrading to a newer version (but not latest)
* php-pear-mail_mime  –  fixed incorrect checksums
* php-pear-net_idna2  –  fixed incorrect checksums
* php-pear-net_socket  –  fixed incorrect checksums
* py-fuse  –  fixed source URL, the package has been moved
* py-genshi  –  already works
* py-libvirt  –  fixed by upgrading to the latest version
* qt5-qtwebkit  –  fixed incorrect checksums
* rt4  –  fixed source URL
* ssh-getkey-ldap  –  fixed silly bug, mea culpa :/
* xdg-utils  –  fixed by replacing links with lynx (needed for building man pages)


jirutka:
Already merged, so closing.

PR #71, opened 2016-05-05
testing/mysecureshell: new abuild

jirutka:
https://github.com/mysecureshell/mysecureshell
A solution which has been made to bring more features to sftp/scp protocol given by OpenSSH.

Fixes http://bugs.alpinelinux.org/issues/5280


ncopa:
applied. thanks!

PR #72, opened 2016-05-05
main/sntpc: fix project url and improve abuild

jirutka:
Fixes http://bugs.alpinelinux.org/issues/5087.


ncopa:
applied. thanks!

PR #74, opened 2016-05-09
testing/R: rewrite abuild

jirutka:
Since Nirosan is not responding to mails for few weeks (Alba Pompeo has tried to contact him), I’m taking maintenance of this package (and I hope that I’ll not regret it 🙀).


jirutka:
This is already merged in 01e2d702cdda59e4678e91a8a21ba9f332329b20.

PR #76, opened 2016-05-17
python: protobuf new aports

little-dude:
py-protobuf needs py-google-apputils to build so CI may fail. I also builds without the `makedepends`, since `setuptools` pulls the package anyway but I thought it would be better to make this explicit.

### DIFF
--- a/testing/py-protobuf/APKBUILD
+++ b/testing/py-protobuf/APKBUILD
@@ -1,0 +1,29 @@
+# Maintainer: Corentin Henry <corentinhenry@gmail.com>
+# Contributor: Corentin Henry <corentinhenry@gmail.com>
+pkgname=py-protobuf
+_pkgname=protobuf
+pkgver=2.6.1
+pkgrel=0
+pkgdesc="Protocol Buffers are Google's data interchange format."
+url="https://github.com/google/protobuf"
+arch="noarch"
+license="BSD"
+depends="python py-six>=1.9"
+makedepends="python-dev py-setuptools py-google-apputils"
+source="http://pypi.python.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
+	cd "$builddir"
+	python setup.py build || return 1
+}
+
+package() {
+	cd "$builddir"
+	python setup.py install --prefix=/usr --root="$pkgdir" || return 1
+}
+
+md5sums="6bf843912193f70073db7f22e2ea55e2  protobuf-2.6.1.tar.gz"
+sha256sums="8faca1fb462ee1be58d00f5efb4ca4f64bde92187fe61fde32615bbee7b3e745  protobuf-2.6.1.tar.gz"
+sha512sums="c345b5b2822e7142e27cd6ff4ca4e8cc307acd3673043428073ed260a899c48ac6afa32e290b2d2f0ee54316d1a0f0ec72287604fb372992715a2191fe5623d4  protobuf-2.6.1.tar.gz"


testing/py-google-apputils/APKBUILD
```diff
@@ -0,0 +1,28 @@
+# Maintainer: Corentin Henry <corentinhenry@gmail.com>
+pkgname=py-google-apputils
+_pkgname=google-apputils
+pkgver=0.4.0
+pkgrel=0
+pkgdesc="Google Application Utilities for Python"
+url="https://github.com/google/google-apputils"
+arch="noarch"
+license="Copyright 2014, Google Inc. All rights reserved."
```
jirutka: This is semantically and even factually incorrect. google-apputils is licensed under [Apache License 2.0](https://github.com/google/google-apputils/blob/master/LICENSE), so there should be `license="ASL-2.0"`.

testing/py-protobuf/APKBUILD
```diff
@@ -0,0 +1,28 @@
+# Maintainer: Corentin Henry <corentinhenry@gmail.com>
+pkgname=py-protobuf
+_pkgname=protobuf
+pkgver=2.6.1
+pkgrel=0
+pkgdesc="Protocol Buffers are Google’s data interchange format."
```
jirutka: It’s better to avoid non-ASCII characters in APKBUILD, such as `’` here.

testing/py-protobuf/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+url="https://github.com/google/protobuf"
+arch="noarch"
+license="Copyright 2014, Google Inc. All rights reserved."
```
jirutka: This is semantically and even factually incorrect. [protobuf is licensed under BSD license](https://github.com/google/protobuf/blob/master/LICENSE), so there should be `license="BSD"`.

jirutka:
Abuild `py-protobuf` is broken, see https://travis-ci.org/alpinelinux/aports/builds/130781573#L421.

testing/py-google-apputils/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends="python"
```
jirutka: google-apputils [has some Python dependencies](https://github.com/google/google-apputils/blob/master/setup.py#L28-L32) that must be also specified in `depends`.

testing/py-protobuf/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends="python"
```
jirutka: Module protobuf [has some Python requirements](https://github.com/google/protobuf/blob/master/python/setup.py#L213) that must be also specified in `depends`. It seems that only `six` and `setuptools` are needed, `ordereddict` is not needed for Python 2.7 and `unittest2` is needed only for tests.

testing/py-protobuf/APKBUILD
little-dude: @jirutka I don't see mention of the BSD license. I copy/pasted the first line of the license file : `Copyright 2014, Google Inc. All rights reserved.`.

I don't have much licensing knowledge, how do you know this is a BSD license?

little-dude:
I put the `py-google-apputils` in a separate PR, since this package won't build without it: https://github.com/alpinelinux/aports/pull/84

jirutka:
How do I know that it's BSD license? I simply googled it.

Missing dependency isn't the reason why this abuild has failed on CI.

little-dude:
> How do I know that it's BSD license? I simply googled it.

Ok. Seems that Archlinux pkgbuild also set it to a BSD license anyway.

> Missing dependency isn't the reason why this abuild has failed on CI.

That's why I made a separate PR for the dependency. We can merge it first and get this one to build.

little-dude:
Build fails due to this: https://github.com/google/google-apputils/issues/4

little-dude:
@jirutka this is building fine now :)

jirutka:
Merged in 2595f01f3adc14265428e9da83772e6b183f3105, thanks!

PR #79, opened 2016-05-20
Fix unresolved references from boost_python libs

AmatCoder:
It is needed to update libtorrent-raster/deluge packages.
https://github.com/arvidn/libtorrent/issues/477


ncopa:
applied. thanks

PR #80, opened 2016-05-20
testing/libtorrent-rasterbar: Update to 1.1.0

AmatCoder:

PR #81, opened 2016-05-20
testing/deluge: Update to 1.3.12

AmatCoder:


AmatCoder:
Travis build fails because #80 needs be merge first.

jirutka:
Please remove whole `prepare()` function. When not defined, then abuild uses `default_prepare` that does exactly the same thing (applying patches).

And you may also rename `_builddir` to `builddir` to be consistent with recent changes in abuild tool.

jirutka:
Please squash commits in this PR into a single commit.

testing/deluge/APKBUILD
```diff
@@ -16,17 +16,7 @@ install=""
...
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
-}
-
+builddir="$srcdir"/$pkgname-$pkgver
 build() {
 	cd "$_builddir"
```
jirutka: You must obviously also rename all occurrences of `$_builddir` to `$builddir`.

AmatCoder:
Done. Thanks for suggestions.

ncopa:
applied. thanks

PR #83, opened 2016-05-22
testing/qemu-init-script: new abuild

jirutka:
https://github.com/jirutka/qemu-openrc
OpenRC init script for QEMU/KVM


ncopa:
applied. thanks!

PR #88, opened 2016-05-25
testing/viewnior: upgrade to 1.6

AmatCoder:


jirutka:
Could you please add to your commit the following fix?

```diff
- depends_dev="gtk+2.0-dev exiv2-dev"
- makedepends="$depends_dev autoconf automake gnome-common libtool"
+ makedepends="gtk+2.0-dev exiv2-dev autoconf automake gnome-common libtool"
```

`depends_dev` defines dependencies for -dev subpackage, but this abuild doesn’t create a -dev subpackage, so it doesn’t make sense to define this variable. Abuild [prints warning](https://travis-ci.org/alpinelinux/aports/builds/132764928#L298) for this, but the original author of this package probably overlooked it.

AmatCoder:
Done. Thanks for review it.

testing/viewnior/APKBUILD
```diff
@@ -1,33 +1,24 @@
...
-		esac
-	done
-	./autogen.sh || return 1
-}
+builddir="$srcdir"/Viewnior-$pkgname-$pkgver
 
 build() {
-	cd "$_builddir"
+	cd "$builddir"
+	./autogen.sh || return 1
```
jirutka: I've missed this line before; it should stay in the `prepare()` function. You can call `default_prepare` inside `prepare` to keep the default behaviour, i.e. apply patches.

```sh
prepare() {
	default_prepare || return 1

	cd "$builddir"
	./autogen.sh
} 
```

AmatCoder:
Fixed and squashed.

jirutka:
Merged in 873a73aff5c81d9390722c2d2853c914d063905a, thanks!

PR #106, opened 2016-06-03
testing/amule: upgrade to 10929

AmatCoder:


AmatCoder:
The Travis Ci build fails because "The log length has exceeded the limit of 4 MB"

7heo:
This is failing because of the build verbosity... Surely there's a way to make it less verbose (it's not even warnings it's mostly notes)

testing/amule/APKBUILD
```diff
@@ -2,33 +2,25 @@
 # Maintainer: August Klein <amatcoder@gmail.com>
 pkgname=amule
 pkgsvn=aMule-SVN-r
```
jirutka: This is not a standard abuild variable, so it must be prefixed with an underscore (`_pkgsvn`).
I would prefer to name it `_pkgname` instead, that's kinda a convention in abuilds.

AmatCoder:
Fixed abuild variable and squashed.
Also I added `-Wno-unused-local-typedefs` to `$CPPFLAGS` to suppress most of warnings.

jirutka:
Merged in ae3a0e238d0b088df86433f4e82243afaef835b4, thanks!

PR #107, opened 2016-06-03
Create README.md

7heo:
I created a README.md file for github, to be more user-friendly.


ncopa:
I think we need to mention `git send-email` to the aports mailing list too. github pull requests are only one way to send patches.

clandmeter:
And maybe an explanation or just mentioning about readonly?

7heo:
yeah that's a good idea.

PR #108, opened 2016-06-03
FIX: Ignore files in build-pkgs (changed_repos)

7heo:
When travis fetches a diff, anything that isn't hidden (starting with a
dot) will be considered as an alpine repository. It turns out that there
might be files in there as well. This commit disables the travis package
testing against those files, as they are not valid repositories.

In addition, this commit fixes a bug making it impossible to contribute
to the github alpine repo unless a package is submitted in the commit.

### DIFF
--- a/.travis/build-pkgs
+++ b/.travis/build-pkgs
@@ -75,5 +75,5 @@ if [ -n "$failed_pkgs" ]; then
 	die "Failed to build packages:$failed_pkgs"
 
 elif [ -z "$successful_pkgs" ]; then
-	die 'No packages found to be build.'
+	print -s1 -c3 'No packages found to be built.'
 fi


7heo:
This PR is to fix issues such as https://travis-ci.org/alpinelinux/aports/builds/135646856.

In a nutshell: if there is no diff involving a package: don't fail as if a package build had failed. If we need more test coverage, let's test more. Otherwise, we don't need more test coverage, so there's no reason to fail.

Barthalion:
As I proposed on IRC, I guess we need a little bit more than that.

If some APKBUILD has been modified, but `pkgver` and `pkgrel` are the same as before the commit, we should run `abuild sanitycheck` on it to catch possible syntax error. That's no perfect, but in my opinion it's easy to catch just with eyes if a pull request really modifies the resulting package. Packages with unmodifies `pkg{ver,rel}` won't be picked up by our build infrastructure anyway (and if it fails in the future, that's life. We can handle it.)

Otherwise script should just `exit 0`, because we don't care about non-APKBUILD files or commits removing packages.

7heo:
As reported by @jirutka, the previous implementation was using `ls -Ap | sed '/\/$/d'` instead of `find -type f`. This is now fixed.

7heo:
@Barthalion This PR fixes the issue where pushing changes not changing a package would be automatically rejected. I think that the issue where a package is modified but the `pkgver` or `pkgrel` haven't been modified should be solved in another PR; unless I am missing something and this current change does affect that behavior.

.travis/build-pkgs
```diff
@@ -8,8 +8,9 @@ set -eu
 # changed/created in the specified revisions.
 changed_repos() {
 	local commit_ish="$1"
+	local files="$(find -type f -maxdepth 1 | sed 's:^\./::' | tr '\n' '|' | sed 's:|$::')"
 
-	git diff-tree --name-only "$commit_ish" | grep -v '^\..*' || echo ''
+	git diff-tree --name-only "$commit_ish" | grep -Ev -e '^\..*' -e $files || echo ''
```
7heo: Adding the `-E` flag is needed since the variable `$files` now contains escaped (`|`) pipes, instead of the escaped pipes (`\|`) the previous expression produced.

.travis/build-pkgs
```diff
@@ -65,11 +66,11 @@ done
...
-	print -s1 -c2 "Successfully build packages:$successful_pkgs\n"
+	print -s1 -c2 "Successfully build packages: $successful_pkgs\n"
 fi
 if [ -n "$failed_pkgs" ]; then
-	die "Failed to build packages:$failed_pkgs"
+	die "Failed to build packages: $failed_pkgs"
 
 elif [ -z "$successful_pkgs" ]; then
-	die 'No packages found to be build.'
+	warn 'No packages found to be built.'
```
7heo: We don't want to fail if no packages have been found to require a build (see above).

.travis/build-pkgs
```diff
@@ -65,11 +66,11 @@ done
 
 printf '\n----\n'
 if [ -n "$successful_pkgs" ]; then
-	print -s1 -c2 "Successfully build packages:$successful_pkgs\n"
+	print -s1 -c2 "Successfully build packages: $successful_pkgs\n"
 fi
 if [ -n "$failed_pkgs" ]; then
-	die "Failed to build packages:$failed_pkgs"
+	die "Failed to build packages: $failed_pkgs"
```
7heo: This is where we want to fail if there is a build problem.

.travis/build-pkgs
```diff
@@ -65,11 +66,11 @@ done
...
 
 elif [ -z "$successful_pkgs" ]; then
```
7heo: Since this is an `elif`, it will NOT execute the warning `if [ -n "$failed_pkgs" ]`. So if we go to this block, we know that:
- there are no failed packages.
- there are no successfully built packages.

Therefore, no package build was triggered.

jirutka:
Please take a look at #115. I haven’t fixed all the issues yet and I’ll add comments here later.

.travis/build-pkgs
```diff
@@ -65,11 +66,11 @@ done
 
 printf '\n----\n'
 if [ -n "$successful_pkgs" ]; then
-	print -s1 -c2 "Successfully build packages:$successful_pkgs\n"
+	print -s1 -c2 "Successfully build packages: $successful_pkgs\n"
```
jirutka: There should not be a space, because if `$successful_pkgs` is not empty, then it always starts with a space. See [line 44](https://github.com/7heo/aports/blob/d418c0d19d1d74a409d11b47fbef900c3d6d13f7/.travis/build-pkgs#L44) and [line 59](https://github.com/7heo/aports/blob/d418c0d19d1d74a409d11b47fbef900c3d6d13f7/.travis/build-pkgs#L59). The same for the `$failed_pkgs` variable.

.travis/build-pkgs
jirutka: This is unnecessarily complicated. There’s a more straightforward and simpler way, see [here](https://github.com/alpinelinux/aports/pull/115/commits/960e0dad422116d9dbba27fa6bc08638e44f3526)

7heo:
PR not relevant anymore, see #115 instead.

PR #111, opened 2016-06-03
testing/gogs: Multiple fixes

7heo:
List of fixes:
- Do not try (and fail) to add the user 'gogs' to the group 'www-data'
  if the latter group doesn't exist.
- Explicitely pass the path to the .ini file to the gogs binary.
- Use openrc-run instead of runscript.
- Patch the list.tmpl template not to cause a fatal error when gogs starts.

### DIFF
--- a/testing/gogs/APKBUILD
+++ b/testing/gogs/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
-# Maintainer:
+# Maintainer: 7heo <7heo@mail.com>
 pkgname=gogs
 pkgver=0.6.1
-pkgrel=4
+pkgrel=5
 pkgdesc="A self-hosted Git service written in Go"
 url="http://gogs.io/"
 arch="all"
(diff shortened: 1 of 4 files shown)


testing/gogs/APKBUILD
```diff
@@ -1,5 +1,5 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
-# Maintainer:
+# Maintainer: 7heo <7heo@mail.com>
 pkgname=gogs
 pkgver=0.6.1
 pkgrel=4
```
jirutka: You must always bump `pkgver`, or `pkgrel` when doing any change in the abuild. That’s why CI fails when no package is build, to prevent this kind of mistakes.

7heo: There's a difference between failing a package build and not building any package.

jirutka: Travis and its GitHub integration has only two states – OK and failed.

jirutka:
Since you use group `www-data` in the APKBUILD, you must declare variable [pkggroups](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkggroups) with that group (similar to pkgusers).

PR #112, opened 2016-06-03
testing/julia and its dependencies

jirutka:
Missing dependencies:

- [x] openlibm
- [x] openspecfun
- [x] dsfmt
- [x] [suitesparse](http://faculty.cse.tamu.edu/davis/suitesparse.html) (the most problematic, https://github.com/tkelman/julia-alpine)
- [x] arpack
- [x] [pcre2](http://pcre.sourceforge.net/) (pcre 10.x)
- [x] [utf8proc](https://github.com/JuliaLang/utf8proc)

TODO

- [x] Fix the Illegal instruction error.
- [x] Fix runtime dependency on symlinks for *.so libs from *-dev packages.

/cc @albap

### DIFF
--- a/testing/julia/APKBUILD
+++ b/testing/julia/APKBUILD
@@ -1,0 +1,187 @@
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+pkgname=julia
+pkgver=0.4.5
+# Keep in sync with deps/libuv.version.
+_libuv_ver=efb40768b7c7bd9f173a7868f74b92b1c5a61a0e
+# Keep in sync with deps/Versions.make.
+_rmathjulia_ver=0.1
+pkgrel=0
+pkgdesc="A high-level, high-performance dynamic language for technical computing"
+url="http://julialang.org"
+arch="x86 x86_64"
+license="MIT"
+depends="arpack
+	fftw>=3.3
+	gmp>=5.0
+	libgit2>=0.21
+	mpfr3
+	openblas
+	openlibm
+	openspecfun>=0.4
+	pcre2>=10.0
+	suitesparse>=4.1
+	"
+depends_dev="arpack-dev
+	dsfmt-dev
+	fftw-dev
+	gmp-dev
+	libgit2-dev
+	libunwind-dev
+	llvm-dev>=3.7
+	mpfr-dev
+	openblas-dev
+	openlibm-dev
+	openspecfun-dev
+	pcre2-dev
+	suitesparse-dev
+	utf8proc-dev>=1.3
+	"
+makedepends="$depends_dev
+	bash
+	cmake
+	dsfmt
+	gfortran
+	linux-headers
+	openssl-dev
+	paxmark
+	perl
+	"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-debug"
+source="https://github.com/JuliaLang/julia/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz
+	libuv-$_libuv_ver.tar.gz::https://api.github.com/repos/JuliaLang/libuv/tarball/$_libuv_ver
+	Rmath-julia-$_rmathjulia_ver.tar.gz::https://api.github.com/repos/JuliaLang/Rmath-julia/tarball/v$_rmathjulia_ver
+	find-syslibs
+	0001-hardened.patch
+	0002-verbose-build.patch
+	0003-no-clean-docs.patch
+	0004-remove-sysctl.h.patch
+	0005-llvm-3.8.patch
+	0006-fix-contrib-install.patch
+	0007-fix-stack-addr.patch
+	0008-ldconfig-compat.patch
+	"
+builddir="$srcdir/$pkgname-$pkgver"
+ldpath="/usr/lib/julia"
+
+prepare() {
+	cd "$builddir"
+
+	cp "$srcdir"/Rmath-julia-$_rmathjulia_ver.tar.gz deps/ || return 1
+
+	# Julia needs patched libuv.
+	cp "$srcdir"/libuv-$_libuv_ver.tar.gz deps/ || return 1
+
+	# Prevent fetching of bundled stuff in the build and package phase.
+	cat <<-EOF > deps/jldownload
+		#!/bin/sh
+		echo "!!! Downloading disabled !!!"
+		echo "Abuild should not fetch any files in the build phase."
+		echo "Add all the needed files to the APKBUILD's source=."
+		exit 1
+	EOF
+
+	cat <<-EOF > Make.user
+		prefix=/usr
+		libexecdir=/usr/lib
+		sysconfdir=/etc
+		DESTDIR="$pkgdir"
+		LIBBLAS=-lopenblas
+		LIBBLASNAME=libopenblas
+		LIBLAPACK=-lopenblas
+		LIBLAPACKNAME=libopenblas
+		USE_SYSTEM_LLVM=1
+		USE_SYSTEM_LIBUNWIND=1
+		USE_SYSTEM_PCRE=1
+		USE_SYSTEM_LIBM=1
+		USE_SYSTEM_OPENLIBM=1
+		USE_SYSTEM_OPENSPECFUN=1
+		USE_SYSTEM_DSFMT=1
+		USE_SYSTEM_BLAS=1
+		USE_SYSTEM_LAPACK=1
+		USE_SYSTEM_FFTW=1
+		USE_SYSTEM_GMP=1
+		USE_SYSTEM_MPFR=1
+		USE_SYSTEM_ARPACK=1
+		USE_SYSTEM_SUITESPARSE=1
+		USE_SYSTEM_LIBUV=0
+		USE_SYSTEM_UTF8PROC=1
+		USE_SYSTEM_LIBGIT2=1
+		USE_SYSTEM_PATCHELF=1
+		USE_LLVM_SHLIB=1
+		VERBOSE=1
+	EOF
+
+	# Pre-SSE2 CPU targets are not supported;
+	# create a generic 32-bit x86 binary.
+	if [ "$CARCH" = x86 ]; then
+		echo "MARCH=i686" >> Make.user
+		echo "JULIA_CPU_TARGET=pentium4" >> Make.user
+	else
+		echo "MARCH=${CARCH/_/-}" >> Make.user
+	fi
+ 
+	default_prepare
+}
+
+build() {
+	cd "$builddir"
+
+	# If we don't clean here, then make install will recompile everything,
+	# don't know why...
+	make clean && make release
+}
+
+package() {
+	cd "$builddir"
+
+	make install || return 1
+	install -D "$srcdir"/find-syslibs "$pkgdir"/usr/share/julia/find-syslibs
+}
+
+debug() {
+	pkgdesc="$pkgdesc (debug build)"
+
+	install -d "$subpkgdir"/usr/bin
+	mv "$pkgdir"/usr/bin/*-debug "$subpkgdir"/usr/bin/ || return 1
+
+	install -d "$subpkgdir"/usr/lib/julia
+	mv "$pkgdir"/usr/lib/julia/*-debug.* "$subpkgdir"/usr/lib/julia/
+}
+
+md5sums="dc2ff07bb6fd5eeae12c7ec41ec38f22  julia-0.4.5.tar.gz
+b94c9164471d0492dce170961d7c65a2  libuv-efb40768b7c7bd9f173a7868f74b92b1c5a61a0e.tar.gz
+4e3f9e41e7b8cd3070225d1f5e8b21d9  Rmath-julia-0.1.tar.gz
+d529d078678c242b6ac61d65ad61e530  find-syslibs
+8a2ced6314457d47739deef08adb9ff4  0001-hardened.patch
+2539b4abe49d669ae01a4239fb14243f  0002-verbose-build.patch
+0fc44cc7b6a0167ee1e36310364e7049  0003-no-clean-docs.patch
+e5a50262e22a97d7f8108e6e8fc160d1  0004-remove-sysctl.h.patch
+aa0a5c577ccd3d0c652d15b6a064594e  0005-llvm-3.8.patch
+ab0e26d5ac79161d7e65481700cf9114  0006-fix-contrib-install.patch
+16f15ab47f1aff789966c90887916e9e  0007-fix-stack-addr.patch
+3cf5ef31fd586ffac9a29cc3eca00e8a  0008-ldconfig-compat.patch"
+sha256sums="441ff61b71bb0ad1d5279156105b20e9d2b59fccf2ad0c390636923d9c37ec25  julia-0.4.5.tar.gz
+872a96b616fc3517526812df11ed188500abff8dd73b75ae92377916b6efd3fe  libuv-efb40768b7c7bd9f173a7868f74b92b1c5a61a0e.tar.gz
+64c1ac8ee12859cf295d23bc1c840ea52e5abbcf4384ffacdaec0fc2246c252a  Rmath-julia-0.1.tar.gz
+b4bdd0dc307a76e51b40aa33de8856e9bd6e3d8be11b4c509d73f05d78a0cb79  find-syslibs
+587f87971d0e8fa65cf06653c56f62db283be102f8ed57712cad30d18dfb6c22  0001-hardened.patch
+892b3e0b3a94ad53a06574851dd397da4293c371ad980a03daa816a0585cb0ee  0002-verbose-build.patch
+8dc6aabe30ff02acfdae7aea3c6ef8fcf397481dfe2a3e2ad7c65bd8d1bbb09a  0003-no-clean-docs.patch
+70b9f2957d818ccd93f64a61f87eb5c7476258905d235c8e34745ec9c9e9a649  0004-remove-sysctl.h.patch
+2fa4fd8aef4a8805919cf66442f8b76f6a47886c1f6d31e7be525aad33c358c0  0005-llvm-3.8.patch
+56ce9bbbc2a623ee06646fe41a9c9170cc2637b6794ea8a3e41e1adb68e720a5  0006-fix-contrib-install.patch
+5688090d48c93ae36b068a454d0c9b11e40ba5722b4e62980bfc4ee5de8caaf7  0007-fix-stack-addr.patch
+7895279378e08f394501af568b08fb59244c272735045b077de1d7ff67810787  0008-ldconfig-compat.patch"
+sha512sums="66af1b1f25350f92c229725391993ccf2ad516c22348d3523a76c110509792c083d7ad0c32058c236900afb0b854e939d151501c2611dfd8afe4b266861287cc  julia-0.4.5.tar.gz
+21bdcffd0c13c864bac6a1f42c5f229a57fd91e124ab4c43bd14ab29e5334bfcbdcdda2e16b3b08e8f79df5ab133bac5cdaeb2a88627fe152dd91d5e18154e61  libuv-efb40768b7c7bd9f173a7868f74b92b1c5a61a0e.tar.gz
+1dd94eca05bc06c3f33e8be5087d3b1742d01c8b38588e0fa79b8f6b9998760bca21bee1b65873825c4dde2d994a8271166bc21f670a34810fa9db598473a887  Rmath-julia-0.1.tar.gz
+0283c7b0ff7dd99ae79f6ddac63513ce7d58ba31de506a3fee07af9b8882ddc275d0f9cb002381ba1e304bcacf252612fa16b21b85667422477e6b945b725899  find-syslibs
+15968d95b09ed0f2315e08a5fd14e798db7fd1faebd1327ceaf7886248b61d2e61f64e726d98cd70cbc2349c66fc3a707298a2481c6abf98c699826f3f2f4ccb  0001-hardened.patch
+266e0170c133e5be98f843adca948a4d13f3549a3a01b0c6220d495149ef26a9a837e9d7b0d9066eb7f8370a46ee5de042882a7dd585ddb245a1573882e32365  0002-verbose-build.patch
+2eae3ea85a56b5deaef464d36f543d60afb2e6a65ee4b57b599dd508046a02179ca53c5efd91351c8d0cbd5a3585b236da9b316e82416f9bff249b8b74ded305  0003-no-clean-docs.patch
+6c111ccdd12af5b00fdf0b7cc8fba11980fceccf7c246b40e791cd1be8e39cd75b8c290c65ebd9b4add60fc242261e17e94807736fce6a96733e9bf18182b489  0004-remove-sysctl.h.patch
+8546ff6c762b2a3a9f15a898e8b95b6cb0dd50a014de8349ee8d94d9663ca5031f775ec4f00e136ae2aef4778a8bdb1074930d10e2e1c40a8ec8db55fb7dc4d8  0005-llvm-3.8.patch
+8a3f14a908e3b16aef0afbbd437322ca2fdbba0bf2738f93704f19b06ef4dfee64f3e3932e4b2e18776333f25962d1b1c62c1638d4ca020570e30ddef1fb56f8  0006-fix-contrib-install.patch
+523a403095160af5f126e8883f356e52b1b0a3056f1f4921c8204fb699a729e2f5537437d00c0edb1af05d1f3c1489edc44149a629edca47bb91361243fb4db3  0007-fix-stack-addr.patch
+ab53aeaf3464236896295eab25092e4881def9e12483728be4d8323e5d3213c8b0fc064e62d21fd049df9acd2976eee032b254f10d2092ac0a4be4ecf9c319af  0008-ldconfig-compat.patch"
--- a/testing/suitesparse/APKBUILD
+++ b/testing/suitesparse/APKBUILD
@@ -1,0 +1,60 @@
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+pkgname=suitesparse
+_pkgname=SuiteSparse
+pkgver=4.5.3
+pkgrel=0
+pkgdesc="A collection of sparse matrix libraries"
+url="http://faculty.cse.tamu.edu/davis/suitesparse.html"
+arch="x86 x86_64"
+license="GPL"
+depends_dev="openblas-dev"
+makedepends="$depends_dev"
+subpackages="$pkgname-dev $pkgname-doc"
+source="http://faculty.cse.tamu.edu/davis/$_pkgname/$_pkgname-$pkgver.tar.gz
+	0001-exclude-metis.patch
+	0002-remove-rpath.patch
+	"
+builddir="$srcdir/$_pkgname"
+
+prepare() {
+	cd "$builddir"
+
+	# OpenBLAS provides also LAPACK.
+	export LAPACK="-lopenblas"
+
+	# gcc fails on internal error when using default -O3.
+	export OPTIMIZATION="-O2"
+
+	# Do not include the Partition module and METIS (it's optional and
+	# Julia doesn't need it).
+	export CFLAGS="$CFLAGS -DNPARTITION"
+	rm -r metis-*
+
+	default_prepare
+}
+
+build() {
+	cd "$builddir"
+
+	make library
+}
+
+package() {
+	cd "$builddir"
+
+	make install \
+		INSTALL="$pkgdir/usr" \
+		INSTALL_INCLUDE="$pkgdir/usr/include/$pkgname" \
+		INSTALL_DOC="$pkgdir/usr/share/doc/$pkgname"
+}
+
+md5sums="8ec57324585df3c6483ad7f556afccbd  SuiteSparse-4.5.3.tar.gz
+308b4f96e82d7f5c92335ca6c7c74251  0001-exclude-metis.patch
+ce5696f96435db0125ac1815888fb7f6  0002-remove-rpath.patch"
+sha256sums="6199a3a35fbce82b155fd2349cf81d2b7cddaf0dac218c08cb172f9bc143f37a  SuiteSparse-4.5.3.tar.gz
+06f825ed80f73c30d71439650b951ea835a6139fe2e30078ed155fdecc7606ab  0001-exclude-metis.patch
+4db04c59039e9fe50f75d00fb662eaffbd76ede498ff8ec8858b697df2679e32  0002-remove-rpath.patch"
+sha512sums="88a5ba6c70e4d66751d1ac32fd8998cde11509e7b034814c34757180db70ce2838ebb77fd7c617bf96e473214c70ccbdd0064b4f5e4dd69ae4c51dfeb9696496  SuiteSparse-4.5.3.tar.gz
+aa2d62a371eefb33f0c81d3f7b50cf36e76c4f010fb5bb424e5e7344a1921cb6a430742ebb92df42cef587ec46ad8d2b777d3d9a62eb4a128e02bee8ad5de272  0001-exclude-metis.patch
+cd40b696515b26c95298b435a5111e582dc42c2fa2e8ef9889558467d375b1d19452838b9463df66958e73ddb0c8c0191936e3e952586a3c7d345edb08fa7e62  0002-remove-rpath.patch"
--- a/testing/arpack/APKBUILD
+++ b/testing/arpack/APKBUILD
@@ -1,0 +1,49 @@
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+pkgname=arpack
+_pkgname=$pkgname-ng
+pkgver=3.3.0
+pkgrel=0
+pkgdesc="Collection of Fortran77 subroutines designed to solve large scale eigenvalue problems"
+url="https://github.com/opencollab/arpack-ng"
+arch="all"
+license="BSD"
+depends=""
+depends_dev="openblas-dev"
+makedepends="$depends_dev autoconf automake gfortran libtool"
+subpackages="$pkgname-dev $pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/opencollab/$_pkgname/archive/$pkgver.tar.gz"
+builddir="$srcdir/$_pkgname-$pkgver"
+
+prepare() {
+	cd "$builddir"
+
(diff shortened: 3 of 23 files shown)


albap:
Hell yeah, great work!
@tkelman should also be interested in this.

tkelman:
The internal compiler error in suitesparse (https://github.com/tkelman/julia-alpine/issues/1#issuecomment-177282137) should probably be looked into from the alpine and/or gcc side.

> utf8proc (multiple variants of this library exist :hankey:)

What do you mean? The Public Software Group were the original authors but no longer had the manpower to maintain the library beyond version 1.1.6. They transferred maintainership over to the JuliaLang project and we've been making releases since then. Are there other forks out there?

I also haven't spun up a docker container to figure out what's wrong with our `contrib/install.sh` script when running on busybox, hopefully it'll be an easy fix once I reproduce it and play with it a bit.

testing/suitesparse/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+source="http://faculty.cse.tamu.edu/davis/$_pkgname/$_pkgname-$pkgver.tar.gz
+	0001-exclude-metis.patch
+	0002-remove-rpath.patch
+	"
+builddir="$srcdir/$_pkgname"
+
+prepare() {
+	cd "$builddir"
+
+	# OpenBLAS also provides LAPACK (?)
```
tkelman: it can, depends how you build openblas

albap: I think Alpine's openblas is built with lapack.
And as I understand openblas is faster and more optimized, so better use it than standard lapack, right?

tkelman: openblas has a few changes to lapack, I don't think they're absolutely critical for performance but sometimes it can be more convenient to have the libblas and liblapack just in the same library (anything that is ever set up to use Intel MKL probably has to support this kind of setup). Some distros are very traditional about how they package separate blas and lapack libraries, maybe for the sake of switching them around with alternatives, but the threading api's in optimized blas/lapack implementations are all different so I don't think alternatives is really all that useful here.

jirutka:
> The internal compiler error in suitesparse (https://github.com/tkelman/julia-alpine/issues/1#issuecomment-177282137) should probably be looked into from the alpine and/or gcc side.

I’ve already solved it – found a workaround, to be more precise – by forcing `-O2` instead of default `-O3`. It’s a problem with too strong optimization.

>> utf8proc (multiple variants of this library exist :hankey:)

> What do you mean? … Are there other forks out there?

I also found http://source.netsurf-browser.org/libutf8proc.git/; Arch Linux and Gentoo uses this fork as the upstream, so I was confused.

> …what's wrong with our `contrib/install.sh` script when running on busybox…

Do you mean that install.sh needs gawk or something else?

tkelman:
> I’ve already solved it – found a workaround, to be more precise – by forcing `-O2` instead of default `-O3`. It’s a problem with too strong optimization.

Should still be reported and looked into somewhere.

> I also found http://source.netsurf-browser.org/libutf8proc.git/

Huh, no kidding. Wonder why they didn't raise issues/PR's. Little annoying for packagers to have to deal with versions that have replaced build systems. Maybe netsurf should maintain their build harness separately and discuss the small API changes so julia and netsurf don't have conflicting dependency versions.

> Do you mean that install.sh needs gawk

Right, that. https://github.com/tkelman/julia-alpine/issues/2#issuecomment-223998252
There's also a potential busybox bug in there but it's not causing a failure if you use gawk.

albap:
> Do you mean that install.sh needs gawk or something else?

I think putting gawk as a makedepend is a bad idea and we should instead wait for a fix on busybox side or adpt install.sh depending on busybox devs answer.

jirutka:
> I think putting gawk as a makedepend is a bad idea…

Why? Are you sure that it’s a bug in Busybox and not simply a missing feature? However, I agree that it’d be better to fix `install.sh` script to be compatible with Busybox’s awk. Meanwhile, we can just use gawk.

tkelman:
I think the awk vs gawk difference and the output from `cp -va` are actually two separate issues. I happened to find the latter while looking into, but the former is what actually causes more serious problems, and I haven't fully debugged.

jirutka:
What’s wrong with output from `cp -va`?

tkelman:
> What’s wrong with output from `cp -va`?

There is none, if what you're copying is a symlink. http://lists.busybox.net/pipermail/busybox/2016-June/084252.html

I finished debugging the awk issue. It's different treatment of `length`. I think there's an easy fix of calling `length($0)` instead.
```
/ # echo foobar | busybox awk '{print substr($0, 2, length-2)}'
awk: cmd. line:1: Unexpected token
/ # echo foobar | busybox awk '{print substr($0, 2, length($0)-2)}'
ooba
/ # echo foobar | gawk '{print substr($0, 2, length-2)}'
ooba
/ # echo foobar | gawk '{print substr($0, 2, length($0)-2)}'
ooba
```

tkelman:
PR submitted: https://github.com/JuliaLang/julia/pull/16784

jirutka:
@tkelman I'm starting to be clueless. 😿 Julia and all its dependencies builds fine, but when I run julia, this happens:

```
$ julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.5
 _/ |\__'_|_|_|\__'_|  |
|__/                   |  x86_64-alpine-linux-musl


signal (4): Illegal instruction
jl_unprotect_stack at /usr/bin/../lib/julia/libjulia.so (unknown line)
jl_unprotect_stack at /usr/bin/../lib/julia/libjulia.so (unknown line)
jl_switchto at /usr/bin/../lib/julia/libjulia.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
_init at /usr/lib/julia/sys.so (unknown line)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f4d6b945d67)
unknown function (ip: 0x7f4d6b9458bf)
__libc_start_main at /lib/ld-musl-x86_64.so.1 (unknown line)
Illegal instruction
```

When I use `--precompiled=no`, then:

```
$ julia --precompiled=no
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.5
 _/ |\__'_|_|_|\__'_|  |
|__/                   |  x86_64-alpine-linux-musl


signal (4): Illegal instruction
jl_unprotect_stack at /usr/bin/../lib/julia/libjulia.so (unknown line)
jl_unprotect_stack at /usr/bin/../lib/julia/libjulia.so (unknown line)
jl_switchto at /usr/bin/../lib/julia/libjulia.so (unknown line)
yieldto at ./task.jl:71
wait at ./task.jl:371
Illegal instruction
```

When I try to remove `/usr/lib/julia/sys.so`, then:

```
$ julia
Illegal option -p
ERROR: system image file "/usr/bin/../lib/julia/sys.so" not found
```

I’ve tried to compile all the packages in this PR and run julia on my machine with grsecurity/PAX kernel inside a LXC container and also compile on Travis and run inside OpenVZ container without grsecurity/PAX on different infra. It’s still the same.

I’ve tried to recompiled OpenBLAS without `DYNAMIC_ARCH`, but with no difference. And also many other tries, changing some julia flags, I don’t remember now.

Do you have any idea why this happens? I can give you access to some container with Alpine, if it helps.

strace logs:
[strace-julia-no-precompiled.txt](https://github.com/alpinelinux/aports/files/301170/strace-julia-no-precompiled.txt)
[strace-julia-without-sys.so.txt](https://github.com/alpinelinux/aports/files/301169/strace-julia-without-sys.so.txt)
[strace-julia.txt](https://github.com/alpinelinux/aports/files/301168/strace-julia.txt)

jirutka:
BTW all the packages from this PR are now available in [my aports repository](https://github.com/jirutka/aports), so you can try it yourself: https://alpine.geeknet.cz/packages/edge/legacy

tkelman:
Oh, or we may need to set JULIA_CPU_TARGET if you're building on a newer system than you're running the binaries?

jirutka:
> Oh, or we may need to set JULIA_CPU_TARGET if you're building on a newer system than you're running the binaries?

I’ve already tried that, with no difference.

jirutka:
Sight, you don’t need a Dockerfile for this, once we have all the packages, it’s trivial to install them.

```sh
echo "http://dl-cdn.alpinelinux.org/alpine/v3.4/main" > /etc/apk/repositories
echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
echo "https://alpine.geeknet.cz/packages/edge/testing" >> /etc/apk/repositories

wget -O "/etc/apk/keys/jakub@jirutka.cz-56d0d9fd.rsa.pub" https://alpine.geeknet.cz/keys/jakub%40jirutka.cz-56d0d9fd.rsa.pub

apk update
apk add julia julia-dev
julia
```

BTW You _must_ install even julia-dev, there’s some bug that I’ll fix later.

tkelman:
putting RUN at the start of those lines, `from: alpine` at the top and calling it a Dockerfile at least grabs the right image and runs things via a familiar and widely-used ui without having to mess with chroots or installing a distro that not many people have heard of... much easier to get upstreams to be able to reproduce musl specific issues that way.

jirutka:
@albap could you please do that? I don’t use and don’t wanna use Docker.

jirutka:
@tkelman It’s still a mystery why it refuses to run without sys.so… it’s only a cache, isn’t it? And where this `Illegal option -p` comes from?

```
$ julia
Illegal option -p
ERROR: system image file "/usr/bin/../lib/julia/sys.so" not found
```

tkelman:
dunno where the -p is coming from, maybe a difference in getopt. the only thing you can really do without a system image present is build a new system image. the repl is written in julia so it won't run without a built standard library present.

albap:
> @albap could you please do that? I don’t use and don’t wanna use Docker.

Done. https://github.com/JuliaLang/julia/issues/16792

jirutka:
Now it starts to be really interesting… I’ve compiled `julia-debug` and it works!

jirutka:
Hm, with julia-debug REPL finally runs, but…

```
$ julia-debug build_sysimg.jl --help
Usage: build_sysimg.jl <sysimg_path> <cpu_target> <usrimg_path.jl> [--force] [--debug] [--help]
   <sysimg_path>    is an absolute, extensionless path to store the system image at
   <cpu_target>     is an LLVM cpu target to build the system image against
   <usrimg_path.jl> is the path to a user image to be baked into the system image
   --debug          Using julia-debug instead of julia to build the system image
   --force          Set if you wish to overwrite the default system image
   --help           Print out this help text and exit

 Example:
   build_sysimg.jl /usr/local/lib/julia/sys core2 ~/my_usrimg.jl --force

 Running this script with no arguments is equivalent to:
Two passes with the same argument (-LowerSIMDLoop) attempted to be registered!

signal (11): Segmentation fault
_ZN4llvm2cl19generic_parser_base10findOptionEPKc at /usr/bin/../lib/libLLVM-3.8.so (unknown line)
_ZN4llvm23SmallVectorTemplateBaseINS_2cl6parserIPKNS_8PassInfoEE10OptionInfoELb0EE4growEm at /usr/bin/../lib/libLLVM-3.8.so (unknown line)
_ZN4llvm12PassRegistry12registerPassERKNS_8PassInfoEb at /usr/bin/../lib/libLLVM-3.8.so (unknown line)
unknown function (ip: 0x7fab6311f13b)
_dl_debug_state at /lib/ld-musl-x86_64.so.1 (unknown line)
Segmentation fault
```

albap:
@jirutka the internal instruction error was fixed on https://github.com/JuliaLang/julia/pull/16813
Does julia work now (master branch)?

testing/julia/APKBUILD
```diff
@@ -0,0 +1,169 @@
...
+	openlibm-dev
+	openspecfun-dev
+	pcre2-dev
+	suitesparse-dev
+	utf8proc-dev>=1.3
+	"
+makedepends="$depends_dev
+	bash
+	cmake
+	gawk
```
albap: gawk is not needed, @tkelman fixed the install.sh issue https://github.com/JuliaLang/julia/pull/16784

jirutka:
@albap I’ve backported that patch to 0.4.5 and it works! 🎉

albap:
Nice, it's getting close!

jirutka:
After [this travis job](https://travis-ci.org/jirutka/aports/builds/136022783) completes, you can try the latest version of the julia package from my repository. Don’t forget to update both julia and julia-dev.

It may segfault because of bug in LLVM, I’ve opened #120 for it.

albap:
Great, I can't wait to test it. As soon as https://github.com/alpinelinux/aports/pull/120 is accepted, I'll do it.

It would also be cool if the container on the testing OpenVZ server you gave access to @yuyichao and @tkelman was updated and they tested julia on alpine for any errors.
I don't have many Julia programs to test and certainly not many complex ones, but I'm sure they as developers know well what to test.

I'm excited.

jirutka:
You can test it even know, you can at least print "Hello world!" in REPL. 😸

I’ve already updated the julia package on the testing server and also installed patched LLVM.

albap:
Got to install julia and julia-dev, neat!
An unrelated alpine bug I had was that ERROR: openblas-dev-0.2.18-r0: trying to overwrite usr/include/cblas.h owned by lapack-dev-3.6.0-r0. and many similar errors. It seems on alpine openblas and lapack can't be installed together (I need lapack for R - I think both R and Julia on alpine should use openblas and we get rid of lapack package that isn't used anywhere?).

Anyway, **Julia is installed!**

Let's do some tests.

yuyichao:
Response to https://github.com/JuliaLang/julia/issues/16792#issuecomment-224456213 since this is fairly off topic there...

> ZSH shell? Could you please describe me steps to reproduce it?

It seems that something in my `.zshrc` triggers it. I haven't reduced it yet....

> Good idea, I’ve added subpackage julia-debug.

FYI, the julia unwinder currently can't make use of this information (so you might see weird names in the backtrace). GDB should though.

yuyichao:
And yet another OT, the `py-pexpect` package seems to be broken since it requires ptyprocess and there doesn't seem to be package for it....

jirutka:
@albap I know about it, openblas also provides lapack. Maybe it'd be better to separate openblas' lapack into a subpackage. Anyway, I'm gonna change R abuild to use openblas instead of lapack.

albap:
So far it seems to work perfectly!
I've found only 1 problem that I actually don't think it's with the alpine julia package, but something else.
I did Pkg.add("Gadfly") and Pkg.add("Cairo"). It worked.
But then:
```julia> using Gadfly
INFO: Precompiling module Gadfly...
```
And it hangs there. It's been almost 10 minutes!

Anyway, looks like Alpine Linux finally got Julia packaged and it's working! At least apparently, let's wait for further tests.
For now I think it could be added to Alpine's /testing repository. Could you do it please @ncopa ?
And I'm really grateful to you @jirutka for the package and @tkelman and @yuyichao for the help. I was trying for a long time to compile Julia on Alpine, found some small issues myself that I fixed, but missed many others. I know this was a huge amount of work and I thank you all. I hope the fixes can be upstreamed and Julia eventually works clean, without any patches.

albap:
So, after installing Gadfly (+DataArrays, RDatasets) there are some issues and I'm not sure if it's due to Julia packaged using musl, etc or a general problem with Gadfly. I think @dcjones can help us here, if he has a minute.

```
julia> using Gadfly, RDatasets

julia> iris = data("datasets", "iris")
ERROR: MethodError: `data` has no method matching data(::ASCIIString, ::ASCIIString)
```

And another one.

```
julia> using Gadfly, DataArrays, RDatasets

julia> plot(dataset("lattice", "singer"), x=:VoicePart, y=:Height, Geom.boxplot)
Error showing value of type Gadfly.Plot:
ERROR: could not spawn `xdg-open /tmp/julia_CbJmMe.html`: no such file or directory (ENOENT)
 in _jl_spawn at ./process.jl:262
```

This second error especially appears to be rare, so it makes me believe it's local.

jirutka:
Hold on, one issue left: Fix runtime dependency on symlinks for *.so libs from *-dev packages. I wanna fix it before merging.

albap:
I found what about the second Gadfly error. It needs `xdg-utils` to be installed. Should it be added as a dependency?

```
julia> using Gadfly, DataArrays, RDatasets

julia> plot(dataset("lattice", "singer"), x=:VoicePart, y=:Height, Geom.boxplot)
awk: cmd. line:13: Unexpected token
```

Maybe another small issue because of busybox awk (alpine's default)?
But the graph is plotted! Nice.

tkelman:
Dependencies of julia packages rather than julia itself should be dealt with in the Julia packages, not here. If it's not a dependency for all usage it doesn't belong here. Try running `Base.runtests()`, I suspect you'll see several issues still.

albap:
@tkelman 
```
julia> Base.runtests()
        From worker 3:       * linalg/qr             in 129.59 seconds, maxrss  446.45 MB
        From worker 3:       * linalg/dense          in  87.33 seconds, maxrss  588.41 MB
        From worker 3:       * linalg/matmul         in  29.36 seconds, maxrss  615.04 MB
        From worker 3:       * linalg/schur          in  13.61 seconds, maxrss  615.04 MB
        From worker 3:       * linalg/special        in   8.52 seconds, maxrss  615.04 MB
        From worker 2:       * linalg/triangular     in 273.66 seconds, maxrss  701.01 MB
        From worker 3:       * linalg/eigen          in   5.16 seconds, maxrss  615.04 MB
        From worker 3:       * linalg/bunchkaufman   in   1.57 seconds, maxrss  616.41 MB
        From worker 2:       * linalg/svd            in  14.02 seconds, maxrss  714.76 MB
        From worker 3:       * linalg/lapack         in  35.01 seconds, maxrss  641.29 MB
        From worker 2:       * linalg/tridiag        in  25.06 seconds, maxrss  727.14 MB
        From worker 3:       * linalg/bidiag         in  27.82 seconds, maxrss  708.50 MB
        From worker 3:       * linalg/pinv           in  16.26 seconds, maxrss  766.57 MB
        From worker 2:       * linalg/diagonal       in  42.33 seconds, maxrss  790.86 MB
        From worker 3:       * linalg/givens         in   4.51 seconds, maxrss  766.57 MB
        From worker 2:       * linalg/cholesky       in  28.87 seconds, maxrss  837.79 MB
        From worker 3:       * linalg/lu             in  38.48 seconds, maxrss  791.15 MB
        From worker 2:       * linalg/symmetric      in  18.78 seconds, maxrss  863.64 MB
        From worker 3:       * linalg/generic        in   8.14 seconds, maxrss  793.10 MB
        From worker 2:       * linalg/uniformscaling in   3.44 seconds, maxrss  870.38 MB
        From worker 3:       * linalg/arnoldi        in  41.47 seconds, maxrss  837.35 MB
        From worker 3:       * keywordargs           in   4.82 seconds, maxrss  837.35 MB
        From worker 2:       * core                  in  59.00 seconds, maxrss  907.50 MB
        From worker 2:       * printf                in  14.62 seconds, maxrss  911.26 MB
        From worker 2:       * char                  in   2.36 seconds, maxrss  916.38 MB
WARNING: iconv not found, skipping unicode tests!
        From worker 2:       * string                in  58.74 seconds, maxrss  984.51 MB
        From worker 2:       * triplequote           in   0.70 seconds, maxrss  985.38 MB
        From worker 2:       * unicode               in  32.62 seconds, maxrss 1007.65 MB
        From worker 3:       * numbers               in 212.04 seconds, maxrss 1072.98 MB
        From worker 3:       * dict                  in  38.09 seconds, maxrss 1102.07 MB
        From worker 3:       * hashing               in  20.81 seconds, maxrss 1133.27 MB
        From worker 3:       * remote                in   1.61 seconds, maxrss 1133.30 MB
        From worker 3:       * iobuffer              in   6.65 seconds, maxrss 1139.16 MB
        From worker 3:       * staged                in   3.53 seconds, maxrss 1145.08 MB
        From worker 2:       * dates                 in 168.72 seconds, maxrss 1132.40 MB
        From worker 2:       * tuple                 in   4.28 seconds, maxrss 1137.44 MB
        From worker 3:       * arrayops              in 134.30 seconds, maxrss 1228.55 MB
        From worker 3:       * reduce                in  15.16 seconds, maxrss 1228.55 MB
        From worker 3:       * reducedim             in  51.89 seconds, maxrss 1228.55 MB
        From worker 3:       * random                in  46.86 seconds, maxrss 1228.55 MB
        From worker 3:       * abstractarray         in  87.11 seconds, maxrss 1308.50 MB
        From worker 3:       * intfuncs              in   2.28 seconds, maxrss 1311.33 MB
        From worker 3:       * simdloop              in   3.48 seconds, maxrss 1314.48 MB
        From worker 3:       * blas                  in  19.05 seconds, maxrss 1332.46 MB
        From worker 3:       * sparse                in 237.37 seconds, maxrss 1440.18 MB
        From worker 3:       * bitarray              in 239.98 seconds, maxrss 1515.61 MB
        From worker 3:       * copy                  in   6.82 seconds, maxrss 1519.68 MB
        From worker 3:       * math                  in  45.53 seconds, maxrss 1550.10 MB
        From worker 3:       * fastmath              in  16.19 seconds, maxrss 1563.57 MB
        From worker 3:       * functional            in   6.02 seconds, maxrss 1564.92 MB
        From worker 3:       * operators             in   2.13 seconds, maxrss 1567.16 MB
        From worker 3:       * path                  in  10.63 seconds, maxrss 1574.89 MB
        From worker 3:       * ccall                 in  11.35 seconds, maxrss 1576.84 MB
        From worker 3:       * parse                 in  11.52 seconds, maxrss 1581.75 MB
        From worker 3:       * loading               in   0.31 seconds, maxrss 1581.75 MB
        From worker 3:       * bigint                in  10.00 seconds, maxrss 1591.75 MB
        From worker 3:       * sorting               in 121.97 seconds, maxrss 1659.71 MB
        From worker 3:       * statistics            in  25.92 seconds, maxrss 1660.71 MB
        From worker 3:       * spawn                       [stdio passthrough ok]
        From worker 3:   in  37.40 seconds, maxrss 1667.51 MB
        From worker 3:       * backtrace             in   1.27 seconds, maxrss 1667.51 MB
        From worker 3:       * priorityqueue         in   4.18 seconds, maxrss 1667.51 MB
        From worker 3:       * file                  in  49.73 seconds, maxrss 1667.51 MB
        From worker 3:       * mmap                  in  21.75 seconds, maxrss 1667.51 MB
        From worker 3:       * version               in   8.16 seconds, maxrss 1667.51 MB
        From worker 3:       * resolve               in  21.65 seconds, maxrss 1667.51 MB
        From worker 3:       * pollfd                in   5.09 seconds, maxrss 1667.51 MB
        From worker 3:       * mpfr                  in  17.26 seconds, maxrss 1667.51 MB
        From worker 3:       * broadcast             in  47.76 seconds, maxrss 1667.51 MB
        From worker 3:       * complex               in  15.71 seconds, maxrss 1667.51 MB
        From worker 3:       * socket                in  17.22 seconds, maxrss 1667.51 MB
        From worker 3:       * floatapprox           in   1.79 seconds, maxrss 1667.51 MB
        From worker 3:       * readdlm               in  71.80 seconds, maxrss 1667.51 MB
The following 'Returned code...' warnings indicate normal behavior:
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
WARNING: Returned code may not match what actually runs.
        From worker 3:       * reflection            in   4.90 seconds, maxrss 1667.51 MB
        From worker 3:       * regex                 in   3.70 seconds, maxrss 1667.51 MB
        From worker 3:       * float16               in   5.44 seconds, maxrss 1667.77 MB
        From worker 3:       * combinatorics         in  11.48 seconds, maxrss 1668.02 MB
        From worker 3:       * sysinfo               in   5.03 seconds, maxrss 1668.02 MB
        From worker 3:       * rounding              in   4.08 seconds, maxrss 1668.02 MB
        From worker 2:       * subarray              in 1456.31 seconds, maxrss 1520.69 MB
        From worker 2:       * mod2pi                in   0.72 seconds, maxrss 1520.69 MB
        From worker 2:       * euler                 in   7.31 seconds, maxrss 1520.69 MB
        From worker 2:       * show                  in  28.21 seconds, maxrss 1520.69 MB
        From worker 2:       * lineedit              in  13.74 seconds, maxrss 1520.69 MB
        From worker 2:       * replcompletions       in  15.99 seconds, maxrss 1520.69 MB
        From worker 2:       * repl                  in  20.71 seconds, maxrss 1520.69 MB
        From worker 2:       * replutil              in   5.50 seconds, maxrss 1520.69 MB
        From worker 2:       * sets                  in  13.06 seconds, maxrss 1520.69 MB
        From worker 2:       * test                  in   1.53 seconds, maxrss 1520.69 MB
        From worker 2:       * goto                  in   0.24 seconds, maxrss 1520.69 MB
        From worker 2:       * llvmcall              in   0.25 seconds, maxrss 1520.69 MB
        From worker 2:       * grisu                 in  22.70 seconds, maxrss 1520.69 MB
        From worker 2:       * nullable              in   8.23 seconds, maxrss 1520.69 MB
        From worker 2:       * meta                  in   1.03 seconds, maxrss 1520.69 MB
        From worker 2:       * profile               in  14.79 seconds, maxrss 1520.69 MB
        From worker 2:       * libgit2               in   0.11 seconds, maxrss 1520.69 MB
        From worker 2:       * docs                  in  14.67 seconds, maxrss 1520.69 MB
        From worker 2:       * markdown              in  17.82 seconds, maxrss 1520.69 MB
        From worker 3:       * ranges                in 194.38 seconds, maxrss 1668.02 MB
        From worker 2:       * base64                in   1.54 seconds, maxrss 1520.69 MB
        From worker 2:       * functors              in   2.61 seconds, maxrss 1520.69 MB
        From worker 3:       * serialize             in  12.22 seconds, maxrss 1668.02 MB
        From worker 3:       * enums                 in   5.72 seconds, maxrss 1668.02 MB
        From worker 2:       * misc                  in 104.51 seconds, maxrss 1585.68 MB
        From worker 2:       * i18n                  in   0.15 seconds, maxrss 1585.68 MB
        From worker 2:       * workspace             in   1.20 seconds, maxrss 1585.68 MB
        From worker 3:       * cmdlineargs           in  92.44 seconds, maxrss 1668.02 MB
Illegal option -p
        From worker 2:       * libdl                 in   1.81 seconds, maxrss 1585.68 MB
        From worker 2:       * intset                in   5.04 seconds, maxrss 1585.91 MB
        From worker 3:       * int                   in   9.24 seconds, maxrss 1668.02 MB
        From worker 2:       * floatfuncs            in   7.35 seconds, maxrss 1587.50 MB
        From worker 2:       * inline                in   1.20 seconds, maxrss 1587.50 MB
        From worker 2:       * fft                   in  73.56 seconds, maxrss 1587.50 MB
WARNING: Forcibly interrupting busy workers
WARNING: rmprocs: process 1 not removed
        From worker 3:       * parallel              in 113.63 seconds, maxrss 1668.02 MB
        From worker 2:       * dsp                   in  54.69 seconds, maxrss 1587.50 MB
        From worker 3:       * examples              in 165.76 seconds, maxrss 1668.02 MB
        From worker 1:       * compile
NOTE: The following 'replacing module' warning indicates normal operation:
WARNING: replacing module Foo4b3a94a1a081a8cb
WARNING: replacing module Foo4b3a94a1a081a8cb
WARNING: replacing module Foo4b3a94a1a081a8cb

NOTE: The following 'LoadError: __precompile__(false)' indicates normal operation
ERROR: LoadError: __precompile__(false) is not allowed in files that are being precompiled
 in _init at /usr/lib/julia/sys.so (repeats 2 times)
while loading /tmp/tmpoGEJkA/Baz.jl, in expression starting on line 1
INFO: Recompiling stale cache file /tmp/tmpoGEJkA/FooBar.ji for module FooBar.

NOTE: The following 'LoadError: break me' indicates normal operation
INFO: Recompiling stale cache file /tmp/tmpPJNebh/FooBar.ji for module FooBar.
ERROR: LoadError: break me
 in _init at /usr/lib/julia/sys.so (repeats 3 times)
while loading /tmp/tmpoGEJkA/FooBar.jl, in expression starting on line 3
 in  91.73 seconds, maxrss  160.58 MB
    SUCCESS

julia>
```

tkelman:
Hey, not bad at all. A warning about iconv, and that `Illegal option -p` again. You can also do `Base.runtests("pkg git")` (or, on master, `Base.runtests("pkg libgit2-online")`) for a few tests that aren't in the default set because they require internet access.

If you really want to throw down a gauntlet you can fork https://github.com/JuliaCI/PackageEvaluator.jl and modify the vagrant image to use alpine instead of ubuntu. But at this point that's overkill, things are looking encouraging and I'd say go ahead and use it for whatever you'd like to. You'll probably find more portability issues in various places in packages or base Julia, we can report and gradually fix them as you find them. For example I have no idea where in the process of creating a plot Gadfly might be calling `awk`, but it should be possible to track that down. Maybe put a wrapper script temporarily on your path that at least tells you what arguments it's getting called with, which may help narrow it down.

jirutka:
I think that I know the source of this mysterious `Illegal option -p` error! Julia uses `ldconfig -p` to find location of *.so libs, but ldconfig on Alpine doesn’t know the option `-p`.

```
$ ldconfig -p
Illegal option -p
```

tkelman:
Should it? I see from http://git.alpinelinux.org/cgit/aports/plain/main/musl/ldconfig that it just "eats" all other options?

jirutka:
@tkelman I dunno? Julia uses it to find location of shared libs, https://github.com/JuliaLang/julia/issues/6742. Do you have any idea how to fix it or how to simply workaround it?

yuyichao:
Sorry I'm not sure how to open an issue (seems disabled?) so I'll just follow up here about the zsh issue.

The reduced `.zshrc` is

```
__yyc_zle_init_func() {
    zle -D zle-line-init
}

zle -N __yyc_zle_init_func
zle -A __yyc_zle_init_func zle-line-init
zle -D __yyc_zle_init_func
```

On the test setup you provide on my account, starting zsh with this `.zshrc` and type `l` (as in the first letter of `ls`) very reliably segfaults zsh. (The full zshrc is [here](https://github.com/yuyichao/dotfile/tree/master/zsh), which is even more likely to segfault).

If this were to be a julia segfault, I would think it is a memory corruption due to oob access. Not sure about what it might be for zsh though.

jirutka:
🎉 I’ve solved (workarounded) the last problem, so the julia package is now complete! 🎉

Julia uses `ldconfig -p` to find system libraries. The problem is that musl’s ldconfig doesn’t support the option `-p`. Moreover, ldconfig relies on `.pc` files that are installed only with `-dev` subpackages.

The workaround lies in using [custom script](https://github.com/jirutka/alpine-aports/blob/f0d2bb2ecf486ca754b40b738b3d784534940093/testing/julia/find-syslibs) that adapts output of scanelf to the format produced by `ldconfig -p` and [patching](https://github.com/jirutka/alpine-aports/blob/f0d2bb2ecf486ca754b40b738b3d784534940093/testing/julia/0008-ldconfig-compat.patch) Julia to use that script instead of `ldconfig -p`.

It seems that it works well, it’s not needed to install `julia-dev` anymore, mysterious `Illegal option -p` is gone and everything seems to be fine! 😸

albap:
:tada: **It's working perfectly** :tada:
I think it's safe to push it to Alpine's /testing repository.
Great work, this was for sure one of the hardest programs to package.
I'm impressed @jirutka 
Thanks **a lot**

jirutka:
@yuyichao, thanks a lot for the steps how to reproduce segfault on ZSH. I’ve opened a [bug report](https://bugs.alpinelinux.org/issues/5699) for it in Alpine’s issue tracking system. That’s also answer to your question about GitHub Issues, we don’t use them, so they are disabled.

jirutka:
Merged into upstream, julia is now available in the official testing repository. 😸 

@albap Julia package in the official repository has lower `pkgrel` than the package in my repository, so it will not be upgraded for you. You must remove my repository from your repositories list and then reinstall julia to get the latest version. However, it’s not necessary, you can also left it as is for now.

albap:
@jakub just so you know, we finally removed all bashisms from Julia with https://github.com/JuliaLang/julia/pull/16825, https://github.com/JuliaLang/julia/pull/16827, https://github.com/JuliaLang/julia/pull/16829 and https://github.com/JuliaLang/julia/pull/16838

Now for the next Julia release (0.4.6) we should be able to remove bash from makedepends :smiley:

tkelman:
you'll want to check against the release-0.4 branch, we don't make releases directly from master. just cherry-picking those commits might be enough, but I would like them to be on master for a little longer so you might need them as local patches until either 0.4.7 or 0.5.0, whichever happens first

albap:
Didn't know about that, thanks for the info.

tkelman:
I reported the suitesparse gcc internal compiler error: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71505

Would've saved some time if creduce (http://embed.cs.utah.edu/creduce/) had been packaged in alpine. It depends on llvm-dev, clang-dev, flex, zlib-dev, and 4 perl modules (optionally 2 more, but the optional ones had some issues). Here's the dockerfile I put together to build it: https://github.com/tkelman/julia-alpine/blob/creduce/Dockerfile

tkelman:
Who's the Alpine maintainer for gcc? They apparently fixed the bug causing the internal compiler error and it's a tiny patch, see https://gcc.gnu.org/viewcvs/gcc/branches/gcc-5-branch/gcc/tree-vect-data-refs.c?r1=237373&r2=237372&pathrev=237373. Would be good to see if applying that to Alpine's gcc allows SuiteSparse to be built at `-O3` again.

PR #113, opened 2016-06-06
testing/prayer: Fix include, so it builds against tidyhtml-5.1.25

ccxcz:
It's just a simple change of header name in upstream tidyhtml.


jirutka:
When you modify an abuild, you must also bump `pkgrev` (when not bumping `pkgver`), otherwise abuild will not be rebuilded.

Also please add comment to `htmltidy_buffio.patch` saying why it’s needed (i.e. for compatibility with recent tidyhtml).

jirutka:
@ccxcz ping

PR #119, opened 2016-06-07
Add php7-apcu, the PHP7 APC user cache #5366

andypost:
That's update based on php5-apcu package

### DIFF
--- a/testing/php7-apcu/APKBUILD
+++ b/testing/php7-apcu/APKBUILD
@@ -1,37 +1,36 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Maintainer: Carlo Landmeter <clandmeter@gmail.com>
-_php=php7
-pkgname=${_php}-apcu
+pkgname=php7-apcu
 _pkgreal=apcu
 # release 5 is php7+
 pkgver=5.1.5
 _pkgver=${pkgver/_rc/RC}
-pkgrel=1
+pkgrel=0
 pkgdesc="PHP extension APC User Cache"
 url="http://pecl.php.net/package/$_pkgreal"
 arch="all"
 license="PHP"
 depends=
-pecldepends="${_php}-dev autoconf"
+pecldepends="php7-dev autoconf"
 makedepends="$pecldepends pcre-dev"
 install=""
 subpackages=""
 source="http://pecl.php.net/get/$_pkgreal-$_pkgver.tgz"
 
-_builddir="$srcdir"/$_pkgreal-$_pkgver
+builddir="$srcdir"/$_pkgreal-$_pkgver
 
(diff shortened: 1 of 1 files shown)


andypost:
Looks build broken somehow
I used to install/compile this extension after patching pecl https://github.com/skilld-labs/docker-php/blob/master/php7/Dockerfile#L52

andypost:
PR updated with ` --with-php-config=php-config7` now it works
probably it makes sense to update syntax to new format `s/_builddir/builddir`

testing/php7-apcu/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: Valery Kartel <valery.kartel@gmail.com>
+# Maintainer: Carlo Landmeter <clandmeter@gmail.com>
+pkgname=php7-apcu
+_pkgreal=apcu
```
jirutka: Please name it rather as `_pkgname`, `_pkgreal` has very short Levenshtein distance from `_pkgrel`. ;)

andypost: That's strange suggestion because `_pkgreal` & `_pkgver` looks a standard all over php* packages

andypost:
maybe better to update all packages to use more readable names? Looks that out of scope if adding new package

testing/php7-apcu/APKBUILD
jirutka: Hm, you’re right, and not even php* packages, but also a perl* packages. Actually there are only slightly more packages that uses `_pkgname`.

I’ll discuss it on IRC later. For now, let’s keep `_pkgreal`.

testing/php7-apcu/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: Valery Kartel <valery.kartel@gmail.com>
+# Maintainer: Carlo Landmeter <clandmeter@gmail.com>
```
jirutka: Is @clandmeter okay with being maintainer of this package?

clandmeter: better to let @andypost claim maintainership (if he want to move it to community)

PS i don't think we have a standard for local based variables, until we do so it shouldn't matter what to use.

andypost:
I changed Maintainer to myself and removed previous contributor cos this package is new

jirutka:
Merged in e07019964b87b8d37c77bd529a3e3c1ad180ffd6, thanks!

andypost:
Yay! so what is next steps to see this in testing?

jirutka:
It should be in there in few minutes.

andypost:
Thanx a lot for guidance, will review other php7 stack needed for Drupal development

PR #117, opened 2016-06-07
main/py-sphinx: fix missing dependency on py-imagesize, upgrade to 1.4.3 and add new abuild main/py-imagesize

jirutka:
🔙 Backport requested in http://bugs.alpinelinux.org/issues/5688.

PR #120, opened 2016-06-07
main/llvm: fix segfault when unpacking aggregates structs

jirutka:
This patch is backported from upstream, see https://github.com/JuliaLang/julia/pull/16503 for more information.

🔙  Backport requested in https://bugs.alpinelinux.org/issues/5690.

⚠️ The build fails due to a missing dependency in py-sphinx, see #117.


jirutka:
Merged in 3ff479a2733ea94d72b79391a915f0ff96c579b9.

PR #122, opened 2016-06-09
testing/gearmand: new aport

leprechau:
Gearman job server and client.

http://gearman.org

### DIFF
--- a/testing/gearmand/APKBUILD
+++ b/testing/gearmand/APKBUILD
@@ -1,0 +1,85 @@
+# Contributor: Aaron Hurt <ahurt@ena.com>
+# Maintainer: Aaron Hurt <ahurt@ena.com>
+pkgname=gearmand
+pkgver="1.1.12"
+pkgrel=2
+pkgdesc="Gearman job server and client"
+url="http://gearman.org"
+arch="all"
+license="BSD"
+depends="libgearman"
+depends_dev="boost-dev gperf libevent-dev util-linux-dev mariadb-dev
+	hiredis-dev openssl-dev sqlite-dev libmemcached-dev"
+makedepends="$depends_dev"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages="$pkgname-doc libgearman libgearman-dev"
+source="https://launchpad.net/${pkgname}/1.2/${pkgver}/+download/${pkgname}-${pkgver}.tar.gz
+	libtest-cmdline.cc.patch
+	$pkgname.initd
+	$pkgname.confd"
+
+_builddir="$srcdir/$pkgname-$pkgver"
+prepare() {
+	local i
+	cd "$_builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
+}
+
+build() {
+	cd "$_builddir"
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--disable-libpq \
+		--disable-libtokyocabinet \
+		--disable-libdrizzle \
+		--enable-ssl \
+		--enable-hiredis \
+		--enable-jobserver=no \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$_builddir"
+	make DESTDIR="$pkgdir" install || return 1
+
+	install -m755 -D "$srcdir"/$pkgname.initd \
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -m644 -D "$srcdir"/$pkgname.confd \
+		"$pkgdir"/etc/conf.d/$pkgname || return 1
+}
+
+libgearman() {
+	pkgdesc="gearman C client library"
+
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/libgearman.so* "$subpkgdir"/usr/lib/
+}
+
+dev() {
+	default_dev
+	depends="libgearman"
+	pkgdesc="libgearman development files"
+}
+
+md5sums="99dd0be85b181eccf7fb1ca3c2a28a9f  gearmand-1.1.12.tar.gz
+3973b88414bdfc06ec5883bab9f837c3  libtest-cmdline.cc.patch
+b12060f98a25bfcb43b31eca7498d2b0  gearmand.initd
+9f79358567f822307f022001372a075c  gearmand.confd"
+sha256sums="973d7a3523141a84c7b757c6f243febbc89a3631e919b532c056c814d8738acb  gearmand-1.1.12.tar.gz
+9e640a145b445d765065e662d5301450a42c321913b3a44a26ecd148e0106350  libtest-cmdline.cc.patch
+24e06809832cad4785a7aacee3c232c7f77e6b178e5d3863219a7f54fcec1f0c  gearmand.initd
+dc84fd32db3310a1d8b9250e9d5a64cf13bbc67c29e8e1f30ed202b5109a629e  gearmand.confd"
+sha512sums="51d1596761c3f3ff3213edb64054fd9942b109383a51e998ff096ea972966dad1780a11814a3d37d4e8396365f402e3e3c9f759502dc3b6da25ea41936204e28  gearmand-1.1.12.tar.gz
+2e333ca902b3cedb7057cda5ab75fd40c3b86a6e3eb3249c0f860388db9ccc7f603d71c047e6c3d7b7750b00e71f25c5a7434751af577443fa393a5396aedcfb  libtest-cmdline.cc.patch
+7ac61fb4ffab2dbc20ccbbd9af9f1ff808f05623923b96368e1a84bcb32de258e3dcd800d4ef4e0e55029c0c9cc38c6857cbe999e7bda351d672622b5efb5071  gearmand.initd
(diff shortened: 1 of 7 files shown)


testing/gearmand/APKBUILD
```diff
@@ -0,0 +1,85 @@
...
+_builddir="$srcdir/$pkgname-$pkgver"
+prepare() {
+	local i
+	cd "$_builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
+}
```
jirutka: This is now default behaviour when `prepare()` function is not declared in APKBUILD, i.e. you can remove it entirely.

testing/gearmand/APKBUILD
```diff
@@ -0,0 +1,85 @@
...
+	hiredis-dev openssl-dev sqlite-dev libmemcached-dev"
+makedepends="$depends_dev"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages="$pkgname-doc libgearman libgearman-dev"
+source="https://launchpad.net/${pkgname}/1.2/${pkgver}/+download/${pkgname}-${pkgver}.tar.gz
+	libtest-cmdline.cc.patch
+	$pkgname.initd
+	$pkgname.confd"
+
+_builddir="$srcdir/$pkgname-$pkgver"
```
jirutka: It should be named `builddir`, this has been changed recently (you should upgrade your abuild package).

testing/gearmand/gearmand.initd
```diff
@@ -0,0 +1,29 @@
...
+	eend $?
+}
+
+stop() {
+	ebegin "Stopping ${name}"
+		start-stop-daemon --stop --quiet \
+			--pidfile /var/run/$name.pid \
+			--exec ${daemon}
+	eend $?
+}
```
jirutka: It’s unnecessary to implement start/stop functions. OpenRC is not stupid as SysV init and provides conventional variables for typical use cases. ;) You can write it witch much less boilerplate:

```sh
#!/sbin/runscript

name="gearmand"
description="Gearman server"

command="/usr/sbin/gearmand"
command_args="-d $gearmand_opts"
command_background="yes"

pidfile="/run/${RC_SVCNAME}.pid"
start_stop_daemon_args="--user=$gearmand_user:$gearmand_group"
```

testing/gearmand/gearmand.pre-install
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+## add gearmand group or use exsiting
+if ! getent group gearmand &>/dev/null; then
+	/usr/sbin/addgroup -S gearmand > /dev/null 2>&1
```
jirutka: It’s unnecessary to use absolute paths for adduser and addgroup.

testing/gearmand/gearmand.post-deinstall
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+## print notice
```
jirutka: Sorry, I couldn’t resist. :)

leprechau: It's just a habit. I tend to comment the thought before I write the code below.

testing/gearmand/gearmand.pre-install
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+## add gearmand group or use exsiting
+if ! getent group gearmand &>/dev/null; then
```
jirutka: Syntax `&>` is a bash-ism. Busybox ash supports it, so it’ll work on Alpine, but we prefer to avoid them for better compatibility, if there’s no a good reason to do the opposite.

Also there’s no need to check if the user/group exists, because adduser/addgroup already does this for you and you’re redirecting their stdout/stderr to /dev/null anyway.

So you can just write the following and it’ll behave exactly the same.

```#!/bin/sh

addgroup -S gearmand >/dev/null 2>&1
adduser -H -S -G gearmand -h /var/lib/gearmand -s /bin/false \
	-g "Gearman Server" gearmand >/dev/null 2>&1

exit 0
```

jirutka: BTW, there’s even example for this on Alpine wiki [here](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#.24pkgname.pre-install).

przemoc: To better reiterate @jirutka's point: Never use bashisms in scripts not using bash in shebang (`#!` line) and never use bash for system scripts, because systems, especially systems valueing small size, may not have full blown shell interpreters like bash or zsh out-of-the-box.

You can find boilerplate in wiki:
https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#.24pkgname.pre-install

It would be nice if you could preserve the argument order too, as it helps in mass processing/maintaining them.

(In future AL wants to go away from adduser/addgroup, but we have to live with it for now.)

testing/gearmand/gearmand.post-deinstall
leprechau: You're not the only one.  Buddy of mine in the office keeps threatening to write a de-commenter for me.

jirutka: Don’t worry, I’ll strip these comments before merging. :)

testing/gearmand/APKBUILD
```diff
@@ -0,0 +1,75 @@
...
+pkgdesc="Gearman job server and client"
+url="http://gearman.org"
+arch="all"
+license="BSD"
+depends="libgearman"
+depends_dev="boost-dev gperf libevent-dev util-linux-dev mariadb-dev
+	hiredis-dev openssl-dev sqlite-dev libmemcached-dev"
+makedepends="$depends_dev"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages="$pkgname-doc libgearman libgearman-dev"
```
jirutka: I’ve discussed this PR on IRC and we agreed that it’ll be better to name these subpackages `gearmand-libs` (*.so libs) and `gearmand-dev` (header files etc.) It’s more common naming schema in Alpine. That said, it may look odd in this case, because of trailing “d” in gearmand that usually means “daemon.” Naming it “gearman” would make more sense, but Gearman’s developers decided to name it “gearmand” for whatever reason and we prefer to name packages the same as the packaged software is officially named. WDYT?

testing/gearmand/APKBUILD
```diff
@@ -0,0 +1,75 @@
...
+build() {
+	cd "$builddir"
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--disable-libpq \
```
jirutka: This disables PostgreSQL support? Why do you prefer MySQL to PostgreSQL for gearmand?

testing/gearmand/APKBUILD
leprechau: The main purpose is the daemon ```gearmand``` but the libs are needed for the SDKs for PHP and other languages.  The website itself is *gearman.org* and the title is just *Gearman* so I think ```gearman-libs```, ```gearman-dev``` and a binary ```gearmand`` provided by the main package sound reasonable.

testing/gearmand/APKBUILD
leprechau: Persistence in general was a late addition to Gearman and when enabled it's most commonly used with SQLite or MySQL in practice.  However I don't know of any reason why a person couldn't use PostreSQL if they wanted.  I'll update the build.

testing/gearmand/APKBUILD
leprechau: For what it's worth Fedora/Redhat have the following packages for ```gearmand```

```gearmand```
```libgearman```
```libgearman-devel```

testing/gearmand/APKBUILD
jirutka: How often is the persistence used in practice? If it’s rare, then it’d be maybe better to disable it, to minimize number of dependencies and size of the package.

leprechau: We don't use persistence in our deployment and it's usually not for short running jobs.  Persistence really only makes sense for long running jobs and then you might be better off using RabbitMQ or something similar.

leprechau:
@jirutka Do you need me to do anything else with this build?  I think i've covered all the points raised but didn't want to leave it sitting if there were other issues I still needed to resolve.

testing/gearmand/APKBUILD
jirutka: Then I think that it would be better also disable MySQL and keep just SQLite that is tiny.

How does it use Redis and Memcached? Are both of them used for different purpose or it’s one, or another?

leprechau: They are all used for persistence of queue data.

jirutka: As an alternative to SQLite, PostgreSQL, MySQL, …, or for different data?

leprechau: As alternatives.  You wouldn't probably persist data with more than one driver at any one time.

jirutka: Okay, these deps (hiredis and libmemcached) are tiny and just client libs.

However, I’d really prefer disabling MySQL, what do you think?

leprechau: Works for me. I'll change it when I get back to a desktop machine.

jirutka:
I’ve fixed some minor issues that I’ve overlooked before and merged it in 1d12b9eb04b0c3f8c63711843ad7ff351af79087.

Thanks for your pull request and sorry that it took so long, I’ve been quite busy in previous days.

leprechau:
Not a problem, thank you for all the input.  It's really helpful getting up and going producing packages that meet the expectations of the community.

PR #123, opened 2016-06-09
testing/glide: new aport

leprechau:
Package Management for Go

https://glide.sh

### DIFF
--- a/testing/glide/APKBUILD
+++ b/testing/glide/APKBUILD
@@ -1,0 +1,28 @@
+# Contributor: Aaron Hurt <ahurt@ena.com>
+# Maintainer: Aaron Hurt <ahurt@ena.com>
+pkgname="glide"
+pkgver="0.10.2"
+pkgrel=0
+pkgdesc="Vendor Package Management for Golang"
+url="https://github.com/Masterminds/glide"
+arch="x86_64"
+license="Masterminds"
+depends=""
+depends_dev=""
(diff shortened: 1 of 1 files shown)


Barthalion:
Uh, just no. I know how ungrateful and tedious job is packaging Go software, but we won't put pre-baked binary as an official package in aports.

leprechau:
Fair enough. I can build it.

-- Aaron

leprechau:
@Barthalion does this work?

testing/glide/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+url="https://github.com/Masterminds/glide"
+arch="all"
+license="MIT"
+depends=""
+makedepends="go"
+install=""
+subpackages=""
+source="$pkgname-$pkgver.tar.gz::https://github.com/Masterminds/glide/archive/$pkgver.tar.gz"
+builddir="$srcdir/go/src/github.com/Masterminds/glide"
+options="!strip"
```
Barthalion: Is there any reason not to strip the binary? I always thought it is possible without causing issues.

leprechau: It's not supported or recommended to strip go binaries with the regular ```strip``` implementation.  You can pass some ```ldflags``` to the build though that have a similar effect.

Barthalion:
Merged in 0dbfd788.

PR #124, opened 2016-06-09
testing/dumb-init: new aport

leprechau:
A minimal init system for Linux containers

https://github.com/Yelp/dumb-init


jirutka:
Travis build failed, because dump-init uses bash in Makefile (for whatever reason), so you have to add `bash` to `makedepends`.

leprechau:
Gotcha, thanks

testing/dumb-init/APKBUILD
```diff
@@ -0,0 +1,30 @@
+# Contributor: ena r&d <r&d@ena.com>
```
jirutka: Is this really a valid mail address?

leprechau: It is valid but it's a group list.  I can update those to point to me directly.

jirutka: That would be better.

testing/dumb-init/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+
+builddir="$srcdir/$pkgname-$pkgver"
+build() {
+	cd "$builddir"
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	install -cpsD -o root -g root -m 655 dumb-init "$pkgdir"/usr/bin/dumb-init || return 1
```
jirutka: Are you sure about 655? Shouldn’t it be 755?

leprechau: Yes, it should be 755

jirutka:
Merged in 12ecd81bbe54bbd873c73c6bb81c41b124cadca5, thanks!

PR #125, opened 2016-06-10
testing/pacparser: new aport

leprechau:
A library to parse proxy auto-config (PAC) files

http://pacparser.manugarg.com


testing/pacparser/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+}
+
+build() {
+	cd "$builddir"
+	make -C src pactester > /dev/null 2>&1 || return 1
+}
+
+package() {
+	cd "$builddir"
+	make -C src install DESTDIR=$pkgdir LIB_PREFIX=$pkgdir/usr/lib > /dev/null 2>&1 || return 1
```
jirutka: Why you’re redirecting stdout/stderr of make to `/dev/null`?

leprechau: No real reason, it's just quite a bit of output building all the components.

jirutka: It's always better to have a lot of output than none. How would you know what's wrong if the package fail to build, when you throw away even error messages?

leprechau: I added it after the build was working.  I was looking at it more as debug output.  I took it off and Travis has rebuilt the package.

testing/pacparser/spidermonkey-make.patch
```diff
@@ -0,0 +1,17 @@
+--- src/spidermonkey/Makefile.orig
```
jirutka: Please add comment to the patch file explaining why is this patch needed.

testing/pacparser/APKBUILD
```diff
@@ -11,12 +11,14 @@ depends=""
...
 subpackages="$pkgname-doc $pkgname-dev"
-source="$pkgname-$pkgver.tar.gz::https://github.com/pacparser/pacparser/archive/$pkgver.tar.gz"
+source="$pkgname-$pkgver.tar.gz::https://github.com/pacparser/pacparser/archive/$pkgver.tar.gz
+spidermonkey-make.patch"
 builddir="$srcdir/$pkgname-$pkgver"
 
 prepare() {
 	cd "$builddir"
 	find . -name \*.git\* -delete
+	patch -p0 -i "$srcdir/spidermonkey-make.patch"
```
jirutka: Abuild knows how to apply patches, just add `default_prepare` to your `prepare()` function and it will automatically apply all *.patch files specified in the `source`.

leprechau: Ahh, thank you.  I was wondering what the best way was to have your own prepare and still get the benefits of the default.

jirutka:
Merged in 4e051e1e7a83787c77db255b7cb24b017419f857. I failed to compile because of https://github.com/pacparser/pacparser/issues/27, so I’ve forced `-j1` to disable parallel build. I’ve also added subpackage `py-pacparser` in f7cec01da5d9d16b5604c3ddf76e10f56cd50bf6.

leprechau:
Great, thank you sir.

jirutka:
It fails not just on armhf, but also x86, so it’s currently built only for x86_64. I’ve reported it to upstream: https://github.com/pacparser/pacparser/issues/77.

PR #128, opened 2016-06-12
testing/creduce: new abuild

jirutka:
/cc @tkelman


tkelman:
Good stuff! I think llvm-libs is needed for some of the passes - does creduce have tests that you're running, or would it help if I posted the real-world preprocessed source and run script that I used yesterday? Or does apkbuild automatically detect linked dependencies?

jirutka:
I’ve tried the [Getting started example](http://embed.cs.utah.edu/creduce/using/) and it seems that it works fine. Abuild automatically detects linked dependencies:

```
so:libLLVM-3.8.so
so:libc.musl-x86_64.so.1
so:libgcc_s.so.1
so:libstdc++.so.6
```

jirutka:
Merged in 1192a785dbaa88255d59f4ea9ff21c63781d2fbc, fb922ad0b0559fad3d45bbb801e27f1e66fbdd29, 7feadd908280effbe37152d7f5ecd7a0b59d030b, c95e18c97c441b3cb5a5d2c4a7a4ef207587d425.

PR #129, opened 2016-06-15
testing/rabbitmq-c: update to version 0.8.0

nathanejohnson:
Also splitting package into -dev, -doc and -bin packages.

### DIFF
--- a/testing/rabbitmq-c/APKBUILD
+++ b/testing/rabbitmq-c/APKBUILD
@@ -1,27 +1,46 @@
 # Contributor: Fabio Ribeiro <fabiorphp@gmail.com>
 # Maintainer: Fabio Ribeiro <fabiorphp@gmail.com>
 pkgname=rabbitmq-c
-pkgver=0.7.1
+pkgver=0.8.0
 pkgrel=0
 pkgdesc="RabbitMQ C client"
 url="https://github.com/alanxz/rabbitmq-c"
+subpackages="$pkgname-bin $pkgname-doc $pkgname-dev"
+depends_dev="openssl-dev popt-dev"
+makedepends="$depends_dev cmake xmlto doxygen"
 arch="all"
 license="MIT"
-source="https://github.com/alanxz/$pkgname/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz"
+source="$pkgname-$pkgver.tar.gz::https://github.com/alanxz/$pkgname/archive/v$pkgver.tar.gz"
 
 builddir="$srcdir"/$pkgname-$pkgver
 
 build() {
 	cd "$builddir"
-	./configure --prefix=/usr || return 1
+	mkdir build && cd build || return 1
+	cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib .. || return 1
 	make || return 1
 }
 
 package() {
-	cd "$builddir"
+	cd "$builddir"/build
 	make DESTDIR="$pkgdir" install || return 1
 }
 
-md5sums="6216c8876299a5efc4ff5ff84dc636d8  rabbitmq-c-0.7.1.tar.gz"
-sha256sums="23df349a7d157543e756acc67e47b217843ecbdafaefe3e4974073bb99d8a26d  rabbitmq-c-0.7.1.tar.gz"
-sha512sums="197b8476373e3a9056d5f24e0d932a37a556a946d2b6471c95e8b0e693b611de51882be4911bd099f2c2e76153225d83cac1bec8106c031058315f062a8a4ab4  rabbitmq-c-0.7.1.tar.gz"
+bin() {
+	pkgdesc="command line utilities for rabbitmq"
+	depends="$pkgname"
+	mkdir -p "$subpkgdir"/usr/bin
+	mv "$pkgdir"/usr/bin "$subpkgdir"/usr/
+}
+
+doc() {
+	pkgdesc="rabbitmq-c documentation and examples"
+	arch="noarch"
+	mkdir -p "$subpkgdir"/usr/share/"$pkgname"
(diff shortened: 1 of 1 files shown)


jirutka:
Please examine the [build log](https://travis-ci.org/alpinelinux/aports/builds/137696175) and read readme of the rabbitmq. They use cmake, but also maintain legacy make/autohells build system. You’re somehow trying to use both, that doesn’t make sense.

nathanejohnson:
I have cleaned up the messiness in the build logs, however I'm not sure what you mean by mixing / matching cmake and autoconf .  I also bumped the pkgrel but I'm not sure if that's required in the github workflow.

testing/rabbitmq-c/APKBUILD
```diff
@@ -1,27 +1,50 @@
...
 license="MIT"
-source="https://github.com/alanxz/$pkgname/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz"
+source="$pkgname-$pkgver.tar.gz::https://github.com/alanxz/$pkgname/archive/v$pkgver.tar.gz"
 
 builddir="$srcdir"/$pkgname-$pkgver
 
 build() {
 	cd "$builddir"
-	./configure --prefix=/usr || return 1
+	rm -rf build && mkdir build && cd build
```
jirutka: There’s no `build` directory in the tarball; why you’re removing directory that doesn’t exist?

testing/rabbitmq-c/APKBUILD
```diff
@@ -1,27 +1,50 @@
...
+	cmake -DCMAKE_INSTALL_PREFIX=/usr ..
 	make || return 1
```
jirutka: You should execute `cmake`, or `make`, not both of them. rabbitmq-c prefers cmake.

testing/rabbitmq-c/APKBUILD
```diff
@@ -1,27 +1,50 @@
...
 }
 
 package() {
-	cd "$builddir"
+	cd "$builddir"/build
 	make DESTDIR="$pkgdir" install || return 1
+	cd "$pkgdir"/usr
+	if [ -d ./lib64 ]; then
+		mv lib64 lib
```
jirutka: Why this?

testing/rabbitmq-c/APKBUILD
```diff
@@ -1,27 +1,50 @@
...
-	./configure --prefix=/usr || return 1
+	rm -rf build && mkdir build && cd build
+	cmake -DCMAKE_INSTALL_PREFIX=/usr ..
 	make || return 1
 }
 
 package() {
-	cd "$builddir"
+	cd "$builddir"/build
 	make DESTDIR="$pkgdir" install || return 1
```
jirutka: Again, do you wanna use cmake build system, or legacy autotools/make build system? It doesn’t make sense to use both at once. Please read https://github.com/alanxz/rabbitmq-c#building-and-installing.

nathanejohnson:
The use of cmake and make are not mutually exclusive and, in fact, go hand in hand.  cmake generates Makefiles.  Automake also generates Makefiles, however I am not using automake with this APKBUILD.  I do, however, use make in conjunction with cmake.  As do most of the aports/main APKBUILD files that use cmake, for what it's worth.

If there is some rule about not using make with cmake, perhaps you could help me with a hint as to how to have the installation go into a fakeroot, as with make DESTDIR ?

testing/rabbitmq-c/APKBUILD
nathanejohnson: This installs to lib64 on x86_64 builds, which doesn't appear to be an appropriate location on alpine.

testing/rabbitmq-c/APKBUILD
nathanejohnson: See comments below in the thread

testing/rabbitmq-c/APKBUILD
nathanejohnson: If you run the build successively, it needs to be removed.  The -f flag ensures it still returns exit status of 0 and goes to the next step.  This is more for me while I was getting the build right.

testing/rabbitmq-c/APKBUILD
nathanejohnson: Again, see below.

testing/rabbitmq-c/APKBUILD
jirutka: Yeah, that’s true, but this is very bad approach to solve it. cmake accepts variable `CMAKE_INSTALL_LIBDIR` that allows to define name or path of the lib directory (example: `cmake -DCMAKE_INSTALL_LIBDIR=lib`).

jirutka:
I owe you an apology. You’re right about cmake vs. make, it’s used in many abuild in the same way. It seems that I kinda didn’t understand what exactly cmake is and rabbitmq-c’s readme has supported me in my wrong idea that these are mutually exclusive.

nathanejohnson:
No worries!  So at this point, I can ditch that rm -rf on the build dir, outside of that do you have any remaining concerns?

nathanejohnson:
One last thing before I rebase / push, do I need to bump pkgrel or keep it at 0?

nathanejohnson:
I added the -DCMAKE_INSTALL_LIBDIR , removed the rm -rf , and reset pkgrel to 0 (I assume that this only needs to be bumped after merge).  Hopefully this addresses your concerns.

jirutka:
Thanks. The last concern/question: I think that `rabbitmq-c-utils` would be better than `rabbitmq-c-bin`. What do you think? (I can change it before merging, not needed to update the PR.)

nathanejohnson:
That works for me.

jirutka:
Merged in 6af1dbb56d77f977fb0a3988940eaa628e068651, thanks!

PR #131, opened 2016-06-18
testing/php5-xdebug: rename abuild and upgrade to 2.4.0

andypost:
This package is broken so let's upgrade it and rename inline with other php ones

PS Related fix https://github.com/alpinelinux/aports/pull/130


jirutka:
The commit message is not right, this is not a new package, but renamed…

jirutka:
Merged in eb116ed3eeaa9c703f1d3c73a8f4a78add925747, thanks!

andypost:
Thanks a lot for quick response!

PR #132, opened 2016-06-18
testing/php5-redis: rename abuild and upgrade to 2.2.8

andypost:
The package is broken http://bugs.alpinelinux.org/issues/5593
Also renamed to be inline with other php5 ones


jirutka:
Merged in 4516b36babb17662e9d904d1cc8de80ee8401f93, thanks!

PR #133, opened 2016-06-18
testing/php7-redis: upgrade to 3.0.0

andypost:
Upgrade is needed because php7 is supported starting from 3.0.0
http://pecl.php.net/package/redis/3.0.0


jirutka:
Merged in 16bf61df0f34abbe376dfd0fd6b53719310964a9, thanks!

PR #134, opened 2016-06-18
testing/php5-memcached: rename abuild

andypost:
Package is broken because .ini and .so files should be in php5
Renamed to be inline with other php5 extensions


jirutka:
Merged in 5f74b23ad5816d7e720f2b7b8d8c62d9eef14615, thanks!

PR #135, opened 2016-06-18
testing/php5-ssh2: rename abuild and upgrade to 0.13

andypost:
Package is broken so fixed and renamed inline with  other php5 ones


jirutka:
Merged in d2ce891ad820709784d0822d578fdcdb6bf03d4e, thanks!

PR #137, opened 2016-06-20
testing/rabbitmq-server: new aport

nathanejohnson:
https://www.rabbitmq.com
RabbitMQ is an open source multi-protocol messaging broker.

### DIFF
--- a/testing/rabbitmq-server/rabbitmq-server.initd
+++ b/testing/rabbitmq-server/rabbitmq-server.initd
@@ -1,0 +1,42 @@
+#!/sbin/openrc-run
+
+pidfile=/run/rabbitmq/rabbitmq.pid
+name=rabbitmq
+daemon=/usr/sbin/rabbitmq-server
+
+startup_log=/var/log/rabbitmq/startup_log
+startup_err=/var/log/rabbitmq/startup_err
+shutdown_log=/var/log/rabbitmq/shutdown_log
+shutdown_err=/var/log/rabbitmq/shutdown_err
+
+depend() {
+	need net localmount
+	after firewall
+}
+
+start() {
+	ebegin "Starting ${name}"
+	# We need to test if /var/run/rabbitmq exists, since scirpt will fail if it does not
+	checkpath -m 700 -o rabbitmq:rabbitmq -d "$(dirname "$pidfile")"
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $startup_log
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $startup_err
+	start-stop-daemon --start \
+		--env RABBITMQ_PID_FILE=$pidfile \
+		--pidfile $pidfile \
+		--stdout $startup_log \
+		--stderr $startup_err \
+		--background $daemon        
+	eend $?
+}
+
+stop() {
+	ebegin "Stopping ${name}"
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $shutdown_log
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $shutdown_err
+	start-stop-daemon --stop --quiet --pidfile $pidfile \
+		--exec /bin/sh -- -c "/usr/sbin/rabbitmqctl \
+		--stop $pidfile > $shutdown_log \
+		2> $shutdown_err"
+	eend $?
+}
+
--- a/testing/rabbitmq-server/rabbitmq-server.pre-install
+++ b/testing/rabbitmq-server/rabbitmq-server.pre-install
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+addgroup -S rabbitmq 2>/dev/null
+adduser -S -D -H -s /bin/false -h /usr/lib/rabbitmq -G rabbitmq \
+		-g 'RabbitMQ Server' rabbitmq 2>/dev/null
+exit 0
--- a/testing/rabbitmq-server/APKBUILD
+++ b/testing/rabbitmq-server/APKBUILD
@@ -1,0 +1,89 @@
+# Contributor: Nathan Johnson <nathan@nathanjohnson.info>
+# Maintainer: Nathan Johnson <nathan@nathanjohnson.info>
+pkgname=rabbitmq-server
+pkgver=3.6.2
+_realver=${pkgver//\./_}
+pkgrel=0
+pkgdesc="RabbitMQ is an open source multi-protocol messaging broker."
+url="http://www.rabbitmq.com/"
+arch="noarch"
+license="MPL 1.1"
+depends="erlang erlang-tools erlang-runtime-tools erlang-stdlib
+	logrotate erlang-ssl erlang-crypto erlang-parsetools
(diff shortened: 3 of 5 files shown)


jirutka:
Merged in , thanks!

_(TODO)_

nathanejohnson:
So just curious, what does "needs-review" tag mean?

jirutka:
It’s basically an information for myself that I should do a code-review ASAP.

testing/rabbitmq/APKBUILD
```diff
@@ -0,0 +1,91 @@
+# Contributor: Nathan Johnson <nathan@nathanjohnson.info>
+# Maintainer: Nathan Johnson <nathan@nathanjohnson.info>
+pkgname=rabbitmq
+_realname=rabbitmq-server
```
jirutka: Why you’ve named it `rabbitmq` instead of `rabbitmq-server`, that seems to be the official name?

testing/rabbitmq/rabbitmq.initd
```diff
@@ -0,0 +1,38 @@
+#!/sbin/openrc-run
+
+pidfile=/var/run/rabbitmq/rabbitmq.pid
```
jirutka: Alpine, as many others, stores pidfiles in the directory `/run` that is mounted as tmpfs. `/var/run` still exists, but it’s just a symlink to `/run` for backward compatibility.

testing/rabbitmq/rabbitmq.initd
```diff
@@ -0,0 +1,38 @@
...
+	need net localmount
+	after firewall
+}
+
+start() {
+	ebegin "Starting ${name}"
+	# We need to test if /var/run/rabbitmq exists, since scirpt will fail if it does not
+	[ ! -e /var/run/rabbitmq ] && \
+		mkdir /var/run/rabbitmq && \
+		chown rabbitmq:rabbitmq /var/run/rabbitmq
```
jirutka: OpenRC provides a handy helper function for this, `checkpath` (see [man openrc-run](http://manpages.org/openrc-run/8)).

```diff
-	# We need to test if /var/run/rabbitmq exists, since scirpt will fail if it does not
-	[ ! -e /var/run/rabbitmq ] && \
-		mkdir /var/run/rabbitmq && \
-		chown rabbitmq:rabbitmq /var/run/rabbitmq
+	checkpath -m 700 -o rabbitmq:rabbitmq -d "$(dirname "$pidfile")"
```

testing/rabbitmq/rabbitmq.initd
```diff
@@ -0,0 +1,38 @@
...
+
+    
+    
+	start-stop-daemon --start \
+		--env RABBITMQ_PID_FILE=$pidfile \
+		--pidfile $pidfile \
+		--exec /bin/sh -- -c "$daemon \
+		> /var/log/rabbitmq/startup_log \
+		2> /var/log/rabbitmq/startup_err 0<&- &"
```
jirutka: I’m quite sure (but not 100%) that output redirection doesn’t work here. Moreover, `start-stop-daemon` provides options for them: `--stdout` and `--stderr`.
Using `&` here to put the process into background is absolutely wrong, it defeats the purpose of start-stop-daemon.

testing/rabbitmq/rabbitmq.initd
nathanejohnson: If this is a rule, there are a lot of APKBUILD files that violate this rule.  It is a trivial change, but it's also a trivial complaint.

testing/rabbitmq/rabbitmq.initd
nathanejohnson: Thanks for the tip.

testing/rabbitmq/rabbitmq.initd
nathanejohnson: This is as close to the upstream systemd script as possible, that was my initial intent.  It's not my favorite init script, but it is functional.

testing/rabbitmq/rabbitmq.initd
jirutka: Yes, that's legacy. It's not a big deal, but it's trivial to fix.

testing/rabbitmq/APKBUILD
nathanejohnson: I looked for other packages named -server and didn't find any.  I happen to like rabbitmq-server name better than rabbitmq personally

testing/rabbitmq/rabbitmq.initd
jirutka: This is not systemd, but OpenRC that is very different. Using `&` here really doesn't make any sense and it's not acceptable.

However, as I looked at RabbitMQ, it'll be a bit more complicated. I'll rewrite this init script myself, so don't make a worry about it.

testing/rabbitmq/APKBUILD
jirutka: W don't have any rule for using or not using `-server` suffix (as I know), but we have a rule to stick with the upstream's name where possible.

testing/rabbitmq/rabbitmq.initd
nathanejohnson: Their shell wrapper doesn't detach from foreground, so I think that's why their init script does the & business.  I do see that start-stop-daemon has a -b option that might make more sense than the , and your suggestion of --stdout and --stderr is a good one.

nathanejohnson:
I made some updates, including:

1. renamed to rabbitmq-server
2. updated init script so that --background, --stdout and --stderr is used with start-stop-daemon versus & and redirection for the start and to use checkpath to ensure path exists, has appropriate ownership etc
3. changed /var/run paths to /run

testing/rabbitmq-server/rabbitmq-server.initd
```diff
@@ -0,0 +1,42 @@
...
+	# We need to test if /var/run/rabbitmq exists, since scirpt will fail if it does not
+	checkpath -m 700 -o rabbitmq:rabbitmq -d "$(dirname "$pidfile")"
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $startup_log
+	checkpath -m 644 -o rabbitmq:rabbitmq --file-truncate -f $startup_err
+	start-stop-daemon --start \
+		--env RABBITMQ_PID_FILE=$pidfile \
+		--pidfile $pidfile \
+		--stdout $startup_log \
+		--stderr $startup_err \
+		--background $daemon
```
ncopa: nitpick: there are whitespace before end-of-line above.

testing/rabbitmq-server/rabbitmq-server.pre-install
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+addgroup -S rabbitmq 2>/dev/null
+adduser -S -D -H -s /bin/false -h /usr/lib/rabbitmq -G rabbitmq \
```
ncopa: I think we should use `/sbin/nologin` as shell instead of  `/bin/false`

jirutka:
@ncopa There are more issues, especially with the init script; I've been working on it, but haven't time to finish it yet. That's why I didn't merge it yet. Well, I'll fix it later.

PR #144, opened 2016-06-26
main/qemu@3.4: fix chown/chmod for qemu-bridge-helper and gid of qemu group

jirutka:
Backport request http://bugs.alpinelinux.org/issues/5804


jirutka:
Merged in 114825bece91175e9c8bcae3c93cde954d3ce984.

PR #147, opened 2016-07-06
New aports for pgcli and mycli including deps

tboerger:


jirutka:
These strange hashes in URLs from PyPI are calculated from the file being downloaded, i.e. it changes with every version. This makes updating abuilds more complicated than usual.

Please replace links like:
```
https://pypi.python.org/packages/c3/24/91f4b4c6bce71c92187aa3919ef42028f4af3b4f1485769709d37237384a/${pkgname}-${pkgver}.tar.gz
```

with:

```
$pkgname-$pkgver.tar.gz::https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz
```

The leading part `$pkgname-$pkgver.tar.gz::` is not necessary, but it’s better to avoid possible name clashes with other packages.

I’ll update the newapkbuild script soon.

For more information, see [pypa/pypi#438](https://bitbucket.org/pypa/pypi/issues/438/backwards-compatible-un-hashed-package).

tboerger:
@jirutka thanks, that was a missing piece. The fixed URLs in older packages didn't work anymore. I will update it now.

PR #151, opened 2016-07-06
testing/gogs: New version 0.9.13, that builds (on GCC 5.3.0)

7heo:
I know that this version isn't the best we can have; especially concerning dependencies (taken from master of relevant repositories for all I know); but I'll fix that in a following commit. This is to have the latest gogs version.


7heo:
The error seems to be related to [TEXTRELs](https://wiki.gentoo.org/wiki/Hardened/Textrels_Guide). Maybe it's related to GCC afterall. Good thing we have travis testing.

7heo:
Builds if sqlite is disabled...

clandmeter:
please rebase and keep the snapshot or provide a better solution to keep reproducible builds.

textrels is caused by go and trying to set no-pie. seems related to https://github.com/golang/go/issues/6940

7heo:
@clandmeter I will rebase and provide a better solution to keep reproducible builds; but not keep the snapshot.

As discussed on IRC, it seems like adding `-buildmode=pie` to `go build` solves the issue, because by default, `go build` tries to add `-r` and `-pie` together when invoking `ld`.

jirutka:
What is the state of this PR? Can it be safely merged?

Barthalion:
We already have 0.9.13 in testing.

7heo:
@jirutka: As @Barthalion stated, 0.9.13 is already in testing, due to @clandmeter upgrading it when fixing the GCC6/Go mess roughly two weeks ago. I needed this PR for later, since this PR was for quality improvements as well as (at the time) a new version. I'll just re-open one.

PR #152, opened 2016-07-11
main/perl: upgrade to 5.24.0

liqa:


jirutka:
Please change commit message to: `main/perl: upgrade to 5.24.0` (hint: commit --amend && git push -f).

jirutka:
⚠️ This upgrade will require to rebuild all perl packages that are not noarch! ⚠️

liqa:
is that message for me or something @ncopa and other devs will do when merging the patch?
because I'm not really sure how to do it…

jirutka:
@liqa This message is just a reminder for us, you can ignore it.

liqa:
ah okay thanks
good luck, I wish we can have updated perl soon
lots of work

apk search -q perl |grep -v "doc" |grep -v "dev" |wc -l
601

but minus the noarch ones, I don't know how to check which ones are

liqa:
what if you just bump all pkgrel at once?
apk search -q perl |grep -v "doc" |grep -v "dev"| xargs apkgrel -a

the problem is that it bumps 'noarch' packages too this way

PR #153, opened 2016-07-11
upgrade mpv to 0.18.1

liqa:


jirutka:
When you upgrade a package, please write commit log in format: `${branch}/${pkgname}: upgrade to ${pkgver}` (e.g. community/mpv: upgrade to 0.18.1) and reset `pkgrel` to 0.

liqa:
okay, I'll do it next times.
question: why does alpine use md5sum**s**, sha256sum**s** and sha512sum**s**. it's just 1 sum each.
shouldn't it be md5sum, sha256sum and sha512sum?

jirutka:
Single checksum is just a special case, see e.g. [julia/APKBUILD](https://github.com/alpinelinux/aports/blob/faac41e938404b68cfe48405cfb2a715228047ac/testing/julia/APKBUILD#L153-L188), there are checksums for ~12 files.

liqa:
aha, thanks for explanation

PR #157, opened 2016-07-12
main/nagios-plugins: split plugins into subpackages and upgrade to 2.1.1

jirutka:


main/nagios-plugins/nagios-plugins.post-upgrade
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+cat <<EOF
+NOTE: Since 2.0.3-r2 all plugins are separated into subpackages
```
Barthalion: This shouldn't be printed unconditionally, but I don't see a way to compare old and new versions. I'll merge it as is.

jirutka: Agree, but as you wrote, don't know how to print it conditionally based on previous version.

PR #170, opened 2016-07-22
main/hunspell: build with ncurses and readline

liqa:
So that it can be used as a stand-alone command line spell checker

### DIFF
--- a/main/hunspell/APKBUILD
+++ b/main/hunspell/APKBUILD
@@ -7,8 +7,8 @@ url="http://hunspell.github.io/"
 arch="all"
 license="GPL2+ LGPL2+ MPL 1.1"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-depends="ncurses readline"
-makedepends="gettext-dev"
+depends=""
+makedepends="gettext-dev ncurses-dev readline-dev"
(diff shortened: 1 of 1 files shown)


main/hunspell/APKBUILD
```diff
@@ -1,13 +1,13 @@
...
 pkgver=1.3.4
-pkgrel=0
+pkgrel=1
 pkgdesc="Spell checker and morphological analyzer library and program"
 url="http://hunspell.github.io/"
 arch="all"
 license="GPL2+ LGPL2+ MPL 1.1"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-depends=""
+depends="ncurses readline"
```
jirutka: This doesn’t work. You want to compile hunspell against ncurses and readline, i.e. you need their header files. These are included in *-dev packages; ncurses-dev and readline-dev in this case. The *-dev packages are needed only when making the package, so they should be declared in `makedepends`. Since abuild detects dynamically linked dependencies automatically, you don’t need to specify ncurses and readline in `depends`.

```diff
- makedepends="gettext-dev"
+ makedepends="gettext-dev ncurses-dev readline-dev"
```

clandmeter:
@liqa can you please squash these commits into a single one?

PR #179, opened 2016-07-27
testing/glide: New version 0.11.1

7heo:


7heo:
As explained [here](https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies):
> Get downloads and installs the packages named by the import paths, along with their dependencies.
> The -d flag instructs get to stop after downloading the packages; that is, it instructs get not to install the packages.

The `-v` flag is undocumented in this go applet, but is documented to increase the verbosity in other applets; and also does so with the `go get` applet.

Glide is a superior way to vendor the dependencies, but we cannot depend on glide to build glide; so I would say that this APKBUILD is the only place where it is acceptable to have `go get -d`.

testing/glide/APKBUILD
```diff
@@ -23,6 +23,7 @@ prepare() {
 build() {
 	export GOPATH="$srcdir/go"
 	cd "$builddir"
+	go get -v -d
```
7heo: As explained [here](https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies):
> Get downloads and installs the packages named by the import paths, along with their dependencies.
> The -d flag instructs get to stop after downloading the packages; that is, it instructs get not to install the packages.

The `-v` flag is undocumented in this go applet, but is documented to increase the verbosity in other applets; and also does so with the `go get` applet.

Glide is a superior way to vendor the dependencies, but we cannot depend on glide to build glide; so I would say that this APKBUILD is the only place where it is acceptable to have `go get -d`.

jirutka: How did this abuild worked before, without `go get`?

7heo: With a newer version of `go` than I was using. Sorry.

PR #180, opened 2016-07-27
FIX: Improve travis with pattern rejection

7heo:
Adds pattern rejection to travis, to ensure that the APKBUILD files do not contain known harmful constructs and expressions.


Barthalion:
This is out of Travis scope. There are some sanity checks in abuild; it should be implemented there instead, so a packager knows she's doing questionable things even if the change wasn't contributed via GitHub.

Barthalion:
Probably checkapk could also be a good place for linting APKBUILDs; it is already used to look for ABI changes after the build, so it would need some re-organization.

7heo:
@Barthalion Makes sense. I'll look into that then.

PR #182, opened 2016-07-28
Syncthing v0.14.2

zburgermeiszter:


jirutka:
@7heo please rewrite this package to use glide for resolving dependencies.

7heo:
@jirutka Will do asap.

7heo:
@jirutka it's much harder than "expected", because they rely on `go generate` in their build process (that tool is intended for development only, not during build time).

jirutka:
Well, @clandmeter already updated syncthing to 0.14.3 (0a9b332b7e1ee8855c47688142e374b637bbbcef), independently of this PR. But thanks for your contribution anyway!

PR #183, opened 2016-07-28
main/apache2: fix missing /run directory

7heo:
When setting up apache2 on a minimal environement (say, a docker image),
/run/apache2 doesn't exist and won't be created. So httpd will silently
fail. In such a case the error will only be visible in the logs (i.e.
nothing in stderr).

### DIFF
--- a/main/apache2/APKBUILD
+++ b/main/apache2/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=apache2
 _pkgreal=httpd
 pkgver=2.4.23
-pkgrel=2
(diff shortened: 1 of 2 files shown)


7heo:
Actually, this is only true in the case when the init system is missing (such as the docker image). Otherwise, `/run/apache2` is created by the init script, with `checkpath --directory $(dirname $PIDFILE)` in `/etc/init.d/apache2`.

7heo:
WONTFIX

PR #184, opened 2016-07-28
main/dnsmasq: improve runscript and abuild

jirutka:


main/dnsmasq/APKBUILD
```diff
@@ -1,50 +1,58 @@
...
-	cd "$srcdir/$pkgname-$pkgver"
+	cd "$builddir"
+
 	sed -i -e 's/^AWK.*/AWK = awk/' Makefile
 	make PREFIX=/usr \
 		CFLAGS="$CFLAGS" \
 		DBUS_LIBS= \
 		DBUS_CFLAGS= \
 		DBUS_MINOR= \
-		all || return 1
```
Barthalion: How is this an improvement?

jirutka: This is just a minor insignificant change. `|| return 1` at the last line of a function doesn’t have any effect.

Barthalion:
Some time ago Valery Kartel submitted [a patch](http://patchwork.alpinelinux.org/patch/2120/) which introduces separate user for dnsmasq. Would be nice to have it here while you're on it.

jirutka:
We already use separate user for dnsmasq. However, this patch reminded me something other that I’d like to put here.

main/dnsmasq/APKBUILD
Barthalion: We already discussed this in past. Having explicit return is better than implicit one.

main/dnsmasq/APKBUILD
```diff
@@ -1,50 +1,55 @@
...
 	$pkgname.initd
 	$pkgname.confd
 	"
+builddir="$srcdir/$pkgname-$pkgver"
+
+prepare() {
+	default_prepare || return 1
+
+	cd "$builddir"
+	sed -i 's/^AWK.*/AWK = awk/' Makefile
```
Barthalion: Is this one still needed? Also both seds lack `|| return 1`.

main/dnsmasq/APKBUILD
```diff
@@ -1,50 +1,54 @@
...
 	$pkgname.initd
 	$pkgname.confd
 	"
+builddir="$srcdir/$pkgname-$pkgver"
+
+prepare() {
+	default_prepare || return 1
+
+	cd "$builddir"
+	sed -Ei 's|#(conf-dir=/etc/dnsmasq.d/,\*\.conf)|\1|' dnsmasq.conf.example
```
Barthalion: Still missing `|| return 1`…

jirutka: Not needed here, this sed is not mandatory. I haven't removed any existing one, this line is new. We don't have any rule saying "put `|| return 1` after every single line of apkbuild".

I should finally fix abuild to run with `set -e` to end this nonsense cargo cult...

Barthalion: Every command is mandatory, and code readability is not cargo cult. The only nonsense is your attitude to any review not coming from you or Natanael.

clandmeter: @jirutka I would also prefer a proper fix in abuild but currently we dont have it. I agree with @Barthalion any cmd in our apkbuild should be mandatory or we should remove it completely. On another note, i would prefer patching config files purely for the sense of readability and making sure it applies properly when bumping version.

PR #185, opened 2016-07-29
testing/gogs: reuse the existing glide files

7heo:
Reuse the glide files that are present in the gogs tarball, update the missing packages/versions with patches.


jirutka:
Upstream's glide files are so incomplete, that it'd be IMO better (easier to maintain) just keep full copy of these files instead of patches.

7heo:
@jirutka the issue would be that then, we don't get upstream's release. We have a non-consistent release with random versions of everything.

jirutka:
Have you tried to tell them about it (i.e. open an issue)?

How have you actually created these patches?

7heo:
No I have not.

To create the patch, I have used the previous versions of the glide files I checked out in git, and the versions from the gogs repository. It's simply a diff of them, keeping the package versions from the gogs one in case there were conflicts.

jirutka:
Take a look at this for inspiration: https://github.com/alpinelinux/aports/commit/10c0bb8397a231f8a6f1db280e67de7bf7668005

7heo:
@jirutka the current gogs package is broken; because of non-matching versions. It builds, but it has execution errors. Can we please merge this PR asap?

7heo:
@jirutka also, about the other PR you've linked, I don't see the difference between it and the current version of gogs (i.e. the one before the PR we're discussing): all the glide files are currently checked in the RCS. I feel that the decision having-full-files vs having-patches in the RCS is independent from having-the-package-work. ;)

Can we merge this and then discuss what makes more sense?

jirutka:
Look closer, the `glide_init` function. I'm just at launch.

7heo:
@jirutka valid point, I did not see it on the patch (@clandmeter) did show it to me last week in a WIP version. My problem stands: current gogs is broken, and I need it. It really stresses me out a LOT, and I fear that I won't be able to calmly discuss cosmetic changes without having a working RCS. Can we merge this asap? :)

PR #186, opened 2016-07-31
main/mariadb: move dialog.so and mysql_clear_password.so to -common

kiwiroy:
This patch fixes a typo and makes the `mariadb-common` package more consistent with MariaDB packaging for CentOS. 

``` bash
$ rpm -qlp http://yum.mariadb.org/10.1/centos/7.1/x86_64/rpms/MariaDB-10.1.16-centos7-x86_64-common.rpm | grep plugin
/usr/lib64/mysql/plugin
/usr/lib64/mysql/plugin/dialog.so
/usr/lib64/mysql/plugin/mysql_clear_password.so
```

This is logical for the client use-case that wishes to use `dialog.so`, for connection to a PAM authenticating server, but not include `mariadb`.

### DIFF
--- a/main/mariadb/APKBUILD
+++ b/main/mariadb/APKBUILD
@@ -125,11 +125,16 @@ _client_libs() {
 }
 
 common() {
-	pkgdesc="MariaDB common files for boh server and client"
+	pkgdesc="MariaDB common files for both server and client"
 	replaces="mysql-common"
 	depends=
-	mkdir -p "$subpkgdir"/usr/share/mysql "$subpkgdir"/etc
+	mkdir -p "$subpkgdir"/usr/share/mysql \
+		"$subpkgdir"/etc \
+		"$subpkgdir"/usr/lib/mysql/plugin
 	mv "$pkgdir"/etc/mysql "$subpkgdir"/etc/ || return 1
+	mv "$pkgdir"/usr/lib/mysql/plugin/dialog.so \
+		"$pkgdir"/usr/lib/mysql/plugin/mysql_clear_password.so \
+		"$subpkgdir"/usr/lib/mysql/plugin/ || return 1
 	local lang="charsets danish english french greek italian korean norwegian-ny
 		portuguese russian slovak swedish czech dutch estonian german
 		hungarian japanese norwegian polish romanian serbian spanish


jirutka:
Are these two plugins (`dialog.so`, `mysql_clear_password.so`) even used for server, or just for client? If it’d be used just for client, then it’d be better to move them to mariadb-client subpackage.

Could you please squash the commits and change the commit message to `main/mariadb: move dialog.so and mysql_clear_password.so to -common` (use force push)?

**Notes for maintainers:**

These files are tiny and don’t pull any additional dependencies, so it’s okay.

```
du usr/lib/mysql/plugin/dialog.so usr/lib/mysql/plugin/mysql_clear_password.so
	8.0K	usr/lib/mysql/plugin/dialog.so
	8.0K	usr/lib/mysql/plugin/mysql_clear_password.so

ldd usr/lib/mysql/plugin/dialog.so
	ldd (0x34878ede000)
	libc.musl-x86_64.so.1 => ldd (0x34878ede000)

ldd usr/lib/mysql/plugin/mysql_clear_password.so
	ldd (0x2dde8ab2000)
```

Other distros:

* Arch Linux doesn’t have -common subpackage, it does have -client, but these plugins are provided by the main (server) package.
* Fedora provides these plugins in the -common package (the same as CentOS).
* Debian provides these plugins in the package libmariadb2, not -common or -client, and it seems that -server does _not_ depend on libmariadb2.

kiwiroy:
@jirutka I can't tell for sure if they are used by the server, but the [knowledge base](https://mariadb.com/kb/en/mariadb/pam-authentication-plugin/) suggests not.

Commits squashed and message corrected.

jirutka:
Debian package also suggests that they are not used by the server, but `libmysqlclient.so`. I also find some issues mentioning this connection.

Could you please move `dialog.so` and `mysql_clear_password.so` to `-client-libs` subpackage and try if it works for you (i.e. you can still connect to a PAM authenticating server)?

ncopa:
Thost libs is used for server replication, so they are needed by both server and client. https://github.com/mariadb/server/commit/b988553c527d0fb55a561dc3a1308052d94ae2a4

kiwiroy:
@ncopa thanks

PR #191, opened 2016-08-02
main/hunspell: upgrade to 1.4.1

liqa:


liqa:
it was reverted back to 1.3.4 why?

jirutka:
Because it breaks ABI and we were close to 3.4.3 release. More in the [IRC log](http://www.irclogger.com/.alpine-devel/2016-08-23#1471955372-1471955872).

liqa:
now that 3.4.3 is released can this package be updated?

ncopa:
It still breaks the ABI so we need rebuild a bunch of packages together with this update, or we need to add a temporary compat package, hunspell1.3 for a migration period. In either case, we need do more than just update it.

PR #195, opened 2016-08-07
testing/julia: use llvm3.7 instead of latest llvm package

jirutka:


jirutka:
Note: rerun after llvm3.7-3.7.1-r3 finally arrive…

PR #198, opened 2016-08-09
testing/openblas: add ILP64 variant (subpackage -ilp64) and some fixes

jirutka:


jirutka:
Travis build failed because the log output exceeded 4MiB limit.

PR #199, opened 2016-08-09
testing/julia: use OpenBLAS ILP64 on x86_64

jirutka:


tmpfile:
cc @Shizmob the new maintainer

PR #200, opened 2016-08-10
main/nginx: use dynamic modules, add new subpackages [WIP]

jirutka:


Barthalion:
I'm not sure what do you have on your to do list with regard to this pull request (as it's marked as WIP without further description), but RTMP module can be build externally since yesterday (and unofficial patch was flying around a little bit longer), additionally there is nginx-naxsi in testing which should be merged into the main package.

Modules could also ship a file similar to what PHP modules do, for example to modules-available directory. And even without it, there should be post upgrade message about breaking changes to nginx configuration.

Barthalion:
And on a side note… I'd be happy if you add echo and headers-more modules too while you're on it.

jirutka:
> RTMP module can be build externally since yesterday (and unofficial patch was flying around a little bit longer)

I know about that latest commits fixing dynamic modules. However, I’m not sure if we can consider it stable enough and already use it in the package.

By the unofficial patch, do you mean [sergey-dryabzhinsky’s](https://github.com/sergey-dryabzhinsky/nginx-rtmp-module) fork? Do you know about someone how uses it daily?

I found two issues/PRs about dynamic modules support and it’s not very clear what is the current state:

* https://github.com/arut/nginx-rtmp-module/pull/786
* https://github.com/arut/nginx-rtmp-module/issues/782

However, I’d love to use that patch, so I can get rid of nginx-rtmp variant and simplify the abuild.

> nginx-naxsi in testing which should be merged into the main package.

Thank’s for the info, I’ll do it.

> Modules could also ship a file similar to what PHP modules do … there should be post upgrade message about breaking changes to nginx configuration.

Yes, that’s what I’m working on now. I haven’t finished it yet because it was too late. 🌃

> I'd be happy if you add echo and headers-more modules too

Okay, I’ll add them.

Thanks for your comments.

jirutka:
* RTMP is updated to 1.1.9 and built as a dynamic module.
* Every subpackage with dynamic module also installs file `/etc/nginx/modules/$mod_name.conf` with `load_module` directive that is loaded by default (same as in Fedora package).
* There is an post-upgrade message that explains changes in the nginx package, it’s shown when upgrading from nginx <1.10.1-r3.
* There are new 3rd party dynamic modules: http-echo, http-headers-more, http-upload-progress, and http-fancyindex.

* * *

Package [testing/nginx-naxsi](https://github.com/alpinelinux/aports/blob/b64569ea04735b7cb38a615ba67f61b1f0cb8927/testing/nginx-naxsi/APKBUILD) contains modules [naxsi](https://github.com/nbs-system/naxsi), [cache_purge](https://github.com/FRiCKLE/ngx_cache_purge), [upstream-fair](https://github.com/gnosek/nginx-upstream-fair), and [http-sysguard](https://github.com/alibaba/nginx-http-sysguard). None of these modules provides support for dynamic loading in stable release. Naxsi is going to support it in upcoming 0.55 (currently in RC2). Other modules are not actively maintained.

Patches for adding dynamic loading exist (it’s usually very simple, just about modifying config file), but there’s zero guarantee (?) that they actually work correctly without any undesirable behaviour. It may be ok-ish if we know that faulty dynamic modules _that are not loaded_ cannot affect stability of nginx, but I haven’t studied nginx code, so don’t know this. Does anyone else know?

So I don’t like adding these modules to the stable nginx package, at least not in this pull request.

PR #202, opened 2016-08-10
testing/openimageio: new aport

Klowner:
new package💃
(depends on opencolorio)

### DIFF
--- a/testing/openimageio/APKBUILD
+++ b/testing/openimageio/APKBUILD
@@ -1,0 +1,67 @@
+# Contributor: Mark Riedesel <mark@klowner.com>
+# Maintainer: Mark Riedesel <mark@klowner.com>
+pkgname=openimageio
+pkgver=1.6.14
+pkgrel=0
+pkgdesc="Image i/o library supporting a multitude of image formats"
+url="https://sites.google.com/site/openimageio/"
+arch="all"
+license="BSD"
+depends=""
+makedepends="cmake giflib-dev zlib-dev boost-dev jpeg-dev libpng-dev tiff-dev
+	glew-dev python-dev>=2.6 python3-dev openexr-dev ffmpeg2.8-dev libraw-dev libwebp-dev freetype-dev
+	qt-dev mesa-dev openssl-dev jasper-dev opencolorio-dev"
+install=""
+subpackages="py2-$pkgname:_python py3-$pkgname:_python $pkgname-dev $pkgname-doc $pkgname-tools"
+source="$pkgname-$pkgver.tar.gz::https://github.com/OpenImageIO/oiio/archive/Release-${pkgver}.tar.gz
+	nocpp-error.patch"
+
+builddir="$srcdir"/oiio-Release-$pkgver
+
+build() {
+	local py2path=$(python2 -c 'import site; print(site.getsitepackages()[0])')
+	local py3path=$(python3 -c 'import site; print(site.getsitepackages()[0])')
+
+	cd "$builddir"
+	mkdir build && cd build
+	cmake .. \
+		-DUSE_OPENSSL=ON \
+		-DUSE_PYTHON=ON \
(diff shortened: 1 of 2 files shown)


testing/openimageio/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+	cd "$builddir"
+	mkdir build && cd build
+	cmake .. \
+		-DUSE_OPENSSL=ON \
+		-DUSE_PYTHON=ON \
+		-DUSE_PYTHON3=OFF \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DPYLIB_INSTALL_DIR=lib/python2.7/site-packages \
+		-DOIIO_BUILD_TESTS=OFF \
+		-DOIIO_BUILD_TOOLS=ON
```
jirutka: Please add `|| return 1` at the end of this line (just this one), to ensure that an error will halt the build (abuild doesn’t run with `set -e` yet).

testing/openimageio/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+
+package() {
+	cd "$builddir"/build
+	make DESTDIR="$pkgdir" install || return 1
+	rm -rf "$pkgdir"/usr/share/fonts/oiio
+}
+
+doc() {
+	default_doc
+}
```
jirutka: `default_doc` is called implicitly when `doc` function is not declared, so you can just omit it.

testing/openimageio/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+
+dev() {
+	default_dev
+}
```
jirutka: `default_dev` is called implicitly when `dev` function is not declared, so you can just omit it.

testing/openimageio/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+
+builddir="$srcdir"/oiio-Release-$pkgver
+
+build() {
+	cd "$builddir"
+	mkdir build && cd build
+	cmake .. \
+		-DUSE_OPENSSL=ON \
+		-DUSE_PYTHON=ON \
+		-DUSE_PYTHON3=OFF \
```
jirutka: Why have you disabled Python 3?

testing/openimageio/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+pkgdesc="Image i/o library supporting a multitude of image formats"
+url="https://sites.google.com/site/openimageio/"
+arch="all"
+license="BSD"
+depends=""
+makedepends="cmake giflib-dev zlib-dev boost-dev jpeg-dev libpng-dev tiff-dev
+	glew-dev python2-dev openexr-dev ffmpeg2.8-dev libraw-dev libwebp-dev freetype-dev
+	qt-dev mesa-dev openssl-dev jasper-dev opencolorio-dev"
+install=""
+subpackages="$pkgname-python2:py2 $pkgname-python3:py3 $pkgname-dev $pkgname-doc $pkgname-utils"
```
jirutka: Python subpackages should be named `py2-$pkgname` and `py3-$pkgname`.

jirutka: You can also use common split function `_python` for both of these subpackages. Then you can use `$subpkgname` in the split function to get name of the specific subpackage and e.g. `${subpkgname:2:1}` to extract `2` or `3`.

Klowner:
It seems there's some sort of issue with the boost packages? I was having this same build issue on my local machine, but after manually rebuilding boost everything seems to have resolved itself. 😕

jirutka:
I'm on vacation now, with almost no internet connection. I'll finish it at the end of week.

Klowner:
enjoy your vacation! ⛵

Klowner:
Anything else left for this? I think I got everything.

PR #203, opened 2016-08-10
testing/opencolorio: new aport

Klowner:
new package! 🌈

### DIFF
--- a/testing/opencolorio/APKBUILD
+++ b/testing/opencolorio/APKBUILD
@@ -1,0 +1,81 @@
+# Contributor: Mark Riedesel <mark@klowner.com>
+# Maintainer: Mark Riedesel <mark@klowner.com>
+pkgname=opencolorio
+pkgver=1.0.9
+pkgrel=0
+pkgdesc="A color management framework for visual effects and animation"
+url="http://opencolorio.org"
+arch="all"
+license="BSD"
+depends=""
+makedepends="cmake python2-dev python3-dev freeglut-dev glew-dev lcms2-dev tinyxml-dev yaml-cpp-dev boost-dev"
+install=""
+subpackages="py2-$pkgname:py2 py3-$pkgname:py3 $pkgname-dev $pkgname-tools"
+source="$pkgname-$pkgver.tar.gz::http://github.com/imageworks/OpenColorIO/archive/v$pkgver.tar.gz
+	yaml-cpp-0.5-support.patch
+	temp-deref-with-yaml-0.5.0.patch"
+builddir="$srcdir"/OpenColorIO-$pkgver
+
+build() {
+	cd "$builddir"
+
+	# Just to be sure we're using alpine dev packages
+	rm -f ext/lcms* ext/tinyxml* ext/yaml*
+
+	local pyver
+	for pyver in python3 python2; do
+		mkdir build-$pyver && cd build-$pyver
+
+		cmake .. \
+			-DCMAKE_INSTALL_PREFIX=/usr \
+			-DPYTHON=/usr/bin/$pyver \
+			-DOCIO_BUILD_TRUELIGHT=OFF \
+			-DOCIO_BUILD_NUKE=OFF \
+			-DOCIO_BUILD_SHARED=ON \
+			-DOCIO_BUILD_STATIC=OFF \
+			-DUSE_EXTERNAL_YAML=TRUE \
+			-DUSE_EXTERNAL_TINYXML=TRUE \
+			-DUSE_EXTERNAL_LCMS=TRUE \
+			|| return 1
+		make || return 1
+		cd ..
+	done
+}
+
+package() {
+	local pyver
+	for pyver in python3 python2; do
+		cd "$builddir"/build-$pyver
+		make DESTDIR="$pkgdir" install
+	done
+}
+
+tools() {
+	pkgdesc="OpenColorIO color management framework tools"
+	mkdir -p "$subpkgdir"/usr/
+	mv "$pkgdir"/usr/bin "$subpkgdir"/usr || return 1
+}
+
+_python() {
+	pkgdesc="Python $1 bindings for OpenColorIO color management framework"
+	depends="python$1"
+	mv "$pkgdir"$2 "$subpkgdir"
+}
+
+py2() {
+	_python 2 $(python2 -c 'import site; print(site.getsitepackages()[0])')
+}
+
+py3() {
+	_python 3 $(python3 -c 'import site; print(site.getsitepackages()[0])')
+}
+
+md5sums="06d0efe9cc1b32d7b14134779c9d1251  opencolorio-1.0.9.tar.gz
+11bb6b587b2e5242f75aca281bab158c  yaml-cpp-0.5-support.patch
+5f13bef26e14b2dd908e69a2bc8fb7a4  temp-deref-with-yaml-0.5.0.patch"
+sha256sums="27c81e691c15753cd2b560c2ca4bd5679a60c2350eedd43c99d44ca25d65ea7f  opencolorio-1.0.9.tar.gz
+b381912a81342e242d1a6ef39cf1556f7b383407af9c1aefbde06e38186777ad  yaml-cpp-0.5-support.patch
+6df29f0d3de06ab0fbc2fa06f6fc7537d0628c4185d7a9f6228b2d0be07917d7  temp-deref-with-yaml-0.5.0.patch"
+sha512sums="ac953ba9904aff44de37cc2ee60dbf524bd86d25f699c1eacaa61ca30fed8f077194d47d34a72c05b706da7e1a3974a988d67d60031d424d91b9240f8ab86ed3  opencolorio-1.0.9.tar.gz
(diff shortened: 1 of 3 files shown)


testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+}
+
+package() {
+	cd "$builddir"/build
+	make DESTDIR="$pkgdir" install
+}
+
+dev() {
+	default_dev
+}
```
jirutka: `default_dev` is called implicitly when `dev` function is not declared, so you can just omit it.

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+makedepends="cmake python2-dev"
+install=""
+subpackages="$pkgname-python $pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::http://github.com/imageworks/OpenColorIO/archive/v$pkgver.tar.gz"
+builddir="$srcdir"/OpenColorIO-$pkgver
+
+build() {
+	cd "$builddir"
+	mkdir build && cd build
+	cmake -DCMAKE_INSTALL_PREFIX=/usr ..
```
jirutka: Please add `|| return 1` at the end of this line, to ensure that an error will halt the build (abuilds don’t run with `set -e` yet).

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+dev() {
+	default_dev
+}
+
+python() {
+	pkgdesc="Python bindings for OpenColorIO color management framework"
+	depends="python2"
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/python2.7 "$subpkgdir"/usr/lib
+}
```
jirutka: OpenColorIO supports both legacy Python and Python 3. Could you please also add a subpackage for Python 3?

Klowner: It looks like I have to build it once for each python version (seems silly), I'll keep digging.

Klowner:
I think I've addressed all your feedback. It seems their build system isn't suited towards building python2 and python3 at the same time, so I'm just building both. If you'd prefer I take another route, let me know.

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,69 @@
...
+pkgver=1.0.9
+pkgrel=0
+pkgdesc="A color management framework for visual effects and animation"
+url="http://opencolorio.org"
+arch="all"
+license="BSD"
+depends=""
+makedepends="cmake python2-dev python3-dev freeglut-dev glew-dev"
+install=""
+subpackages="$pkgname-python2:py2 $pkgname-python3:py3 $pkgname-dev $pkgname-utils"
```
jirutka: Python subpackages should be named `py2-$pkgname` and `py3-$pkgname`.

jirutka:
Thanks for addressing the issues.

I’ve looked at opencolorio more closely. It bundles bunch of dependencies, namely lcms, tinyxml and yaml (and some others used only for document generation). Luckily these can be simply replaced with system-provided, you have to just pass these options to cmake:

```
-DUSE_EXTERNAL_YAML=TRUE
-DUSE_EXTERNAL_TINYXML=TRUE
-DUSE_EXTERNAL_LCMS=TRUE
```

… and delete the files, just to be sure: `rm -f ext/lcms* ext/tinyxml* ext/yaml*`.

We have a package for tinyxml and lcms, but not yet for yaml-cpp.

According to Fedora package [OpenColorIO](http://pkgs.fedoraproject.org/cgit/rpms/OpenColorIO.git/tree/OpenColorIO.spec?h=f25) it may be also needed to use `-DOCIO_USE_SSE=OFF` for non-x86_64 platforms (but I haven’t checked out OpenColorIO docs about this).

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,69 @@
...
+source="$pkgname-$pkgver.tar.gz::http://github.com/imageworks/OpenColorIO/archive/v$pkgver.tar.gz"
+builddir="$srcdir"/OpenColorIO-$pkgver
+
+build() {
+	cd "$builddir"
+
+	for pyver in python3 python2; do
```
jirutka: This variable should be declared as local (`local pyver`), to not pollute global namespace of the script.

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,69 @@
...
+
+_python() {
+	pkgdesc="Python $1 bindings for OpenColorIO color management framework"
+	depends="python$1"
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/python$2 "$subpkgdir"/usr/lib/
+}
+
+py2() {
+	PY2_VERSION=$(python -c 'import sys; print("%i.%i" % (sys.version_info.major, sys.version_info.minor))')
```
jirutka: Maybe it’d be better (simpler) to use `python -c 'import site; print(site.getsitepackages()[0])'`, this directly prints location of the `site-packages` directory (e.g. `/usr/lib/python2.7/site-packages`).

jirutka: Or you can entirely avoid it as in your second package using `mv "$pkgdir"/usr/lib/python$1?? "$subpkgdir"/usr/lib/` (or simply `$pkgdir"/usr/lib/python$1*`).

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,69 @@
...
+	for pyver in python3 python2; do
+		mkdir build-$pyver && cd build-$pyver
+
+		cmake .. \
+			-DCMAKE_INSTALL_PREFIX=/usr \
+			-DPYTHON=/usr/bin/$pyver \
+			-DOCIO_BUILD_TRUELIGHT=OFF \
+			-DOCIO_BUILD_NUKE=OFF \
+			|| return 1
+		make
```
jirutka: Okay, I had to be more accurate previously. Every statement inside a function that may fail (and should not) should be followed by `|| return 1` (until we change abuild to run `set -e`). It’s not needed after the last statement of a function, because shell has implicit return. For example `cd ..`, `mkdir -p` etc. normally don’t fail, so not needed here, but `make` can fail and it should halt the built in that case.

Klowner:
Looks like they're using a `yaml-cpp > 0.3.0` which appears to be different from the `yaml-dev` in Alpine. I suppose I could make an `APKBUILD` for that too?

jirutka:
> I suppose I could make an APKBUILD for that too?

Yes, exactly. Since it’s a dependency of this package, add it to this pull request.

testing/opencolorio/APKBUILD
Klowner: Oh yes, that's much nicer.

Klowner:
Think that's everything, do you see anything else I can improve?

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,77 @@
...
+pkgver=1.0.9
+pkgrel=0
+pkgdesc="A color management framework for visual effects and animation"
+url="http://opencolorio.org"
+arch="all"
+license="BSD"
+depends=""
+makedepends="cmake python2-dev python3-dev freeglut-dev glew-dev lcms2-dev tinyxml-dev yaml-cpp-dev boost-dev"
+install=""
+subpackages="py2-$pkgname:py2 py3-$pkgname:py3 $pkgname-dev $pkgname-utils"
```
jirutka: It’d be better to rename opencolorio-utils to opencolorio-tools. This name uses Fedora and Debian, so it’d be more clear for users coming from main-stream distros.

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,77 @@
...
+source="$pkgname-$pkgver.tar.gz::http://github.com/imageworks/OpenColorIO/archive/v$pkgver.tar.gz
+	yaml-cpp-0.5-support.patch"
```
jirutka: Please add also [this patch](https://github.com/imageworks/OpenColorIO/commit/1384029df00525c479e2c7e52fdfa294413fceb6) that fixes dereference of temporary when yaml-cpp 0.5.x is used.

jirutka:
I’ve merged testing/yaml-cpp in 47c4c32a7954941a3250746a1e4910ce55875910.

testing/opencolorio/APKBUILD
```diff
@@ -0,0 +1,77 @@
+# Contributor: Mark Riedesel <mark@klowner.com>
+# Maintainer:
```
jirutka: Why you haven’t declared yourself as a maintainer?

Klowner: I must have misunderstood the purpose of the maintainer field. The simple example ( https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Simple ) also contains a blank maintainer field.

jirutka: Aha. Have you used `newapkbuild` when creating abuild? This automatically fills Maintainer.

Maintainer is a person who’s responsible for maintaining the package. That person i.a. gets mails from http://pkgs.alpinelinux.org/ when someone flags the package as outdated or something. Maintainer can be anyone, not just Alpine core or trusted devs.

Klowner:
Good catch on finding that temp dereference patch. It's now included.

PR #206, opened 2016-08-15
main/php5: Change php5-module.conf to use SetHandler

pawciobiel:
Avoids serving non php files by php handler.


jirutka:
You have to update checksums for `php7-module.conf` (run `abuild checksum` inside this package directory) and also increment `pkgrel`.

ncopa:
finally merged with 91d57c3ab0335169b5805e0bfb8f9af826500837. Thanks!

PR #207, opened 2016-08-15
testing/php7: Change php7-module.conf to use SetHandler

pawciobiel:
Avoids serving non php files by php handler.
Deny access to *.phps files.
Deny access to files without filename e.g. .php3
Disable running PHP scripts in user directories by default.


jirutka:
You have to update checksums for `php7-module.conf` (run `abuild checksum` inside this package directory) and also increment `pkgrel`.

andypost:
revision could be done after https://github.com/alpinelinux/aports/pull/218

PR #212, opened 2016-08-15
main/boost: improve abuild

jirutka:


main/boost/APKBUILD
```diff
@@ -1,68 +1,61 @@
...
 pkgname=boost
 pkgver=1.60.0
-
-_x=${pkgver%%.*}	# strip .y.z
-_y=${pkgver%.*}		# strip .z
-_y=${_y#*.}		# strip x.
-_z=${pkgver##*.}	# strip x.y.
-_ver=${_x}_${_y}_${_z}
-
+_pkgver="${pkgver//./_}"
```
jirutka: This syntax is not POSIX, but ash supports it.

przemoc suggested on IRC that we should avoid it and use `_pkgver="$(echo "$_pkgver" | sed "s/\./_/g")"` instead. It’ll not be much slower when using Busybox (no external binary invocation). If Alpine switch to something else, then it will still work, just be a little bit slower.

ncopa: I think i prepare the ash syntax.

PR #216, opened 2016-08-18
testing/blender: new aport

Klowner:
Lots of subpackages in this one. I'm not sure what the preferred way for handling alternate binaries is; for the headless build I renamed the binary to `blender-headless`, although I'm not sure why someone would want headless and regular builds installed at the same time. Is this maybe a use-case for `provides`?

Thanks!

(depends on openimageio)

### DIFF
--- a/testing/blender/APKBUILD
+++ b/testing/blender/APKBUILD
@@ -1,0 +1,122 @@
+# Contributor: Mark Riedesel <mark@klowner.com>
+# Maintainer: Mark Riedesel <mark@klowner.com>
+pkgname=blender
+pkgver=2.77a
+pkgrel=0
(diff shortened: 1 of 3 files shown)


Klowner:
Probably also want to have opensubdiv http://graphics.pixar.com/opensubdiv/docs/intro.html and CUDA support.

Klowner:
Waiting on `opensubdiv`, `ilmbase`, and `openexr`, otherwise it should be all good to go.

Klowner:
#226 is the opensubdiv pr, for reference

fabled:
Committed, with follow-up upgrade and minor fixes.

PR #217, opened 2016-08-18
Updates py-lxml to 3.6.3, adds py3-lxml

funkyfuture:
- adds missing python dependency
- more explicit `pkgdesc`
- changes to a simpler source location
- checksums are generated from a gpg-verified download

### DIFF
--- a/main/py-lxml/APKBUILD
+++ b/main/py-lxml/APKBUILD
@@ -1,29 +1,53 @@
 # Contributor: Francesco Colista <fcolista@alpinelinux.org>
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=py-lxml
-_pkgname=lxml
-pkgver=3.6.0
+_pkgname="${pkgname#py-}"
+pkgver=3.6.3
 pkgrel=0
 pkgdesc="Python LXML Library"
+_pkgdesc=$pksdesc
 url="http://lxml.de/"
-arch="all"
+arch="noarch"
 license="BSD"
 depends=""
-makedepends="python-dev libxml2-dev libxslt-dev py-setuptools"
-install=""
-source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+makedepends="python-dev python3-dev libxml2-dev libxslt-dev py-setuptools"
+subpackages="py2-$_pkgname:_py2 py3-$_pkgname:_py3"
+source="${url}files/$_pkgname-$pkgver.tgz"
 
-_builddir="$srcdir"/$_pkgname-$pkgver
+_builddir="$srcdir/$_pkgname-$pkgver"
 build() {
 	cd "$_builddir"
 	python setup.py build || return 1
+	python3 setup.py build || return 1
 }
 
 package() {
-	cd "$_builddir"
-	python setup.py install --prefix=/usr --root="$pkgdir" || return 1
+	mkdir -p "$pkgdir"
 }
 
-md5sums="5957cc384bd6e83934be35c057ec03b6  lxml-3.6.0.tar.gz"
-sha256sums="9c74ca28a7f0c30dca8872281b3c47705e21217c8bc63912d95c9e2a7cac6bdf  lxml-3.6.0.tar.gz"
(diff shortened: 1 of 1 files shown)


jirutka:
We don’t create separate APKBUILDs for py2 and py3 variants, these should be subpackages in a single APKBUILD. See [py-setproctitle](https://github.com/alpinelinux/aports/blob/master/community/py-setproctitle/APKBUILD) for an example.

funkyfuture:
that's cool. i will amend it this evening.

some questions / remarks concerning your example:

- at the top arch is noarch, in _py it's all. is this intended?
- setting the specific interpreter as dependency is missing, isn't it?
- line 21 missing the orred return 1
- the proper term in the description would be package, not module (even though it contains only one module)

wouldn't it be cool if pip's download command was used und er the hood?

funkyfuture:
i adjusted the APKBUILD, but am stuck with this error, which i can't figure out. i'm using andyshinn's docker image to test the build.

```
running install_egg_info
running egg_info
writing requirements to src/lxml.egg-info/requires.txt
writing src/lxml.egg-info/PKG-INFO
writing top-level names to src/lxml.egg-info/top_level.txt
writing dependency_links to src/lxml.egg-info/dependency_links.txt
reading manifest file 'src/lxml.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'src/lxml.egg-info/SOURCES.txt'
Copying src/lxml.egg-info to /home/builder/package/pkg/py-lxml/usr/lib/python2.7/site-packages/lxml-3.6.3-py2.7.egg-info
running install_scripts
>>> py2-lxml*: Preparing subpackage py2-lxml...
/usr/bin/abuild: cd: line 2246: can't cd to /home/builder/package/pkg/py2-lxml
>>> ERROR: py2-lxml*: Missing /home/builder/package/pkg/py2-lxml
>>> ERROR: py-lxml*: prepare_subpackages failed
>>> ERROR: py-lxml: all failed
>>> py-lxml: Uninstalling dependencies...
```

would that be merged into another tree than *edge*? in that case the `depends` value in line 40 would be wrong.

jirutka:
Sorry for late response, we have been very busy with cleaning the repo and preparing aarch64.

> - at the top arch is noarch, in _py it's all. is this intended?

Yes. py-* has `noarch` (it’s an empty package), py2-* and py3-* have `all`.

> - line 21 missing the orred return 1

It’s the last statement of the function, so it has no effect here (shell has implicit return).

> - the proper term in the description would be package, not module (even though it contains only one module)

You’re right.

> wouldn't it be cool if pip's download command was used und er the hood?

No, because we want to verify checksums.

> i adjusted the APKBUILD, but am stuck with this error…

There should be `--root="$subpkgdir"`, not `--root="$pkgdir"`.

* * *

I’ve already added py3 and upgraded lxml to 3.6.4 in #243.

funkyfuture:
thanks for the feedback anyway.

>> wouldn't it be cool if pip's download command was used under the hood?

> No, because we want to verify checksums.

i don't know about the mechanisms of abuild, but i didn't mean to skip checksum-tests. the idea is motivated by the fact the one wouldn't have to care about a package's source as the PyPI is very well informed about this.

> There should be --root="$subpkgdir", not --root="$pkgdir".

d'oh, obviously. ;-)

i'll extend the python package guide in wiki when i got my first package build for both versions of Python working.

jirutka:
> the idea is motivated by the fact the one wouldn't have to care about a package's source as the PyPI is very well informed about this.

This would just increase complexity of abuild and I don’t see any benefits. Currently the only (extra) thing needed for python abuilds to fetch the source tarball is this line:

```sh
```

And it’s the same for basically all python abuilds.

> i'll extend the python package guide in wiki when i got my first package build for both versions of Python working.

I did it yesterday, if you mean [this example](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python).

You may be also interested in [this page](https://wiki.alpinelinux.org/wiki/TODO:py3_packages) with list of python abuilds that don’t provide py3- variant yet and instructions how to convert them. 😺

funkyfuture:
perfect. :+1: 

the wording of the last section's headline on the second page is a little confusing as it's not clear what is meant with packages (alpine vs. python). and may imply that these packages don't support Python 3.

PR #224, opened 2016-08-24
testing/vulkan-headers: new abuild

Klowner:
this is also an optional build dep for `testing/glfw`


ncopa:
Whats up with the `_pkgver`? It looks like `$pkgver` is totally unused.

I have maintenance scripts that will compare `pkgver` for newer releases and other scripts that modifies `pkgver` to upgrade the package, but those scripts would only change the version of the apk. the content would stay unmodified.

There must be some connection with the `$pkgver` and what is actually shipped in the package.

Klowner:
I think my concern was that it didn't match the common versioning of the other alpine packages; is it alright to use the weird long datestamp version for the `$pkgver`?

jirutka:
You can’t use `1.0-core-20160930` as pkgver, but `1.0.20160930` is not a problem. Alpine follows [Gentoo versioning format](https://devmanual.gentoo.org/ebuild-writing/file-format/).

Klowner:
cool, I can work with that

Klowner:
Betterish?

Klowner:
They started using normal versioning! Updated, also requires tinyxml2 (also added)

testing/vulkan-headers/APKBUILD
```diff
@@ -0,0 +1,75 @@
...
+url="https://www.khronos.org/vulkan/"
+arch="noarch"
+license="custom"
+depends=""
+makedepends="asciidoc tinyxml2-dev gcc sed python3"
+install=""
+subpackages="$pkgbase-doc $pkgbase-html-doc:_htmldoc"
+archive_name="Vulkan-Docs"
+source="$pkgname-$pkgver.tar.gz::https://github.com/KhronosGroup/Vulkan-Docs/archive/v${pkgver}-core.tar.gz
+	https://raw.githubusercontent.com/KhronosGroup/Vulkan-Hpp/master/VulkanHppGenerator.cpp"
```
fabled: This seems to download git master version of the file, which is since changed. Please fix this by one:
1. Package Vulkan-Hpp separately, and makedepend on it (preferred)
1. Add a Vulkan-Hpp release tarball as dependency here, and use files from it
1. Using download url including git hash, in that case the downloaded filename should be versioned too

nmeum:
@Klowner gentle reminder that there are outstanding reviews.

Klowner:
Oh shoot, so sorry, totally slipped under my radar. I'll try to get on this soon.

PR #226, opened 2016-08-25
testing/opensubdiv: new aport

Klowner:
Pixar's OpenSubdiv (with CUDA and OpenCL disabled), dependency for Blender


Klowner:
fwiw, I built Blender against this and it works.

PR #239, opened 2016-08-29
community/openexr: re-add, add maintainer, clean up

Klowner:


jirutka:
Do you use it? If you’ve checked that it works, then we should move it to the community repo.

Klowner:
Yep, it works. I built Blender against it and successfully rendered to an EXR file.

Klowner:
Changed to community/openexr

Klowner:
moving testing/ilmbase to community as well, since it is a dependency

PR #246, opened 2016-08-31
unmaintained/qtserialport: update to 5.6.0 and move to testing/qt5-qtserialport

w1r0x:
http://qt-project.org/wiki/QtSerialPort
Interface to serial ports for Qt application


testing/qt5-qtserialport/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+depends="qt5-qtbase"
+depends_dev="$depends"
+makedepends="$depends_dev qt5-qtbase-dev"
+install=""
+subpackages="$pkgname-dev"
+
+_ver=${pkgver/_/-}
+_ver=${_ver/beta0/beta}
+_ver=${_ver/rc0/rc}
+_V=${_ver/rc/RC}
```
jirutka: Why do you have these replacements here? I don’t see any versions with these suffixes on https://download.qt.io/official_releases/qt/.

testing/qt5-qtserialport/APKBUILD
```diff
@@ -0,0 +1,43 @@
+# Contributor: Denis Ryabyy <vv1r0x@gmail.com> 
+# Maintainer: Denis Ryabyy <vv1r0x@gmail.com>
+pkgname=qt5-qtserialport
+pkgver=5.6.0
+pkgrel=1
```
jirutka: It’s a new aport, so pkgrel should be 0. See [pkgrel on wiki](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgrel).

testing/qt5-qtserialport/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+        cd "$_builddir/install_root/usr"
+        mkdir $pkgdir/usr -p
+        mv lib $pkgdir/usr/
+}
+
+dev () {
+       cd "$_builddir/install_root/usr"
+       mkdir $subpkgdir/usr -p
+       mv include $subpkgdir/usr/
+}
```
jirutka: Please use tabs for indentation in APKBUILDs.

testing/qt5-qtserialport/APKBUILD
w1r0x: I took for a base APKBUILD file old "unmaintained/qtserialport/APKBUILD" so it is from there. Just fix that.

Is that ok that I'm deleting "unmaintained/qtserialport/APKBUILD" in this PR?

testing/qt5-qtserialport/APKBUILD
w1r0x: I've increased it every time when I test build. Now I know how to do it right way, so fixed.

testing/qt5-qtserialport/APKBUILD
w1r0x: fixed

w1r0x:
>depends="qt5-qtbase"
depends_dev="qt5-qtbase-dev"
makedepends="$depends $depends_dev"

I need some help with the depends block.
Is this correct?

When I'm installing "qt5-qtserialport-5.6.0-r0.apk" ( not the dev one ). I'm getting "qt5-qtbase-dev" as its depence. But it shouldn't be there.

testing/qt5-qtserialport/APKBUILD
jirutka: Nope, this is not from [unmaintained/qtserialport/APKBUILD](https://github.com/alpinelinux/aports/blob/master/unmaintained/qtserialport/APKBUILD).

Well, if you haven’t written it from scratch, then you should rather update unmaintained/qtserialport and then move it into testing (or vice versa), instead of adding it as a new aport.

jirutka:
>> depends="qt5-qtbase"
>> depends_dev="qt5-qtbase-dev"
>> makedepends="$depends $depends_dev"
>
>Is this correct?

No, `$depends` should not be in `makedepends`, this is implicit.

> When I'm installing "qt5-qtserialport-5.6.0-r0.apk" ( not the dev one ). I'm getting "qt5-qtbase-dev" as its depence.

That’s weird, it shouldn’t pull `qt5-qtbase-dev`.

w1r0x:
Thank you. I'll check it with fixed $makedepends and try to find what cause qt5-qtbase-dev to be depence.

testing/qt5-qtserialport/APKBUILD
w1r0x: I wrote it taking some other APKBUILD file as example. I though it was unmaintained/qtserialport/APKBUILD. But I think it should be better to update old and move. Thank you.

w1r0x:
```
+_ver=${pkgver/_/-}
+_ver=${_ver/beta0/beta}
+_ver=${_ver/rc0/rc}
+_V=${_ver/rc/RC}

```
I got it from qt5-qtbase
https://github.com/alpinelinux/aports/blob/1b77b99057f204718bab6468f87b16ac07190aae/main/qt5-qtbase/APKBUILD

w1r0x:
Just resolved depends problem. It was not all dev files moved to dev package. So autodepends was added. Now it just fine.

jirutka:
I’ve looked at this more deeply and still can’t understand why you’re doing it so complicated. Everything what you need in the `build` function is `cd "$builddir"; make INSTALL_ROOT="$pkgdir" install`. Explicit `dev` function is not needed at all. This is even in the original abuild that you have read.

PR #248, opened 2016-08-31
Update community/R

artemklevtsov:
Move some deps to depends_dev (required for the several popular packages from CRAN). Remove enabled by default configure opts with the same result. Add `/etc/R/Rprofile.site` and symlink.
Also I think we should use `--without-x` `configure` option.
To significantly reduce the size of package we can build it with the internal BLAS implementation.

### DIFF
--- a/community/R/APKBUILD
+++ b/community/R/APKBUILD
@@ -1,9 +1,10 @@
+# Contributor: Artem Klevtsov <a.a.klevtsov@gmail.com>
 # Contributor: Nirosan <pnirosan@gmail.com>
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=R
 pkgver=3.3.1
-pkgrel=2
+pkgrel=3
 pkgdesc="Language and environment for statistical computing"
 url="https://www.r-project.org"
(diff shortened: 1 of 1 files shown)


artemklevtsov:
Added `-D_DEFAULT_SOURCE` to CFLAGS.

jirutka:
> Move some deps to depends_dev (required for the several popular packages from CRAN).

I’ve adopted this in cc89c3ccaf385a1a949533916923406fe2a6f024.

> Remove enabled by default configure opts with the same result.

IIRC default value of these options is autodetect, i.e. they are enabled when required library is found in the system. This can cause many troubles, explicit is better here.

> Also I think we should use `--without-x` configure option.

No, this is enabled intentionally – it’s needed to plot graphs.

> To significantly reduce the size of package we can build it with the internal BLAS implementation

This would hurt performance.

> Added `-D_DEFAULT_SOURCE` to CFLAGS: the `readxl` package compilation failed without it.

What exactly does this flag do? I can’t find any information about it, just that it’s somehow related to glibc.

* * *

All other changes in the first commit are really not for better.

artemklevtsov:
> What exactly does this flag do? I can’t find any information about it, just that it’s somehow related to glibc.
I think this would be helpful: http://stackoverflow.com/a/29201732/1863950

PR #261, opened 2016-09-03
community/xpra: update to 0.17.5

itoffshore:
community/xpra: update to 0.17.5


community/xpra/APKBUILD
```diff
@@ -27,7 +28,7 @@ prepare() {
 }
```
jirutka: You can remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

PR #263, opened 2016-09-03
community/ffmpegthumbnailer: fix $builddir

itoffshore:
changes `$_builddir` -> `$builddir`


itoffshore:
this has already been committed - closing

community/ffmpegthumbnailer/APKBUILD
```diff
@@ -23,13 +23,13 @@ prepare() {
 }
```
jirutka: You can also remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

PR #265, opened 2016-09-03
main/unbound: improve update-unbound-root-hints

itoffshore:
`nf_conntrack_ftp` is needed for passive `ftp` to work through
`iptables` with a default deny policy.


ncopa:
passive ftp is not an unbound specific issue so I wonder why this is fixed in unbound script.

I'd prefer a comment that explains how to fix it properly (add it to `/etc/modules`)

ncopa:
Please rebase fixes in the future. I did it for you this time.

(I did `git rebase -i` and changed the 2 last commits as `fixup`)

PR #266, opened 2016-09-03
community/autossh: update README.alpine

itoffshore:
community/autossh: update README.alpine


community/autossh/APKBUILD
```diff
@@ -27,7 +27,7 @@ prepare() {
 }
```
jirutka: You should also remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

PR #269, opened 2016-09-03
community/syslog-summary: new aport

itoffshore:
also includes it's dependent library: `py-magic`

### DIFF
--- a/community/psad/APKBUILD
+++ b/community/psad/APKBUILD
@@ -7,91 +7,96 @@ pkgdesc="3 lightweight system daemons that analyze iptables log messages to dete
 url="http://cipherdyne.org/psad/"
 arch="all"
 license="GPL"
-depends="perl iptables ip6tables ssmtp mailx psmisc perl-bit-vector perl-date-calc perl-iptables-chainmgr \
-		perl-iptables-parse perl-net-ipv4addr perl-unix-syslog net-tools whois"
+depends="perl iptables ip6tables ssmtp mailx psmisc perl-bit-vector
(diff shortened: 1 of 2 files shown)


community/syslog-summary/APKBUILD
```diff
@@ -0,0 +1,55 @@
...
+
+prepare() {
+        local i
+	cd "$builddir"
+        for i in $source; do
+                case $i in
+                *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+                esac
+        done
+}
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

community/py-magic/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+pkgver=0.4.12
+pkgrel=0
+pkgdesc="A python wrapper for libmagic"
+url="https://pypi.python.org/pypi/python-magic/"
+arch="noarch"
+license="MIT"
+depends="python libmagic"
+makedepends="python2-dev python3-dev py-setuptools"
+install=""
+options="!strip"
```
jirutka: Why this package needs stripping to be disabled?

itoffshore: no point in stripping symbols from python scripts

jirutka: If it does not contain any binaries, then stripping is not performed anyway; I don’t see a point in explicit disable.

PR #270, opened 2016-09-03
community/logcheck: add busybox compatibility / move to community

itoffshore:
I've been running `logcheck` for 6 weeks with `syslog-summary` (https://github.com/alpinelinux/aports/pull/269) without any problems. Tested with both busybox `syslog` & `socklog`.


community/logcheck/logcheck.cron.sample.d
```diff
@@ -0,0 +1,35 @@
...
+ 
+@reboot         logcheck    if [ -x /usr/sbin/logcheck ]; then nice -n10 /usr/sbin/logcheck -R; fi
+2 * * * *       logcheck    if [ -x /usr/sbin/logcheck ]; then nice -n10 /usr/sbin/logcheck; fi
+
+# EOF
+
+
+## busybox crond reboot script [ /etc/periodic/reboot/logcheck ]
+
+#!/bin/sh
```
jirutka: How do these script snippets inside cron file work? I’ve never seen it.

community/perl-ipc-signal/APKBUILD
```diff
@@ -1,37 +1,33 @@
...
-makedepends="perl-dev $cpanmakedepends"
+depends=""
+makedepends="perl-dev"
 subpackages="$pkgname-doc"
 source="http://search.cpan.org/CPAN/authors/id/R/RO/ROSCH/$_pkgreal-$pkgver.tar.gz"
-
-_builddir="$srcdir/$_pkgreal-$pkgver"
+builddir="$srcdir/$_pkgreal-$pkgver"
 
 prepare() {
```
jirutka: Please add `default_prepare || return 1` at top of custom `prepare` function.

community/perl-mime-construct/APKBUILD
```diff
@@ -1,36 +1,34 @@
...
 depends="perl $cpandepends"
-makedepends="perl-dev $cpanmakedepends"
+makedepends="perl-dev"
 subpackages="$pkgname-doc"
 source="http://search.cpan.org/CPAN/authors/id/R/RO/ROSCH/$_pkgreal-$pkgver.tar.gz"
-
-_builddir="$srcdir/$_pkgreal-$pkgver"
+builddir="$srcdir/$_pkgreal-$pkgver"
 
 prepare() {
```
jirutka: Please add `default_prepare || return 1` at top of custom `prepare` function.

community/perl-proc-waitstat/APKBUILD
```diff
@@ -1,37 +1,34 @@
...
 depends="$cpandepends"
-makedepends="perl-dev $cpanmakedepends"
+makedepends="perl-dev"
 subpackages="$pkgname-doc"
 source="http://search.cpan.org/CPAN/authors/id/R/RO/ROSCH/$_pkgreal-$pkgver.tar.gz"
-
-_builddir="$srcdir/$_pkgreal-$pkgver"
+builddir="$srcdir/$_pkgreal-$pkgver"
 
 prepare() {
```
jirutka: Please add `default_prepare || return 1` at top of custom `prepare` function.

community/logcheck/logcheck.cron.sample.d
itoffshore: Those lines are from the original sample by @ncopa - I think they may work with another cron (not busybox `crond`).

Let me know if you want if removed - my busybox version has been running ok for weeks now

jirutka: I’d just like to know how it works, when it is executed; I’m curious and can’t figure out right keywords to google it. :)

If it doesn’t work on Busybox crond, then it should be mentioned in the comment.

itoffshore:
`default_prepare` added to the 3 `perl` pkgs. 

I also changed the `cron` script / example a little so they work with `linux-pam`.

itoffshore:
this no longer uses the busybox initscript change that was not accepted. 

the sample file shows an `/etc/local.d` script

PR #271, opened 2016-09-03
main/busybox-initscripts: add @reboot support

itoffshore:
adds @reboot support to busybox crond if enabled in it's confd.


jirutka:
@ncopa, what do you think about this patch, is it okay?

ncopa:
I'm sceptic.

Why is something `reboot` stored under `periodic`? Are reboots expected to be periodic?

Why does something called `reboot` happens on all boot not only  *re*boot?

Why should something you have marked to be run on reboots run if you restart `crond` service?

How is this different from `/etc/init.d/local` which seems to do the same thing, run custom scripts at startup?

Are there other distros doing something similar?

itoffshore:
altered the `initd` slightly to give the [same functionality in Debian's `crond`](http://www.unixdaemon.net/linux/how-does-cron-reboot-work/) to only run on reboots.

jirutka:
This can’t work, `/run` is tmpfs, so it’s empty at each startup. You can’t distinguish between boot and *re*boot using this approach.

itoffshore:
it is because `/run` is a `tmpfs` that it does work (run once at startup)

jirutka:
> … (run once at startup) 

→ “Why does something called `reboot` happens on all boot not only  *re*boot?”

itoffshore:
```
 string         meaning
 ------         -------
 @reboot        Run once, at startup.
 @yearly        Run once a year, "0 0 1 1 *".
 @annually      (same as @yearly)
 @monthly       Run once a month, "0 0 1 * *".
 @weekly        Run once a week, "0 0 * * 0".
 @daily         Run once a day, "0 0 * * *".
 @midnight      (same as @daily)
 @hourly        Run once an hour, "0 * * * *".
```

PR #277, opened 2016-09-05
community/modemmanager: update to 1.6.0

itoffshore:
also updates `libqmi` & `libmbim`


community/libmbim/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=libmbim
-pkgver=1.12.4
+pkgver=1.6.0
```
jirutka: Ehm, this not an upgrade, but downgrade… ;)

itoffshore: well spotted - `libmbim` & `libqmi` both fixed

PR #278, opened 2016-09-05
community/runit: improve initd

itoffshore:
if `socklog` is detected this prevents busybox `syslog` starting even
when disabled (due to `crond` needing `logger`).

### DIFF
--- a/community/runit/runit.initd
+++ b/community/runit/runit.initd
@@ -8,7 +8,10 @@ description="starts and monitors a collection of runsv(8) processes"
 retry="SIGHUP/5"
 
 depend() {
-    need localmount
-    after firewall
+	need localmount
+	after firewall
+	if [ -x /sbin/socklog ]; then
+		provide logger
+	fi
 }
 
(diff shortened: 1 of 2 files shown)


community/runit/runit.initd
```diff
@@ -8,7 +8,10 @@ description="starts and monitors a collection of runsv(8) processes"
 retry="SIGHUP/5"
 
 depend() {
-    need localmount
-    after firewall
+	need localmount
+	after firewall
+	if [ -x /sbin/socklog ]; then
+		provide logger
+	fi
```
jirutka: I don’t understand this. `/sbin/socklog` is provided by the package socklog, not runit, and runit does not depend on socklog, so what does it do here?

itoffshore: `runit` runs the `socklog` service (`socklog` does not have an initd)

without this in the `runit` initd - busybox's syslog will be started (even if it is disabled in `rc-update`) - because `crond` requires `logger` & the only thing that provides it is busybox syslog `initd`

if `/etc/init.d/runitd` provides `logger` (if `socklog` is detected) - busybox syslog correctly does not get started by `openrc`.

jirutka: Aha, I understand it now. Thanks for clarification.

PR #280, opened 2016-09-06
testing/rspamd: update to 1.3.5 + config fixes

fatalbanana:
- Update to 1.3.5
- /var/lib/rspamd/rspamd_dynamic is a _file_ that might be created by web interface so do not precreate it as a directory.
- Changing path to control socket breaks rspamadm control commands: don't do it.
- Perl is a build dependency
- Hiredis is _bundled_ so is not a dependency (it's patched)

I've fixed my edits now. ;) Feedback welcome.

### DIFF
--- a/testing/rspamd/APKBUILD
+++ b/testing/rspamd/APKBUILD
@@ -1,9 +1,10 @@
 # Maintainer: Valery Kartel <valery.kartel@gmail.com>
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Contributor: Nathan Angelacos <nangel@alpinelinux.org>
+# Contributor: Andrew Lewis <nerf@judo.za.org>
 pkgname=rspamd
-pkgver=1.3.4
-pkgrel=1
+pkgver=1.3.5
+pkgrel=0
 pkgdesc="Rapid spam filtering system"
 url="https://rspamd.com"
 arch="x86_64 x86 armhf"
@@ -13,7 +14,7 @@ pkggroups="rspamd"
 depends=""
 depends_dev=""
 makedepends="$depends_dev cmake libressl-dev libevent-dev glib-dev gmime-dev
-	lua5.1-dev lua5.1 sqlite-dev hiredis-dev file-dev pcre-dev ragel"
+	lua5.1-dev lua5.1 sqlite-dev perl file-dev pcre-dev ragel"
 install="$pkgname.pre-install"
 subpackages="$pkgname-doc $pkgname-web $pkgname-client"
 source="https://rspamd.com/downloads/$pkgname-$pkgver.tar.xz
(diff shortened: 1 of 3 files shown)


fatalbanana:
I've updated the PR to fix the custom config rather than blowing it away in hopes this may be less contentious (not that I've picked up any hints of contention) & so make it easier to merge the rest of the changes. I believe the custom config is of dubious usefulness and will create problems in the long-run/is already doing as much, but, well, it's there now, so maybe it's not so nice to get rid of it. Assigning controller config to web interface seems logically wrong & I understand this to be the whole point of the custom config & the -web package.

Feedback welcome.

jirutka:
> Hiredis is _bundled_ so is not a dependency

That’s not good. Is there any reason why it _must_ be bundled (i.e. they rely on some custom patches)? Could you remove it and use system-provided hiredis?

fatalbanana:
Reason hiredis is bundled is because system hiredis on platforms like Debian is too old. There is currently no possibility to use unbundled copy though that should be doable.

jirutka:
> There is currently no possibility to use unbundled copy ~~though that should be doable~~ (actually it is patched).

I don’t understand… it is patched for what? To allow, or disallow using system-provided dependency?

testing/rspamd/APKBUILD
```diff
@@ -35,6 +35,7 @@ build() {
 		-DRSPAMD_USER=$pkgusers \
 		-DRSPAMD_GROUP=$pkggroups \
 		-DENABLE_HIREDIS=ON \
+		-DENABLE_LUAJIT=OFF \
```
jirutka: Wouldn’t it be better to replace lua5.1 with luajit for this package?

jirutka:
What are the differences between default and our custom configs? Could you please paste the default config here?

testing/rspamd/APKBUILD
fatalbanana: Maybe not on Alpine. :) https://github.com/alpinelinux/aports/commit/8bdc98d68351be28bf94e42b1fc6aa62cfe7a206#diff-0faa079b2a5bf9f91d71fe9aef8837f6

jirutka: Hm, it seems that @ncopa didn’t want to disable PaX memory protection for rspamd for some reason…?

fatalbanana:
Default config is here: https://github.com/vstakhov/rspamd/blob/master/conf/rspamd.sysvinit.conf

Difference is only in how worker configuration is included into config (& in include option which is updated in this PR).

testing/rspamd/APKBUILD
fatalbanana: "for server software we prefer to have PaX" :)

jirutka: This is not a binary option, it’s possible to disable specific PaX protections for specific binaries. We do it even for nginx.

fatalbanana:
Rebased to resolve conflicts.

PR #282, opened 2016-09-07
community/salt: Update and refactoring based on official SRPM

coredumb:


community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
 builddir="$srcdir"/$pkgname-$pkgver
+prepare() {
+    local i
+    cd "$builddir"
+    for i in $source; do
+        case $i in
+        *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+        esac
+    done
+}
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
+324e687e8b0eaccb799b4a719448341a4e09cda3add102890beddaf371ac42957bbc1622f145b2b86420061655bb27f27879302b5b34c6eee1987d2dcae6b8ef  salt-master.initd
 0051e13351cef8db81dc075a194bb384723f07d5591b5b4d3f3adf4180afaf7beced470ab79ceca9d1ec0dae62dbd72084eb76af009fc78411a011050a94a2ed  salt-minion.confd
+c6634a592c6f3e65dd2b704cb500486bf8004c5b287d4a4d42b09af36ef129c59d6a89f005af058cf7911e8587d927b3db931186569084f13ebaca56f6ef93a0  salt-minion.initd
+bafc6ea10cdafd0aef868feb35aecbe4ae6a7dff0ae42862bded85715ad763eb89e1ed27437866a7e5f2b9f7064e3c2a3fb59814487744ba4227238d95cf3818  salt-syndic.confd
 d71133e834685304e0167554035ebbc861252f972bbe981cc71e45b70f15d94a28a02a369463c9a641372919689f96b62a0408b14f824ad986d536e52b1e5ec0  salt-syndic.initd
-bafc6ea10cdafd0aef868feb35aecbe4ae6a7dff0ae42862bded85715ad763eb89e1ed27437866a7e5f2b9f7064e3c2a3fb59814487744ba4227238d95cf3818  salt-syndic.confd"
+2ae35d146ee44154ed8ad910fe0afd9854bf7ded21683e0d06edcae4f194823e172a21147fec5b21c967efcfcf5d1d05cf4696c71fe32cf6ba5c49c4e0b643be  0001-alpine-support.patch"
```
jirutka: APKBUILDs should use tabs for indentation.
Tip: Use [editorconfig](http://editorconfig.org/).

community/salt/0001-alpine-support.patch
```diff
@@ -0,0 +1,601 @@
...
++++ b/salt/modules/gentoo_service.py
+@@ -31,6 +31,8 @@ def __virtual__():
+     '''
+     if __grains__['os'] == 'Gentoo' and not salt.utils.systemd.booted(__context__):
+         return __virtualname__
++    if __grains__['os'] == 'Alpine':
++        return __virtualname__
+     return (False, 'The gentoo_service execution module cannot be loaded: '
+             'only available on Gentoo/Open-RC systems.')
+
```
jirutka: Have you sent this patch to upstream?

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
+master() {
+    pkgdesc="Management component for salt, a parallel remote execution system"
+    depends="$pkgname"
+
+    init_copy master
+    conf_copy master
+
+    mkdir -p ${subpkgdir}/usr/bin || return 1
+    mv ${pkgdir}/usr/bin/salt ${subpkgdir}/usr/bin || return 1
+    for i in cp key master run unity
```
jirutka: Local variables inside functions should be always declared as `local` (e.g. `local i; for i in …`).

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
+    cd "$builddir"
+    python2 setup.py install --root=$pkgdir/ --optimize=1 || return 1
+    mkdir -p ${pkgdir}/var/log/$pkgname || return 1
 }
 
-md5sums="27a9bb34c205ffe07e95b316e2131177  salt-2016.3.2.tar.gz
-0b40fd1b49ff4c2f34382fd69c341c34  salt-master.initd
+init_copy() {
+    type="$1"
+    mkdir -p ${subpkgdir}/etc/init.d/ || return 1
```
jirutka: Please always quote variables that may possible contain spaces, like `srcdir`, `pkgdir` and `subpkgdir`.

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
-	cp ${srcdir}/${pkgname}-${pkgver}/conf/master ${pkgdir}/etc/salt/
-	cp ${srcdir}/${pkgname}-${pkgver}/conf/minion ${pkgdir}/etc/salt/
+    cd "$builddir"
+    python2 setup.py install --root=$pkgdir/ --optimize=1 || return 1
+    mkdir -p ${pkgdir}/var/log/$pkgname || return 1
 }
 
-md5sums="27a9bb34c205ffe07e95b316e2131177  salt-2016.3.2.tar.gz
-0b40fd1b49ff4c2f34382fd69c341c34  salt-master.initd
+init_copy() {
```
jirutka: Helper functions should be prefixed with underscore (e.g. `_init_copy`).

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
+    type="$1"
+    mkdir -p ${subpkgdir}/etc/init.d/ || return 1
+    mkdir -p ${subpkgdir}/etc/conf.d/ || return 1
+
+    cp ${srcdir}/salt-$type.initd ${subpkgdir}/etc/init.d/salt-$type || return 1
+    cp ${srcdir}/salt-$type.confd ${subpkgdir}/etc/conf.d/salt-$type || return 1
+
+    chmod +x ${subpkgdir}/etc/init.d/salt-$type || return 1
```
jirutka: These lines:

```sh
mkdir -p ${subpkgdir}/etc/init.d/ || return 1
cp ${srcdir}/salt-$type.initd ${subpkgdir}/etc/init.d/salt-$type
chmod +x ${subpkgdir}/etc/init.d/salt-$type
```

should be replaces by more convenient `install -m 755 -D "$srcdir"/salt-$type.initd "$subpkgdir"/init.d/salt-$type`.

The same applies for confd.

community/salt/APKBUILD
```diff
@@ -1,67 +1,166 @@
...
+        mv ${pkgdir}/usr/bin/salt-$i ${subpkgdir}/usr/bin || return 1
+    done
+}
+
+syndic() {
+    pkgdesc="Master-of-master component for Salt, a parallel remote execution system"
+    depends="$pkgname-master"
+
+    mkdir -p ${subpkgdir}/usr/bin || return 1
+    mv ${pkgdir}/usr/bin/salt-syndic ${subpkgdir}/usr/bin || return 1
```
jirutka: These two lines can be replaced by `install -m 755 -D "$pkgdir"/usr/bin/salt-syndic "$subpkgdir"/usr/bin/salt-syndic`.

community/salt/0001-alpine-support.patch
jirutka: Great! 👍 Please add link to the PR into the patch’s header.

coredumb:
Should be all good now

community/salt/APKBUILD
jirutka: Mea culpa, I’ve been probably too tired when writing this comment. These two lines _cannot_ be replaces by `install …`, because `install` copies files, but we need to move them here.

jirutka:
There’s no point in separating `syndic`, `api`, and `ssh` into subpackages, at least not in this way. They contain only single tiny file that looks like this:

```sh
#!/usr/bin/python2
from salt.scripts import salt_syndic
if __name__ == '__main__':
    salt_syndic()
```

…one tiny config file and maybe a runscript. They all depends only on salt or salt-master, so it does not reduce dependencies either. Thus there’s no benefit.

Moreover, salt package is quite huge, 46.6 MiB huge, so even benefit of other subpackages (except -doc) with few dozens of kilobytes is questionable. Think about confused users that install salt and find out that something is missing here.

coredumb:
I kinda disagree... "apk search salt" will show all the sub packages... Aren't users looking at what's available before installing a package? I don't think there's a need to worry about confusion here.

Now about the need for separation - apart that the official packages are split like that - it's alway nice to not have things that are not needed isn't it? It's about functionalities more than saving space or whatever.
Why would I want the binary and config file of a functionality I don't use ?

same point for all the subpackages, it's a functionality split not a size split

coredumb:
As py2-libcloud is still in testing, it has not been included and thus the whole salt-cloud shouldn't even work... Would it be better to remove it completely ?

jirutka:
> the official packages are split like that

Aha, this completely changes the situation! However, what official packages do you mean?

coredumb:
https://repo.saltstack.com/yum/redhat/latest/x86_64/2016.3/SRPMS/salt-2016.3.3-2.el7.src.rpm 
From their repo

jirutka:
> As py2-libcloud is still in testing…

So let’s move it to the community.

> it's a functionality split not a size split

The problem is that the functionality is de facto still here, you just removes the entry point. It’s quite a fraud in my eyes.

However, I didn’t know that it’s the upstream’s decision to split them like this. I was just afraid that it’s a single package on other systems and so splitting it would confuse people without any real benefit (like when Ubuntu removes virtualenv from Python 3.x, that’s directly against upstream’s will and it basically breaks the expected behaviour).

So after your explanation I have no problem with the subpackages. I’ve checked also Fedora’s package and they do it in the same way (even with syndic subpackage).

community/salt/APKBUILD
```diff
@@ -1,67 +1,146 @@
...
 arch="noarch"
 license="ASL 2.0"
-depends="py-tornado py2-markupsafe py2-futures" 
+depends="py-tornado py2-yaml py-jinja2 py2-markupsafe py2-msgpack py2-crypto py2-zmq
+	py2-six py2-requests py2-futures procps"
 depends_dev=""
-makedepends="$depends_dev python2-dev py-zmq
-	py2-crypto py2-psutil py-jinja2 py-yaml py-msgpack"
+makedepends="$depends_dev python2-dev py-tornado py2-yaml py-jinja2 
+	py2-markupsafe py2-msgpack py2-crypto py2-zmq py2-requests py2-futures"
```
jirutka: `makedepends` should not repeat dependencies from `depends`, this is implicit. The original package was correct.

community/salt/APKBUILD
```diff
@@ -1,67 +1,146 @@
...
-	cp ${srcdir}/${pkgname}-${pkgver}/conf/master ${pkgdir}/etc/salt/
-	cp ${srcdir}/${pkgname}-${pkgver}/conf/minion ${pkgdir}/etc/salt/
+	python2 setup.py install --root="$pkgdir"/ --optimize=1 || return 1
+	mkdir -p "${pkgdir}"/var/log/$pkgname || return 1
 }
 
-md5sums="27a9bb34c205ffe07e95b316e2131177  salt-2016.3.2.tar.gz
-0b40fd1b49ff4c2f34382fd69c341c34  salt-master.initd
+_init_copy() {
+	type="$1"
```
jirutka: `local`…

PR #283, opened 2016-09-08
community/syslog-summary: new aport

itoffshore:
also includes it's dependent library: `py-magic` (without `!strip`)

### DIFF
--- a/community/py-magic/APKBUILD
+++ b/community/py-magic/APKBUILD
@@ -1,0 +1,51 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-magic
+_pkgname=${pkgname#py-*}
+_pkgreal=${pkgname/py/python}
+pkgver=0.4.12
+pkgrel=0
+pkgdesc="A python wrapper for libmagic"
+url="https://pypi.python.org/pypi/python-magic/"
+arch="noarch"
+license="MIT"
+depends="libmagic"
+makedepends="python2-dev python3-dev py-setuptools"
+subpackages="py3-${pkgname/py-/}:_py3 py2-${pkgname/py-/}:_py2"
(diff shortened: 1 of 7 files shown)


jirutka:
New aports should always go to the testing repository first.

itoffshore:
Ordinarily yes - but I've been using it on 2 servers + 8 containers for 3 weeks with no problems on both busybox `syslog` & `socklog`.

It's pretty well tested ;o)

jirutka:
Okay then.

community/py-magic/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+pkgname=py-magic
+_pkgname=${pkgname#py-*}
+_pkgreal=${pkgname/py/python}
+pkgver=0.4.12
+pkgrel=0
+pkgdesc="A python wrapper for libmagic"
+url="https://pypi.python.org/pypi/python-magic/"
+arch="noarch"
+license="MIT"
+depends="python libmagic"
```
jirutka: `python` is old name of the package `python2`, they are de facto synonyms, so this would make py3-magic depend on python2 that doesn’t make sense. Please look at https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python.

itoffshore: well spotted - s/be fixed now

community/syslog-summary/APKBUILD
```diff
@@ -0,0 +1,43 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=syslog-summary
+pkgver=1.14
+pkgrel=0
+pkgdesc="Summarize the contents of a syslog log file / useful with logcheck"
+url="https://packages.debian.org/sid/syslog-summary"
```
jirutka: This doesn’t look like the upstream, it’s a package for https://github.com/dpaleino/syslog-summary. Why do you download it from Debian instead of the upstream repository?

itoffshore: I didn't realise it was on github - s/be fixed now

PR #286, opened 2016-09-12
testing/php7-memcached: fixed to use static source

aleksijohansson:
Introduced _pkgcommit variable to tie the source to a specific commit instead of rolling branch as it was. This should provide a fix requested in https://github.com/alpinelinux/aports/commit/3f354e53c053b41e775a8a88d7b1bef9cec205f6. Updated the source and checksums to the latest available.


testing/php7-memcached/APKBUILD
```diff
@@ -15,7 +16,7 @@ pecldepends="php7-dev autoconf zlib-dev"
 makedepends="$pecldepends libmemcached-dev"
 install=""
 subpackages=""
-source="$pkgname-$pkgver.tar.gz::https://github.com/$_pkgrepo-dev/$_pkgrepo/archive/$_pkgbranch.tar.gz"
+source="$pkgname-$pkgver.tar.gz::https://github.com/$_pkgrepo-dev/$_pkgrepo/archive/$_pkgcommit.tar.gz"
 
 builddir="$srcdir"/$_pkgrepo-$_pkgbranch
```
Moeser: shouldn't the builddir also match the commit package?  I think that's why travis builds are failing.  try changing $_pkgbranch to $_pkgcommit on the builddir= line

Moeser:
Also it looks like the php7 package was moved out of testing/ and into community/

pcambra:
Sorry if this is an out of the blue question, but why has the php7-memcached been moved to unmaintained? I'm using it in my builds atm and not sure how to workaround the issue.

ncopa:
> @pcambra Sorry if this is an out of the blue question, but why has the php7-memcached been moved to unmaintained?

Did you know that with git you can add comments to commits? With the command `git log -- path/to/relevant` you can get a list of the commits and the comments.

If you are uncomfortable with the command line you can also use the web interface in github. Browse to the directory you are interested in (eg [unmaintained/php7-memcached](https://github.com/alpinelinux/aports/tree/master/unmaintained/php7-memcached)) up to the right you have a button named [history](https://github.com/alpinelinux/aports/commits/master/unmaintained/php7-memcached). When you click this you will get the history of the directory. From there you can click on the commit 3f354e53. You will find the answer there about who did it and why it was done.

I hope that helps.

ncopa:
I think it would be nice if https://github.com/php-memcached-dev/php-memcached/issues/225 was solved before we move it back to testing.

pcambra:
@ncopa I know about git history and the issue is linked above. I am just not familiar with the alpine process, I just build my images with it so I lack understanding about the reasons on why this was the case. It'd seem that there's something missing in php7-memcached itself that need to be corrected, is that the reason?

ncopa:
@pcambra I think what happened was that the aarch64 build broke because when it tried to download the source it got the latest git which did not correspond with the checksum in the APKBUILD. @clandmeter then moved it to "unmaintained" because it got in the way for the build server.

The suggested fix (to specify a given git commit) will solve the checksum issue but I don't think it is a good solution since there is no clear definition what version `3.0.0b1` really is. (Imagine a future bug report saying `php7-memcached-3.0.0b1` has a problem, and someone else claims it cannot be reproduced on version `3.0.0b1` - because they happen to have a newer snapshot)

So i think upstream need to create a tag, even if they don't make proper release, as php-memcached-dev/php-memcached#225 suggests

Moeser:
> @pcambra I am just not familiar with the alpine process, I just build my images with it

I'm on this comment thread because lack of php7-memcached was blocking some docker image builds.  Your comment makes me think you're here for the same reason.  I worked around it by just compiling php7-memcached during the docker build, instead of installing from the package.  Here's an example, in case it helps you work around the problem in the meantime: https://gist.github.com/Moeser/e2dcd64b0e07076b36f8337530d731f3

@ncopa If this merge resolves the floating commit issue, does that mean php7-memcached gets moved back to the testing/ repo?  I ask because there's still some php7-* packages in testing/ (as of this writing), but the base php7 package already moved to community/.  Will the php7 packages continue to be spread across multiple repos?  What's preventing php7 and related packages from entering the main repo?

pcambra:
Thanks @Moeser, I was working on that too, here's my version :) https://gist.github.com/pcambra/2ce7914e8e7099ee4c30b681753fe8b4

aleksijohansson:
Wow, it has been active here. Thanks @pcambra and @Moeser for the workarounds. I'll fix the PR to not fail while building. I'm really hoping we could then move php7-memcached back to testing even though I agree that it would be best to have a tag from upstream as @ncopa pointed out.

aleksijohansson:
> The suggested fix (to specify a given git commit) will solve the checksum issue but I don't think it is a good solution since there is no clear definition what version 3.0.0b1 really is. (Imagine a future bug report saying php7-memcached-3.0.0b1 has a problem, and someone else claims it cannot be reproduced on version 3.0.0b1 - because they happen to have a newer snapshot)

I don't think this is entirely true if we fix it to a specified commit and stick with it. And isn't that what the testing repo is for, testing?

testing/php7-memcached/APKBUILD
aleksijohansson: You're absolutely right, thanks!

aleksijohansson: This is now fixed.

andypost:
Looks sticking on commit is a only way now

testing/php7-memcached/APKBUILD
```diff
@@ -4,6 +4,7 @@ pkgname=php7-memcached
 _pkgreal=memcached
 _pkgrepo=php-memcached
 _pkgbranch=php7
+_pkgcommit=583ecd68faec886ac9233277531f78fb6e2043c7
 pkgver=3.0.0b1
```
jirutka: This is a snapshot, so the version should be something like `3.0_pre20160923`. And you should also try to request a release from the upstream, at least open an issue on their issue tracking…

aleksijohansson: Thanks for the input. I'll change the version and update the PR.

aleksijohansson:
@jirutka Just when I was wondering what's wrong. Thanks!

testing/php7-memcached/APKBUILD
```diff
@@ -5,7 +5,7 @@ _pkgreal=memcached
 _pkgrepo=php-memcached
 _pkgbranch=php7
 _pkgcommit=583ecd68faec886ac9233277531f78fb6e2043c7
-pkgver=3.0.0_pre160808
+pkgver=3.0.0_pre1
```
jirutka: Have you read the build log?! https://travis-ci.org/alpinelinux/aports/builds/162148558#L283

aleksijohansson: @jirutka Just found out that the checksums are missing and they were hardcoded to the previous version number format. Fixing now.

testing/php7-memcached/APKBUILD
```diff
@@ -5,7 +5,7 @@ _pkgreal=memcached
 _pkgrepo=php-memcached
 _pkgbranch=php7
 _pkgcommit=583ecd68faec886ac9233277531f78fb6e2043c7
-pkgver=3.0.0pre20160808
+pkgver=3.0.0_pre160808
```
jirutka: I’ve omitted the last digit for the purpose. What if the upstream someday release 3.0.0_beta? Then you will have a problem, because `3.0.0_beta < 3.0.0_pre160808`. But `3.0.0_beta > 3.0_pre160808`.

aleksijohansson: @jirutka Ok, understood. I'll omit the last digit

ncopa:
I'm ok with the `3.0.0_pre<date>`.

testing/php7-memcached/APKBUILD
```diff
@@ -37,6 +37,6 @@ package() {
 	echo "extension=$_pkgreal.so" > "$pkgdir"/etc/php7/conf.d/20_$_pkgreal.ini
 }
 
-md5sums="081d43caf5a5cb4f70348280014e3da9  php7-memcached-3.0.0b1.tar.gz"
-sha256sums="a770003ebc6bb2a4976dd9ebaa9304552575c5ed9e0c998d941bdd087fee106e  php7-memcached-3.0.0b1.tar.gz"
-sha512sums="3e38885bb43aa26a69d101ad890fe73a4f545c76210450da25c067699b06c2ce7c94f900f8956ff38b5d68376bc2fadfd8c18e32a83df907a3437698c1c6ae84  php7-memcached-3.0.0b1.tar.gz"
+md5sums="081d43caf5a5cb4f70348280014e3da9  php7-memcached-$pkgver.tar.gz"
```
jirutka: Checksums are automatically filled by running `abuild checksum`, they are never edited by hand. Have you read https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#source?

aleksijohansson: No I have not, thanks for the info. I'm totally new to Apline packages, just trying to get this thing sorted out.

ncopa:
> @jirutka What if the upstream someday release 3.0.0_beta?

This is ofc a good point, and same applies to `_alpha`.

aleksijohansson:
The build seems to have some issues. Getting the same ones locally. Figuring out why.

aleksijohansson:
This should now be ready to be merged. Package version is now 3.0_pre20160808 to identify the snapshot used to build the package.

aleksijohansson:
@ncopa @jirutka Is there anything more I could do to get this PR forward and merged?

PR #296, opened 2016-09-14
testing/rundeck: new aport (v2.7.2)

jbguerraz:
testing/rundeck: new aport (v2.6.9)

### DIFF
--- a/testing/rundeck/APKBUILD
+++ b/testing/rundeck/APKBUILD
@@ -1,0 +1,76 @@
+# Contributor: Jean-Baptiste Guerraz <jbguerraz@gmail.com>
+# Maintainer: Jean-Baptiste Guerraz <jbguerraz@gmail.com>
+pkgname=rundeck
+pkgver=2.7.2
+pkgrel=0
+pkgdesc="Job scheduler and runbook automation. Enable self-service access to existing scripts and tools."
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+url="http://rundeck.org"
+arch="all"
+license="ASL 2.0"
+depends="openjdk8-jre-base ncurses5-libs findutils"
+makedepends="openjdk8 bash"
+install="$pkgname.pre-install"
+source="
+	$pkgname-$pkgver.tar.gz::https://github.com/rundeck/${pkgname}/archive/v${pkgver}.tar.gz
+	rundeck.initd
+	0001-Fix-paths.patch
+	"
+builddir="$srcdir"/$pkgname-$pkgver
+
+prepare() {
+	cd "$builddir"
+	patch -p0 -i "$srcdir"/0001-Fix-paths.patch || return 1
+}
+
+build() {
+	export JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk
+	export TERM=dumb
+	cd "$builddir"
+	bash -c "$builddir/gradlew build"
+}
+
+package() {
+	cd "$builddir"
+	local jvers=$pkgver-SNAPSHOT
+	install -m755 -D "$srcdir"/$pkgname.initd \
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -d -o $pkgusers -g $pkggroups \
+		"$pkgdir"/var/lib/rundeck/.ssh \
+		"$pkgdir"/var/lib/rundeck/exp/webapp \
+		"$pkgdir"/var/lib/rundeck/data \
+		"$pkgdir"/var/lib/rundeck/work \
+		"$pkgdir"/var/lib/rundeck/logs \
+		"$pkgdir"/var/lib/rundeck/libext \
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/tmp/rundeck/pluginJars \
+		"$pkgdir"/var/log/rundeck \
+		"$pkgdir"/var/rundeck/projects || return 1
+	install -d "$pkgdir"/etc/rundeck || return 1
+	cp rundeck-launcher/rundeck-jetty-server/build/libs/rundeck-jetty-server-"$jvers".jar \
+	   "$pkgdir"/var/lib/rundeck/bootstrap || return 1
+	cp rundeck-launcher/launcher/build/launcher-contents/lib/*.jar \
+	   "$pkgdir"/var/lib/rundeck/bootstrap/ || return 1
+	cp plugins/*-plugin/build/libs/*-plugin-"$jvers".jar \
+	   "$pkgdir"/var/lib/rundeck/libext || return 1
+	cp core/src/main/resources/com/dtolabs/launcher/setup/templates/admin.aclpolicy.template \
+	   "$pkgdir"/etc/rundeck/admin.aclpolicy || return 1
+	cp core/src/main/resources/com/dtolabs/launcher/setup/templates/apitoken.aclpolicy.template \
+	   "$pkgdir"/etc/rundeck/apitoken.aclpolicy || return 1
+	cp -R packaging/root/etc/rundeck/* \
+	   "$pkgdir"/etc/rundeck || return 1
+	unzip -oq rundeckapp/target/rundeck-"$jvers".war -d "$pkgdir"/var/lib/rundeck/exp/webapp/ || return 1
+	ln -s /etc/rundeck/log4j.properties .; mv log4j.properties "$pkgdir"/var/lib/rundeck/exp/webapp/WEB-INF/classes
+}
+md5sums="442308215ea6694acfa246bc2dac7988  rundeck-2.7.2.tar.gz
+bb9144fe807710dce7e2cb11b9178036  rundeck.initd
+39b94911994f2b704651e7eceaf5daee  0001-Fix-paths.patch"
+sha256sums="0c33c5094652c444daa10e6e739adc9eb0bdf1c78542e5dcec26d252141f6f4b  rundeck-2.7.2.tar.gz
+3274697c8c6b96a37d6f446e9c86137d6876b31619df039996eca1b4a18e10d0  rundeck.initd
+282af0d5608eef31407eb76ef8b07671678c3df91e5a84011aff83532dc33a70  0001-Fix-paths.patch"
+sha512sums="2371cb1389a141d86e80b6181a867ed4627a85e6cb170b6668c32c6627185a04e848895555bc3bcb06ebfbfb9ff5b2ee59a4cbe92b5976a4c8f0b9c035dec1d8  rundeck-2.7.2.tar.gz
+21a12c6cbe0a7542c3011b4d997aecf40f7f2bcd91292aae116d2eadfa59fc46cad7624cb69dd69144d650224d61e02a612955fd931f128bdf2acf967c6d00bf  rundeck.initd
+4fb28203d502518c1d4639f72550e4a93b0573bab06f815401951dac4f0cb4f0b474c0cd70491840e3aa521017eb7c964280d1d15d0b259b74825f2566d9ec90  0001-Fix-paths.patch"
--- a/testing/rundeck/rundeck.initd
+++ b/testing/rundeck/rundeck.initd
@@ -1,0 +1,23 @@
+#!/sbin/openrc-run
+
+# Rundeck init.d file for alpine linux.
+
+. /etc/rundeck/profile >/dev/null 2>&1
+
+name="rundeck"
+command="${JAVA_HOME:-/usr}/bin/java"
+command_args="${RDECK_JVM} -cp ${BOOTSTRAP_CP} com.dtolabs.rundeck.RunServer /var/lib/rundeck ${RDECK_HTTP_PORT}"
+command_background="yes"
+pidfile="/run/$name/$name.pid"
+daemonlog="/var/log/$name/$name.service.log"
+start_stop_daemon_args="-1 ${daemonlog} -2 ${daemonlog} --user $name:$name"
+
+depend() {
+	need net
+	after firewall
+}
+
+start_pre() {
+	checkpath -m 700 -o $name:$name -d ${pidfile%/*}
+	checkpath -m 700 -o $name:$name -d /tmp/$name
+}
--- a/testing/rundeck/0001-Fix-paths.patch
+++ b/testing/rundeck/0001-Fix-paths.patch
@@ -1,0 +1,13 @@
+--- packaging/root/etc/rundeck/original.framework.properties
++++ packaging/root/etc/rundeck/framework.properties
+@@ -21,8 +21,8 @@
+
+ framework.projects.dir=/var/rundeck/projects
+ framework.etc.dir=/etc/rundeck
+-framework.var.dir=/var/lib/rundeck/var
+-framework.tmp.dir=/var/lib/rundeck/var/tmp
++framework.var.dir=/var/rundeck
++framework.tmp.dir=/var/tmp/rundeck
+ framework.logs.dir=/var/lib/rundeck/logs
+ framework.libext.dir=/var/lib/rundeck/libext
+
--- a/testing/rundeck/rundeck.pre-install
+++ b/testing/rundeck/rundeck.pre-install
@@ -1,0 +1,4 @@
+#!/bin/sh
+addgroup -S rundeck 2>/dev/null
+adduser -S -D -H -s /bin/false -G rundeck -g rundeck rundeck 2>/dev/null
+exit 0


testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+url="http://rundeck.org"
+arch="all"
+license="ASL 2.0"
+depends="openjdk8 bash ncurses5-libs"
+depends_dev=""
+makedepends="$depends_dev"
+install="$pkgname.pre-install"
+subpackages=""
+source="
+	https://github.com/${pkgname}/${pkgname}/archive/v${pkgver}.zip
```
7heo: could you please prepend `$pkgname-$pkgver::` to that line?

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+pkgname=rundeck
+pkgver=2.6.9
+pkgrel=0
+pkgdesc="Job scheduler and runbook automation. Enable self-service access to existing scripts and tools."
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+url="http://rundeck.org"
+arch="noarch"
+license="ASL 2.0"
+depends="openjdk8 bash ncurses5-libs"
```
jirutka: Does it really need complete JDK? IMO it needs only headless JRE, so `openjdk8-jre-base` would be more appropriate as runtime dependency (and `openjdk8` for build).

jirutka: Why does it need ncurses libs?

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/lib/rundeck/work \
+		"$pkgdir"/var/lib/rundeck/logs \
+		"$pkgdir"/var/lib/rundeck/cli \
+		"$pkgdir"/var/lib/rundeck/libext \
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/lib/rundeck/var/tmp/pluginJars \
+		"$pkgdir"/var/log/rundeck \
+		"$pkgdir"/tmp/rundeck \
```
jirutka: `/tmp` is usually a tmpfs (i.e. it’s cleaned on each restart), so it doesn’t make sense to create it here. This directory should be created by init script.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+	   "$pkgdir"/var/lib/rundeck/libext || return 1
+	cp core/src/main/resources/com/dtolabs/launcher/setup/templates/admin.aclpolicy.template \
+	   "$pkgdir"/etc/rundeck/admin.aclpolicy || return 1
+	cp core/src/main/resources/com/dtolabs/launcher/setup/templates/apitoken.aclpolicy.template \
+	   "$pkgdir"/etc/rundeck/apitoken.aclpolicy || return 1
+	cp -R packaging/root/etc/rundeck/* \
+	   "$pkgdir"/etc/rundeck || return 1
+	find core/src/main/resources/com/dtolabs/rundeck/core/cli/templates ! -name '*.bat' -a -type f \
+	     | sed -e /rd-check/d -e /rd-setup/d \
+	     | xargs -I Z cp Z "$pkgdir"/usr/bin
```
jirutka: This installs files: `dispatch`, `rd-acl`, `rd-jobs`, `rd-project`, `rd-queue`, `run`. The first one and the last one has too general name to be in `/usr/bin`, it’d be better to prefix them by `rd-` as others.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/lib/rundeck/logs \
+		"$pkgdir"/var/lib/rundeck/cli \
+		"$pkgdir"/var/lib/rundeck/libext \
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/lib/rundeck/var/tmp/pluginJars \
+		"$pkgdir"/var/log/rundeck \
+		"$pkgdir"/tmp/rundeck \
+		"$pkgdir"/var/rundeck/projects \
```
jirutka: What’s that directory, why is directly under `/var` and not `/var/lib`?

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -m644 -D "$srcdir"/$pkgname.confd \
+		"$pkgdir"/etc/conf.d/$pkgname || return 1
+	install -d -o $pkgusers -g $pkggroups \
+		"$pkgdir"/usr/bin \
+		"$pkgdir"/var/lib/rundeck/.ssh \
+		"$pkgdir"/var/lib/rundeck/exp/webapp \
+		"$pkgdir"/var/lib/rundeck/data \
+		"$pkgdir"/var/lib/rundeck/work \
+		"$pkgdir"/var/lib/rundeck/logs \
```
jirutka: Logs should be always in `/var/log` (i.e. `/var/log/$pkgname/` in this case). If Rundeck doesn’t allow to change the logs directory, then create a symlink.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/lib/rundeck/cli \
+		"$pkgdir"/var/lib/rundeck/libext \
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/lib/rundeck/var/tmp/pluginJars \
```
jirutka: This should be probably in `/var/tmp/$pkgname`.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/rundeck/projects \
+		"$pkgdir"/etc/rundeck || return 1
+	cp -R core/build/tools/lib/*.jar \
+	   "$pkgdir"/var/lib/rundeck/cli || return 1
+	cp core/build/libs/rundeck-core-"$JVERS".jar \
+	   "$pkgdir"/var/lib/rundeck/cli || return 1
+	cp rundeck-launcher/rundeck-jetty-server/build/libs/rundeck-jetty-server-"$JVERS".jar \
+	   "$pkgdir"/var/lib/rundeck/bootstrap || return 1
+	cp rundeck-launcher/launcher/build/launcher-contents/lib/*.jar \
+	   "$pkgdir"/var/lib/rundeck/bootstrap/ || return 1
```
jirutka: This directory contains `jna-3.2.2.jar` that bundles a native library (`libjnidispatch.so`), compiled against glibc, so it _may_ not work on Alpine correctly. Moreover, this makes the package arch-dependent and limits it only to x86 and x86_64.

There’s a package [java-jna](https://pkgs.alpinelinux.org/package/edge/community/x86_64/java-jna), compiled for all archs, but Rundeck still uses old 3.x version. However, I hope that it may not be a problem to create a package for this old version.

This JAR is also in `/var/lib/rundeck/cli` and `/var/lib/rundeck/exp/webapp/WEB-INF/lib`, but in newer version 4.1.0.

You can read [this article](https://github.com/jruby/jruby/wiki/JRuby-on-Alpine-Linux) for more information about this topic.

testing/rundeck/rundeck.confd
```diff
@@ -0,0 +1,4 @@
+# Rundeck conf.d file for alpine linux
+
+. /etc/rundeck/profile
+daemon_opts="${RDECK_JVM} -cp ${BOOTSTRAP_CP} com.dtolabs.rundeck.RunServer /var/lib/rundeck ${RDECK_HTTP_PORT}"
```
jirutka: These two lines should be definitely in `rundeck.initd`, not here.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+}
+
+package() {
+	cd "$builddir"
+	install -m755 -D "$srcdir"/$pkgname.initd \
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -m644 -D "$srcdir"/$pkgname.confd \
+		"$pkgdir"/etc/conf.d/$pkgname || return 1
+	install -d -o $pkgusers -g $pkggroups \
+		"$pkgdir"/usr/bin \
```
jirutka: `/usr/bin` definitely shouldn’t be owned by `rundeck` and `rundeck`.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/lib/rundeck/cli \
+		"$pkgdir"/var/lib/rundeck/libext \
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/lib/rundeck/var/tmp/pluginJars \
+		"$pkgdir"/var/log/rundeck \
+		"$pkgdir"/tmp/rundeck \
+		"$pkgdir"/var/rundeck/projects \
+		"$pkgdir"/etc/rundeck || return 1
```
jirutka: Is it really needed to be writable for the `rundeck` user?

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+pkgver=2.6.9
+pkgrel=0
+pkgdesc="Job scheduler and runbook automation. Enable self-service access to existing scripts and tools."
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+url="http://rundeck.org"
+arch="noarch"
+license="ASL 2.0"
+depends="openjdk8 bash ncurses5-libs"
+depends_dev=""
```
jbguerraz: Ok. Thx)

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+makedepends="$depends_dev"
+install="$pkgname.pre-install"
+subpackages=""
+source="
+	$pkgname-$pkgver.tar.gz::https://github.com/rundeck/${pkgname}/archive/v${pkgver}.tar.gz
+	rundeck.initd
+	rundeck.confd
+	"
+builddir="$srcdir"/$pkgname-$pkgver
+JVERS=$pkgver-SNAPSHOT
```
jbguerraz: Ok. Thx)

testing/rundeck/APKBUILD
jbguerraz: Ok. Thx)

testing/rundeck/APKBUILD
jbguerraz: This is a place where rundeck store its jobs logs; the rundeck logs are stored in /var/logs/rundeck
FYI I kept it all from the deb/rpm rundeck official packaging stuff https://github.com/rundeck/rundeck/blob/master/packaging/Makefile

Despite that, wondering if I should use $pkgname instead of hardcoded "rundeck" considering the rundeck provided default configuration files uses "rundeck" in their paths (wouldn't be an issue as the package is named rundeck but in absolute theory we could rename the package and by doing so, break the rundeck install) : https://github.com/rundeck/rundeck/blob/master/packaging/root/etc/rundeck/framework.properties

testing/rundeck/APKBUILD
jbguerraz: That also comes from the official rundeck packaging stuff and default configurations : 
https://github.com/rundeck/rundeck/blob/master/packaging/Makefile
https://github.com/rundeck/rundeck/blob/master/packaging/root/etc/rundeck/framework.properties

testing/rundeck/APKBUILD
jbguerraz: Ok. Thx)

testing/rundeck/APKBUILD
jbguerraz: It's where rundeck store the projects definitions : https://github.com/rundeck/rundeck/blob/master/packaging/root/etc/rundeck/framework.properties#L22

testing/rundeck/APKBUILD
jbguerraz: agreed but I don't feel hacking the official rundeck stuff. I would open an issue on their github.

testing/rundeck/rundeck.confd
jbguerraz: Ok. thx)

testing/rundeck/APKBUILD
jbguerraz: Ok. I've updated arch (to "all")

testing/rundeck/APKBUILD
jbguerraz: indeed that is enough! thx! :)

testing/rundeck/APKBUILD
jbguerraz: Considering those tools are now deprecated (as said there : https://github.com/rundeck/rundeck/issues/2081) I'll just remove those lines (71-73)

testing/rundeck/APKBUILD
jbguerraz: nope. should not! updated that.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+		"$pkgdir"/var/lib/rundeck/bootstrap \
+		"$pkgdir"/var/lib/rundeck/libext/cache \
+		"$pkgdir"/var/lib/rundeck/var \
+		"$pkgdir"/var/lib/rundeck/var/tmp/pluginJars \
+		"$pkgdir"/var/log/rundeck \
+		"$pkgdir"/tmp/rundeck \
+		"$pkgdir"/var/rundeck/projects \
+		"$pkgdir"/etc/rundeck || return 1
+	cp -R core/build/tools/lib/*.jar \
+	   "$pkgdir"/var/lib/rundeck/cli || return 1
```
jbguerraz: Agreed. That is also related to the deprecated CLI. removing that directory as well then.

testing/rundeck/APKBUILD
jirutka: > This is a place where rundeck store its jobs logs…

Then it’s okay.

> …wondering if I should use $pkgname instead of hardcoded "rundeck"

This is not needed, especially if changing these paths would also require modifying non-parametrized configs.

testing/rundeck/APKBUILD
jirutka: That doesn’t mean that it’s correct. It’s not uncommon that upstream violates standard directory structure etc. Look at the GitLab for example, they install everything into `/home/git`…

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,68 @@
...
+build() {
+	cd "$builddir"
+	export JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk
+	export TERM=dumb
+	bash -c "$builddir/gradlew build"
+}
+
+package() {
+	cd "$builddir"
+	JVERS=$pkgver-SNAPSHOT
```
jirutka: This should be declared as `local` and preferably lowercase (i.e. `local jvers="$pkgver-SNAPSHOT"`).

testing/rundeck/rundeck.initd
```diff
@@ -0,0 +1,23 @@
+#!/sbin/openrc-run
+
+# Rundeck init.d file for alpine linux.
+
+. /etc/rundeck/profile &>/dev/null
```
jirutka: `&>` is not valid per POSIX-sh, this is bashism.

testing/rundeck/APKBUILD
jbguerraz: Thx. Fixed!

testing/rundeck/APKBUILD
jbguerraz: Ok. Added a patch and changed APK accordingly (/var/rundeck instead of /var/lib/rundeck/var and /tmp/rundeck instead of /var/lib/rundeck/tmp)

testing/rundeck/rundeck.initd
jbguerraz: Fixed :) thx

testing/rundeck/APKBUILD
jbguerraz: Great.

jbguerraz:
Updated to latest version (2.7.1)

andypost:
We using this build and I'd like to get this in
Looks everything in requested changes fixed

7heo:
Aside from the minor change that might be required if my comment is right, LGTM.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,76 @@
...
+url="http://rundeck.org"
+arch="all"
+license="ASL 2.0"
+depends="openjdk8-jre-base ncurses5-libs findutils"
+makedepends="openjdk8 bash"
+install="$pkgname.pre-install"
+source="
+	$pkgname-$pkgver.tar.gz::https://github.com/rundeck/${pkgname}/archive/v${pkgver}.tar.gz
+	rundeck.initd
+	0001-Fix-pathes.patch
```
jbguerraz: I'm such a frog! ;)
Fixed.

jbguerraz:
any news ? :)

jbguerraz:
Updated to latest version (2.7.2)

7heo:
> opened on Sep 14, 2016 by jbguerraz

That will be half a year, in a month. @jirutka Anything left to do here? :)

andypost:
Can we get this in?

jbguerraz:
@jirutka : maybe now ? :D or maybe someone else can review it so we could get it in ?

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,76 @@
...
+# Maintainer: Jean-Baptiste Guerraz <jbguerraz@gmail.com>
+pkgname=rundeck
+pkgver=2.7.2
+pkgrel=0
+pkgdesc="Job scheduler and runbook automation. Enable self-service access to existing scripts and tools."
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+url="http://rundeck.org"
+arch="all"
+license="ASL 2.0"
```
kaniini: license should be Apache-2

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,76 @@
...
+prepare() {
+	cd "$builddir"
+	patch -p0 -i "$srcdir"/0001-Fix-paths.patch || return 1
+}
+
+build() {
+	export JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk
+	export TERM=dumb
+	cd "$builddir"
+	bash -c "$builddir/gradlew build"
```
kaniini: this requires an internet connection to build the package, in violation of alpine policy.

generally Alpine packages are not supposed to download things inside the APKBUILD, and should have the downloaded files declared in `$source` and `$sha512sums` as we download them outside the build environment and share the source files.

it also downloads a lot of packages (200+!), many i know for a fact should be separate from this one, like grails.  additionally, i'd be willing to bet not all the downloaded packages are Apache-2.  we need to solve this policy violation before we can accept the package.

testing/rundeck/rundeck.initd
```diff
@@ -0,0 +1,23 @@
+#!/sbin/openrc-run
+
+# Rundeck init.d file for alpine linux.
+
+. /etc/rundeck/profile >/dev/null 2>&1
```
kaniini: this should be using conf.d settings, not sourcing a separate file in my opinion

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,76 @@
...
+source="
+	$pkgname-$pkgver.tar.gz::https://github.com/rundeck/${pkgname}/archive/v${pkgver}.tar.gz
+	rundeck.initd
+	0001-Fix-paths.patch
+	"
+builddir="$srcdir"/$pkgname-$pkgver
+
+prepare() {
+	cd "$builddir"
+	patch -p0 -i "$srcdir"/0001-Fix-paths.patch || return 1
```
kaniini: patches need to be -p1, and you do not need this prepare function, it will patch it on its own as long as the patch is in `$sources`

testing/rundeck/APKBUILD
kaniini: it also runs unit tests -- these should be postponed until the check() function is called.

testing/rundeck/APKBUILD
```diff
@@ -0,0 +1,76 @@
...
+	bash -c "$builddir/gradlew build"
+}
+
+package() {
+	cd "$builddir"
+	local jvers=$pkgver-SNAPSHOT
+	install -m755 -D "$srcdir"/$pkgname.initd \
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -d -o $pkgusers -g $pkggroups \
+		"$pkgdir"/var/lib/rundeck/.ssh \
```
kaniini: why are we creating a .ssh directory?  it seems sketchy.

testing/rundeck/APKBUILD
kaniini: also the testsuite it runs throws a lot of exceptions...

kaniini:
there's lots to do.  it's downloading hundreds of files during the build process for one, solve that and resubmit for review.

testing/rundeck/APKBUILD
andypost: yep https://github.com/rundeck/rundeck#license

andypost: but alpine suggests to use in this case `ASL 2.0` https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license

kaniini: fair enough.  we used to do it the other way :P

clandmeter:
Closing due to inactivity from contributor.

PR #306, opened 2016-09-15
community/caddy: upgrade to 0.9.1

7heo:
@jirutka as requested ;)


7heo:
@jirutka could you please restart that build? `gopkg.in` was unaccessible during the build, causing it to fail.

PR #308, opened 2016-09-16
testing/cassandra-cpp-driver: new aport

gena01:
Adding cassandra-cpp-driver 2.4.3


gena01:
@ncopa Thanks for merging. Saw you disable arm (which is hard for me to test). There does seem to be a PR open that's like a one liner (if we need to) to get this compiling everywhere (hopefully)

PR #316, opened 2016-09-21
testing/criu: new aport

itoffshore:
A utility for the live checkpoint / restore of `lxc` containers.

Requires the following additional kernel modules to be enabled:

```
CONFIG_CHECKPOINT_RESTORE=y
CONFIG_FHANDLE=y
```

Optional (incremental dumps):

`CONFIG_MEM_SOFT_DIRTY=y`

### DIFF
--- a/testing/criu/APKBUILD
+++ b/testing/criu/APKBUILD
@@ -1,0 +1,49 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=criu
+pkgver=2.6
+pkgrel=0
+pkgdesc="A Checkpoint/Restore utility for Linux in Userspace"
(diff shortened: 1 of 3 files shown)


jirutka:
@ncopa What do you think about enabling `CONFIG_CHECKPOINT_RESTORE` and `CONFIG_FHANDLE` in Alpine kernel?

clandmeter:
Merged as somebody can also use their own kernel with these changes applied.

jirutka:
@itoffshore It fails on armhf, could you please look at it? http://build.alpinelinux.org/buildlogs/build-edge-armhf/testing/criu/criu-2.8-r0.log

itoffshore:
@jirutka - the instructions for [building on `armhf`](https://criu.org/Installation#Non-standard_compilation) do not seem to be current - there are no `deps` or `criu-aarch64` in the sources.

PR #317, opened 2016-09-21
community/numix-themes: update to 2.6.4

itoffshore:
community/numix-themes: update to 2.6.4

### DIFF
--- a/community/numix-themes/APKBUILD
+++ b/community/numix-themes/APKBUILD
@@ -1,30 +1,19 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=numix-themes
 _pkgname=numix-gtk-theme
-pkgver=2.6.1
+pkgver=2.6.4
 pkgrel=0
 pkgdesc="A modern flat theme with a combination of light and dark elements, GTK 2 and 3"
 url="http://shimmerproject.org/project/Numix/"
 arch="noarch"
 license="GPLv3"
-makedepends="$depends_dev"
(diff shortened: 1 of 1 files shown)


community/numix-themes/APKBUILD
```diff
@@ -1,55 +1,45 @@
...
 	local _i
 	shift; shift
 	mkdir -p "$subpkgdir"/usr/share/themes/Numix
 	for _i; do
 		msg "moving $_i"
-		mv "$pkgdir"/usr/share/themes/Numix/$_i  \
+		mv "$pkgdir"/usr/share/themes/Numix/src/$_i  \
 			"$subpkgdir"/usr/share/themes/Numix || return 1
 	done
+	if rmdir "$pkgdir"/usr/share/themes/Numix/src 2>/dev/null; then return 0; fi
```
jirutka: What the hack is the purpose of this?!

itoffshore: it was the only way to remove an empty `src` directory in the top level of `numix-themes`

It cannot be removed until after `_mv()` has finished - so it cannot be removed in `package()`

jirutka: No, I mean the `if` statement, not `rmdir`.

itoffshore: using `rmdir` on it's own would be detected as an error by `abuild` if it failed (which it would several times until everything was moved).

enclosing `rmdir` in an `if` lets the package build.

jirutka: Aha, I understand it now. You actually need to run it after the last split function, before that it will and should fail. That’s actually not bad approach.

Well, the common idiom for this is `rmdir "$pkgdir"/usr/share/themes/Numix/src 2>/dev/null || true`. Also please add a comment explaining that it removes an empty directory on the last run.

PR #325, opened 2016-09-25
community/py-lz4: update to 0.8.2 / create python3 & tests subpkgs

itoffshore:
community/py-lz4: update to 0.8.2 / create python3 & tests subpkgs


community/py-lz4/APKBUILD
```diff
@@ -1,40 +1,53 @@
...
+	install_if="$pkgname=$pkgver-r$pkgrel $python"
 	cd "$builddir"
-	python setup.py install --prefix=/usr --root="$pkgdir" || return 1
+	$python setup.py install --prefix=/usr --root="$subpkgdir"
+}
+
+_py2() {
+	_py python2
+	replaces="$pkgname"
+}
```
jirutka: Please follow [APKBUILD snippet on wiki](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python). `_py python2` is after `replaces=`, as the _last statement_ of the function, for a good reason. If it’s not the last, as in your case, and not followed by `|| return 1`, then it may silently fail (abuild doesn’t run with `set -e` yet). If it’s the _last statement_, as on the wiki, then the `_py()`’s return value is implicitly returned from the `_py2()` function and so if `_py()` returns non-zero status, it halts the abuild.

community/py-lz4/APKBUILD
```diff
@@ -1,40 +1,53 @@
...
 }
 
 package() {
+	mkdir -p "$pkgdir"
+}
+
+_py() {
+	local python=$1
+	pkgdesc="$pkgdesc - $python"
+	depends=$2
```
jirutka: What is `$2` doing here? You never pass the second argument to this function. There should be `depends="$python"`.

itoffshore: fixed

community/py-lz4/APKBUILD
```diff
@@ -1,40 +1,53 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=py-lz4
-pkgver=0.7.0
+pkgver=0.8.2
 _pkgname=${pkgname#py-}
 pkgrel=0
 pkgdesc="LZ4 Bindings for Python"
 url="https://pypi.python.org/pypi/lz4"
 arch="all"
```
jirutka: Metapackage should be `noarch`.

itoffshore: sorry - fixed

PR #328, opened 2016-09-27
testing/packer: new aport

7heo:


testing/packer/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+pkgname=packer
+pkgver=0.10.2
+pkgrel=0
+pkgdesc="Packer is a tool for creating identical machine images for multiple platforms from a single source configuration."
+url="https://www.packer.io/"
+arch="x86_64"
+license="MPL 2.0"
+depends=""
+depends_dev=""
+makedepends="$depends_dev go go-tools"
```
jirutka: This package doesn’t provide -dev subpackage, so `depends_dev` should not be declared here, nor used in `makedepends`. aport tool even warns about this.

7heo: Sorry, took the terraform APKBUILD as a template, maybe I shouldn't have.

7heo: @jirutka Done

testing/packer/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: 7heo <7heo@mail.com>
+# Maintainer: 7heo <7heo@mail.com>
+pkgname=packer
+pkgver=0.10.2
+pkgrel=0
+pkgdesc="Packer is a tool for creating identical machine images for multiple platforms from a single source configuration."
+url="https://www.packer.io/"
+arch="x86_64"
```
jirutka: Why just x86_64, it doesn’t work on other archs?

7heo: From the terraform APKBUILD, same as the other point.

testing/packer/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+makedepends="go go-tools"
+install=""
+subpackages=""
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/mitchellh/$pkgname/archive/v$pkgver.tar.gz"
+
+builddir="$srcdir"/$pkgname-$pkgver
+
+build() {
+	cd $srcdir
+	export GOPATH="$startdir"
```
jirutka: I would personally prefer the approach that @gena01 used in https://github.com/alpinelinux/aports/commit/5bef15666ec397f281fa09fa57b059549902f144#diff-09aa8ce633e2619a7d2132841839fce1R19, it looks less hack-ish for me. WDYT?

jirutka:
@7heo so what about this PR?

PR #338, opened 2016-09-29
community/salt: Fix in pkg.install as well to have a working pkg.late…

coredumb:
…st working state


jirutka:
I’ve merged commit with the same message just a while ago. o.O

coredumb:
pkg.latest_version vs pkg.install. The previous fix was incomplete and this part was overlooked.

jirutka:
Okay. For the next time, please keep title (i.e. the first line) of a commit message ≤ 72 chars long. It’s quite hard to shorten it when I don’t know what it actually means. :)

PR #339, opened 2016-09-29
testing/logcheck-busybox: new aport

itoffshore:
The version of `logcheck` in main requires `bash` GNU `grep` / Debian `run-parts` to work. It also cannot be called from an init script as `openrc` will cause busybox `grep` & `run-parts` to be used by `logcheck`. 

`logcheck-busybox` is patched to use the busybox builtin tools so it works out of the box.

includes:

* `/etc/init.d/logcheck` to provide `@reboot` functionality to `logcheck` with busybox `crond`

* `/etc/periodic/hourly/logcheck`

### DIFF
--- a/testing/logcheck-busybox/00-busybox-compat.patch
+++ b/testing/logcheck-busybox/00-busybox-compat.patch
@@ -1,0 +1,100 @@
+--- logcheck-1.3.17/src/logcheck
++++ logcheck-1.3.17/src/logcheck.new
+@@ -73,7 +73,7 @@
+ CONFFILE="/etc/logcheck/logcheck.conf"
+ STATEDIR="/var/lib/logcheck"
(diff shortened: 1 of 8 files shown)


itoffshore:
I'm going to rewrite logcheck to remove the perl dependencies

itoffshore:
https://github.com/alpinelinux/aports/pull/866 needs to be pulled for this to build.

itoffshore:
have revised & rebased with a much simpler `initd`

itoffshore:
I've been using this patched `logcheck` for 2 years now with no problems in lots of containers.

These patches would be better to apply to the main `logcheck` package. The logcheck package should function with the standard `busybox` tools which it currently does not.

itoffshore:
I will just add these changes to `main/logcheck`

PR #340, opened 2016-09-30
testing/waitforit: new aport

maxcnunes:
https://github.com/maxcnunes/waitforit

A tool written in Golang to wait until an address become available.

http://lists.alpinelinux.org/alpine-aports/3489.html

### DIFF
--- a/testing/waitforit/APKBUILD
+++ b/testing/waitforit/APKBUILD
@@ -1,0 +1,41 @@
+# Contributor: Max Claus Nunes <maxcnunes@gmail.com>
+# Maintainer: Max Claus Nunes <maxcnunes@gmail.com>
+pkgname=waitforit
+pkgver=1.3.2
+pkgrel=0
+pkgdesc="Wait until an address become available"
+url="https://github.com/maxcnunes/waitforit"
+arch="all"
+license="MIT"
+depends=""
+depends_dev=""
+makedepends="$depends_dev go"
+install=""
+subpackages=""
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/maxcnunes/waitforit/archive/v$pkgver.tar.gz"
+
+builddir="${srcdir}/${pkgname}-${pkgver}"
+_godir="${srcdir}/go"
+_gourl="github.com/maxcnunes/waitforit"
+
+build() {
+	mkdir -p ${_godir}/bin ${_godir}/src/github.com/maxcnunes
+	ln -sf ${builddir} ${_godir}/src/github.com/maxcnunes/waitforit
+
+	cd ${_godir}/src/${_gourl}
+	export GOPATH="${_godir}"
+
+	go get -d -v ./...
+	go install -v ${_gourl} || return 1
+}
+
+package() {
+	cd "$_godir"/bin
+
+	install -Dm755 waitforit \
+		"${pkgdir}/usr/bin/waitforit" || return 1
+}
+
+md5sums="1d7d2d25386cd732c03fb63324ba3d15  waitforit-1.3.2.tar.gz"
+sha256sums="8359692a2f63b5a13e831919f6ac51cf98ec7eb7b6e07038566c5ae21f40abe3  waitforit-1.3.2.tar.gz"
+sha512sums="c50933cab19e160ea591cc2ee7793167b918c7c1dd94fcdf6f70c9744e92ce8287fad1ec135a325fc256fa3af8182d104595829c0627fd956f2a5b50ed2155a6  waitforit-1.3.2.tar.gz"


jirutka:
@7heo Another Go stuff here. :/ Could you please review it?

7heo:
`END=$(($(date "+%s+$TIMEOUT"))); while [ $(date "+%s") -lt $END ]; do nc -z -w1 $HOST $PORT && break; done`.

Saved you 158 lines of Go, 41 lines of APKBUILD, and a package to maintain. You're welcome.

maxcnunes:
Sure. Pretty straight forward solution. But waitforit also works with checking over HTTP status.
Also waitforit has a pretty and unique goal while nc does that and much more. There are good and bad points about that. The good point is that it is simple to use it.
**waitforit**
```
waitforit -host=google.com -port=90 -timeout=20
```

**nc**
```
END=$(($(date "+%s+$TIMEOUT"))); while [ $(date "+%s") -lt $END ]; do nc -z -w1 $HOST $PORT && break; done
```
I don't mind maintaining a code with 158 lines if using it is simpler then other solutions. I use waitforit in almost all my Docker containers. So is a only code to maintaining that can be reused everywhere.

7heo:
@maxcnunes that is incorrect. The equivalent of:

    HOST=google.com; PORT=90; TIMEOUT=20; END=$(($(date "+%s+$TIMEOUT"))); while [ $(date "+%s") -lt $END ]; do nc -z -w1 $HOST $PORT && break; done

currently is:

```
sudo apk add go; echo 'package main

import (
	"errors"
	"flag"
	"fmt"
	"log"
	"net"
	"net/http"
	"regexp"
	"strconv"
	"time"
)

// VERSION is definded during the build
var VERSION string
var debug *bool

const regexAddressConn string = `^([a-z]{3,}):\/\/([^:]+):?([0-9]+)?$`
const regexPathAddressConn string = `^([^\/]+)(\/?.*)$`

func logDebug(msg interface{}) {
	if *debug {
		log.Print(msg)
	}
}

// Connection data
type Connection struct {
	Type   string
	Scheme string
	Port   int
	Host   string
	Path   string
}

func buildConn(host string, port int, fullConn string) *Connection {
	if host != "" {
		return &Connection{Type: "tcp", Host: host, Port: port}
	}

	if fullConn == "" {
		return nil
	}

	res := regexp.MustCompile(regexAddressConn).FindAllStringSubmatch(fullConn, -1)[0]
	if len(res) != 4 {
		return nil
	}

	port, err := strconv.Atoi(res[3])
	if err != nil {
		port = 80
	}

	hostAndPath := regexp.MustCompile(regexPathAddressConn).FindAllStringSubmatch(res[2], -1)[0]
	conn := &Connection{
		Type: res[1],
		Port: port,
		Host: hostAndPath[1],
		Path: hostAndPath[2],
	}

	if conn.Type != "tcp" {
		conn.Scheme = conn.Type
		conn.Type = "tcp"
	}

	if conn.Scheme == "https" {
		conn.Port = 443
	}

	return conn
}

func pingTCP(conn *Connection, timeoutSeconds int) error {
	timeout := time.Duration(timeoutSeconds) * time.Second
	start := time.Now()
	address := fmt.Sprintf("%s:%d", conn.Host, conn.Port)
	logDebug("Dial address: " + address)

	for {
		_, err := net.DialTimeout(conn.Type, address, time.Second)
		logDebug("ping TCP")

		if err == nil {
			logDebug("Up")
			return nil
		}

		logDebug("Down")
		logDebug(err)
		if time.Since(start) > timeout {
			return err
		}

		time.Sleep(500 * time.Millisecond)
	}
}

func pingHTTP(conn *Connection, timeoutSeconds int) error {
	timeout := time.Duration(timeoutSeconds) * time.Second
	start := time.Now()
	address := fmt.Sprintf("%s://%s:%d%s", conn.Scheme, conn.Host, conn.Port, conn.Path)
	logDebug("HTTP address: " + address)

	for {
		resp, err := http.Get(address)

		if resp != nil {
			logDebug("ping HTTP " + resp.Status)
		}

		if err == nil && resp.StatusCode < http.StatusInternalServerError {
			return nil
		}

		if time.Since(start) > timeout {
			return errors.New(resp.Status)
		}

		time.Sleep(500 * time.Millisecond)
	}
}

func main() {
	fullConn := flag.String("full-connection", "", "full connection")
	host := flag.String("host", "", "host to connect")
	port := flag.Int("port", 80, "port to connect")
	timeout := flag.Int("timeout", 10, "time to wait until port become available")
	printVersion := flag.Bool("v", false, "show the current version")
	debug = flag.Bool("debug", false, "enable debug")

	flag.Parse()

	if *printVersion {
		fmt.Println("waitforit version " + VERSION)
		return
	}

	conn := buildConn(*host, *port, *fullConn)
	if conn == nil {
		log.Fatal("Invalid connection")
	}

	logDebug("Waiting " + strconv.Itoa(*timeout) + " seconds")
	if err := pingTCP(conn, *timeout); err != nil {
		log.Fatal(err)
	}

	if conn.Scheme != "http" && conn.Scheme != "https" {
		return
	}

	if err := pingHTTP(conn, *timeout); err != nil {
		log.Fatal(err)
	}
}

0' > /tmp/file.go && go build /tmp/file.go -o /usr/local/bin/waitforit && chmod a+x /usr/local/bin/waitforit && waitforit -host=google.com -port=90 -timeout=20
```

And if this PR goes in:

    sudo apk add waitoforit && waitforit -host=google.com -port=90 -timeout=20

This might also pull some dependencies in later, we don't know. It won't work without internet access, for the `apk add` part.

And that is because `nc` is a busybox applet, not a binary. There is nothing to install, no root rights required, no internet access required. Also, if you want to compare invocation to invocation (after adding the function to your .${SHELL}rc, here goes:

nc
---
    waitforit google.com 90 20

go
---
    waitforit -host=google.com -port=90 -timeout=20

maxcnunes:
This tool was originally created to be used on Docker. To wait for external services before start running the process. So this is always installed during the image build. Depend on internet at that time (for most cases) is not a problem. About depending on third parties to build the tool on Alpine. That could be avoided if instead of building the tool during the package installation I could just download the binary from my repository (would still depend on curl ou wget to download, but it is something that already occur in my Dockerfile nowadays). But based in the docs I have read and the others APKBUILD I have seen, it seems is not a common approach.

But if you think it should not be included in the Alpine repository. That is ok, I will just let people know it is not available in the Alpine repository and should be installed through the other approaches: https://github.com/maxcnunes/waitforit#installing-with-a-dockerfile

7heo:
@maxcnunes I perfectly understand the use of it. A shell function will do the same and will be portable, will not require a binary and will work on all UNIXes (and even windows 10 now that it has bash). I think this goes against the philosophy behind the distribution, which is why I am against that PR. Sorry, and thanks for submitting a PR.

albap:
> I think this goes against the philosophy behind the distribution, which is why I am against that PR

Are you serious?
The guy contributed a working APKBUILD for a package he finds useful and wants to maintain and you guys are rejecting because you don't like the software?
But PHP and other stuff remains supported?
You're confusing the community repo with installed-by-default packages.
If he was proposing a PR for replacing openrc with systemd, okay that's against the philosophy behind the distribution.
But his PR is for a community APKBUILD for a software he uses.

I'm also 100% against this package and would never use it myself, but it's ridiculous to not merge his PR because of personal opinions like this.

7heo:
@albap The PR is still open afaik. I just gave my opinion.

clandmeter:
Lets not discuss the usefulness of every app please.

@7heo as you are our go expert, is this apkbuild correctly written to be merged (as im seeing go get here)

Vanuan:
@7heo 
>     HOST=google.com; PORT=90; TIMEOUT=20; END=$(($(date "+%s+$TIMEOUT"))); while [ $(date "+%s") -lt $END ]; do nc -z -w1 $HOST $PORT && break; done

This is not fair. Putting code in one line doesn't make it shorter. Here's a more realistic one:

```
waitforit () {
  HOST=$1
  PORT=$2
  TIME_OUT=($3 || 30)
  END=$(($(date "+%s+$TIME_OUT")))
  while [ $(date "+%s") -lt $END ]
    do nc -z -w1 $HOST $PORT && break
  done
  return $END
}
waitforit google.com 80 20
```

And here's the reality: https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh
Look how many stars does this script have!

But even that doesn't cover all the features the go package provides.
What about checking if http code is 200? How long should be your one liner to provide that?

If `nc` had an option to wait for a port to become open that would've been more fair. But it's not the case.

7heo:
> This is not fair. Putting code in one line doesn't make it shorter.

I don't know. That's how I wrote it. I didn't "make it" that way for the sake of anything else.

> And here's the reality: https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh

I'm not even going to start commenting on that script. Not the place, not the point.

> Look how many stars does this script have!

Popularity isn't a good metric for correctness, and "Doing the right things" ™. [NodeOS](https://github.com/NodeOS/NodeOS) has order of magnitude more likes than Alpine Linux. Yet I'm not going to touch it, even with a ten foot pole.

> What about checking if http code is 200?

[curl](https://curl.haxx.se/) does that pretty well AFAIK.

> How long should be your one liner to provide that?

A few more lines I presume. But not more than a few.

> If `nc` had an option to wait for a port to become open that would've been more fair. But it's not the case.

That is the magic with non-blocking I/O. You can implement the blocking behavior yourself with a loop!

---

So yeah. No matter what arguments you throw at this, it isn't going to make me like it, or even consider it worth existing. Now, I am really busy, I took the time I had this week for this PR answering you, so I will check later if the PR fits our requirements for merging, as @clandmeter requested.

Thanks for your contribution.

Vanuan:
> curl does that pretty well AFAIK.

You see, curl is another binary you need to install. And it's quite heavier than this waitforit binary. Who's minimalist now?

7heo:
> it's quite heavier than this waitforit binary

173.9K vs 4.5M. You should check your facts before next time.

jirutka:
It has been already merged in c03980ca31c00cde5c00e29b4b264e533430d9ca.

PR #341, opened 2016-09-30
main/ipset: improve initd / add conf.d

itoffshore:
The existing initd loads ip list items individually which delays booting by 1-2 minutes using
the ip lists from iblocklist.com (approximately 140,000 ip addresses)

This patch rewrites `initd` & generates / formats a temporary file for `ipset` to read as a file.
It also saves `ipset` rules to a file for reading in `start()` as done by Gentoo.

* `initd` implements all of http://ipset.netfilter.org/tips.html

* sets are now loaded from `ipset.d` files in a fraction of a second.

* adds optional support to manage `iptables` / `ip6tables` rules [ disabled by default ]

* adds optional support to add `iptables` / `ip6tables` logging [ disabled by default ]

* adds a '`load`' option to load any `ipset.d` file not currently loaded.

* adds an '`info`' option to show loaded sets / `iptables` rules.

To view available options run:

`rc-service ipset describe`

### DIFF
--- a/main/ipset/ipset.initd
+++ b/main/ipset/ipset.initd
@@ -1,13 +1,17 @@
 #!/sbin/openrc-run
 # Init script for ipset
-# Copyright (C) 2012 Kaarle Ritvanen
+# Copyright (C) 2016  Stuart Cardall
 # Licensed under the terms of the GPL2
 
 description="Manage IP sets in the Linux kernel"
+description_info="List active IP sets / raw iptables rules"
 description_save="Save firewall IP sets"
-description_reload="Load firewall IP sets"
+description_load="Load inactive ipset.d files"
+description_unload="Unload / remove specific IP set"
+description_reload="Reload all firewall IP sets"
 
-extra_started_commands="save reload"
+extra_commands="info"
+extra_started_commands="save load unload reload"
 
 IPSET=/usr/sbin/ipset
 DIR=/etc/ipset.d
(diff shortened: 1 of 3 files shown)


itoffshore:
you can try the revised [confd](http://it-offshore.co.uk/downloads/ipset.confd) & [initd](http://it-offshore.co.uk/downloads/ipset.initd) with an example [iblocklist](http://it-offshore.co.uk/downloads/iblocklist.tar.xz)

* the existing `ipset` initd will take 5 minutes to reload the list - my revised script takes 1 second
* implements http://ipset.netfilter.org/tips.html

rnalrd:
@kunkku you are the package maintainer. Any comments?
Thanks

kunkku:
I would prefer improving the script's performance without changing the file format in an incompatible way. I pushed an optimized version of the current script, which should bring the execution time down to a few seconds. It can still be optimized further if needed.

This pull request contains several new, mutually independent features. I think you should split each feature to its own pull request to make them easier to review. However, I'm not sure if the function for creating iptables/ip6tables rules belongs to this package. Any opinions on this?

itoffshore:
@kunkku I tried your updated script & it no longer delays booting by several minutes - but your revised functions do not allow any comments to exist in the `ipset.d` files (my functions filter out comments).

My `initd` should not have introduced any incompatibilites (as long as the `ipset.d` files have a valid header) - it is based on the gentoo `initd` with some extra functions for unloading specific ipsets & loading inactive sets (rather than reloading all sets to add a new set). 

Without the `iptables` functions the service cannot be stopped:
```
root@alpine [~]# rc-service ipset.apk-new stop
 * Flushing firewall IP sets ...
ipset v6.32: Set cannot be destroyed: it is in use by a kernel component
ipset v6.32: Set cannot be destroyed: it is in use by a kernel component
ipset v6.32: Set cannot be destroyed: it is in use by a kernel component
ipset v6.32: Set cannot be destroyed: it is in use by a kernel component
ipset v6.32: Set cannot be destroyed: it is in use by a kernel component                                                                                    [ !! ]
 * ERROR: ipset.apk-new failed to stop
```
You also need to maintain the rules somewhere else for `ipset` to even work which seems less than intuitive.

kunkku:
The current implementation allows comments in the files, starting with the `#` character.

Your patch changes the behavior of the `save` command. The current implementation saves each set to a dedicated file. The use of set-specific files was specifically requested by @fabled when the script was introduced.

What is the use case for stopping the `ipset` service while `iptables` is running?

The dependencies of the `ipset` service have been set such that it starts before `iptables`. Therefore, the logic in the `add_ip` function will not work properly when invoked from `start` as the set-related rules are not yet in place.

itoffshore:
I start `ipset` before `iptables` too without any problems. I've been using my `initd` on multiple servers for about a year blocking around `150,000` ip's. The extra functions give me the functionality I need for using custom lists easily + updating the lists from [iblocklist](https://www.iblocklist.com/) daily.  

I tried your `initd` again & some sets failed to load. I'm going to keep using my `initd` & doing things the gentoo way with a single save file. 

If anyone is interested in the extra `info` / `load` / `unload` & `iptables` functions [see my github](https://github.com/itoffshore/alpine-linux-scripts).

main/ipset/ipset.initd
```diff
@@ -59,57 +98,141 @@ stop() {
...
+	rm -f $tmpfile
+}
+unload() {
+	local name= ldsets=$(sets) i=0 choice=
+	if [ -n "$ldsets" ]; then
+		for name in $ldsets; do
+			i=$(( $i + 1 ))
+			einfo "($i): $name"
+		done
+		printf "\nSelect ipset to destroy? [1 - $i]: "; read choice
```
brimston3: From a cursory search, this is the only init script in aports that has an interactive terminal read with no timeout. The nearest other script is cryptsetup/dmcrypt.initd .

main/ipset/ipset.initd
```diff
@@ -59,57 +98,141 @@ stop() {
...
+	if head -n 1 $1 |grep -q "family inet6"; then
+		echo ip6tables
+	else
+		echo iptables
+	fi
+}
+
+add_ip() {
+	local name= ipt_cmd=
+	if echo "${IPT_MANAGE}" | egrep -vq '(^y|^Y)'; then
```
brimston3: This option really means "install blacklist rules and only single entry or range blacklist-type sets are allowed." Could this all be a separate aport providing the management scripts and separate storage location for this type of blacklist? In combination with SAVE_ON_STOP, this option can make a mess of sets generated by fail2ban's iptables-ipset-proto4.conf for port-specific bans.

itoffshore: ok - will put the extra functions into a new aport `ipset-blacklist` with just an `initd`

main/ipset/ipset.initd
itoffshore: will use `read -t 100`

main/ipset/ipset.initd
itoffshore: @brimston3 @kunkku - see https://github.com/alpinelinux/aports/pull/1212

* the `save` function only saves the sets in `/etc/ipset.blacklist` so it should not affect any other sets
* `unload()` now uses `read -t 100`

PR #345, opened 2016-10-03
testing/logstash: update to version 2.4.0

nathanejohnson:
Substantial rework of aport


fabled:
The logs indicating it's downloading ruby binaries, and using gem to bundle lot of ruby dependencies. This is against the packaging principle that we can reproduce build of same version.

ncopa:
seem like we have version 5.2.2 in aports now.

PR #346, opened 2016-10-03
community/salt: Replace deps migrated to py2/py3 subpkg

coredumb:


jirutka:
BTW This split of some python packages to py2/py3 is backward compatible. When you `apk add py-tornado`, apk automatically installs py2-tornado if you have python2 installed and py3-tornado if you have python3 installed (or nothing if you don’t have any python installed).

The problem with py-tornado was that I’ve renamed dependency py-backports_abc to py2-backports_abc without actually renaming or splitting py-backports_abc.

PR #352, opened 2016-10-04
testing/php-couchbase: merge php5/php7-couchbase into single abuild

jirutka:
It’s based on our approach for multiversion Lua packages. What do you think @nathanejohnson,  @ncopa, @clandmeter?


nathanejohnson:
@jirutka I like the idea, however I will need some education on how to make that work.

jirutka:
It already works, Travis failed just because the repository was out-of-sync and this relies on some recent change in the php5 package.

PR #355, opened 2016-10-06
community/php-xdebug: merge php5/php7-xdebug into single abuild

andypost:
I think this extension is battle tested enough


jirutka:
Please merge it with php5-xdebug into a single abuild, take 1958a7f678c088978c6190cfb53461bb902d4fd6 as a template.

andypost:
Merged! Thanx for nice example

community/php-xdebug/APKBUILD
```diff
@@ -0,0 +1,68 @@
...
+	pkgdesc="$pkgdesc (for PHP $ver)"
+	arch="all"
+	depends="$depends php$ver"
+	install_if="$pkgname=$pkgver-r$pkgrel php$ver"
+	[ "$ver" = 5 ] && replaces="$pkgname"
+
+	make -C "$builddir-$ver" INSTALL_ROOT="$subpkgdir" install || return 1
+
+	install -d "$subpkgdir"/etc/php$ver/conf.d || return 1
+	echo "extension=$_pkgname.so" > "$subpkgdir"/etc/php$ver/conf.d/$_pkgname.ini
```
andypost: This probably should be removed to prevent enabling the extension by default because it affects performance and require manual configuration in 90% of cases.
But this breaks backward compatibility...

andypost:
I can't get how to get rid of `php-xdebug` itself and what will be installed with `apk add php-xdebug`

community/php-xdebug/APKBUILD
jirutka: To be honest, I don’t know why it is here and what exactly does it do, but it’s currently in all (?) php packages.

jirutka:
> I can't get how to get rid of `php-xdebug` itself and what will be installed with `apk add php-xdebug`

You can’t and should not, it’s here for purpose. ;) It works as a virtual package. When you have php5 installed and run `apk add php-xdebug`, then it automatically installs `php5-xdebug` (see `install_if` variable). And similarly for php7. If you have both php versions installed, then it installs both xdebug subpackages.

community/php-xdebug/APKBUILD
andypost: As official docs tells https://github.com/xdebug/xdebug#xdebug this line should be added manually
And should be `zend_extension=xdebug.so`

So will reroll

andypost:
> then it automatically installs php5-xdebug

Is there a way to `cat ";zend_extension=xdebug.so" > /etc/php$ver/conf.d/$_pkgname.ini` only when installed as dependency of `php-xdebug`?

jirutka:
> Is there a way to `cat ";zend_extension=xdebug.so" > /etc/php$ver/conf.d/$_pkgname.ini` only when installed as dependency of `php-xdebug`?

No. Don’t worry about it, just add post-upgrade message that it has been changed. See [this](https://github.com/alpinelinux/aports/blob/master/main/nginx/nginx.post-upgrade) for example.

PR #357, opened 2016-10-06
main/lua-yaml: upgrade to 1.1.2

jirutka:
@ncopa Latest version of lua-yaml requires new dependency – lua-[lub](http://doc.lubyk.org/lub.html). What to do in this case? Add new aport lua-lub to the main, or move lua-yaml to testing or community?


ncopa:
I am ok to add lua-lub directly to main in this case. :+1:

PR #358, opened 2016-10-07
testing/cockroach: new aport

nathanejohnson:
https://www.cockroachlabs.com/
An open source, survivable, strongly consistent, scale-out SQL database.


testing/cockroach/force_autoreconf.patch
```diff
@@ -0,0 +1,11 @@
+--- a/Makefile
++++ b/Makefile
+@@ -217,6 +217,8 @@
+ # of them (or their dependencies) change.
+ .bootstrap: $(GITHOOKS) $(GLOCK) GLOCKFILE
+ 	@unset GIT_WORK_TREE; $(GLOCK) sync github.com/cockroachdb/cockroach
++	#Force autoconf to run, otherwise won't work on musl for instance
```
knz: Could you investigate this further and detail to us (cockroach labs) what the issue is. We'd be keen to ensure the build goes smoothly with musl too.

testing/cockroach/ldflags_fix.patch
```diff
@@ -0,0 +1,18 @@
+--- a/Makefile
++++ b/Makefile
+@@ -78,12 +78,12 @@
+ build: install
+ 
+ .PHONY: install
+-install: LDFLAGS += $(shell GOPATH=${GOPATH} build/ldflags.sh)
++install: GOLDFLAGS := $(shell GOPATH=${GOPATH} build/ldflags.sh)
```
knz: What's the issue with LDFLAGS?

testing/cockroach/force_autoreconf.patch
nathanejohnson: From what I can tell, it looks like you're shipping with artifacts generated from an autogen run.  This works well for glibc, but is broken on musl.

testing/cockroach/ldflags_fix.patch
nathanejohnson: LDFLAGS has implications for gcc and pals.  GOLDFLAGS is something I picked that isn't used by anything else.

knz: But that's the thing, right. Don't we want to propagate these LDFLAGS to the compiled C/C++ dependencies?

knz: Oh ok never mind me then. Thanks for the heads up.

jirutka:
@7heo Uh, another Go stuff, could you please review it?

testing/cockroach/cockroach.post-deinstall
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+echo "Please manually remove the 'cockroach' user and group if no longer needed" >&2
+
+exit 0
```
jirutka: This is really unnecessary post-deinstall message.

rnalrd:
It downloads most of the dependencies during build. This is not a [reproducible build](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Go).
Besides several tests fails.

nathanejohnson:
@rnalrd, it uses dependency management, so it is downloading deps but locked to a specific sha.  Which tests are you seeing build?  I kind of gave up on this a while ago, have not revisited it since nobody seemed to want to review it.

kaniini:
The dependency management should not be handled at build-time.  It is plausible that we may disable network access to builders while they are building packages in the future in order to enforce this.  @7heo has a few packages which handle the dependency management as a separate step.

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,95 @@
...
+	mv "$srcdir/$pkgname-$_pkgrealver" "$builddir"
+	default_prepare
+}
+
+build() {
+	export GOPATH="$srcdir/go"
+	cd $builddir
+	glide install
+	# force an auto reconf, since it seems this ships with some
+	# artificats from an autoconf run for a glibc distro
```
knz: Please file an issue on our side at @cockroachdb for this. This is not the right approach.

knz:
The dependency gathering step is clearly separated: the make target `.bootstrap` contains all this work (and excludes the building proper). Why don't you capture the result of running `make .bootstrap` once as the source package?

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,95 @@
...
+	mkdir "$srcdir/go/bin"
+	mkdir "$srcdir/go/localpkg"
+	mv "$srcdir/$pkgname-$_pkgrealver" "$builddir"
+	default_prepare
+}
+
+build() {
+	export GOPATH="$srcdir/go"
+	cd $builddir
+	glide install
```
tamird: why are you running `glide install`? this is not required; all the dependencies ship in the `vendor` submodule.

testing/cockroach/APKBUILD
tamird: Yeah, this is not correct. For one, we already ship various artifacts for various platform; it would be trivial to add one for alpine/musl.

That said, I've built this against musl myself, so I don't know why this is required.

testing/cockroach/ldflags_fix.patch
tamird: This has been fixed upstream, FWIW. https://github.com/cockroachdb/cockroach/pull/14257

nathanejohnson:
@tamird @knz This PR has laid dormant for nearly 6 months with no input, I am on vacation this week so I'm not going to be able to address any of your concerns until at the earliest next week.  glide install is a pretty standard way of honoring a glide.lock, but it seems you guys aren't expecting to use this flow (though I found very little in the way of documentation, so this is what I came up with).  If there is interest from the cockroach team in taking this over, I would be more than happy to close this PR and let you guys put something out there.

benesch:
Hi @nathanejohnson—thanks so much for the initial version of this! I recently landed some improvements to our packaging flow for Homebrew that I think will be quite helpful for this package as well. In short, we now publish a release tarball that includes pre-vendored dependencies (no additional download required) and hardcoded build information.

I'm going to try my hand at updating this PR. I've never used Alpine Linux, though, so if I get stuck I'd love to work with you next week, if you're back from vacation, to push this through.

benesch:
Done! Please take a look at #1183 when you have a minute.

fabled:
Superceded by PR #1183

PR #360, opened 2016-10-11
community/salt: upstream update

WnP:
see: https://github.com/saltstack/salt/pull/36904


jirutka:
@coredumb, is this update okay?

coredumb:
@jirutka Yes It's OK and required.

PR #366, opened 2016-10-13
testing/mediainfo: new aport

albap:
Note - I don't want to be contributor or maintainer so it'd be nice if someone who likes mediainfo could take care of.
Thanks.

### DIFF
--- a/testing/mediainfo/APKBUILD
+++ b/testing/mediainfo/APKBUILD
@@ -12,20 +12,18 @@ makedepends=""
 install=""
 subpackages=""
 source="https://mediaarea.net/download/binary/mediainfo/$pkgver/MediaInfo_CLI_${pkgver}_GNU_FromSource.tar.xz"
-builddir="$srcdir/$pkgname-$pkgver"
+builddir="$srcdir"
 
 build() {
-	cd "$builddir"
-	sh CLI_Compile.sh
+	cd "$builddir"/MediaInfo_CLI_GNU_FromSource
+	sh ./CLI_Compile.sh
 }
 
 package() {
-	cd "$builddir"/MediaInfo/Project/GNU/CLI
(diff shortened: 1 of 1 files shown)


testing/mediainfo/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+pkgrel="0"
+pkgdesc="Unified display of the most relevant technical and tag data for video and audio files."
+url="https://mediaarea.net/"
+arch="all"
+license="Simplified BSD"
+depends=""
+makedepends=""
+install=""
+subpackages=""
+source="https://mediaarea.net/download/binary/mediainfo/$pkgver/MediaInfo_CLI_$pkgver_GNU_FromSource.tar.xz"
```
jirutka: There’s no such variable `$pkgver_GNU_FromSource`… You must wrap it in braces in this case: `MediaInfo_CLI_${pkgver}_GNU_FromSource`.

BTW It’s really not needed nor desirable to write checksums by hand. That’d be really annoying and unnecessary, so we have a tool for that – `abuild checksum`. Have you read [guide](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#APKBUILD_variables.2Ffunctions) on the Alpine wiki?

albap:
@coreyjonoliver wants to maintain this?

BTW I looked at http://git.alpinelinux.org/cgit/aports/tree/testing/libmediainfo/APKBUILD for you and there are mistakes.
Website is wrong, license is wrong, version is old.

Since you maintain libmediainfo I think you could want to maintain mediainfo too?

coreyjonoliver:
@albap I'd be willing to also maintain mediainfo.

Also, thanks for pointing out the mistakes in libmediainfo. When I have some time, I'll submit a patch to address them.

albap:
Added you as the maintainer - nice.

PR #367, opened 2016-10-13
community/emacs cleaned up testing&unmaintained, bumped v to 25.1, added n…

stef:
…ox, x11 and gtk2 variants


ncopa:
It fails to build because it depends on `community/webkit2gtk`. We have the following options:
- move emacs to community
- find someone who want maintain webkit2gtk and move it to main
- split emacs into 2 apkbuilds, where one is in main, without webkit2gtk and the toerh in community

stef:
oh noez. webkit is not an active dependency. i'll remove it.

stef:
looks clean now.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+	imagemagick-dev glib-dev fontconfig-dev libpng-dev
+	libxml2-dev pango-dev tiff-dev libjpeg-turbo-dev ncurses-dev
+	ncurses-libs gnutls-dev libxaw-dev"
+subpackages="$pkgname-doc $pkgname-nox $pkgname-x11 $pkgname-gtk2"
+source="ftp://ftp.gnu.org/gnu/emacs/emacs-$pkgver.tar.xz
+        gamedir.patch
+        noaslr-dump.patch
+        musl.patch"
+
+_builddir="$srcdir/emacs-$pkgver"
```
jirutka: Please rename it to `builddir`. It has been changed quite recently, `builddir` is now an abuild-defined variable.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+prepare() {
+	local i
+	cd "$_builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
+}
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+		--without-sound \
+		--without-x \
+		--without-file-notification \
+		|| return 1
+	make
+}
+
+build() {
+	cd "$_builddir"
+	[[ -d nox ]] || {
```
jirutka: `[[` is not defined by POSIX-sh, it’s a so called _bashism_ and must be avoided in APKBUILDs.

Also I don’t really understand the purpose of this line here.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+	   cp -a nox gtk2
+	   cp -a nox x11
+   }
+   build_nox
+   build_x11
+   build_gtk2
+}
+
+package() {
+   mkdir -p "$pkgdir" && return 0
```
jirutka: This is nonsense. `&& return 0` is invoked only when `mkdir -p` returns 0 and since this is the last statement in the function, shell implicitly returns exit code of the last statement, i.e. this is like writing `true || true`.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+	mv "$subpkgdir"/usr/share/man/man1/ctags.1.gz "$subpkgdir"/usr/share/man/man1/ctags.emacs.1.gz
+   # only keep info and man directories, all other is in the specific package
+   rm -rf "$subpkgdir"/usr/bin
+   rm -rf "$subpkgdir"/usr/bin
+   rm -rf "$subpkgdir"/usr/lib
+   rm -rf "$subpkgdir"/usr/share/appdata
+   rm -rf "$subpkgdir"/usr/share/applications
+   rm -rf "$subpkgdir"/usr/share/emacs
+   rm -rf "$subpkgdir"/usr/share/icons
+   rm -rf "$subpkgdir"/var
```
jirutka: `rm` accepts one _or more_ paths, so you can (and should) give it all these paths at once instead of multiple calls.

```sh
rm -rf "$subpkgdir"/usr/bin \
	"$subpkgdir"/usr/bin \
	"$subpkgdir"/usr/lib \
	"$subpkgdir"/usr/share/appdata \
	"$subpkgdir"/usr/share/applications \
	"$subpkgdir"/usr/share/emacs \
	"$subpkgdir"/usr/share/icons \
	"$subpkgdir"/var
```

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+		|| return 1
+	make
+}
+
+build() {
+	cd "$_builddir"
+	[[ -d nox ]] || {
+      mkdir -p nox
+	   mv * nox || true
+	   cp -a nox gtk2
```
jirutka: Unfortunately APKBUILD does not run with `set -e` (yet, it’s on our TODO list to change it), so every statement that may fail, but should not fail, and is not the last statement of the function, should end with `|| return 1`.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+	# remove conflict with ctags package
+	mv "$subpkgdir"/usr/bin/ctags "$subpkgdir"/usr/bin/ctags.emacs
+	rm -rf "$subpkgdir"/usr/share/info
+	rm -rf "$subpkgdir"/usr/share/man
+
+	# fix user/root permissions on usr/share files
+	find "$subpkgdir"/usr/share/emacs/ -exec chown root:root {} \;
+	find "$subpkgdir"/usr/lib -perm -g+s,g+x ! -type d -exec chmod g-s {} \;
+	# fix perms on /var/games
+	chmod 775 "$subpkgdir"/var/games
```
jirutka: Why the heck emacs uses `/var/games`? o.O

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+	CFLAGS=-fno-pie \
+	LDFLAGS=-no-pie \
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--libexecdir=/usr/lib \
+		--localstatedir=/var \
+		--without-makeinfo \
```
jirutka: There’s a huge code duplication, it’d be better to add custom build function with shared parts and just pass additional flags to it.

main/emacs/APKBUILD
```diff
@@ -0,0 +1,186 @@
...
+		--with-gameuser=:games \
+      --with-x-toolkit=athena \
```
jirutka: This APKBUILD is full of mixed tabs & spaces, please use only tabs for indentation.

main/emacs/APKBUILD
stef: i followed the instructions here: https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#package.28.29

but you're right, will remove it. but perhaps also the wiki needs updating?

main/emacs/APKBUILD
stef: there is some games bundled, i think it's used for the highscores.

stef:
thanks for the review! will adopt things.

clandmeter:
It looks good except few things.

1. I think its better if you prefix the build functions with an underscore.
2. Please move autoreconf to prepare function (there is no need  to run it 3x)
3. Do we really want this in main?
4. the main package provides nothing. I guess this is not what the user would expect. 
5. is there a reason you use double tabs to ident?

stef:
this PR is superseded by #526.

PR #369, opened 2016-10-14
testing/hugo: Updated to 0.17

tboerger:
This currently doesn't work because it's missing the dependencies. The dependencies are defined at https://github.com/spf13/hugo/blob/v0.17/vendor/vendor.json but how do you handle the dependency retrieve now?

### DIFF
--- a/testing/hugo/APKBUILD
+++ b/testing/hugo/APKBUILD
@@ -1,41 +1,39 @@
 # Contributor: Thomas Boerger <thomas@webhippie.de>
 # Maintainer: Thomas Boerger <thomas@webhippie.de>
 pkgname=hugo
-pkgver=0.16
+pkgver=0.17
 pkgrel=0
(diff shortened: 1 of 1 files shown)


jirutka:
@7heo could you please help with this?

7heo:
Hey guys, I'll happily do that as soon as I can, but at the moment I am swamped like crazy. I'll keep you posted once I can help with that.

tboerger:
If you can tell me the preferred way to package the dependencies I can continue on that. The old package version just fetched everything via go get, but I know we want to avoid that.

errm:
Would it not make sense to install the dependencies with go vendor using the upstream Makefile.  These deps are only needed at build time, right?

tboerger:
The old version used go get to fetch the dependencies, but is not reliable. So we want a dependency archive like it's done for other go packages

clandmeter:
i guess we can this.

PR #370, opened 2016-10-15
testing/vault: bumping to 0.6.2

gena01:
* Build latest 0.6.2 release
* Fix "mlock" permission in init.d script so vault can start
  as "vault" user.
* Adding cleanup script to stop the vault service on package
  removal
* Ading cleanup script to remove vault user and group on package
  removal

### DIFF
--- a/testing/vault/vault.pre-deinstall
+++ b/testing/vault/vault.pre-deinstall
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+# Stop vault service
+rc-service vault stop 2>/dev/null
+exit 0
+
--- a/testing/vault/APKBUILD
+++ b/testing/vault/APKBUILD
@@ -1,15 +1,15 @@
 # Contributor: Christian Kampka <christian@kampka.net>
 # Maintainer:
 pkgname=vault
-pkgver=0.6.1
-pkgrel=1
+pkgver=0.6.2
+pkgrel=0
 pkgdesc="Vault is a tool for securely accessing secrets."
 url="https://www.vaultproject.io/"
 arch="all"
 license="MPL-2.0"
 depends=""
 makedepends="go"
-install="$pkgname.pre-install"
+install="$pkgname.pre-install $pkgname.pre-deinstall"
 pkgusers="vault"
 pkggroups="vault"
 options="!strip"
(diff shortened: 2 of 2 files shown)


testing/vault/vault.post-deinstall
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Delete vault group and user
+delgroup vault 2> /dev/null
+deluser vault 2> /dev/null
```
jirutka: This is not a good idea. You can’t know if the vault user/group has been created by the package or not; user can use it even for something else.

testing/vault/vault.pre-deinstall
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# Stop vault service
+rc-service vault stop &> /dev/null
```
jirutka: `&>` is not POSIX-sh compatible, it’s a _bashism_. I also don’t think that it’s needed, user should now that (s)he should stop the service before uninstalling…

testing/vault/vault.initd
```diff
@@ -10,9 +10,10 @@ pidfile="/run/${SVCNAME}.pid"
 
 start_pre() {
 	checkpath -f -m 0644 -o ${SVCNAME}:${SVCNAME} "$VAULT_LOG_FILE"
+	setcap cap_ipc_lock=+ep $(readlink -f $(which vault))
```
jirutka: This should not be set in runscript, but in APKBUILD.

testing/vault/vault.post-deinstall
gena01: I understand, will take it out. (Actually copied from consul package, might need to fix that there as well)

testing/vault/vault.pre-deinstall
gena01: Well if they don't stop it and they have it running then it keeps on running and then if it's killed and added back again "service vault status" shows crashed. I could update this not to be a bashism.

This one was also copied from consul. Maybe another thing to fix there.

testing/vault/vault.initd
gena01: Yeah wasn't sure about this one, but it did fix the mlock issue. Let me update the commit.

P.S. "setcap" line came from official docs and is needed is vault is not running as root user. See here: https://www.vaultproject.io/docs/config/

gena01:
Updated. Let me know if this looks better.

testing/vault/vault.initd
jirutka: I understand why it is used and I didn’t mean that you should not use `setcap` at all, just that it should be set in APKBUILD, not post-install script. abuild and apk handles these attributes (if not for you, then use the latest abuild and apk versions).

gena01: @jirutka fixed. Sorry didn't know I could to setcap in APKBUILD just like install.

jirutka:
Merged in f625013a34ed3b3129cacf92a271ebaf15dd1d0b, thanks!

PR #373, opened 2016-10-17
main/xf86-video-qxl: add spiceqxland doc subpkgs

xming:


main/xf86-video-qxl/APKBUILD
```diff
@@ -36,6 +38,18 @@ build() {
...
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--infodir=/usr/share/info \
+		--localstatedir=/var \
 		|| return 1
 	make || return 1
```
jirutka: This is almost certainly wrong. The second run of make just overwrites all the files generated by the previous run.

xming: nope it doesn't, there is not make clean in between, the only file needed is spiceqxl_drv.so and it is preserved.

fabled: Why like this? The configure allows building both in one go:

>   --enable-xspice[=(no|yes|only)]  Build the xspice driver
>                    no      (default) Do not build
>                    yes     Build in addition to qxl
>                    only    Only build xspice

PR #374, opened 2016-10-17
testing/darktable: new aport

Ikke:
Musl support was recently added, so there is no tag to use yet.

### DIFF
--- a/testing/darktable/APKBUILD
+++ b/testing/darktable/APKBUILD
@@ -1,0 +1,39 @@
+# Contributor: Kevin Daudt <ops@ikke.info>
+# Maintainer: Kevin Daudt <ops@ikke.info>
+_tag=release-2.2.0rc0
+pkgname=darktable
+pkgver=2.2.0_rc0
+pkgrel=0
+pkgdesc="an open source photography workflow application and raw developer"
+url="http://www.darktable.org/"
+arch="all"
+license="GPL3"
+depends="libxslt gtk+3.0-dev libgphoto2-dev libwebp-dev
+	 lensfun-dev librsvg-dev sqlite-dev curl-dev jpeg-dev tiff-dev
+	 lcms2-dev json-glib-dev exiv2-dev lua5.2-dev pugixml-dev
+	 colord-dev colord-gtk-dev openexr-dev"
+makedepends="cmake intltool"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
+source="https://github.com/darktable-org/darktable/archive/$_tag.tar.gz"
+builddir="$srcdir/$pkgname-$_tag"
+
+prepare() {
+	cd "$builddir"
+	mkdir -p build
+}
+
+build() {
+	cd "$builddir/build"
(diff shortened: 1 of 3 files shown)


Ikke:
darktable fails to build because it relies on one of the packages in this commit

jirutka:
Ahaa, this is not a Travis issue! darktable depends on pugixml-dev, but pugixml abuild does not provide -dev subpackage. That’s why it failed.

testing/darktable/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+builddir="$srcdir/$pkgname-$_commit"
+
+prepare() {
+	cd "$builddir"
+	mkdir -p build
+}
+
+build() {
+	cd "$builddir/build"
+	cmake -DCMAKE_INSTALL_PREFIX=/usr .. || exit 1
```
jirutka: APKBUILD should never `exit` the abuild, there should be `return 1` instead.

testing/pugixml/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+url="http://pugixml.org"
+arch="all"
+license="MIT"
+makedepends="cmake"
+source="http://github.com/zeux/pugixml/releases/download/v1.7/pugixml-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	cmake -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=YesPlease scripts/ || exit 1
```
jirutka: APKBUILD should never `exit` the abuild, there should be `return 1` instead.

testing/pugixml/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Kevin Daudt <ops@ikke.info>
+# Maintainer: Kevin Daudt <ops@ikke.info>
+pkgname=pugixml
+pkgver=1.7
+pkgrel=0
+pkgdesc="Light-weight, simple and fast XML parser for C++ with XPath support"
+url="http://pugixml.org"
+arch="all"
+license="MIT"
+makedepends="cmake"
```
jirutka: You need to add `subpackages="$pkgname-dev"` here.

testing/darktable/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+# Maintainer: Kevin Daudt <ops@ikke.info>
+_commit=de8ac28bbb45db1b3eb9b0d51d5ec1288cdfc9dc
+pkgname=darktable
+pkgver=2.1.0.1928
+pkgrel=1
+pkgdesc="an open source photography workflow application and raw developer"
+url="http://www.darktable.org/"
+arch="all"
+license="GPL3"
+depends="libxslt gtk+3.0-dev libgphoto2-dev libwebp-dev
```
jirutka: This doesn’t look right. Why darktable depends on -dev packages (i.e. header files and similar) in runtime?

PR #375, opened 2016-10-17
Add py-six and py-futures as dependencies

jaustinpage:
I didn't get these, this should make the package good. py-futures is edge, so this cannot be backported...


jaustinpage:
I think this package is also a candidate for community/edge at this point, let me know what I should do to make that happen. Thanks!

jirutka:
> I didn't get these…

Yes, because you haven’t added them and I haven’t verified it.

The futures package is needed only for Python 2.

> I think this package is also a candidate for community/edge at this point

Have you verified that it really works?

jaustinpage:
"works on my machine" :-)

jaustinpage:
Ohh, it occurs to me that I have not tested python3.

PR #376, opened 2016-10-18
testing/protobuf: new aport (v3.1.0)

yzlin:
protobuf 3.1.0 released.


jirutka:
This package already exists ([main/protobuf](https://pkgs.alpinelinux.org/package/edge/main/x86_64/protobuf)), it just needs to be updated.

PR #379, opened 2016-10-20
testing/xdotool: move from unmaintained

stwa:


jirutka:
This program has a new home at https://github.com/jordansissel/xdotool with some new releases. Could you please also update it to the latest version?

stwa:
Thanks for the info, I updated the pull request.

PR #380, opened 2016-10-20
testing/xbindkeys: move from unmaintained

stwa:


jirutka:
@stwa Do you have any idea why this package fails on armhf in the install phase? There’s a [log](http://build.alpinelinux.org/buildlogs/build-edge-x86/testing/xbindkeys/xbindkeys-1.8.6-r0.log). There’s no such problem on x86, x68_64 and aarch64, only armhf. /cc @clandmeter

stwa:
Hmm, looks like things are called twice after entering the fakeroot?
xbindkeys_show failed to install while xbindkeys does.
I have no idea.

PR #381, opened 2016-10-20
testing/ttf-inconsolata: move from unmaintained

stwa:


testing/ttf-inconsolata/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+# Maintainer: Stefan Wagner <stw@bit-strickerei.de>
+pkgname=ttf-inconsolata
+pkgver=20160814
+pkgrel=0
+pkgdesc="Inconsolata fonts"
+url="https://www.google.com/fonts/specimen/Inconsolata"
+license="OFL"
+arch="noarch"
+source="https://github.com/google/fonts/raw/master/ofl/inconsolata/Inconsolata-Bold.ttf
+	https://github.com/google/fonts/raw/master/ofl/inconsolata/Inconsolata-Regular.ttf"
```
jirutka: This always fetches the latest version from the primary branch (master in this case), so build may not be reproducible. Please use URL with reference to specific commit (SHA1 hash); declare variable `_gitref` under the `pkgver` and use it in `source` URLs.

stwa:
Thank you, I have added the latest gitref.

PR #383, opened 2016-10-20
main/mkinitfs: add support for resume

stwa:
This patch tries to add resume from hibernation (aka suspend to disk)
support to mkinitfs. The swap partition can be provided via the resume
option.

I know this feature is mostly needed on destops/notebooks, therefore I'm
not sure if there is any interest for Alpine Linux. The default Alpine Linux
kernel has no hibernation support compiled in, so you need to build
a custom kernel in order to test this.

These links helped me on the topic:
* https://wiki.gentoo.org/wiki/Custom_Initramfs/Hibernation
* https://git.archlinux.org/mkinitcpio.git/tree/hooks/resume

As this might be a dangerous feature I would like to get your opinions.

### DIFF
--- a/main/mkinitfs/0001-resume-support.patch
+++ b/main/mkinitfs/0001-resume-support.patch
@@ -1,0 +1,42 @@
+Common subdirectories: a/features.d and b/features.d
+diff -u a/initramfs-init.in b/initramfs-init.in
+--- a/initramfs-init.in	2016-06-15 12:57:27.000000000 +0200
++++ b/initramfs-init.in	2016-08-16 00:30:42.936992624 +0200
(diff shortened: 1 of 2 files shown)


stwa:
I can only build this package if I ignore dependencies. Looks like the dependencies apk-tools>=2.0 and lddtree>=1.25 aren't resolved properly by abuild.

jirutka:
mkinitfs is Alpine project, so please open pull request with your patch in [alpinelinux/mkinitfs](https://github.com/alpinelinux/mkinitfs) repository.

stwa:
Opened new pull request in mkinitfs repo: https://github.com/alpinelinux/mkinitfs/pull/6

PR #384, opened 2016-10-20
Fix readline_* functions on PHP7 (issue 5670)

1ma:
The bug is documented in the issue tracker[1], basically all `readline_*` functions are broken under PHP7, thus any code that relies on them (like the Boris REPL) cannot work on Alpine. Fortunately they _do_ work in PHP5.

Comparing both APKBUILDs I noticed that in PHP5 the readline library is embedded into the php executable rather than distributed as a separate `php5-readline` package[2].

As of now this patch just removes the `shared` modifier of the `--with-readline` option to do the same in PHP7, and it seems to work. However, there's still a couple of issues with it.
- The apache module also has a `--with-readline=shared` flag. Editing it just like the other one broke my build.
- The `php7-readline` package still exists, even though it is now useless. I tried removing it from the `_exts` list[3], but this change also broke the build.
- Maybe there was a good reason to not embed readline into php in the first place. Maybe the patch should fix the `php7-readline` package instead of making it redundant. At the very least, thin base binaries seem to align more closely with the whole Alpine philosophy.
- I've noticed some discrepancies between the PHP5 and PHP7 build scripts. The former does not embed readline into the fpm daemon nor the apache module, but the latter does.

I'd appreciate some help to turn this patch into a proper fix. Thanks!

[1] https://bugs.alpinelinux.org/issues/5670
[2] https://github.com/alpinelinux/aports/blob/master/main/php5/APKBUILD#L258
[3] https://github.com/alpinelinux/aports/blob/master/community/php7/APKBUILD#L41

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -167,7 +167,7 @@ build() {
 		--enable-fpm \
 		--enable-embed \
 		--with-litespeed \
-		--with-readline=shared \
+		--with-readline \
 		|| return 1
 }
 


jirutka:
ℹ️ I’ve sent email to the package’s maintainer with link to this PR.

jirutka:
@andypost, could you please help with this PR/issue?

andypost:
Hi, I'm afk now but I seen that php7-readline broken. Will check tomorrow, but current fix makes a lot of sense.
I think it needs to check how shared library for readline used and is there dev dependency

andypost:
Really strange difference between php5 & php7, this extension has little usage in sapi & fpm because mostly useful with cli. So +1 to keep it separate.

andypost:
I checked official php docker and while building `readline` it shows
```
checking for libedit readline replacement... yes, shared
configure: error: Please reinstall libedit - I cannot find readline.h
```
So looks this extension just needs `libedit-dev` to build

andypost:
I came to the same - no way to keep `php7-readline` separate in #598

andypost:
This pr could be closed because  #598 accepted

PR #388, opened 2016-10-24
community/gogs: upgrade to 0.9.97

7heo:
With a little cleanup in the APK.

Note to the reviewer: please check the new way to install go packages (i.e. rebuilding the go directory structure correctly and build from there, instead of symlinking).


community/gogs/APKBUILD
```diff
@@ -17,9 +17,14 @@ source="${pkgname}-${pkgver}.tar.gz::https://github.com/gogits/$pkgname/archive/
...
-	glide.yaml.patch
 	glide.lock.patch"
-builddir="$srcdir/${pkgname}-${pkgver}"
+builddir="$srcdir/github.com/gogits/$pkgname"
+
+unpack() {
+	mkdir -p "$srcdir/github.com/gogits/"
+	tar -C "$srcdir/github.com/gogits/" -zxf $srcdir/${pkgname}-${pkgver}.tar.gz
+	mv "$srcdir/github.com/gogits/${pkgname}-${pkgver}" $builddir
+}
```
jirutka: Why this? Default `unpack` doesn’t work here?

7heo: No, because the default `unpack` doesn't unpack to a different place than `$srcdir/${pkgname}-${pkgver}` AFAIK. Am I wrong?

jirutka: It does not, but you can just move it to different location in `prepare()`. This is definitely better than overriding default `unpack`.

7heo: Ok. Will perform the requested change asap.

PR #397, opened 2016-11-01
main/spamassassin: typo in initd script

stwa:
There is a typo in setting the default value for command_args. Therefore, command_args can not be overwritten.


jirutka:
Please update checksums (`abuild checksum`) and bump `pkgrel`.

stwa:
Well, this is embarrassing... Fixed.

PR #356, opened 2016-10-06
testing/daemontools: move from unmaintained

leite:

PR #399, opened 2016-11-03
testing/daemontools: move from unmaintained

hanez:
Fixed svscan.initd to use openrc-run, a small typo fix in APKBUILD and set r1 to r2


hanez:
It seems there is a pull request existing already (https://github.com/alpinelinux/aports/pull/356). This one contains some fixes.

clandmeter:
I'm in favor of committing this as it does fix some small bugs, @leite are you ok for @hanez taking maintainership? 

@hanez can you cleanup this apkbuild? the _builddir should be renamed to builddir and you can remove the prepare function as this is now by default run inside abuid.

albap:
Can't a package have 2 maintainers?
I think I've seen some before that did...
I think it'd be good for packages if they could be maintained by more than 1 person.

clandmeter:
@albap no, packages should not have more then 1 maintainer. This has been discussed before and some of our infra does not like multiple maintainers.

leite:
No problem. @hanez could you take over?

I'm already happy someone looked at it, thanks @hanez!

hanez:
Hi, i will take that over... I am at home again in about 12 hours...

hanez:
I have created a new pull request at #529 . Dont know if this was the right way...

hanez:
So, I closed pull request #529 cause of some mistakes I did make. I created a new pull request #534 for this one.

It contains the fixes metioned above. Hopefully this okay now...

PR #403, opened 2016-11-06
testing/terraform: updating to 0.7.11

gena01:
* Updating to latest stable release
* Cleaning up APKBUILD

### DIFF
--- a/testing/terraform/APKBUILD
+++ b/testing/terraform/APKBUILD
@@ -1,75 +1,38 @@
 # Contributor: Thomas Boerger <thomas@webhippie.de>
+# Contributor: Gennady Feldman <gena01@gmail.com>
 # Maintainer: Thomas Boerger <thomas@webhippie.de>
 pkgname=terraform
-pkgver=0.6.16
-pkgrel=1
+pkgver=0.7.11
+pkgrel=0
 pkgdesc="Building, changing, and combining infrastructure safely and efficiently"
 url="https://www.terraform.io/"
 arch="x86_64"
 license="MPL 2.0"
 depends=""
-depends_dev=""
-makedepends="$depends_dev go go-tools bash"
+makedepends="go"
 install=""
-subpackages=""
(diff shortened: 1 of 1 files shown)


testing/terraform/APKBUILD
```diff
@@ -1,75 +1,44 @@
...
+unpack() {
+	if [ -z "$force" ]; then
+		verify || return 1
+		initdcheck || return 1
+	fi
 
+	mkdir -p "$projdir"
+	msg "Unpacking sources..."
+	tar -C "$projdir" --strip-components=1 -xzf $pkgname-$pkgver.tar.gz || return 1
+}
```
jirutka: Why do you define custom `unpack` function? You can (and should) just move the unpacked files in `prepare` function.

testing/terraform/APKBUILD
```diff
@@ -1,75 +1,44 @@
...
-	go install -v ${_gourl} || return 1
+	cd "$projdir"
 
-	for binary in ${_providers}; do
-		go install -v ${_gourl}/builtin/bins/${binary} || return 1
-	done
-
-	for binary in ${_provisioners}; do
-		go install -v ${_gourl}/builtin/bins/${binary} || return 1
-	done
```
jirutka: What happened to all these providers and provisioners? Upstream has merged them into single bloated binary?

gena01: It's a single binary now.

testing/terraform/APKBUILD
jirutka: This does not make it cleaner. `unpack` function should not be normally override with custom implementation without calling `default_unpack`, it may cause many problems to future.

gena01:
@jirutka updated:
1. took out the custom unpack method
2. bumped the version to 0.7.10 that just came out.

gena01:
@clandmeter thank you for the review. Updated PR. (and learned something new)

/cc @jirutka

clandmeter:
Why do you have tar in makedepends?

gena01:
@clandmeter forgot to take it out. :( Should be fixed now.

PR #405, opened 2016-11-06
community/py-psutil: upgrade to 5.0.0

fabaff:


community/py-psutil/APKBUILD
```diff
@@ -14,8 +14,7 @@ install=""
 subpackages="py3-$_pkgname:_py3 py2-$_pkgname:_py2"
 source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
```
jirutka: Unfortunately only .zip is available for 5.0.0. :(

fabaff: Thanks.

PR #406, opened 2016-11-06
testing/libmilter: set default pthread stack size

stwa:
This patch tries to fix various crashes for applications depending on
libmilter by setting the stack size for pthreads to 8MB. The default stack size
for musl libc is set to 80KB whereas glibc has it set to 8MB. This causes
problems when a large amount of memory is allocated on the stack.

For example, opendkim allocates blocks of 64KB multiple times, which causes
libmilter (and therefore opendkim) to crash.
Maybe a stack size of 1MB or 2MB would be sufficient, but as opendkim
depends on the default glibc behavior, 8MB should be safe.

I know this patch is kind of ugly, a better solution may be to file
a request for opendkim to allocate large blocks of memory on the heap.
But as libmilter/opendkim are fairly unusable at the moment, I suggest
to apply this patch as long as these packages are in testing.

* refs: #6360

### DIFF
--- a/testing/libmilter/APKBUILD
+++ b/testing/libmilter/APKBUILD
@@ -5,7 +5,7 @@ pkgname=libmilter
 _pkgname="sendmail"
 pkgver=1.0.2
 _pkgver=8.15.2
-pkgrel=1
+pkgrel=2
 pkgdesc="Sendmail Mail Filter API"
 url="http://www.sendmail.com/sm/open_source"
 arch="all"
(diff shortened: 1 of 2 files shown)


rofl0r:
that is very very costly. it is almost certain that a limit of 1MB would be more than sufficient.
(as the maintainer of sabotage linux, i've encountered only one application so far that exceeds 80KB, and it works happily with 128KB).

PR #410, opened 2016-11-07
testing/libmilter: set default pthread stack size to 256 KB

stwa:


jirutka:
Please update the value also in description in the patch.

stwa:
Description updated.

PR #412, opened 2016-11-07
main/py-pbr: enable py3

fabaff:


main/py-pbr/APKBUILD
```diff
@@ -7,24 +7,49 @@ pkgrel=0
...
+builddir="$srcdir"/$_pkgname-$pkgver
+prepare() {
+	local i
+	cd "$builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
+}
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

main/py-pbr/APKBUILD
```diff
@@ -7,24 +7,49 @@ pkgrel=0
 pkgdesc="Python Build Reasonableness"
 url="https://pypi.python.org/pypi/pbr"
 arch="noarch"
-license="APACHE2.0"
-depends="python2"
-depends_dev=""
-makedepends="$depends_dev python2-dev py-setuptools"
-install=""
-subpackages="$pkgname-dev"
+license="ASL 2.0"
```
jirutka: Please add dash: ASL-2.0.

fabaff: The [documentation](http://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license) says otherwise. When I started to clean-up the license tags on a large scale a while back we agreed to follow the naming convention of Fedora/Arch Linux.

main/py-pbr/APKBUILD
fabaff: Still seems pretty common to use `prepare()`, though. 3/4 of the packages in `main` are using this. I doubt that all of those packages needs special threat for their patches.

It looks a bit like that there should be a massive clean-up.

jirutka: Applying of patches was added into the default `prepare()` function some time ago, see [the code](https://github.com/alpinelinux/abuild/blob/69fd2290feaf7b8ae3d5cbaa2d1f6dda4915d71d/abuild.in#L566-L584). We’re cleaning abuilds as part of update of each package, as you can see in many recent commits. The same applies to `_builddir` → `builddir`.

main/py-pbr/APKBUILD
fabaff: I'm not sure where else the license information is used but the package DB seems to grab the full entry. Talking about separation...`MIT BSD` and `GPLv2 and GPLv2+ and GPLv3 and MIT` :wink:

main/py-pbr/APKBUILD
fabaff: For the later PRs I'm removing the `prepare()` and updating the builddir entry.. Thanks for pointing that out.

main/py-pbr/APKBUILD
jirutka: Currently it’s not used anywhere else, but that doesn’t mean that it will not be used in future…

> `GPLv2 and GPLv2+ and GPLv3 and MIT`

Nope, nope, nope.

PR #413, opened 2016-11-07
main/py-funcsigs: enable py3

fabaff:

PR #414, opened 2016-11-07
main/py-mock: enable py3

fabaff:
Depends on #412 and #413.


main/py-mock/APKBUILD
```diff
@@ -27,15 +25,37 @@ prepare() {
 }
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

fabaff: See comment on #412.

PR #415, opened 2016-11-07
testing/py-m2crypto: upgrade to 0.25.1

fabaff:


testing/py-m2crypto/APKBUILD
```diff
@@ -2,30 +2,58 @@
...
+builddir="$srcdir"/$_pkgname-$pkgver
+prepare() {
+	local i
+	cd "$builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
+}
```
jirutka: Please remove `prepare` function; applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared.

fabaff: See comment on #412.

jirutka:
Unfortunately this library is not compatible with Python 3 yet, see https://gitlab.com/m2crypto/m2crypto/issues/114.

PR #434, opened 2016-11-07
main/py-pillow: upgrade to 3.4.2 and enable py3

fabaff:
Required by #440.


jirutka:
This package installs [some python scripts](https://pkgs.alpinelinux.org/contents?repo=main&page=4&arch=x86_64&branch=edge&name=py-pillow) to `/usr/bin`, so it’s not so easy to make it for both Python versions. Are these scripts in /usr/bin really needed? They have extension `.py`, so it’s quite weird to put them into /usr/bin.

PR #454, opened 2016-11-07
community/py3-yarl: new aport

fabaff:
The module provides handy URL class for url parsing and changing.

http://yarl.readthedocs.io/en/latest/

### DIFF
--- a/community/py3-yarl/APKBUILD
+++ b/community/py3-yarl/APKBUILD
@@ -1,0 +1,28 @@
+# Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=py3-yarl
+_pkgname=yarl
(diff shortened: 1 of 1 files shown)


jirutka:
New aports must go into the testing repository first.

fabaff:
If a package needs to go to `testing` first, how is the update of aiohttp handled? In the past packages in `main` and I assume that this would cover `community` too shouldn't depend on those in `testing`.

jirutka:
> If a package needs to go to `testing` first, how is the update of aiohttp handled?

If this package is needed for another one, that is already in the community repository, then you should write this in the commit message.

fabaff:
My bad. Seems that referencing GitHub PRs doesn't work well in that case.

jirutka:
My bad, I overlooked that you’ve referenced it. However, it still should be mentioned in the commit message, because comments on GitHub are not visible in git log.

PR #455, opened 2016-11-07
community/py3-aiohttp: upgrade to 1.1.1

fabaff:
Depends on #454 and #453.

Required by #506.

### DIFF
--- a/community/py3-aiohttp/APKBUILD
+++ b/community/py3-aiohttp/APKBUILD
@@ -2,17 +2,14 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=py3-aiohttp
 _pkgname=aiohttp
-pkgver=0.22.5
+pkgver=1.1.1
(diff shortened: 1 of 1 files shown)


jirutka:
This package has been already updated in f0a940643e0382f675a11574a56e1b8743d20ace.

fabaff:
But the dependencies (yarl and async-timeout) are missing...i doubt that f0a9406 will work as expected.

jirutka:
Aha, you’re right, thanks for pointing to it!

jirutka:
I’ve fixed it in 6b8defef2021a135e49853745f6e7c20ab9a7341 and wrote fcolista to be more careful about dependencies of Python packages.

fabaff:
This was easy to say for me here as missed it for the Fedora package too.

jirutka:
I thought that Fedora has somehow automated checks of dependencies for Python packages, don’t they?

PR #461, opened 2016-11-08
main/py-pep8: enable py3

fabaff:


jirutka:
pep8 has been renamed to pycodestyle. See https://github.com/PyCQA/pycodestyle.

fabaff:
Now there is a permission error.

```bash
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1580, in _get
    with open(path, 'rb') as stream:
PermissionError: [Errno 13] Permission denied: '/usr/lib/python3.5/site-packages/pycodestyle-2.1.0-py3.5.egg-info/PKG-INFO'
```

jirutka:
testing/pep8 is a duplicate of main/pyp-pep8, isn’t it?

fabaff:
Seems so, but not sure. I guess that with `replaces=` we can play it safe.

main/py-pep8/APKBUILD
```diff
@@ -1,3 +1,6 @@
+#
+# Replaced by pycodestyle
```
fabled: If it's truly replaced, we should delete the whole APKBUILD.

main/pycodestyle/APKBUILD
```diff
@@ -0,0 +1,49 @@
+# Contributor: Peter Bui <pnutzh4x0r@gmail.com>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=pycodestyle
+pkgver=2.1.0
+pkgrel=0
+pkgdesc="A Python style guide checker"
+url="http://pep8.readthedocs.io/"
+arch="noarch"
+license="MIT"
+replaces=py-pep8
```
fabled: Should we also add `provides="py-pep8=$pkgver"` ?

testing/pep8/APKBUILD
```diff
@@ -1,3 +1,6 @@
+#
+# Replaced by py-pep8
```
fabled: Again, let's just remove this instead.

main/pycodestyle/APKBUILD
```diff
@@ -0,0 +1,49 @@
+# Contributor: Peter Bui <pnutzh4x0r@gmail.com>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=pycodestyle
```
fabled: Could this go in community, or is there good reasons for it to be in main?

PR #466, opened 2016-11-08
testing/perl-string-random: new aport

fabaff:


testing/perl-string-random/APKBUILD
```diff
@@ -0,0 +1,31 @@
+# Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=perl-string-random
+_realname=String-Random
+pkgver=0.29
+pkgrel=0
+pkgdesc="Perl extension to parse, manipulate, and lookup IP network blocks"
```
mwgamera: pkgdesc left from different package

fabled: We have apkbuild-cpan tool to automaticall create APKBUILD from CPAN. Why not use that?

PR #470, opened 2016-11-08
unmaintained/dhcpdump: move to community

fabaff:


jirutka:
This software is not maintained by upstream for 12 years…

fabled:
Is still useful? I would expect even tcpdump to supercede anything like this...
I'm happy to still merge this if it works. My complaint is though, that the new filename is wrong it's being renamed as `unmaintained/dhcpdump/APKBUILD → community/dhcpdump` while it should be `community/dhcpdump/APKBUILD `. Please fix this and squash the two commits.

PR #481, opened 2016-11-09
testing/py-typing: new aport

fabaff:
Type Hints for Python
https://pypi.python.org/pypi/typing/

Required by #506.


jirutka:
This package is a _backport_ of the typing module from Python 3.5 for older Python versions, so it makes sense only for Python 2 in Alpine. Thus I’ve changed it to py2-only package: 3c3f2b98432734b33225e4a3f8299bfa9abdd25b.

PR #492, opened 2016-11-09
community/py-tempita: clean-up

fabaff:


jirutka:
Are you sure that it works with Python 3? This project seems to be abandoned, project’s website http://pythonpaste.org/tempita/ is down and there’s no URL to sources on PyPI.

fabaff:
No, it doesn't. Looks like that I was caught by the [fork](https://github.com/gjhiggins/tempita) trap. This code looks like py3 compliant.

PR #501, opened 2016-11-09
testing/radare2 bumped rel to 1.0

stef:
had to add a guard to make the linker happy, sent the patch also upstream: https://github.com/radare/radare2/pull/6165


stef:
upstream merged it

stef:
upstream released new version, using that now.

clandmeter:
Can you cleanup this apkbuild, rename builddir and remove prepare. Looks like package() also has some leftovers.

PR #506, opened 2016-11-09
testing/home-assistant: new aport

fabaff:
A Home automation platform
https://home-assistant.io

### DIFF
--- a/testing/home-assistant/APKBUILD
+++ b/testing/home-assistant/APKBUILD
@@ -1,0 +1,30 @@
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=home-assistant
+_pkgname=homeassistant
+pkgver=0.32.2
+pkgrel=0
+pkgdesc="A Home automation platform"
+url="https://home-assistant.io"
+arch="noarch"
+license="MIT"
+depends="py3-requests py3-yaml py3-tz py3-pip py3-jinja2 py3-voluptuous py3-typing
+		py3-aiohttp py-async-timeout py3-aiohttp-cors py3-colorlog py3-distro py2-sqlalchemy
+		py3-astral py3-batinfo py3-beautifulsoup4 py3-zeroconf py3-cpuinfo py3-fixerio
+		py3-dweetpy py3-snmp py3-paho-mqtt"
+makedepends="python3-dev"
+source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
+	cd "$builddir"
+	python3 setup.py build || return 1
+}
+
+package() {
+	cd "$builddir"
+	python3 setup.py install --prefix=/usr --root="$pkgdir" || return 1
+}
+
+md5sums="82169d6794ba8b71767d0382362197ed  homeassistant-0.32.2.tar.gz"
+sha256sums="419e1707bd7048787568a776d00995a4c662898ce6990d537a99550cf4e530b1  homeassistant-0.32.2.tar.gz"
+sha512sums="edf0ec239df78dbe0f95e696c46a1954063b0278c9303648b5272d6155bfc2b6748c570fbca893dfc71a192118862561ddbe9026d1675b945653a99fd27b09f1  homeassistant-0.32.2.tar.gz"


clandmeter:
I'm interested in trying this out, but im currently using a Zwave usb dongle with domoticz.
Would be nice if py-openzwave could be added. I gave it a try but it seems more complicated with cython and make it use the already in aports located openzwave libs.

fabaff:
At the moment using `pip` is the way to go and a `venv`. There are some people who are using HA on Alpine Linux. Providing aports would save time during the installation as compiling is time-consuming on a single-board computer.

clandmeter:
Ive been playing with hass for a while now, and i must say its all a bit confusing for me.

Let me first outline what i think we want in Alpine aports repos regarding python.

1. We should supply an python application with limited dependencies. This means it should start with its default configuration, or the apkbuild should limit its configuration so it can start with less by default.
2. When a user would want to install additional features based up on python modules, they should manually install them with pip.
3. An exception of the above would be if one of the modules is arch depended and need a build system to create them. We would be able to supply them from our aports repository.

Having the above in mind, I have tried to limit the dependencies in hass aport and came up with [this](http://tpaste.us/3gDm) (which already holds one dirty hack). So it looks like the hard dependencies are needed to bootstrap the application. After bootstrap it will check the missing modules and will automatically run pip and install missing deps to the config deps folder. Actually it looks at the current version installed and will try to install the latest version from pip's database. This is where I currently run into issues. If one of the arch depended deps is updated in pip's database it will try to install it and of course it will fail and disable this part inside hass. 

Is there a way we can do as followed:

1. Keep deps to a minimum (like in my apkbuild)
2. Manually install arch depended modules with apk
3. Let hass manage the missing deps itself within its config folder
4. Don't let hass update arch depended modules inside its config folder (or any modules we installed with apk)

clandmeter:
@fabaff could you please explain me what you are trying to achieve with this PR? You mention yourself its impossible to package hass and instead one should use pip and venv.

I tried adding hass to testing but I have found no way to make it work properly so instead im using a combination of pip3 and some arch depended modules via apk.

You currently have [63](https://github.com/alpinelinux/aports/pulls/fabaff) PR's open of which the majority is python related. Are most of these somehow related to hass?

cc @jirutka

fabaff:
> @fabaff could you please explain me what you are trying to achieve with this PR? You mention yourself its impossible to package hass and instead one should use pip and venv.

It's possible but as you already know it will require some work. From my point of view the repositories and the package management tools of a distribution are the tools to manage software when possible. `pip`, `gem`, and `nmp` are third-party tools which should only be used as "a last resort". I think of them as `./configure && make && make install`. They all do their job but not always in the way they should.

> I tried adding hass to testing but I have found no way to make it work properly so instead im using a combination of pip3 and some arch depended modules via apk.

At the moment I haven't spent much thought on how a combination of `apk` and `pip` could look like but I tend to agree with your points. An optimal solution would be that all dependencies are in sync (it's kind hard to archive). 

> You currently have 63 PR's open of which the majority is python related. Are most of these somehow related to hass?

No, there are updates for Sphinx-related stuff and a lot of updates for tools to check Python code for  PEP8 or PEP257.

jirutka:
>  `pip`, `gem`, and `nmp` are third-party tools which should only be used as "a last resort". I think of them as `./configure && make && make install`. 

You’re wrong. These are build and (most importantly) dependency management tools for particular languages/ecosystem. They do a better job in their ecosystem than general dependency management tools in linux distributions (as long as you do not need “native extensions”). Therefore the comparison with `./configure && make && make install` is invalid.

Creating and maintaining gazillion distro-specific packages for Python modules, Rubygems, …, installing them globally (!) and resolving version conflicts is nothing but totally insane. It cost a lot of effort, creates many problems with version conflicts, compatibility, stability… and the result for _applications_ (not system tools etc.) is usually not worth it.

The generally recommended way of installing Python _applications_ is to use virtualenv/venv, i.e. install python dependencies locally, isolated from global modules. The same approach is used even for Ruby, Node.js and many others.

When I have to deal with complex Python/Ruby/Node.js applications, I usually make an aport “bundle” (see examples in [my repo](https://github.com/jirutka/aports#bundles)). Dependency bundling is generally forbidden in Linux distributions, so I keep such aports only in my personal repository (at least for now).

* * *

With all respect, I’m personally against adding such complex Python/Ruby/alike applications with dozens of noarch dependencies into official aports tree. It’s too expensive to maintain them and we don’t have enough maintainers to handle this in long-term.

fabled:
Seems we have now `testing/home-assistant`. Not sure if it's this, based on this or something different. Could you please check that out, and follow up with PRs based on that? I'm closing this one for now. Thanks!

PR #507, opened 2016-11-09
community/zstd initial commit

stef:
a new and efficient compressor.


jirutka:
New packages must go to the testing repository first. The only exception is when the package is needed as a dependency for other package that is already in community or main.

PR #515, opened 2016-11-10
community/py-sqlalchemy-migrate: enable py3 and fix deps

fabaff:
tempita is needed by py-sqlalchemy-migrate. It seems possible to get it working with `2to3`. I can't say how good but at least there was no traceback. 

Successor of #492 (sort of).


community/py-tempita/APKBUILD
```diff
@@ -3,23 +3,21 @@
...
+subpackages="py3-${pkgname/py-/}:_py3 py2-${pkgname/py-/}:_py2"
 source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
-
 builddir="$srcdir"/$_pkgname-$pkgver
 
 build() {
 	cd "$builddir"
 	python2 setup.py build || return 1
+	2to3 -w .
+	python3 setup.py build || return 1
```
jirutka: `setup.py build` installs files into the same location when run with python2 and python3. The result is that your py2 package contains files converted by 2to3.

PR #517, opened 2016-11-10
testing/xkcdpass

fabaff:


community/xkcdpass/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=xkcdpass
+pkgver=1.9.0
+pkgrel=0
+pkgdesc="Generate secure multiword passwords/passphrases"
+url="https://github.com/redacted/XKCD-password-generator"
+arch="noarch"
+license="BSD"
+depends="py3-xkcdpass"
```
jirutka: Pardon, this comment should be for `subpackages="py3-$pkgname:py3"`, not this line.  And also I was too hurry, this actually may not be entirely bad solution. I must think about it for a while.

jirutka:
**Note for myself**: Think about pros/cons of this approach.

fabaff:
ncopa suggested this way for another aport (`awake`) when I created it. I'm not sure about Debian but in Fedora we do it that way.

jirutka:
Could you please send me a link to some Fedora package which do that and also support both py2 and py3?

fabaff:
- http://pkgs.fedoraproject.org/cgit/rpms/awake.git/tree/awake.spec
- http://pkgs.fedoraproject.org/cgit/rpms/python-mutagen.git/tree/python-mutagen.spec
- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=xkcdpass
- http://pkgs.fedoraproject.org/cgit/rpms/pytest.git/tree/pytest.spec (only py2/py3 as the source takes care about the distinction)

My statement was not completely right. The way it's done depends on the packager and the reviewer. Or sometimes something started was a lib and a cli tool was added later. I think that the decision should be made on the base of what's likely be used more often. `xkcdpass` as an alternative to create a password on the command-line while the module is pretty "exotic".

PR #520, opened 2016-11-10
community/tor: build against openssl

obi12341:
https://www.torproject.org
Anonymous network connectivity

Tor should be compiled against openssl-dev instead of libressl-dev, because libressl seems to lack accelerated support for the NIST P-224 and P-256 groups (openssl "enable-ec_nistp_64_gcc_128" compile flag). Using openssl will prevent this warning on tor start: 
"We were built to run on a 64-bit CPU, with OpenSSL 1.0.1 or later, but with a version of OpenSSL that apparently lacks accelerated support for the NIST P-224 and P-256 groups. Building openssl with such support (using the enable-ec_nistp_64_gcc_128 option when configuring it) would make ECDH much faster."
And, as warning is telling, will improve tor performance, especially if you are running a high-traffic exit-,middle- or bridge-node.

### DIFF
--- a/community/tor/APKBUILD
+++ b/community/tor/APKBUILD
@@ -9,7 +9,7 @@ arch="all"
 license="BSD"
 pkgusers="tor"
 depends=""
-makedepends="linux-headers bash libevent-dev libressl-dev ca-certificates
+makedepends="linux-headers bash libevent-dev openssl-dev ca-certificates
 	zlib-dev"
 install="$pkgname.post-upgrade $pkgname.pre-install"
 subpackages="$pkgname-doc"


jirutka:
LibreSSL and OpenSSL currently cannot be installed together. However, there are more packages that does not work _fully_ with LibreSSL (e.g. stunnel), so we must solve this somehow.

/cc @ncopa @clandmeter

albap:
The software that (currently) doesn't work with libressl because of missing features, could have openssl embedded into them (static linked) and be independent of the system SSL of choice (until libressl adds the needed features and we go back to normal). Static linked binaries for these very specific cases. What you think?

albap:
@voidlinux uses Tor with Libressl, maybe it's not too big of a problem?
https://github.com/voidlinux/void-packages/blob/master/srcpkgs/tor/template
OpenBSD does it too since they don't even have an option.

obi12341:
@albap i guess there is not problem, if you just use `tor` to connect to the tor network and browse the web. It can get a problem, if you want to run a tor node for other people with a lot of traffic.

obi12341:
Will alpinelinux stay with libressl by default?

clandmeter:
@jirutka what is the problem with installing libressl and openssl alongside each other? They install just fine together.

skarnet:
Do we have benchmarks for the performance loss? I'm willing to believe it's significant, but is this a real-life issue that people have indeed experienced and that is actively harming their ability to run a Tor node?
My preferred solution would be to contact the libressl authors and ask them whether there's a good reason for not implementing enable-ec_nistp_64_gcc_128. If it's just lack of time, the accelerated support will come at some point. If there's a more fundamental reason, then we can think about implementing a workaround.

ncopa:
we use libressl because we believe its more secure than openssl, so this is a question if you want trade security for performance, which I don't think we want to do in this case.

You should report the issue to libressl.

PR #522, opened 2016-11-12
community/libreoffice: Support more languages

albap:


clandmeter:
If you don't bump pkgrel, nothing will happen. (see travis log).

clandmeter:
please squash commits

albap:
I use github interface and don't know how to squash.
Searching on stackoverflow I found that there's a new feature from March 2016 that lets you squash.
https://stackoverflow.com/questions/35948131/github-squash-commits-from-web-interface-on-pull-request-after-review-comments
If you select the Squash and Merge option all should be fine.
Does it work?

clandmeter:
We do not use the github interface, nor do we use the github git repository.
This is also mentioned in the github index page: 
**Readonly mirror of aports (http://git.alpinelinux.org/cgit/aports)** 

If you want to contribute to aports, you will need to have a build environment to test build your patches/PR's. If you are not able to set this up, the its better to create an issue on our bug tracker and request the prefereer changes.

PR #538, opened 2016-11-19
unmaintained/scanssh: move to community

fabaff:


jirutka:
The last release of this software is 11 years old and there’s no public issue tracker.

fabaff:
That's unfortunate...sure, it's nice to have an active upstream. We all know that an issue tracker is not a guarantee that bug get fixed or issues addressed.

jirutka:
> We all know that an issue tracker is not a guarantee that bug get fixed or issues addressed.

Of course, but at least it gives a chance that when someone discover a bug, it’d be reported, so we can somehow judge in what condition the code is. In this situation we now only one thing, that the code is totally unmaintained and we have no idea how many dangerous security bugs are there.

Therefore I don’t consider reasonable to distribute such software. I’d also strongly discourage from using it, but that’s of course your decision. You can always build the package yourself if you need it.

fabaff:
> The last release of this software is 11 years old and there’s no public issue tracker.

Which release is 11 years old? 

- Latest chntpw release is from 2014-02-01
- scanssh was last releases at on 2013-04-04

jirutka:
http://monkey.org/~provos/scanssh/:

> scanssh-2.1.tar.gz [GPG sig] - Release 2005-03-05 New

fabaff:
So then it's time for an update...https://github.com/ofalk/scanssh/releases

community/scanssh/APKBUILD
```diff
@@ -26,6 +26,6 @@ package() {
 	make DESTDIR="$pkgdir" install || return 1
 }
 
-md5sums="9fab4253b56b2d15367d4872b370cdcb  scanssh-2.1.tar.gz"
-sha256sums="057eec87edafbbe5bc22960cbac53e3ada0222400d649a5e2f22cc8981f5b035  scanssh-2.1.tar.gz"
-sha512sums="1ddff6f8e51735bab28032c8d7cbb641ff0cd19a404a27ab489589ef1e30e3721dd486854eb4fef15368cc53f10c9c2650ffaf7fa7f767dc2202fcc5540ae59c  scanssh-2.1.tar.gz"
+md5sums="e991586bb2e2a6e8ae212d7a28ad3b8d  #scanssh-2.1.1.tar.gz"
```
jirutka: This is not very good file name…

fabaff: The source was completely wrong.

jirutka:
I’ve already merged this one in 32e65351bf37498cdecd385118dc7ca3eaf13115 and 4223e13a1a03b7ee3a4462e50640e229f8b9c573.

PR #541, opened 2016-11-21
add contributing guidelines for Github

clandmeter:
These items are just on the top of my head which i recently bumped into. Lets clean this up and get it merged so contributors can at least have some basic info on how to contribute via github.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -20,7 +20,7 @@ This means you care about the quality of your contributions.
 1. [Fork](https://help.github.com/articles/fork-a-repo/) our Alpine Linux aports repository.
 2. Clone your copy of aports `git clone https://github.com/username/aports`.
 3. Create a feature branch `git checkout -b my_new_feature`.
-4. Make your desired changes and commit them with a [correct commit message](https://git-scm.com/book/ch5-2.html).
+4. Make your desired changes and commit them with a [correct commit message](https://git-scm.com/book/ch5-2.html#Commit-Guidelines).
   * Begin the commit message with a single short line summarizing the changes, followed by a blank line and then a more thorough description. Examples of correctly formated summary lines are:
     * When adding a new aport: testing/apkname: new aport
     * When modifying an aport: testing/apkname: short description about changes
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -30,7 +30,7 @@ This means you care about the quality of your contributions.
 7. If all is ok your PR will be merged but if a developer asks for changes please do as followed.
   * Make the requested changes.
   * Add your file(s) to git and commit (we will squash your commits if needed).
-  * Push your changes `git push --force origin my_new_feature`.
+  * Push your changes `git push origin my_new_feature`.
 8. Goto #6.
 
 ### Submitting a package with new dependencies
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,40 @@
+# How to contribute
+
+We are really glad you are reading this contribution guide. This means you care about the quality of your contributions.
+
+## Submitting patches via Github
+
+### Prerequisite for submitting patches/PR's
+
+1. Basic knowledge about how to us Git(hub)
+  * Properly setting up git with `git config`
+  * Know how to create a commit (for details see below).
+  * Know how to work with git history (rebasing your commits).
+  * Know how to create Pull Requests
+2. Have access to an Alpine Linux [development envirorment](https://wiki.alpinelinux.org/wiki/Developer_Documentation#Development).
+3. Please __do not__ submit PR's via Github's web interface. You should have a working [development envirorment](https://wiki.alpinelinux.org/wiki/Developer_Documentation#Development) available and submit your commits from the git interface.
+
+### Creating a Pull Request (PR)
+
+1. [Fork](https://help.github.com/articles/fork-a-repo/) our alpinelinux aports repository.
+2. Clone your copy of aports `git clone https://github.com/username/aports`
+3. Create a feature branch `git checkout -b my_new_feature`
+4. Make your desired changes and commit them with a correct commit message.
+  * Begin the commit message with a single short (less than 78 character) line summarizing the changes, followed by a blank line and then a more thorough description. Examples of correct formated summary lines are:
+    * When adding a new aport: testing/apkname: new aport
+    * When modifying an aport: testing/apkname: short description about changes
+  * If needed provide a proper formatted (line wrapped) description of what your patch will do. You can provide a description in the PR, but you must include a message for this specific commit in the commit description. If in the future we would like to distance ourselfs from Github the PR information could be lost.
+5. Open the aports repository in your github.com and switch to your feature branch. You should now see an option to create your PR. [More info](https://help.github.com/articles/creating-a-pull-request/)
+6. Wait for an Alpine Linux developer to review your changes
+7. If all is ok your PR will be merged but if a developer asks for changes please do as followed.
+  * Make the requested changes.
+  * add your file to git and `git commit --amend` (this keeps the commit log clean).
+8. Goto #6.
+
+### Submitting a package with new dependencies
+
+When you want to submit a package including its new dependencies to our repository, you should bundle these commits into a single PR. This is needed so our [CI](https://en.wikipedia.org/wiki/Continuous_integration) will first build the dependencies after which it will build the parent package. Failing to include __new__ dependencies will fail the CI tests.
+
+### Clean-up a Pull Request (PR)
+
+If by some mistake you end up with multiple commits in your PR and one of our developers asks you to squash you commits please do __NOT__ create a new pull request instead please follow [this rebase tutorial](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages).
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,40 +1,45 @@
 # How to contribute
 
-We are really glad you are reading this contribution guide. This means you care about the quality of your contributions.
+We are really glad that you are reading this contribution guide.
+This means you care about the quality of your contributions.
 
 ## Submitting patches via Github
 
 ### Prerequisite for submitting patches/PR's
 
-1. Basic knowledge about how to us Git(hub)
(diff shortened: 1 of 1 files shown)


clandmeter:
@jirutka any comments?

jirutka:
@clandmeter, sorry, haven’t read it yet, but it’s on my TODO list.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+2. Have access to an Alpine Linux [development envirorment](https://wiki.alpinelinux.org/wiki/Developer_Documentation#Development).
+3. Please __do not__ submit PR's via Github's web interface. You should have a working [development envirorment](https://wiki.alpinelinux.org/wiki/Developer_Documentation#Development) available and submit your commits from the git interface.
+
+### Creating a Pull Request (PR)
+
+1. [Fork](https://help.github.com/articles/fork-a-repo/) our alpinelinux aports repository.
+2. Clone your copy of aports `git clone https://github.com/username/aports`
+3. Create a feature branch `git checkout -b my_new_feature`
+4. Make your desired changes and commit them with a correct commit message.
+  * Begin the commit message with a single short (less than 78 character) line summarizing the changes, followed by a blank line and then a more thorough description. Examples of correct formated summary lines are:
```
jirutka: Actually it should be less than 69 characters. GitHub truncates any subject line longer than 69 characters with an ellipsis. I should also note that this is not “invention” of GitHub, it originates in git itself.

BTW The body of commit message should be wrapped at 72 characters, but that’s not so important as 69 characters in title.

jirutka:
```diff
- Examples of correct formated summary lines are:
+ Examples of correctly formated summary lines are:
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+    * When adding a new aport: testing/apkname: new aport
+    * When modifying an aport: testing/apkname: short description about changes
+  * If needed provide a proper formatted (line wrapped) description of what your patch will do. You can provide a description in the PR, but you must include a message for this specific commit in the commit description. If in the future we would like to distance ourselfs from Github the PR information could be lost.
+5. Open the aports repository in your github.com and switch to your feature branch. You should now see an option to create your PR. [More info](https://help.github.com/articles/creating-a-pull-request/)
```
jirutka:
```diff
- Open the aports repository in your github.com…
+ Open your copy of the aports repository at github.com…
```

jirutka: > …switch to your feature branch.

Actually it’s not even needed to switch branch, GitHub usually shows the suggestion for Pull request even when you’re in the default branch.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+6. Wait for an Alpine Linux developer to review your changes
```
jirutka: Missing dot…

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+7. If all is ok your PR will be merged but if a developer asks for changes please do as followed.
```
jirutka: I don’t know English rules for using punctuation very well, but I feel that this sentence really needs some.
I suggest:

```diff
- If everything is ok, your PR will be merged. But if a developer asks for changes, please do as followed.
+ If all is ok your PR will be merged but if a developer asks for changes please do as followed.
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+  * Make the requested changes.
```
jirutka: You should also explicitly mention that the change should be done on the branch used in the PR, in the local clone of the repository.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+  * add your file to git and `git commit --amend` (this keeps the commit log clean).
```
jirutka: add -> Add

jirutka: There’s final step missing: `git push --force origin my_new_feature`.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+8. Goto #6.
```
jirutka: GOTO is evil! 😸

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Submitting a package with new dependencies
+
+When you want to submit a package including its new dependencies to our repository, you should bundle these commits into a single PR. This is needed so our [CI](https://en.wikipedia.org/wiki/Continuous_integration) will first build the dependencies after which it will build the parent package. Failing to include __new__ dependencies will fail the CI tests.
+
+### Clean-up a Pull Request (PR)
+
+If by some mistake you end up with multiple commits in your PR and one of our developers asks you to squash you commits please do __NOT__ create a new pull request instead please follow [this rebase tutorial](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages).
```
jirutka: It’d be better to split this sentence into two after “a new pull request”.

.github/CONTRIBUTING.md
jirutka: Actually I’m more happy when contributors add new commits instead of amending, because then it’s easier to review the changes they did. Then I squash it myself before merging.

clandmeter: OK, but what happens when you squash them yourself? the squashed commits its yours or you squash them into the first one and keep author and subject the same?

jirutka: The second option, authorship of the commit is preserved. It’s basically the same as amend, I do it all the time. And I wrote [github-pr-closer](https://github.com/jirutka/github-pr-closer) with this in mind. ;)

.github/CONTRIBUTING.md
```diff
@@ -1,40 +1,45 @@
...
-1. [Fork](https://help.github.com/articles/fork-a-repo/) our alpinelinux aports repository.
-2. Clone your copy of aports `git clone https://github.com/username/aports`
-3. Create a feature branch `git checkout -b my_new_feature`
-4. Make your desired changes and commit them with a correct commit message.
-  * Begin the commit message with a single short (less than 78 character) line summarizing the changes, followed by a blank line and then a more thorough description. Examples of correct formated summary lines are:
+1. [Fork](https://help.github.com/articles/fork-a-repo/) our Alpine Linux aports repository.
+2. Clone your copy of aports `git clone https://github.com/username/aports`.
+3. Create a feature branch `git checkout -b my_new_feature`.
+4. Make your desired changes and commit them with a [correct commit message](https://git-scm.com/book/ch5-2.html#Commit-Guidelines).
+  * Begin the commit message with a single short line summarizing the changes, followed by a blank line and then a more thorough description. Examples of correctly formated summary lines are:
```
jirutka: Why have you removed required length limit of the commit title?

.github/CONTRIBUTING.md
```diff
@@ -1,40 +1,45 @@
...
-5. Open the aports repository in your github.com and switch to your feature branch. You should now see an option to create your PR. [More info](https://help.github.com/articles/creating-a-pull-request/)
-6. Wait for an Alpine Linux developer to review your changes
+5. Open your copy of the aports repository at github.com and switch to your feature branch. You should now see an option to create your PR. [More info](https://help.github.com/articles/creating-a-pull-request/)
+6. Wait for an Alpine Linux developer to review your changes.
 7. If all is ok your PR will be merged but if a developer asks for changes please do as followed.
   * Make the requested changes.
-  * add your file to git and `git commit --amend` (this keeps the commit log clean).
+  * Add your file(s) to git and commit (we will squash your commits if needed).
+  * Push your changes `git push --force origin my_new_feature`.
```
jirutka: `--force` is not needed since you’ve removed requirement for amending commits.

.github/CONTRIBUTING.md
```diff
@@ -1,40 +1,45 @@
...
+This means you care about the quality of your contributions.
 
 ## Submitting patches via Github
 
 ### Prerequisite for submitting patches/PR's
 
-1. Basic knowledge about how to us Git(hub)
-  * Properly setting up git with `git config`
+1. Basic knowledge about how to use Git(hub)
+  * [Properly](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup) setup your git client.
```
jirutka: “setup” is [not a verb](https://en.wiktionary.org/wiki/setup#Verb), it should be “set up”.

.github/CONTRIBUTING.md
clandmeter: I didn't mention its not allowed either and thus --force works in both cases.

.github/CONTRIBUTING.md
jirutka: 69 characters is not about opinion, it’s the exact _hard limit_ assumed by GitHub and git tool. “50 characters or less” is commonly suggested soft limit, but doesn’t make sense for us since we prefix commit message with full name of the package.

Also note that there’s different limit for commit title and body. The body should be wrapped to 72 characters. As I know it doesn’t have any consequences in git tool or GitHub, it’s just a soft limit. However, it’s not just some made up number, it has specific reason.

For reference see e.g. http://chris.beams.io/posts/git-commit/ (scroll to section “2. Limit the subject line to 50 characters”).

.github/CONTRIBUTING.md
jirutka: `git push --force` is needed _only_ when you need to rewrite history of the branch. It must be used with caution, when you really know what you’re doing. It’s harmful to recommend it as default way to git beginners. One of the reasons for not using `--amend` or rebase by contributors is to avoid mistakes e.g. when the contributor accidentally push from wrong branch etc. The main reason is to keep history of the PR easier to track when doing review, i.e. avoid rewriting it.

.github/CONTRIBUTING.md
clandmeter: Can you show me an example of exact hard limit of 69 chars?

clandmeter: Lets not sigh each other, there is no need for that.

Im going to merge this as I cannot find any proof of any specific limit. If I look at the gitlog on github it shows variable cut offs depending on the text. I would also like to keep a limit which is not only based on github but more in general git orientated (as i already linked to the git handbook page).

Thanks for your review.

PR #545, opened 2016-11-22
testing/radicale: fix permissions for config files

stwa:
Radicale is started as user radicale, so the config should be readable by this
user. Currently the config files aren't world readable, which makes the
application crash on startup.

### DIFF
--- a/testing/radicale/APKBUILD
+++ b/testing/radicale/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=radicale
 pkgver=1.1.1
-pkgrel=1
+pkgrel=2
 pkgdesc="A simple CalDAV (calendar) and CardDAV (contact) server"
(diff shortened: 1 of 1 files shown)


jirutka:
> Radicale is started as user radicale, so the config should be readable by this user.

Does these files contain some sensitive data like passwords? If so, then they should by owned by root:radicale and chmod 640. Otherwise root:root and 644 is okay.

stwa:
They don't, maybe if you configure it to use LDAP authentication and provide a bind password (which isn't necessary, depending on your LDAP config). Then the file should be owned by radicale. What do you think? Change it?

jirutka:
As I see at http://radicale.org/user_documentation/#idmain-configuration-file, it may contain credentials for database, right? Then I’d prefer to change it: `install -m640 -g radicale config "$pkgdir"/etc/radicale/`.

stwa:
I only know of ldap_password, but I set the group to radicale and pushed the change so it's always “good“ per default.

PR #546, opened 2016-11-22
community/lego: revamping the APKBUILD

7heo:


jirutka:
You have to add `glide` to `makedepends`. ;)

clandmeter:
As discussed on IRC, I don't agree to set [$GOPATH](https://golang.org/cmd/go/#hdr-GOPATH_environment_variable) to $startdir

Nitpicking: There is no need to launch child processes (dirname) when you can use parameter expansion ie ${builddir%/*}

I think we need to use the vendor tools in prepare() so we are able to modify the source. Also the new idea about removing network from build() would prevent any vendor tools from working properly in build()

Regarding glide_init, it would be nice if you could add some comments about what you are trying to fix. For instance I don't see you are using the generated patch files?

This approach also breaks the usage of abump. Although its the responsibility of the person using abump to check if its safe to use it, we should probably add some functionality to just block the usage on such apkbuilds.

7heo:
@jirutka ;-) Right.

7heo:
@clandmeter wrote:
> As discussed on IRC, I don't agree to set [`$GOPATH`](https://golang.org/cmd/go/#hdr-GOPATH_environment_variable) to `$startdir`.

The reason invoked for that is that with `$GOPATH` set to `$startdir`, it is more likely to have files written to `$startdir` during the build (which should not happen), **if the tool the `APKBUILD` uses has a broken design**.

But I would like to point that any broken tool could write in any parent directory, nothing prevents that at all. And so, adding layers of directories just to lower the chances of problems to happen seems to me as an anti-solution.

The reason why I have set `$GOPATH` to `$startdir` in all my go-related `APKBUILD` manifests so far is that I see no reason to not to reuse an existing directory, which *has the same semantic value* (to contain the source).

The problem that arises with this practice (that @clandmeter pointed out) has to be solved indeed, but a in different way:

1. By changing the behavior of the broken tool via the use of invocation time flags.
2. By restricting the rights of `abuild` (and its children processes) during the execution of `abuild` as @fabled wants to implement.

To be fair, I believe that implementing point two will very likely bring out a LOT of `APKBUILD` manifests that write outside of their directory (in the user `$HOME` for instance), and so this very specific case is part of a larger problem, that we should not fix on a case-by-case basis.

----

@clandmeter wrote:
> Nitpicking: There is no need to launch child processes (dirname) when you can use parameter expansion ie ${builddir%/*}

I will check if that is POSIX, and if it is, I will use it. Thanks for pointing it out.

----

@clandmeter wrote:
> I think we need to use the vendor tools in prepare() so we are able to modify the source. Also the new idea about removing network from build() would prevent any vendor tools from working properly in build()

Very good point, I will change that. Thanks.

----

@clandmeter wrote:
> Regarding glide_init, it would be nice if you could add some comments about what you are trying to fix. For instance I don't see you are using the generated patch files?

I wrote the `glide_init` function to be re-usable in other `APKBUILD` manifests. Therefore it contains code that isn't used in this precise instance.

----

@clandmeter wrote:
> This approach also breaks the usage of abump. Although its the responsibility of the person using abump to check if its safe to use it, we should probably add some functionality to just block the usage on such apkbuilds.

How? I am sorry, I do not follow.

clandmeter:
@7heo when you write something that is reusable then you should try to document it. I know you have something against our wiki, but if you like to contribute to Alpine please use the tools we currently provide. If you have something against them you are free to create other solutions, and if other people agree with these changes/solutions we can also provide the infrastructure to serve them.

@7heo for abump check [abump](https://github.com/alpinelinux/abuild/blob/master/abump.in)
When you bump a package it will not automagically run `glide_init` and thus the result could be the version is updated without updated deps.

7heo:
> when you write something that is reusable then you should try to document it. I know you have something against our wiki, but if you like to contribute to Alpine please use the tools we currently provide.

@clandmeter I have nothing against our wiki, I have something about the software it uses. Wikis are great; but editing complex technical documents in an EULA-sized textarea that also requires me to heavily use the mouse is retarded.

Also, the content of our wiki is in a database that runs in a database engine, while perfectly sane and good tools exist to use flat files and git. I do not understand why I have to explain/defend this with people using *alpine*.

----

> When you bump a package it will not automagically run glide_init and thus the result could be the version is updated without updated deps.

I was not aware of what abump was doing. I always did that manually. But you are right, we should add some hooks so abump fires specific functions. Maybe we could even deport the `glide_init` function (since it can be reused) in a lang-related file that would be "used" by abuild (as in `use golang` for example) and hook it with abump from there?

PR #547, opened 2016-11-22
testing/ejabberd: fixes for libressl

stwa:
This fixes problems with libressl for ejabberd.

I know this patch is kind of ugly, but make deps needs to run before any
dependencies can be patched (dependencies are pulled from git by the Makefile).

* refs #6483

### DIFF
--- a/testing/ejabberd/APKBUILD
+++ b/testing/ejabberd/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: John Regan <john@jrjrtech.com>
 pkgname=ejabberd
 pkgver=16.09
-pkgrel=1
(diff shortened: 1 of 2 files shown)


clandmeter:
I think [this](http://tpaste.us/2bRW) patch against yours should be a bit more sane.

Actually we prefer no source changes in the build function at all. All of that should happen before or inside the prepare function. So maybe its better to add all deps to source and move them to deps folder.

stwa:
Much better! Should I push your change to this PR?

clandmeter:
If you agree you can just commit --amend it.

stwa:
Done, thanks for your help!

PR #552, opened 2016-11-24
community/libreoffice: enable Java

staalmannen:
Changed APKGBUILD to utilize Java. Builds but installing extensions that depend on Java still fail.


community/libreoffice/APKBUILD
```diff
@@ -30,16 +30,44 @@ case $pkgver in
...
+	$_addsrcurl/798b2ffdc8bcfe7bca2cf92b62caf685-rhino1_5R5.zip                              
+	$_addsrcurl/edc4d741888bc0d38e32dbaa17149596-source-sans-pro-2.010R-ro-1.065R-it.tar.gz
+	$_addsrcurl/7a15edea7d415ac5150ea403e27401fd-open-sans-font-ttf-1.10.tar.gz              
+	$_addsrcurl/eeb2c7ddf0d302fba4bfc6e97eac9624-libbase-1.1.6.zip
+	$_addsrcurl/8ce2fcd72becf06c41f7201d15373ed9-librepository-1.1.6.zip                     
+	$_addsrcurl/f94d9870737518e3b597f9265f4e9803-libserializer-1.1.6.zip
+	$_addsrcurl/907d6e99f241876695c19ff3db0b8923-source-code-pro-2.030R-ro-1.050R-it.tar.gz
+	$_addsrcurl/97b2d4dba862397f446b217e2b623e71-libloader-1.1.6.zip                         
+	$_addsrcurl/ace6ab49184e329db254e454a010f56d-libxml-1.1.7.zip
+	http://dev-www.libreoffice.org/extern/185d60944ea767075d27247c3162b3bc-unowinreg.dll"
```
jirutka: What is this?! How have you generated this list? How to proceed when updating the package?

community/libreoffice/APKBUILD
```diff
@@ -198,6 +226,16 @@ prepare() {
 	sed -e 's/@HAVE_GCC_VISIBILITY_BROKEN@/TRUE/g' -i config_host.mk.in
 
 	NOCONFIGURE=1 ./autogen.sh
+
+	mkdir -p "$srcdir"/ext_src
+
+	for t in $_ext_source; do
```
jirutka: Local variables must be declared as `local`. Just use `i` that is already defined in this function.

community/libreoffice/APKBUILD
```diff
@@ -1,8 +1,8 @@
-# Contributor: Timo Teräs <timo.teras@iki.fi>
-# Maintainer:
+# Contributor: Jens Staal <staal1978@gmail.com>
+# Maintainer: Timo Teräs <timo.teras@iki.fi>
```
jirutka: Have you asked @fabled if he wants to be maintainer of this package?

community/libreoffice/APKBUILD
staalmannen: These are the tarballs automatically downloaded by configure if --disable-fetch-external is not defined. configure then also downloads libreoffice-translations-$pkgver.tar.xz, but chokes on this tarball during extraction (!).
This is why the rest of the tarballs are downloaded and processed separately from libreoffice-translations.

Configure/make will demand that they are downloaded though (for example, it will stop if the dll is not present)

community/libreoffice/APKBUILD
staalmannen: OK so I will just use local t on top of the "for t in" function (re-using the i variable feels less good... perhaps just me)

community/libreoffice/APKBUILD
staalmannen: So when the next version is being packaged, the easiest is to first empty _ext_source and remove --disable-fetch-external. Then when(if) the build stops, list the tarballs in $srcdir/ext_src

fabled:
A derivative of this was applied in c4eabe8910112097afa9e6672192570d8a78ae54.

PR #553, opened 2016-11-24
testing/wine-staging: new port

staalmannen:
Wine-staging with gallium nine patches for those that want more bleeding-edge wine functionality


ncopa:
i think this is a good idea

PR #554, opened 2016-11-24
testing/mesa: new aport

staalmannen:
Mesa 13.0.1 for the very latest graphics drivers (experimental)

EDIT: since this basically is an upgrade of main/mesa - should I basically just submit it as a patch against that package instead?


testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+source="https://mesa.freedesktop.org/archive/${pkgver}/mesa-$pkgver.tar.xz
+	glx_ro_text_segm.patch
+	musl-fixes.patch
+	"
+
+_dri_driverdir=/usr/lib/xorg/modules/dri
+_dri_drivers="r200,radeon,nouveau,swrast"
+_gallium_drivers="r300,r600,radeonsi,nouveau,svga,swrast,virgl"
+
+_builddir="$srcdir/mesa-$pkgver"
```
jirutka: Please rename it to `builddir` (without underscore). This has been changed some time ago, many packages still use old approach, but new ones should not.

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+
+_intel_dri=
+case "$CARCH" in
+x86*)
+	_dri_drivers="${_dri_drivers},i915,i965"
+	subpackages="$subpackages $pkgname-dri-intel:intel"
+	export _endian="LITTLEENDIAN_CPU"
```
jirutka: Why do you export this variable?

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+	;;
+esac
+
+prepare() {
+	cd "$_builddir"
+	for i in $source; do
+		case $i in
+		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+		esac
+	done
```
jirutka: Please replace this for loop with `default_prepare || return 1` and move `cd $builddir` under it.

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+	
+	#hack to fix ENDIAN_CPU issue
+	_annoying_header="${_builddir}/src/amd/addrlib/inc/chip/r800/si_gb_reg.h"
```
jirutka: Variables used only in a single function should be declared as `local` and without underscore prefix.

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+	echo -e "#define $_endian 1\n$(cat $_annoying_header)" > $_annoying_header
```
jirutka: Please use `printf` instead of `echo -e`.

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+	return 0
+	# http://cgit.freedesktop.org/mesa/mesa/commit/?id=44ec468e8033553c26a112cebba41c343db00eb1
+	# https://code.google.com/p/chromium/issues/detail?id=412089
+#	local i
+#	install -d "$subpkgdir"/usr/lib/gallium-pipe
+#	for i in "$@"; do
+#		mv "$pkgdir"/usr/lib/gallium-pipe/pipe_$i.* \
+#			"$subpkgdir"/usr/lib/gallium-pipe/ || return 1
+#	done
+}
```
jirutka: Why is this commented code here?

testing/mesa/APKBUILD
```diff
@@ -0,0 +1,233 @@
...
+
+ati() {		_mv_dri radeon_dri r200_dri r300_dri r600_dri radeonsi_dri \
+		&& _mv_vdpau r300 r600 radeonsi \
+		&& _mv_gpipe r300 r600; }
+intel() {	_mv_dri i915_dri i965_dri; }
+nouveau() {	_mv_dri nouveau_dri nouveau_vieux_dri \
+		&& _mv_vdpau nouveau \
+		&& _mv_gpipe nouveau; }
```
jirutka: Please format these as normal functions, this is not very readable.

testing/mesa/APKBUILD
staalmannen: OK I will change accordingly

testing/mesa/APKBUILD
staalmannen: This is from main/mesa. This mesa-testing package is basically a copy of main/mesa with some minor adjustments to build the latest mesa.

testing/mesa/APKBUILD
staalmannen: OK I will put a local there

testing/mesa/APKBUILD
staalmannen: yet again, it comes from the main/mesa package. I did not want to deviate too much from that package because the general idea would be that one could copy the testing/mesa package to main/mesa when deemed stable/useable enough.

Alternatively, I just change this "new port" to a patch to main/mesa for an upgraded mesa package. It is after all on edge and not in a stable branch.

testing/mesa/APKBUILD
staalmannen: OK.

testing/mesa/APKBUILD
staalmannen: OK this was a "fossil" from when I tried to overide a failure of the build to recognize which endian there was. I can remove the export.

tmpfile:
Mesa on edge is already at 17.1.2.
I don't think you need this.

nmeum:
> EDIT: since this basically is an upgrade of main/mesa - should I basically just submit it as a patch against that package instead?

Yes, but since this pull request is pretty old now and `main/mesa` is at 17.X these days I will close this.

PR #555, opened 2016-11-24
testing/wine-mono: new port

staalmannen:
http://wiki.winehq.org/Mono
Open source .NET replacement

### DIFF
--- a/testing/wine-mono/APKBUILD
+++ b/testing/wine-mono/APKBUILD
@@ -4,7 +4,7 @@ pkgver=4.6.3
 pkgrel=0
 pkgdesc="Wine's built-in replacement for Microsoft's .NET Framew"
 url="http://wiki.winehq.org/Mono"
-arch="any"
+arch="x86 x86_64"
 license="GPL"
 depends="wine"
 makedepends=""
@@ -14,7 +14,7 @@ source="https://dl.winehq.org/wine/wine-mono/${pkgver}/wine-mono-${pkgver}.msi"
 
 _builddir=
 prepare() {
-	mkdir -p "$srcdir"
+	return 0
 }
 
 build() {
--- a/testing/wine-mono/APKBUILD
+++ b/testing/wine-mono/APKBUILD
@@ -1,0 +1,32 @@
+# Maintainer: Jens Staal <staal1978@gmail.com>
+pkgname=wine-mono
+pkgver=4.6.3
+pkgrel=0
+pkgdesc="Wine's built-in replacement for Microsoft's .NET Framew"
+url="http://wiki.winehq.org/Mono"
+arch="any"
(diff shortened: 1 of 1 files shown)


testing/wine-mono/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+license="GPL"
+depends="wine"
+makedepends=""
+install=""
+subpackages=""
+source="https://dl.winehq.org/wine/wine-mono/${pkgver}/wine-mono-${pkgver}.msi"
+
+_builddir=
+prepare() {
+	mkdir -p "$srcdir"
```
jirutka: `$srcdir` always exists, what’s the point of this line?

testing/wine-mono/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+pkgrel=0
+pkgdesc="Wine's built-in replacement for Microsoft's .NET Framew"
+url="http://wiki.winehq.org/Mono"
+arch="any"
+license="GPL"
+depends="wine"
+makedepends=""
+install=""
+subpackages=""
+source="https://dl.winehq.org/wine/wine-mono/${pkgver}/wine-mono-${pkgver}.msi"
```
jirutka: How is this binary built? Is it really platform-independent?

testing/wine-mono/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Maintainer: Jens Staal <staal1978@gmail.com>
+pkgname=wine-mono
+pkgver=4.6.3
+pkgrel=0
+pkgdesc="Wine's built-in replacement for Microsoft's .NET Framew"
+url="http://wiki.winehq.org/Mono"
+arch="any"
```
jirutka: This is not a valid type. If this binary is platform-independent, then it should be `noarch`.

jirutka:
@ncopa @clandmeter Is it okay to package binaries like this one?

testing/wine-mono/APKBUILD
staalmannen: OK thanks I will change.

testing/wine-mono/APKBUILD
staalmannen: no point at all. I copied the APKBUILD from wine_gecko and this must have stayed there... I will remove it.

ncopa:
yes, its accptable since its a windows binary to run within wine. We don't dont have any way to build windows binaries on alpine (and I see no reason to why we should implement it).

jirutka:
I'm quite sceptic about usefulness of such package. Also is it legally okay?

ncopa:
fixed with 32b83bc18acf619bd9f4695370f35b3f6c9af6e3

PR #557, opened 2016-11-27
testing/influxdb: new aport

stwa:


testing/influxdb/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+	ln -s "$builddir" "$GOPATH/src/github.com/influxdata/influxdb"
+
+	cd "$GOPATH/src/github.com/influxdata/influxdb"
+	go get -v -d ./... || return 1
+
+	git --git-dir="$GOPATH/src/collectd.org/.git" \
+		--work-tree="$GOPATH/src/collectd.org" \
+		checkout -f tags/v"$_collectdver" || return 1
+
+	go install -v -ldflags="-X main.version=$pkgver" ./... || return 1
```
clandmeter: You should only use go build, as [go install](https://golang.org/cmd/go/#hdr-Compile_and_install_packages_and_dependencies) will also pull in dependencies.

testing/influxdb/influxdb.initd
```diff
@@ -0,0 +1,17 @@
...
+command_background="yes"
+
+depend() {
+	need net
+	after firewall
+}
+
+start_pre() {
+	checkpath -d -o influxdb:influxdb -m755 /var/lib/influxdb
+	checkpath -d -o influxdb:influxdb -m755 /var/log/influxdb
```
clandmeter: Nitpicking.. but [checkpath](http://manpages.org/openrc-run/8#examples) accepts multiple directories.

testing/influxdb/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+_collectdver=0.2.0
+
+build() {
+	export GOPATH="$srcdir/gopath"
+
+	mkdir -p "$GOPATH/src/github.com/influxdata"
+	ln -s "$builddir" "$GOPATH/src/github.com/influxdata/influxdb"
+
+	cd "$GOPATH/src/github.com/influxdata/influxdb"
+	go get -v -d ./... || return 1
```
stwa: Thanks for your feedback! Hm, I can only find gogs and caddy in the repos, where gogs looks like it uses glide as default dependency management tool and caddy provides glide.yml with the APKBUILD. Should I better create glide.yml on the fly or also git commit it?

testing/influxdb/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+makedepends="go asciidoc xmlto"
+pkgusers="influxdb"
+pkggroups="influxdb"
+install="$pkgname.pre-install"
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.zip::https://github.com/influxdata/$pkgname/archive/v$pkgver.zip
+	influxdb.initd
+	influxdb.logrotate"
+builddir="$srcdir/$pkgname-$pkgver"
+options="!strip"
```
stwa: I did not know that, most of the packages in the repos still have !strip (at least the ones I looked at), so sorry :)

testing/influxdb/APKBUILD
```diff
@@ -0,0 +1,72 @@
...
+
+	mkdir -p "$GOPATH/src/github.com/influxdata"
+	ln -s "$builddir" "$GOPATH/src/github.com/influxdata/influxdb"
+
+	cd "$GOPATH/src/github.com/influxdata/influxdb"
+	go get -v -d ./... || return 1
+
+	git --git-dir="$GOPATH/src/collectd.org/.git" \
+		--work-tree="$GOPATH/src/collectd.org" \
+		checkout -f tags/v"$_collectdver" || return 1
```
clandmeter: I thought i already commented on this but why do you specifically need this tag?

testing/influxdb/APKBUILD
clandmeter: https://github.com/golang/go/issues/1242

testing/influxdb/APKBUILD
clandmeter: There are a few possible situations which we want to document see: https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Go

For now it would be nice if you could work something out within prepare to make sure we always fetch the same sources. (im not sure how well glide works with Godeps file)

testing/influxdb/APKBUILD
stwa: Because as you said, I don't use vendoring properly and the version on master isn't compatible with influxdb 1.1.0. Try to work out a glide based solution now...

stwa:
I made some changes and use glide to convert the provided Godeps file. Not sure about the version patch though, which is needed because else the build.py uses git information from my aports repo.

jirutka:
@7heo, could you please review this package once you have some time?

7heo:
yep. :)

clandmeter:
@stwa can you take a look at this [patch](http://tpaste.us/AWvY) against yours?

1. The provided godeps file is not complete so it also fetches other projects master branches.  We can overcome that by providing a glide_init function which we can run each time we bump the version.
2. I agree patching the build script is not the best approach so I just cloned the specific tag in the unpack function.
3. I made some changes to the gopath behavior

stwa:
@clandmeter Again, thanks for your help on this!

1. I agree as it makes the builds reproducible, but we still depend on master branches each time we bump the package. I guess we can't fix that. Anyway, providing glide.yml absolutely makes sense then.
2. It's best just to clone the git repo, but it feels needless to get the source code twice. Will there be a option in the future to tell abuild about git-urls directly?
3. Much better :)

I will commit your changes later.

clandmeter:
How would adding "gituri support to abuild" fix fetching the source twice?

stwa:
I meant something like git clone instead of wget in case one of the sources is known to be a git repository.

clandmeter:
Can you remove the patch?

stwa:
Done

PR #558, opened 2016-11-27
main/openrc: Let modloop workable in loop device

dlintw:
When boot in grub2 as (loop) device, the cmdline will cause the modloop fail, this patch fixed it.


jirutka:
Could you please provide me an example of `/proc/cmdline` when loop device is used?

dlintw:
for example in grub2's cfg file
```
menuentry '[loopback]alpine x86_64' {
        set isofile='/boot/iso/alpine-3.3.3-x86_64.iso'
        loopback loop $isofile
        linux (loop)/boot/vmlinuz-grsec modloop=/boot/modloop-grsec modules=loop,squashfs,sd-mod,usb-storage quiet
        initrd (loop)/boot/initramfs-grsec
}

```
Then the /proc/cmdline would be

BOOT_IMAGE=(loop)/boot/vmlinux-grsec modloop=/boot/modloop-grsec modules=loop,squashfs,sd-mod,usb-storage quiet

ncopa:
what is the error message you get? Is it the `eval` that fails or is it failing later, not finding the mod loop image?

dlintw:
The error message is:
/lib/rc/sh/openrc-run.sh: eval: line 1: syntax error: unexpected "("
* ERROR: modloop failed to start

main/openrc/modloop.initd
```diff
@@ -10,7 +10,7 @@ depend() {
 
 # read kernel options
 init_KOPT() {
-	eval set -- $(cat /proc/cmdline 2>/dev/null)
+	eval set -- $(cat /proc/cmdline 2>/dev/null|sed -e 's/(loop)//g')
```
fabled: @ncopa Why is eval used here first place? Isn't `set` enough? Could we just drop the `eval` instead of adding sed kludge?

ncopa: i think we can drop the `eval` yes. It was originally so we would handle options with spaces like `acpi_osi="!Windows 2006"`

jirutka:
Could you please properly rebase this PR against master and force push, to clean the mess?

dlintw:
I forgive this patch

PR #563, opened 2016-11-29
community/py-configargparse: upgrade to 0.11.0

fabaff:


community/py-configargparse/APKBUILD
```diff
@@ -3,13 +3,12 @@
...
-pkgver=0.10.0
-pkgrel=1
+pkgver=0.11.0
+pkgrel=0
 pkgdesc="A drop-in replacement for argparse"
-url="https://github.com/zorro3/ConfigArgParse"
+url="https://github.com/bw2/ConfigArgParse"
 arch="noarch"
 license="MIT"
-depends=""
```
jirutka: This variable is referenced in the `_py()` function (`depends="$depends $python"`), so if you want to remove it, you must remove `$depends` in `_py()` as well.

PR #565, opened 2016-11-29
testing/py-ujson: upgrade to 1.35 and enable py3

fabaff:


testing/py-ujson/APKBUILD
```diff
@@ -2,40 +2,44 @@
...
+_py() {
+	local python=$1
+	pkgdesc="$pkgdesc - $python"
+	install_if="$pkgname=$pkgver-r$pkgrel $python" 	
+	cd "$builddir"
+	$python setup.py install --prefix=/usr --root="$subpkgdir"
+}
+
+_py2() {
+	_py python2
```
jirutka: This may silently fail. There’s a reason why this is always the last statement in [examples on wiki](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python).

PR #568, opened 2016-11-29
testing/grafana: new aport

stwa:


stwa:
After #557 has been merged, I applied some of the improvements also to this package.

clandmeter:
I've tried to build this thing locally but im not so sure I like the build scripts on this one.

1. It downloads all sorts of build dependencies in build function
2. it tries to exec one of the downloads deps but fails and then tries to build it which pulls in multiple headers to $HOME...

Are we sure we want something like this in aports? I don't think i want to burn my hands on this.

jirutka:
I agree with @clandmeter.

stwa:
> It downloads all sorts of build dependencies in build function

I could move _npm install_ to prepare(), but that's the problem with npm: some dependencies are just downloaded, others are compiled (e.g. node-sass) and compiling stuff in prepare() feels wrong, too. I see no way to split this into two tasks.

> it tries to exec one of the downloads deps but fails and then tries to build it which pulls in multiple headers to $HOME...

That's phantomjs. I have no idea why they force someone (to try) to use their prebuilt binaries. As this dependency is only used for testing I could strip it from the shipped _package.json_.

> Are we sure we want something like this in aports? I don't think i want to burn my hands on this.

Welcome to the javascript build toolset :( _burn my hands_ sounds a bit strong, it could be worse and the devs from Grafana did a lot to clean up their build with the 4.0.x release (I had things like out-of-memory exceptions before when compiling the typescript stuff).
Anyway, I understand your objections, feel free to just close this PR.

clandmeter:
@stwa Sorry about my strong words, I do appreciate your contributions.
Does any other distribution package grafana?

jirutka:
@stwa If you need this package, then you can always create your own Aports repository – for packages that don’t fulfil requirements of the official repository. I wrote a [guide how to setup aports repository](https://github.com/jirutka/aports#how-to-setup-your-own-repository) with GitHub, Travis CI (for building packages) and some server with SSH access for hosting binary packages.

stwa:
@clandmeter No problem, it's always good to discuss things. Grafana is available in Arch, Gentoo, Ubuntu, just to name three big distributions.

@jirutka I already set up my own Aports repository a few months ago thanks to your guide (so this might be the right place to say thank you!).

I have this package in my custom repository, so we can close this PR and add the package anytime as soon as needed. I really don't *have to* maintain this package, just thought I contribute because I created it for my needs. Less packages, less work.

PR #573, opened 2016-11-30
testing/rkt: new aport

brosner:
https://github.com/coreos/rkt
container engine for Linux designed to be composable, secure, and built
on standards

### DIFF
--- a/testing/rkt/APKBUILD
+++ b/testing/rkt/APKBUILD
@@ -1,0 +1,46 @@
+# Contributor: Brian Rosner <brian@brosner.com>
+# Maintainer: Brian Rosner <brian@brosner.com>
+pkgname=rkt
+pkgver=1.20.0
+pkgrel=0
+pkgdesc="container engine for Linux designed to be composable, secure, and built on standards"
+url="https://github.com/coreos/rkt"
+arch="x86_64"
+license="ASL 2.0"
+depends=""
+depends_dev=""
+makedepends="$depends_dev coreutils grep gzip diffutils cpio go bc linux-headers wget file gnupg autoconf automake bash openssl-dev acl-dev xz sed"
+subpackages=""
+source="rkt-$pkgver.tar.gz::https://github.com/coreos/rkt/archive/v$pkgver.tar.gz"
+builddir="$srcdir/rkt-$pkgver"
+
+build() {
+	cd "$builddir"
+	./autogen.sh
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--enable-tpm=auto \
+		--enable-sdjournal=auto \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	mkdir -p "$pkgdir"/var/lib/$pkgname
+	install -Dm755 "build-rkt-$pkgver"/target/bin/rkt "$pkgdir"/usr/bin/rkt
+	mkdir -p "$pkgdir"/usr/lib/rkt/stage1-images
+	for flavor in fly coreos kvm; do
+		install -Dm644 "build-rkt-$pkgver"/target/bin/stage1-${flavor}.aci "$pkgdir"/usr/lib/rkt/stage1-images/stage1-${flavor}.aci
+	done
+}
+
+md5sums="15b28632ef2c50b473de59fb1bd9a85e  rkt-1.20.0.tar.gz"
+sha256sums="3b0a08971cc22004fe8367f8d3e35d5f66fcd0191802b4e35c5c273c1f772096  rkt-1.20.0.tar.gz"
+sha512sums="37c46a044838660eb22d74ff5cc16c7d9a29b178a5ccece97d484e4aebfd587a7558d665d2a2162adee859135330d4504a4dca60faf4ce5e3552633924c3b960  rkt-1.20.0.tar.gz"
+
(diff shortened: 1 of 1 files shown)


testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
+# Contributor: Brian Rosner <brian@brosner.com>
+# Maintainer: Brian Rosner <brian@brosner.com>
+pkgname=rkt
+pkgver=1.20.0
+pkgrel=0
+pkgdesc="container engine for Linux designed to be composable, secure, and built on standards"
+url="https://github.com/coreos/rkt"
+arch="x86_64"
```
jirutka: It doesn’t work on other platforms?

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+license="ASL 2.0"
+depends=""
+depends_dev=""
+makedepends="$depends_dev coreutils grep gzip diffutils cpio go bc linux-headers wget file gnupg autoconf automake bash openssl-dev acl-dev xz sed"
```
jirutka: This package does not provide -dev subpackage, so it’s not useful to declare and use `depends_dev` variable. abuild even warns about this. Please remove it and also wrap line on 80 characters.

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+subpackages=""
+source="rkt-$pkgver.tar.gz::https://github.com/coreos/rkt/archive/v$pkgver.tar.gz"
+builddir="$srcdir/rkt-$pkgver"
+
+build() {
+	cd "$builddir"
+	./autogen.sh
```
jirutka: Unfortunately APKBUILD does not run with `set -e` (yet, it’s on our TODO list to change it), so every statement that may fail, but should not fail, and is not the last statement of the function, should end with `|| return 1`.

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+		--localstatedir=/var \
+		--enable-tpm=auto \
+		--enable-sdjournal=auto \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	mkdir -p "$pkgdir"/var/lib/$pkgname
```
jirutka: This line is unnecessary, `"$pkgdir"/usr/lib/rkt/stage1-images` is created on line 39 (`install -D`).

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+	install -Dm755 "build-rkt-$pkgver"/target/bin/rkt "$pkgdir"/usr/bin/rkt
+	mkdir -p "$pkgdir"/usr/lib/rkt/stage1-images
+	for flavor in fly coreos kvm; do
```
jirutka: Variables used only in a single function should be declared as `local`, i.e.:

```sh
local flavor; for flavor in fly coreos kvm; do
...
```

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+		--enable-sdjournal=auto \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	mkdir -p "$pkgdir"/var/lib/$pkgname
+	install -Dm755 "build-rkt-$pkgver"/target/bin/rkt "$pkgdir"/usr/bin/rkt
+	mkdir -p "$pkgdir"/usr/lib/rkt/stage1-images
```
jirutka: This line is unnecessary, `"$pkgdir"/usr/lib/rkt/stage1-images` is created on line 39 (`install -D`).

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+	for flavor in fly coreos kvm; do
+		install -Dm644 "build-rkt-$pkgver"/target/bin/stage1-${flavor}.aci "$pkgdir"/usr/lib/rkt/stage1-images/stage1-${flavor}.aci
```
jirutka: Also too long line.

testing/rkt/APKBUILD
jirutka: Does it really need all these GNU utilities?

jirutka: We have switched to LibreSSL (since v3.5), so openssl-dev should be replaced with libressl-dev (if possible).

testing/rkt/APKBUILD
brosner: I suspect it does, but I am not able to test to confirm.

testing/rkt/APKBUILD
brosner: I can give this a try and report back if it works.

testing/rkt/APKBUILD
brosner: Noted. Will make the change. Thanks.

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+	./autogen.sh || return 1
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--enable-tpm=auto \
+		--enable-sdjournal=auto \
```
jirutka: Please do not rely on `auto`, replace these two with: ``--disable-tpm --disable-sdjournal`.

testing/rkt/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	install -Dm755 "build-rkt-$pkgver"/target/bin/rkt "$pkgdir"/usr/bin/rkt
+	local flavor; for flavor in fly coreos kvm; do
```
jirutka: I looked into what these “flavours” are and I think that it’d make sense to separate them into subpackages. What do you think?

clandmeter:
Closed due to inactivity.

PR #578, opened 2016-12-04
Switch to apparently more available SourceForge 1st-party mirror for xz

ostera:
Hi!

Over the past few days I've been battling this particular mirror (tukaani) as it's very very flaky. Perhaps it could be considered to use the sourceforge mirror that seems to be a first-party one? 😄


clandmeter:
I verified the url to this file and it works as expected also other dists like fedora and gentoo use the same so im going to close this.

PR #591, opened 2016-12-09
testing/vault: upgrading to 0.6.3

gena01:
* Updating to 0.6.3.
* Cleaning up vault.initd

### DIFF
--- a/testing/vault/vault.initd
+++ b/testing/vault/vault.initd
@@ -2,10 +2,11 @@
 
 VAULT_LOG_FILE="/var/log/${SVCNAME}.log"
 
+description="Vault is a tool for securely accessing secrets"
 command=/usr/sbin/${SVCNAME}
 command_args="${vault_opts}"
 command_background="true"
-start_stop_daemon_args="--user ${SVCNAME}:${SVCNAME}  --stdout $VAULT_LOG_FILE --stderr $VAULT_LOG_FILE"
+start_stop_daemon_args="-w 100 --user ${SVCNAME}:${SVCNAME}  --stdout $VAULT_LOG_FILE --stderr $VAULT_LOG_FILE"
 pidfile="/run/${SVCNAME}.pid"
 
 start_pre() {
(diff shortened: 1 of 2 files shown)


testing/vault/vault.initd
```diff
@@ -2,17 +2,28 @@
...
+
+start() {
+	ebegin "Starting ${SVCNAME}"
+	start-stop-daemon --start --quiet \
+		rm --pidfile /var/run/${name}.pid \
+		--user ${daemon_user} --group ${daemon_group} \
+		-b --stdout $VAULT_LOG_FILE --stderr $VAULT_LOG_FILE \
+		-k 027 --exec ${daemon} -- ${vault_opts}
+	eend $?
+}
```
jirutka: Why you did that?

gena01: I had issues with service start in my testing. It would assume that the service started and then report that it crashed (instead of start failed).

In my setup I setup "consul" service as backend. So if "consul" is not running vault should fail to start.

jirutka: Hm, `-b` is `--background`, so this should do exactly the same thing as the original code. o.O You’ve just moved pidfile to `/var/run` (but that should be symlink to `/run`) and added `-k 027` (that you can simply add to `start_stop_daemon_args=`). Also `rm` is certainly invalid here (`rm --pidfile /var/run/${name}.pid \`), I wonder how it can even work…

jirutka: Wait a moment, but you’ve also removed `pidfile` variable and `command`, so the default `stop` function cannot work!

gena01:
@jirutka updated. 

Adding a 100 ms wait seems to work for me. Also added a description.

testing/vault/vault.initd
gena01: I think i copied it from another APKBUILD. :( 

Changed things back and got it working in the original form. See updated commit. Are there docs on this?

gena01:
@jirutka any other changes you need me to make before this can be merged?

PR #593, opened 2016-12-10
testing/mapnik: new aport

mjuenema:
Added mapnik, mapnik-dev and mapnik-doc packages (http://mapnik.org/).


mjuenema:
Oops, forgot to remove using ccache for compiling.

PR #587, opened 2016-12-08
community/php7: Upgrade to 7.0.14

andypost:
Security release http://php.net/archive/2016.php#id2016-12-08-1

PR #595, opened 2016-12-10
community/php7: fix build for aarch64

andypost:
Follow-up for https://github.com/alpinelinux/aports/pull/587#issuecomment-265750816

Both asm commands works only with registers according docs
* https://developer.arm.com/docs/dui0801/latest/a64-general-instructions/mul
* https://developer.arm.com/docs/dui0801/latest/a64-general-instructions/smulh

So according patch https://bugs.php.net/bug.php?id=70015 and http://stackoverflow.com/questions/36763385/strange-behaviour-of-clang-assembler I added patch to change asm instructions to multiply with registers

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=php7
 _pkgreal=php
 pkgver=7.0.14
-pkgrel=1
+pkgrel=2
 pkgdesc="The PHP language runtime engine - 7th branch"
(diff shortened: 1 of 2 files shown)


andypost:
Related http://stackoverflow.com/questions/36763385/strange-behaviour-of-clang-assembler

andypost:
PHP upstream PR https://github.com/php/php-src/pull/2236

PR #597, opened 2016-12-10
main/cmake: upgrade to 3.7.1

pbregener:


ncopa:
we wait with this til after 3.5-stable is branched

pbregener:
Alright, I opened #602 to at least upgrade to the latest patch release of CMake 3.6 in the meantime

PR #598, opened 2016-12-10
community/php7: bundle php7-readline extension

andypost:
There's no way to keep `php7-readline` extension shared

So I suggest to include it into build like Redhat doing https://apps.fedoraproject.org/packages/php/sources/spec/
Also official PHP for Alpine docker image following the same pattern
https://github.com/docker-library/php/blob/master/7.0/alpine/Dockerfile#L114

Follow-up to https://github.com/alpinelinux/aports/pull/384

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=php7
 _pkgreal=php
 pkgver=7.0.14
-pkgrel=2
+pkgrel=3
 pkgdesc="The PHP language runtime engine - 7th branch"
 url="http://www.php.net/"
 arch="all"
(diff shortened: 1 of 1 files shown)


andypost:
Removal of readline-dev does not help
```
checking for libedit readline replacement... no
checking for readline support... yes
configure: error: Please reinstall readline - I cannot find readline.h
>>> ERROR: php7: all failed
```

andypost:
Changed `--with-readline=shared` to `--with-libedit` according to docs http://php.net/manual/en/readline.installation.php

andypost:
after local build still `PHP Warning:  PHP Startup: Invalid library (maybe not a PHP library) 'readline.so' in Unknown on line 0`

andypost:
Looking at https://apps.fedoraproject.org/packages/php/sources/spec/ I removed `readline` and added `libedit` so package will be bundled

jirutka:
Could you please add some summary of the required change into the commit message?

andypost:
updated commit message

PR #599, opened 2016-12-10
testing/mapnik: new aport

mjuenema:
Added Mapnik (http://mapnik.org/) packages.

### DIFF
--- a/testing/mapnik/APKBUILD
+++ b/testing/mapnik/APKBUILD
@@ -17,13 +17,13 @@ source="https://github.com/mapnik/mapnik/releases/download/$pkgver/mapnik-$pkgve
 builddir="$srcdir/$pkgname-$pkgver"
 build() {
 	cd "$builddir"
-	python scons/scons.py configure PREFIX=/usr CXX="ccache g++"
-	python scons/scons.py           PREFIX=/usr CXX="ccache g++" 
+	python scons/scons.py configure PREFIX=/usr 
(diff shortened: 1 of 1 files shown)


mjuenema:
The compile time for Mapnik is too long and the Travis-CI job fails.

testing/mapnik/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+builddir="$srcdir/$pkgname-$pkgver"
+build() {
+	cd "$builddir"
+	python scons/scons.py configure PREFIX=/usr 
+	python scons/scons.py           PREFIX=/usr
+}
+
+package() {
+	cd "$builddir"
+	python scons/scons.py install   PREFIX=/usr DESTDIR="$pkgdir"
```
fabled: For some reason, a rebuild of everything seems to happen here. It double the build time. Any ideas why it might happen?

testing/mapnik/APKBUILD
```diff
@@ -0,0 +1,44 @@
+# Contributor: Markus Juenemann <markus@juenemann.net>
+# Maintainer: Markus Juenemann <markus@juenemann.net>
+pkgname=mapnik
+pkgver=v3.0.12
```
fabled: Generally we don't want 'v' in the version number. Could we remove it here, and add it elsewhere? Or introduce some _pkgver=v$pkgver and use _pkgver elsewhere.

PR #600, opened 2016-12-11
testing/peervpn: new aport

mjuenema:
Added package for PeerVPN (https://peervpn.net/)


testing/peervpn/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+pkgdesc="PeerVPN builds virtual ethernet networks between multiple computers"
+url="https://peervpn.net"
+arch="all"
+license="GPL3"
+depends="openssl zlib"
+makedepends="$depends openssl-dev zlib-dev linux-headers"
+install=""
+subpackages=""
+source="
+        $url/files/$pkgname-0-044.tar.gz
```
fabled: We normally don't use $url here, because $url goes to package database. Maybe just expand it here?
You could also use ${pkgver/./-} or similar for the version.

testing/peervpn/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+	peervpn.initd
+	"
+
+# peervpn has a strange versioning scheme.
+builddir="$srcdir/$pkgname-0-044"
+
+build() {
+
+	# Must copy these into fakeroot?
+	install -m 755 $pkgname.initd "$builddir"/$pkgname.initd
```
fabled: This should not be needed in build step. Install it directly in package().

testing/peervpn/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+package() {
+ 	cd "$builddir"
+
+	install -Dm755 $pkgname "$pkgdir"/usr/sbin/$pkgname \
+		|| return 1
+
+	install -Dm600 $pkgname.conf "$pkgdir"/etc/$pkgname.conf \
+		|| return 1
+
+	install -Dm755 $pkgname.initd "$pkgdir"/etc/init.d/$pkgname \
```
fabled: Use $srcdir for the source file so it get's picked out without the extra install step in build().

jirutka:
What the heck happened to this PR? Why are there dozens of unrelated commits?

mjuenema:
No idea, it somehow got completely mangled. I am going to create a completely new branch and PR to untangle it all.

jirutka:
Please **don’t** create a new PR, just force push into this one.

mjuenema:
I might have pulled from the wrong repo and polluted the branch with commits from other branches.

PR #608, opened 2016-12-14
testing/hplip: upgrade to 3.16.11, enable scanning

vakartel:
sane support, Bug #6520
lite install (printer and scanner drivers only)
udev-related data moved to -udev subpackage
removed python dependency


vakartel:
found an existing unmaintained/sane port. I'll use it instead of creating new one.

So this PR have no sense for now

PR #609, opened 2016-12-14
community/postsrsd: new aport

Ikke:
Daemon to apply SRS for postfix, which is necessary when relaying messages and still want SPF to pass.

### DIFF
--- a/community/postsrsd/APKBUILD
+++ b/community/postsrsd/APKBUILD
@@ -1,0 +1,52 @@
+# Contributor: Kevin Daudt <ops@ikke.info>
+# Maintainer: Kevin Daudt <ops@ikke.info>
+pkgname=postsrsd
+pkgver=1.4
+pkgrel=0
+pkgdesc="Postfix Sender Rewriting Scheme daemon"
+url="https://github.com/roehling/postsrsd"
+arch="all"
+license="GPL2"
+depends=""
+makedepends="cmake help2man"
+pkgusers="postsrsd"
+install="$pkgname.pre-install"
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/roehling/postsrsd/archive/$pkgver.tar.gz
+	10-fix-defaults.patch
+	postsrsd.initd"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+
+	mkdir build
+	cd build
+
+	cmake .. -DCMAKE_INSTALL_PREFIX=/usr/ \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DGENERATE_SRS_SECRET=OFF \
+		-DCONFIG_DIR=/etc/postsrsd || return 1
+	make all
+}
+
+package() {
+	cd "$builddir"
+
+
+	make DESTDIR="$pkgdir/" install
+
+	install -d "$pkgdir/etc/postsrsd/"
+	install -Dm0644 build/postsrsd.default "$pkgdir/etc/conf.d/postsrsd"
+	install -Dm0755 "$srcdir/postsrsd.initd" "$pkgdir/etc/init.d/postsrsd"
+}
+
+md5sums="cb6e13a06d6bbd7d383efb9bbf0867ce  postsrsd-1.4.tar.gz
+a4104dbe7c6bc88de6a1fa29ee82e531  10-fix-defaults.patch
+575d5c5563def2aabdd752abc54b5e45  postsrsd.initd"
+sha256sums="f3230d57c36ba4688cc3beb90fbb0f199a06381e4df2edbc4ef541a9e8a574ff  postsrsd-1.4.tar.gz
+4e8093e8a90f499402f17ea9be253c499f47218bc7918d58825f6a1e2daa2543  10-fix-defaults.patch
+e59e9193998041023f6ef2740fd8ddaa2903d2a88f39e14f638c5ca2ce023b3b  postsrsd.initd"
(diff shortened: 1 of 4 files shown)


community/postsrsd/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+	postsrsd.init"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+
+	mkdir build
+	cd build
+
+	cmake .. -DCMAKE_INSTALL_PREFIX=/usr/ -DCMAKE_BUILD_TYPE=Release -DGENERATE_SRS_SECRET=OFF -DCONFIG_DIR=/etc/postsrsd
```
clandmeter: Please try to break lines at 80 chars. For readability you could also break after each cmake options.
And always return 1 if its not the last command in a function (we do not set -e).

community/postsrsd/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+	make all
+}
+
+package() {
+	cd "$builddir"
+
+
+	make DESTDIR="$pkgdir/" install
+
+	install -Dm0400 -o postsrsd -g root /dev/null "$pkgdir/etc/postsrsd/postsrsd.secret"
```
clandmeter: This part will prevent your post-install to run?
You probably don't have to specify -g root as the current user/group is already root.
Is this the correct permission to use? I would think 0600 would be more logical.

community/postsrsd/postsrsd.post-install
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Generate a secret that's needed for postsrsd to work
+if [ ! -s /etc/postsrsd/postsrsd.secret ]
+then
+    dd if=/dev/urandom bs=18 count=1 status=none | base64 > /etc/postsrsd/postsrsd.secret
```
clandmeter: Please see comment above.

community/postsrsd/postsrsd.post-install
```diff
@@ -0,0 +1,8 @@
...
+fi
+
```
clandmeter: install scripts should always exit 0

community/postsrsd/postsrsd.pre-install
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+addgroup -S postsrsd 2>/dev/null
+adduser -S -D -H -s /bin/false -G postsrsd -g postsrsd postsrsd 2>/dev/null
```
clandmeter: install scripts should always exit 0

community/postsrsd/postsrsd.init
```diff
@@ -0,0 +1,12 @@
+#!/sbin/openrc-run
+
+description="Postfix SRS Daemon"
+pidfile="/var/run/postsrsd.pid"
```
clandmeter: the default run path is now /run

community/postsrsd/APKBUILD
Ikke: What do you mean with the first question? I copied the 0400 over from the example I used, but I agree, 0600 makes more sense.

community/postsrsd/postsrsd.post-install
Ikke: Just to be sure, what comment are you referring to here?

community/postsrsd/APKBUILD
clandmeter: Oh sorry I read it as -e...

What about adding both logic to the init script with pre_install, i think we do similar for sshd.

community/postsrsd/postsrsd.post-install
clandmeter: Sorry I read it as -e...

Ikke:
@clandmeter Latest push should have fixed the remarks you made. Please let me know what you think.

community/postsrsd/postsrsd.initd
```diff
@@ -0,0 +1,29 @@
+#!/sbin/openrc-run
+
+description="Postfix SRS Daemon"
+pidfile="/run/postsrsd.pid"
+command="/usr/sbin/postsrsd"
+command_args="-D -d ${SRS_DOMAIN} -s ${SRS_SECRET} -f ${SRS_FORWARD_PORT}
+              -r ${SRS_REVERSE_PORT} -p \"${pidfile}\" -u ${RUN_AS}
+              -c \"${CHROOT}\" -a \"${SRS_SEPARATOR}\""
```
jirutka: You should set sensible defaults in initd file, if possible.
When you define so many custom variables, then you should also add commented confd file.
Please prefer lower case variable names.

Ikke: @jirutka Those defaults (names) are coming from the [project](https://github.com/roehling/postsrsd/blob/master/init/postsrsd.default.in). Most of the defaults are defined in there too. The only thing you are required to set yourself is the SRS domain.

Do you think it makes sense to copy these defaults over in the initd file?

clandmeter: @Ikke I think what @jirutka means is if you comment one of the options in the confd file the initd will surely fail because it will apply an option without a value. You can use "Parameter Substitution" to solve this and let the confd file to override it.

A user should never have to edit the init.d file and should only edit the confd file. So if some of the options should be able to be removed you will have to add the logic to the initd file depending if the option is set or not.

Ikke: @jirutka @clandmeter I've updated the postsrsd confd file to include defaults now. The only thing we cannot provide defaults for is `SRS_DOMAIN`, users have to provide that their selves.

The capitalization of the variables comes from upstream, so I'm just referencing them.

clandmeter:
@Ikke what is the reason to directly move this to community instead of testing?

Ikke:
@clandmeter I'm already using the package myself and verified it works. If you still think it's better to have it in testing first, I'll move it there.

clandmeter:
if you verified its correct then im ok for adding it to community.

PR #620, opened 2016-12-20
testing/ceph: upgrade to 11.1.1

dx9:


testing/ceph/APKBUILD
```diff
@@ -114,59 +117,73 @@ _sysconfdir=/etc
 _udevrulesdir=/etc/udev/rules.d
 _python_sitelib=/usr/lib/python2.7/site-packages
 
-
-_builddir="$srcdir"/$pkgname-$pkgver
+_builddir=$srcdir/$pkgname-$pkgver
```
dx9: 😄  nope this if cruft, builddir works fine now. tks

testing/ceph/APKBUILD
```diff
@@ -114,59 +117,69 @@ _sysconfdir=/etc
...
 package() {
-	cd "$_builddir"
-	cd build
-	make DESTDIR="$pkgdir" install
+	cd $builddir/build
+	make DESTDIR=$pkgdir install || return 1
 	cd ..
 
-	rm -f "$pkgdir"$_sysconfdir/init.d/ceph
+	rm -f $pkgdir$_sysconfdir/init.d/ceph
```
jirutka: Why have you removed quotes around pkgdir? Variables that contain externally provided paths (like `builddir`, `pkgdir`, `srcdir`, …) should be quoted.

testing/ceph/APKBUILD
```diff
@@ -178,44 +191,18 @@ base() {
...
-
-	mkdir -p "$subpkgdir"$_sysconfdir/logrotate.d
-	mv "$pkgdir"$_sysconfdir/logrotate.d/ceph \
-		"$subpkgdir"$_sysconfdir/logrotate.d/ceph
-
-	mkdir -p "$subpkgdir"$_python_sitelib
-	mv "$pkgdir"$_python_sitelib/ceph_detect_init* "$subpkgdir"$_python_sitelib
-	mv "$pkgdir"$_python_sitelib/ceph_disk* "$subpkgdir"$_python_sitelib
-
+	_pkg $_docdir/ceph sample.ceph.conf sample.fetch_config
```
jirutka: Unfortunately APKBUILD does not run with `set -e` (yet, it’s on our TODO list to change it), so every statement that may fail, but should not fail, and is not the last statement of the function, should end with `|| return 1`.

testing/ceph/APKBUILD
```diff
@@ -89,16 +88,20 @@ subpackages="
...
+	py-rgw:py_rgw
 	libradosstriper
 	libradosstriper-dev:libradosstriper_dev
 	py-rbd:py_rbd
 	py-cephfs:py_cephfs
 "
 
-if [ "$WITH_TESTS" = ON ]; then
+if [ -n "$CEPH_TEST_PKG" ]; then
 	subpackages="$subpackages $pkgname-test:ceph_test"
```
jirutka: Why is this conditional?

testing/ceph/APKBUILD
```diff
@@ -114,59 +117,69 @@ _sysconfdir=/etc
...
-	   -DWITH_PYTHON3=OFF \
-	   -DWITH_TESTS=$WITH_TESTS \
-	   || return 1
-   make
+	export CEPH_BUILD_VIRTUALENV=$builddir
+
+	mkdir -p $builddir/build
+	cd $builddir/build
+
+	if [ -n "$MAKE_CHECK" ] || [ -n "$CEPH_TEST_PKG" ]; then
```
jirutka: Why is this conditional?

testing/ceph/APKBUILD
dx9: right, my bad

testing/ceph/APKBUILD
dx9: exit 1 is called in _pkg on failure. Is that ok?

testing/ceph/APKBUILD
dx9: So ya this is backwards. Test package should build by default.

testing/ceph/APKBUILD
dx9: Ya this is not right. WITH_TESTS=off is faster, however that's irrelevant if test package is built by default. I'll ditch the "WITH_TEST" build optionality.

testing/ceph/APKBUILD
jirutka: Definitely not, I overlooked that. Please never call `exit` in APKBUILD.

dx9: ok

andypost:
latest release is 12.2.4 also depends on boost, ref #2960

PR #624, opened 2016-12-21
main/postgresql: add support for PL/Perl, PL/Python and PL/Tcl

GJRTimmer:
https://www.postgresql.org/
A sophisticated object-relational DBMS


GJRTimmer:
Additional PL/* Languages as separate installable package(s)
- PL/Perl
- PL/Python
- PL/Tcl

GJRTimmer:
@jirutka Review information

You can find the builded packages at my test repository: http://pkgs.timmertech.nl/main
If you don't want to install with the parameter --allow-untrusted you can find the public key of the APKINDEX @ http://pkgs.timmertech.nl/keys

# Key Download
```bash
wget http://pkgs.timmertech.nl/keys/gjr.timmer%40gmail.com-5857d36d.rsa.pub
```
# Install packages: (handy for copy-paste :-) )
```bash
apk add postgresql postgresql-plperl postgresql-plpython postgresql-pltcl
```

# To activate the extensions
1) Create postgresql database with ```initdb```
2) Run postgres
3) login into database with ```psql```
- PL/Perl: ```CREATE EXTENSION plperl;```
- PL/Python: ```CREATE EXTENSION plpython2u;```
- PL/Tcl: ```CREATE EXTENSION pltcl;```

# Example for PL/Tcl
```sql
CREATE FUNCTION tcl_max(integer, integer) RETURNS integer AS $$
    if {$1 > $2} {return $1}
    return $2
$$ LANGUAGE pltcl STRICT;
```

# Final note:
I hope this additional information will help you with your review.
extension-list
create-extension-success

jirutka:
The added PL modules also install the following files that ends in -contrib subpackage. I’d prefer to move them into PL subpackages, WDYT?
 
```
/usr/lib/postgresql/hstore_plperl.so
/usr/lib/postgresql/hstore_plpython2.so
/usr/lib/postgresql/ltree_plpython2.so
/usr/share/postgresql/extension/hstore_plperl--1.0.sql
/usr/share/postgresql/extension/hstore_plperl.control
/usr/share/postgresql/extension/hstore_plperlu--1.0.sql
/usr/share/postgresql/extension/hstore_plperlu.control
/usr/share/postgresql/extension/hstore_plpython2u--1.0.sql
/usr/share/postgresql/extension/hstore_plpython2u.control
/usr/share/postgresql/extension/hstore_plpython3u--1.0.sql
/usr/share/postgresql/extension/hstore_plpython3u.control
/usr/share/postgresql/extension/hstore_plpythonu--1.0.sql
/usr/share/postgresql/extension/hstore_plpythonu.control
/usr/share/postgresql/extension/ltree_plpython2u--1.0.sql
/usr/share/postgresql/extension/ltree_plpython2u.control
/usr/share/postgresql/extension/ltree_plpython3u--1.0.sql
/usr/share/postgresql/extension/ltree_plpython3u.control
/usr/share/postgresql/extension/ltree_plpythonu--1.0.sql
/usr/share/postgresql/extension/ltree_plpythonu.control
```

PL/Python supports both Legacy Python and Python 3, but currently it’s built only for Legacy Python. Does the build system allows to build two flavours of PL/Python in once? If not, would you mind to change it to Python 3 only?

Thanks for very thorough review information!

main/postgresql/APKBUILD
```diff
@@ -119,6 +122,46 @@ contrib() {
 	rmdir -p "$subpkgdir"/usr/share/doc/postgresql || true
 }
 
+plperl() {
+	depends="postgresql perl"
```
jirutka: Dependency on perl is automatically detected by abuild via dynamic links, so it’s not needed (and desirable) to specify it here. The same applies even for plpython and pltcl.

```
>>> postgresql-plperl*: Tracing dependencies...
	postgresql
	perl
	busybox
	so:libc.musl-x86_64.so.1
-->	so:libperl.so  <--
```

GJRTimmer: I will remove the hardcoded depends. [DONE]

To my knowledge it can only be build against 1 version, I will start research to see If both are possible, If not I will switch it to Python3.

I concur with you about the ```-contrib```, the question that remains is, either a ```postgresql-pl-contrib``` or ```postgresql-plperl-contrib``` etc.. WDYT ?

GJRTimmer:
@jirutka I'm reading through the postgresql documentation, so far it will be possible to compile both Pyhton2 and Python3, however just a headsup, PostgreSQL still considers python2 to be default, regardless of what the OS independent /usr/bin/python points to.

This means that if a user loads the extension ```plpythonu``` it will load the python2 variant.

Reference: https://www.postgresql.org/docs/9.0/static/plpython-python23.html

I will attempt to compile both, this will require a double pass compilation, something I have never done before on alpine.

GJRTimmer:
@jirutka about your contrib idea, I really like it, the question that remains is, should we go for:
Option 1:
  - ```postgresql-pl-contrib```

Option 2:
  - ```postgresql-plperl-contrib```
  - ```postgresql-plpython2-contrib```
  - ```postgresql-plpython3-contrib```
  - ```postgresql-pltcl-contrib```

Either one is fine by me, but because users might only install for example the ```postgresql-pltcl``` extension, going with option 2 makes more sense to me, WDYT ?

GJRTimmer:
@jirutka 
I had to split the PL/Python into separate build(s), to compile for Python2 and Python3, created additional sub -dev and -contrib.

To make everything nice and consistent I've done the same for the other packages.

Packages:
  - postgresql-plperl
  - postgresql-plperl-dev
  - postgresql-plperl-contrib
  - postgresql-plpython2
  - postgresql-plpython2-dev
  - postgresql-plpython2-contrib
  - postgresql-plpython3
  - postgresql-plpython3-dev
  - postgresql-plpython3-contrib
  - postgresql-pltcl

For PL/Tcl there is no -contrib and -dev; this is correct and has been verified.

Confirmation & test see screenshot below.

pl-languages-loaded

GJRTimmer:
@jirutka All done, request review / approval.

GJRTimmer:
final fix between postgresql-plpython2-contrib and postgresql-plpython3-contrib

@jirutka could you please review ?

jirutka:
I'm sorry for long delay, I'm very busy now. :(

tl;dr: I really don't like splitting postgresql pkg into multiple APKBUILDs, that's wrong approach and I think that not really needed in this case. But don't worry about it, I'll try to rewrite your solution into single APKBUILD.

jirutka:
I'm sorry for long delay, I'm very busy now. :(

tl;dr: I really don't like splitting postgresql pkg into multiple APKBUILDs, that's wrong approach and I think that not really needed in this case. But don't worry about it, I'll try to rewrite your solution into single APKBUILD.

GJRTimmer:
how is that possible; we need  separate compilation pass, with it's own full APK process to complete the entire build ?

If you can do this, I really want to learn how, because If tried multiple solutions before resulting to this, I agree it's not pretty. But otherwise the correct packages will not be compiled.

Is there any wiki / knowledge base online where I can learn to perform a double APK build process including the rootpkg process before the second compilation pass ?

Good luck, I'm really looking forward to the solution.

I've just created a PR for  why I started upgrading the base postgres first and that is postgres-xl, scalable postgres cluster, I want to be able to use it within a docker container with alpine. But I had to do this one first.

That PR https://github.com/alpinelinux/aports/pull/783 is also multiple APKBUILD files which I will update as soon I've learned from you how.

GJRTimmer:
@jirutka I've played around with PL/Tcl it turn out that if you want to use the pltcl_* scripts within this package it has a tcl dependency on the tcl library Pgtcl, (the capital is part of tcl, not required for building), so I've added the APKBUILD file for this dependency.

jirutka:
I’ve fixed and merged your pgtcl abuild into testing and gonna move it into main before merging this update.

I started with refactoring of this PR in #805 (still WIP).

jirutka:
Replaced by #805.

PR #631, opened 2016-12-22
unmaintained/sane: upgrade to 1.0.25 and move to testing

vakartel:
add maintainer
add post-install group 'scanner' and user 'saned' creation
add 'saned' init.d script
move 'saned' network server binary to -saned subpackage
move binaries to -utils subpackage
backend drivers are split to subpackages

PR #632, opened 2016-12-22
testing/hplip: upgrade to 3.16.11

vakartel:
add scanner (sane backend) support, close #6520
make lite install (libs and drivers only)
remove fax-, hal-, systemd-, dbus- related trash
remove python dependency
move libs to subpackage


vakartel:
checks have failed because travis do not see unmaintained/sane package. 
https://github.com/alpinelinux/aports/pull/631

clandmeter:
Did you read our guidelines? It should be provided when you create a new PR:

https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md

* Please bundle deps into the same PR so travis can build it properly.
* Also move sane directly to testing (in a separate commit) same PR.
* Add your changes commit to sane also to this PR (easier for us to read changes when split from the move).
* Add your hplip upgrade/changes commit.

This way travis will work and we will be able to apply it to aports in one shot.

vakartel:
It builds well on my localhost but travis check failed.
I think travis ignores packs from unmaintained for its checks.

vakartel:
close/open to trigger travis check

PR #633, opened 2016-12-23
community/opendkim: fixes in build, close #6122

vakartel:
fix user creation and db directory permissions
fix init.d script - explictly set process-owner and pid-file.
move non-server binaries and perl-depended script to -utils subpackage


community/opendkim/opendkim.post-install
```diff
@@ -0,0 +1,23 @@
...
+	chmod 600 $keyfile
+	chown opendkim $keyfile
+	[ "$domain" ] || exit 0
+	echo -en "$selector._domainkey IN TXT (\n\"v=DKIM1; k=rsa;\"\n\"p=" > $keyname.txt
+	grep -v ^- $keyname.public | tr -d '\n' >> $keyname.txt
+	echo -e "\"\n) ; --- DKIM key $selector for $domain" >> $keyname.txt
+	rm $keyname.public
+fi
+
+exit 0
```
jirutka: The post-install script is run right after the package is installed, i.e. _before_ user have a chance to configure it. Therefore post-install script is a not a right place for this. It’d be better to create a standalone setup script or add setup function into the init script.

community/opendkim/opendkim.initd
```diff
@@ -1,16 +1,17 @@
 #!/sbin/openrc-run
 
+owner=opendkim
 pidfile=/run/opendkim/opendkim.pid
-
+cfgfile=/etc/opendkim/opendkim.conf
 command=/usr/sbin/opendkim
-command_args="-f ${EXTRA_OPTS}"
-command_background=yes
+command_args="$command_args -u $owner -P $pidfile"
```
jirutka: Why have you changed it? The current code does not work? It’s generally better and conceptually cleaner to let the daemonization job and handling of pidfile to rc system. It’s just a relict of shitty SysVinit that many programs automatically daemonizes itself and creates pidfiles (actually the whole concept of pidfile is just a legacy).

community/opendkim/APKBUILD
```diff
@@ -3,55 +3,63 @@
...
-install="$pkgname.pre-install"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-libs $pkgname-dbg $pkgname-utils"
+install="$pkgname.pre-install $pkgname.post-install"
 source="http://downloads.sourceforge.net/project/$pkgname/$pkgname-$pkgver.tar.gz
-	opendkim.conf
-	opendkim.initd"
-builddir="$srcdir/$pkgname-$pkgver"
+	$pkgname.confd
+	$pkgname.initd"
+builddir="$srcdir"/$pkgname-$pkgver
```
jirutka: Why have you changed this line?

community/opendkim/APKBUILD
```diff
@@ -3,55 +3,63 @@
...
+		--sysconfdir=/etc/$pkgname \
 		|| return 1
 	make || return 1
 }
 
 package() {
 	cd "$builddir"
-
+	# add libressl to depends. need for post-install script
+	depends="$depends libressl"
```
jirutka: Please do not override `depends` variable inside a `package()` function.

community/opendkim/opendkim.post-install
vakartel: I do it the same as in other distros. Private key file must exists to start service.
The standalone perl-based key generation script 'opendkim-genkey' is on opendkim-utils subpackage.
Maybe the better way to move it to post-upgrade step to regenerate keys with "apk fix opendkim"

community/opendkim/opendkim.initd
vakartel: Oh, I thought exactly the opposite. Everywhere I tried to make the service itself generate pidfile.
bad news for me ... ((

community/opendkim/opendkim.post-install
jirutka: You cannot generate private key _before_ opendkim is configured for particular domain(s). It’s totally okay that opendkim fails to start when it’s not configured. Alpine is not Debian, we do not foolishly auto-start services right after installation, before user can configure them.

No, `apk fix opendkim` is not intended for generating configuration!

community/opendkim/APKBUILD
vakartel: I used old-style apkbuild template. As I see in current one the whole line is quoted.

community/opendkim/APKBUILD
vakartel: really do not need it here. Both the main package and -utils subpackage depends on libressl

PR #641, opened 2016-12-26
testing/postgresql-cron: new aport

GJRTimmer:
https://github.com/citusdata/pg_cron
cron-based job scheduler for PostgreSQL (9.5 or higher)


testing/postgresql-cron/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+url="https://github.com/citusdata/pg_cron"
+arch="all"
+license="custom"
+depends="postgresql"
+makedepends="postgresql-dev"
+install="postgresql-cron.post-install"
+pkgusers="postgres"
+pkggroups="postgres"
+subpackages=""
+source="https://github.com/citusdata/pg_cron/archive/v1.0.0-rc.1.tar.gz"
```
jirutka: Source files must have unique name across all aports. Also do not hard-code version number, use variable, to make bumping version easier.

```diff
- source="https://github.com/citusdata/pg_cron/archive/v1.0.0-rc.1.tar.gz"
+ source="$pkgname-$pkgver.tar.gz::https://github.com/citusdata/pg_cron/archive/v$_pkgver.tar.gz"
```

testing/postgresql-cron/APKBUILD
```diff
@@ -0,0 +1,43 @@
+# Contributor: G.J.R. Timmer <gjr.timmer@gmail.com>
+# Maintainer:
+pkgname=postgresql-cron
+pkgver=1.0.0
```
jirutka: Aports must track upstream version and that’s 1.0.0-rc.1, not 1.0.0. However, `1.0.0-rc.1` is not valid version format in APK, `pkgver` must be `1.0.0_rc1` in this case.

You can do this

```sh
pkgver=1.0.0_rc1
_pkgver=${pkgver/_rc/-rc.}
```
and use `_pkgver` inside `source` and `builddir`.

testing/postgresql-cron/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+license="custom"
+depends="postgresql"
+makedepends="postgresql-dev"
+install="postgresql-cron.post-install"
+pkgusers="postgres"
+pkggroups="postgres"
+subpackages=""
+source="https://github.com/citusdata/pg_cron/archive/v1.0.0-rc.1.tar.gz"
+
+builddir="$srcdir/pg_cron-1.0.0-rc.1"
```
jirutka: The same as above, use variables.

testing/postgresql-cron/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+
+prepare() {
+        default_prepare || return 1
+        cd "$builddir"
+
+	# Remove -Werror from Makefile
+	# Required to compile on alpine, to ignore compilation warnings
+        sed "s/-Werror //" -i Makefile || return 1
```
jirutka: Please use only tabs for indentation.

testing/postgresql-cron/APKBUILD
```diff
@@ -0,0 +1,43 @@
+# Contributor: G.J.R. Timmer <gjr.timmer@gmail.com>
+# Maintainer:
+pkgname=postgresql-cron
```
jirutka: We prefer to follow the upstream name where possible, and that’s pg_cron, not postgresql-cron.

GJRTimmer:
@jirutka Review changes completed, check verify, do you also require an additional squash of commits ?

PR #642, opened 2016-12-26
main/linux-grsec: Add sdhci-acpi module

Nicop06:
I need this driver on my Asus C300 to detect internal hard drive. Although installing Linux is not standard on this device, there might be other devices with similar hardware that require it.

### Testing

The hard drive is a _MMC drive SEM32G_. I tested by building a custom x86_64 iso with the modified kernel and the drive is now correctly detected as `/dev/mmcblk0` (module auto loaded at boot time).

### What wasn't tested?

* I didn't test he x86 build.
* I didn't complete the installation process as the setup freezes, I will still need to investigate. I don't think this is a kernel problem.


Nicop06:
This is weird, Travis doesn't seem to detect the change. Did I forget something? I read a few documents and I can't figure out what I've missed.

PR #647, opened 2016-12-28
testing/iprange: new aport

jonnyt886:
https://github.com/firehol/iprange
manages IP ranges (part of FireHOL)

### DIFF
--- a/testing/iprange/APKBUILD
+++ b/testing/iprange/APKBUILD
@@ -8,25 +8,14 @@ url="https://github.com/firehol/iprange"
 arch="all"
 license="GPL2"
 depends=""
-depends_dev=""
-makedepends="$depends_dev"
 install=""
 subpackages="$pkgname-doc"
 source="https://firehol.org/download/iprange/releases/v1.0.3/iprange-1.0.3.tar.bz2"
 
-_builddir="$srcdir"/$pkgname-$pkgver
-prepare() {
-	local i
-	cd "$_builddir"
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
-}
+builddir="$srcdir"/$pkgname-$pkgver
 
 build() {
-	cd "$_builddir"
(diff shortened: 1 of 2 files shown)


testing/iprange/APKBUILD
```diff
@@ -0,0 +1,56 @@
...
+arch="all"
+license="GPL2"
+depends=""
+depends_dev=""
+makedepends="$depends_dev"
+install=""
+subpackages="$pkgname-doc"
+source="https://firehol.org/download/iprange/releases/v1.0.3/iprange-1.0.3.tar.bz2"
+
+_builddir="$srcdir"/$pkgname-$pkgver
```
fabled: We use ''builddir" instead now.

testing/iprange/APKBUILD
```diff
@@ -0,0 +1,56 @@
...
+prepare() {
```
fabled: abuild in 3.5-stable and edge now provides default_prepare that applies patches. This can be removed.

testing/iprange/APKBUILD
```diff
@@ -0,0 +1,56 @@
...
+		|| return 1 
+	make || return 1 
+}
+
+package() {
+	cd "$_builddir"
+	make DESTDIR="$pkgdir" install || return 1
+}
+
+#doc() {
```
fabled: Default_doc is also provided so this can be removed.

testing/iprange/APKBUILD
```diff
@@ -0,0 +1,56 @@
...
+pkgname=iprange
+pkgver=1.0.3
+pkgrel=0
+pkgdesc="manages IP ranges (part of FireHOL)"
+url="https://github.com/firehol/iprange"
+arch="all"
+license="GPL2"
+depends=""
+depends_dev=""
+makedepends="$depends_dev"
```
jirutka: There’s no -dev subpackage, so variable `depends_dev` should not be declared and used. abuild even warns about this…

jonnyt886:
bump - changes addressed

PR #662, opened 2016-12-28
main/poppler: upgrade to 0.50.0

misery:


jirutka:
⚠️ **Note for maintainers:** There are some so name differences.

PR #663, opened 2016-12-28
main/nettle: upgrade to 3.3

misery:


jirutka:
⚠️ **Note for maintainers:** There are some so name differences.

misery:
http://lists.lysator.liu.se/pipermail/nettle-bugs/2016/003145.html

The shared library names are libnettle.so.6.3 and libhogweed.so.4.3, with sonames still libnettle.so.6 and libhogweed.so.4. It is intended to be fully binary compatible with nettle-3.1.

PR #668, opened 2016-12-30
testing/awscli: upgrade to 1.11.34, switch to python3

vakartel:


fabled:
Looks good, could you flatten the changes py-rsa to be one commit?

PR #684, opened 2017-01-03
testing/idris: new aport

mitchty:
Add testing/idris, a functional programming language with dependent types. Built with libffi and gmp bindings. Doesn't depend on haskell/ghc after its built, but does need ghc to be built.


mitchty:
Updated to the release of 1.0 today, also switched the compilation to just use cabal and not any cabal freeze files.

mitchty:
A note on this port, I'm using cabal to do what it does in downloading and installing haskell dependencies. I know thats generally frowned upon in build(), however, we would be haing to add approximately 100 transitive dependencies to sources.

In talk on irc, there is desire for languages with their own package managers, n.b. rust, to not have to have each language dependency replicated in apk.

So in a way my question on this port is this OK for an initial testing port? Or shall I rework this to work with cabal freeze files to basically save the cabal package managers dependency resolution to make sure that exactly only the versions we tested with are installed?

Note, no matter what if we us cabal to do package dependency, we need to have it download a view of the current package db from hackage.haskell.org. Either we make an APKBUILD that is a snapshot of the hackage package db, or we just download the latest each time. But one way or another cabal needs knowledge of every haskell module to reliably build a dependency graph.

Or as stated we can go the freeze file rout which would be conceptually similar to the ruby bundler method of calculate dependencies once then force cabal to use that.

There are pros and cons to both approaches, but I know @jirutka was looking to do some work relating to rust and cargo, haskell  built packages in this regard would behave similarly.

But since this is a compiler that happens to be implemented in haskell, I'm hoping this hacky way of doing things is acceptable for at least an initial port even with its possible flaws in how it gets built.

jirutka:
It fails to build on my system. Probably due to grsecurity?

```
Preprocessing library libffi-0.1...
creating dist/build/Foreign/LibFFI
creating dist/build/Foreign
creating dist/build/Foreign/LibFFI
/usr/bin/hsc2hs '--cc=/usr/bin/gcc' '--ld=/usr/bin/gcc' '--cflag=-fno-stack-protector' '--lflag=-fno-stack-protector' '--cflag=-D__GLASGOW_HASKELL__=800' '--cflag=-Dlinux_BUILD_OS=1' '--cflag=-Dx86_64_BUILD_ARCH=1' '--cflag=-Dlinux_HOST_OS=1' '--cflag=-Dx86_64_HOST_ARCH=1' '--cflag=-Idist/build/autogen' '--cflag=-include' '--cflag=dist/build/autogen/cabal_macros.h' '--lflag=-L/usr/lib/../lib' '--lflag=-Wl,-R,/usr/lib/../lib' '--lflag=-lffi' '--lflag=-lffi' '--cflag=-I/usr/lib/ghc-8.0.2/bytestring-0.10.8.1/include' '--cflag=-I/usr/lib/ghc-8.0.2/base-4.9.1.0/include' '--cflag=-I/usr/lib/ghc-8.0.2/integer-gmp-1.0.0.1/include' '--cflag=-I/usr/lib/ghc-8.0.2/include' '--lflag=-L/usr/lib/ghc-8.0.2/bytestring-0.10.8.1' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/bytestring-0.10.8.1' '--lflag=-L/usr/lib/ghc-8.0.2/deepseq-1.4.2.0' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/deepseq-1.4.2.0' '--lflag=-L/usr/lib/ghc-8.0.2/array-0.5.1.1' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/array-0.5.1.1' '--lflag=-L/usr/lib/ghc-8.0.2/base-4.9.1.0' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/base-4.9.1.0' '--lflag=-L/usr/lib/ghc-8.0.2/integer-gmp-1.0.0.1' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/integer-gmp-1.0.0.1' '--lflag=-lgmp' '--lflag=-L/usr/lib/ghc-8.0.2/ghc-prim-0.5.0.0' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/ghc-prim-0.5.0.0' '--lflag=-L/usr/lib/ghc-8.0.2/rts' '--lflag=-L/usr/lib/../lib' '--lflag=-Wl,-R,/usr/lib/ghc-8.0.2/rts' '--lflag=-Wl,-R,/usr/lib/../lib' '--lflag=-lm' '--lflag=-lrt' '--lflag=-ldl' '--lflag=-lffi' '--lflag=-lpthread' -o dist/build/Foreign/LibFFI/Internal.hs Foreign/LibFFI/Internal.hsc
hsc2hs: dist/build/Foreign/LibFFI/Internal_hsc_make: runProcess: runInteractiveProcess: exec: permission denied (Permission denied)
cabal: Leaving directory '/tmp/cabal-tmp-24502/libffi-0.1'
Updating world file...
cabal: Error: some packages failed to install:
libffi-0.1 failed during the building phase. The exception was:
ExitFailure 1
```

mitchty:
Gah, grsecurity, my alpine vm is running the stock kernel. Guessing its as simple as setting TMPDIR so cabal doesn't build outside of the build dir. I'll build another alpine linux vm with grsecurity and test.

jirutka:
It executes some files from /tmp?

mitchty:
Sort of, cabal uses TMPDIR for builds or if not set /tmp and when building the libffi module that module executes an internal command (hsc2hs is the haskell c to haskell utility) to convert c to haskell bindings via another command that it builds at runtime.

Also I apparently am running the grsec kernel and can't reproduce your error, is there some configuration for grsecurity I need to replicate that? I updated the APKBUILD locally to set TMPDIR for build() and prepare, but don't see anything in dmesg denying execution. Is it simply kernel.grsecurity.tpe* settings?

```
$ uname -a
Linux alpine 4.4.59-0-grsec #1-Alpine SMP Tue Apr 4 08:26:29 GMT 2017 x86_64 Linux
```

jirutka:
I’m sorry, the error was cause by a misconfiguration on my side. It builds without error.

* * *

Why it installs all libraries into directory `/usr/lib/x86_64-linux-ghc-8.0.2`? Does it mean that it must be rebuilt when ghc is updated?

```
    └── share
        └── x86_64-linux-ghc-8.0.2
            └── idris-1.0
                ├── docs
                │   ├── base
                │   │   ├── IdrisDoc
                │   │   ├── docs
                │   │   │   ├── Control.Arrow.html
                │   │   │   ├── Control.Catchable.html
                │   │   │   ├── Control.IOExcept.html
```

Documentation pages should be moved into -doc subpackage. Does Idris needs them on this path? It’d be more clear to move them into `/usr/share/doc/$pkgname`.

testing/idris/APKBUILD
```diff
@@ -0,0 +1,44 @@
+#-*-mode: Shell-script; coding: utf-8;-*-  
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+pkgname=idris
+pkgver=1.0
+pkgrel=0
+pkgdesc="Idris - A Language with Dependent Types"
+url="http://www.idris-lang.org"
+arch="x86_64 armhf"
+license="bsd3"
+depends="libffi gmp ncurses libffi-dev gmp-dev"
```
jirutka: Please remove libffi, gmp and ncurses from depends, these are auto-discovered by abuild, see:

```
>>> idris*: Tracing dependencies...
	libffi
	gmp
	ncurses
	libffi-dev
	gmp-dev
	so:libc.musl-x86_64.so.1
	so:libffi.so.6
	so:libgmp.so.10
	so:libncursesw.so.6
	so:libz.so.1
```

Does Idris really need development files of libffi and gmp?

testing/idris/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+makedepends="bash ghc ghc-dev cabal libffi gmp libffi-dev gmp-dev ncurses-dev zlib-dev"
```
jirutka: Subpackage -dev should always depend on the origin package (e.g. libffi-dev on libffi), so it’s not needed to add both libffi and libffi-dev to makedepends, only libffi-dev. The same applies for gmp, but currently not for ghc (ghc pkg should be fixed).

testing/idris/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+builddir="$srcdir/Idris-dev-$pkgver"
+
+prepare() {
+	cd "$builddir"
+	(
+		export HOME="$builddir/deps"
+		cabal update || return 1
+		cabal install libffi -v || return 1
+		cabal install --dependencies-only -v || return 1
+	) || return 1
```
jirutka: It’s not needed to spread `|| return 1` across abuilds anymore (yey!), abuild finally runs with `set -e`. :)

testing/idris/APKBUILD
```diff
@@ -0,0 +1,44 @@
+#-*-mode: Shell-script; coding: utf-8;-*-  
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+pkgname=idris
+pkgver=1.0
+pkgrel=0
+pkgdesc="Idris - A Language with Dependent Types"
```
jirutka: Please change to: `pkgdesc="A Language with Dependent Types"`.

mitchty:
> Why it installs all libraries into directory /usr/lib/x86_64-linux-ghc-8.0.2? Does it mean that it must be rebuilt when ghc is updated?
...
> Documentation pages should be moved into -doc subpackage. Does Idris needs them on this path? It’d be more clear to move them into /usr/share/doc/$pkgname.

Simply the default for cabal in general is the short answer, the basic reason is you could have multiple ghc installs for different architectures installing to a shared location and not have them conflict. I can update the document directories easily enough. Not entirely sure how much control I'll have over the haddock documents for html however, I'll poke around a bit.

Actually, its likely we would want to restrict the version of ghc it builds against in general. I'll add that in to require ghc 8.0.2 only for builds for now.

testing/idris/APKBUILD
mitchty: I'll double check on ffi, gmp probably not, likely me being overzealous.

testing/idris/APKBUILD
jirutka: I've already fixed ghc-dev, now it depends on ghc.

testing/idris/APKBUILD
mitchty: Looks like I was wrong its the other way around, idris needs gmp-dev when built with that as a feature but not libffi-dev to build ffi bindngs. I could probably leave out gmp as a feature if we wanted, but don't see much point.

```
/usr/lib/gcc/x86_64-alpine-linux-musl/6.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -lgmp
FAILURE: "gcc" ["-O2","-fwrapv","-fno-strict-overflow","-DHAS_PTHREAD","-DIDRIS_ENABLE_STATS","-I.","./testlib.o","/tmp/idris1669437588406334850.c","-L/usr/share/idris/rts/","-lidris_rts","-lgmp","-lpthread","-I/usr/share/idris/rts/","-lm","-I.","-I.","-I/usr/share/idris/libs/prelude","-I/usr/share/idris/libs/base","-o", "ffitest"]
# grep include  /tmp/idris1669437588406334850.c
#include "testlib.h"
#include "math.h"
#include "idris_rts.h"
#include "idris_bitstring.h"
#include "idris_stdfgn.h"
#include "idris_gmp.h"
#include "idris_opts.h"
#include "idris_rts.h"
#include "idris_stats.h"
#include <Windows.h>
```

Appears that its coming from the idris_gmp.h header.

jirutka: Then keep gmp-dev in depends.

mitchty:
Ok should have everything addressed with this amended version. I'm trying to get the test suite to run but there is a disconnect between how the makefile works and how I had to set this up for cabal. So this may/may not be fully functional. Also on the todo list is to build and install emacs mode files etc...

But nothing that prevents using idris as a compiler/repl based on my testing so far. I didn't test out the node.js part of idris however but I doubt its broken.

I had to do a bit of hacking to get the docs to install into where the doc sub package can find them and how cabal assumes layouts should be.

I'm also completely removing anything that ghc would care about for using idris as a shared library. Didn't see much point and doubt most people want to use idris as a haskell shared library anyway. Plus it shaved off 50MiB on the overall install.

mitchty:
Curious if the prior changes are acceptable or if there is anything further to be done.

jirutka:
I’ve improved it a bit, please take a look.

testing/idris/APKBUILD
```diff
@@ -0,0 +1,116 @@
...
+	# Cabal implicitly loads cabal.config from the project's directory.
+	ln -sf "$srcdir"/cabal.config .
+
+	# Create Cabal sandbox that will be used for all subsequent cabal
+	# invocations. This is something like Python virtual env.
+	cabal sandbox init
+
+	# Fetch all dependencies.
+	# XXX: This should always fetch the same versions as defined in
+	# our cabal.config, but Cabal does not verify packages integrity!
```
mitchty: This is incorrect, hackage and cabal implement TUF The Update Framework.

https://www.well-typed.com/blog/2015/04/improving-hackage-security/

Unless what you're referring to here via package integrity is comparing a sha sum or digest, in which case yes that isn't directly performed. Though obliquely in that if the package changes, it will no longer pass verification due to not having been signed by hackage at both the index level and/or the package level.

mitchty:
Overall looks good, was a bit busy lately so late on my review. I'm just testing out a build on armhf as I didn't see it here just x86_64:
https://pkgs.alpinelinux.org/packages?name=idris&branch=&repo=&arch=&maintainer=

testing/idris/APKBUILD
mitchty: Actually I was wrong, the sha256sum's of the package uploaded to hackage is validated as well as more. The overall process amounts to:
- hackage mirrors have a signing key
- when a new version is uploaded it gets signed, and checked at that point and its sizes added to the 01-index.tar.gz file. Which is a collection of versions+sha256sum's and is also signed
- Cabal knows of the public signing keys used in hackage, so when it downloads anything it compares that with what the signed index.tar.gz file has at download time

Example, if we look at:
https://hackage.haskell.org/01-index.tar.gz

We have for idris/1.0:
```
{"signatures":[],"signed":{"_type":"Targets","expires":null,"targets":{"<repo>/package/idris-1.0.tar.gz":{"hashes":{"md5":"10bd38b0d8bb23ee57d0bd4784f541bd","sha256":"b06852c83e307c9398fa33e031ca1fbe3c4f927496e7f5b00f13b4ed19f82beb"},"length":1595719}},"version":0}}
```

I'm still learning about TUF and its intricacies but as implemented in cabal/hackage there is definitely integrity checking.

https://github.com/haskell/hackage-security/blob/master/README.md

PR #697, opened 2017-01-06
testing/dosbox: update to SVN-r4003, move to testing, take ownership

skrzyp:
This update fixes the problem with PaX while executing writeable memory when DOSBox is run with "dynamic" CPU core.
Enabled back "dynamic" CPU core to improve DOSBox performance and reduce CPU overhead.

While I'm not a fan of using "snapshot" versions, especially when DOSBox has actual releases, please note that the last release was published in 2010, but development of DOSBox is still very active, the commit used as snapshot for this APKBUILD was made about 2 days ago. 

Before merging, please test on actual desktop machines, other than my laptop, with grsec and vanilla kernel.
To prove it's working, here are some screenshots:

dsj.exe
mario.exe


testing/dosbox/APKBUILD
```diff
@@ -0,0 +1,61 @@
...
+_cvstag=4003
+pkgver=r${_cvstag}
+pkgrel=0
+pkgdesc="x86/DOS emulator with sound and graphics"
+url="http://dosbox.com"
+arch="all"
+license="GPLv2+"
+depends="hicolor-icon-theme"
+depends_dev="libpng-dev sdl-dev alsa-lib-dev glu-dev"
+makedepends="$depends_dev paxctl autoconf automake"
```
fabled: Use paxmark instead of paxctl.

testing/dosbox/APKBUILD
```diff
@@ -0,0 +1,61 @@
...
+		--infodir=/usr/share/info \
+		--localstatedir=/var \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$_builddir"
+	make DESTDIR="$pkgdir" install || return 1
+	paxctl -c -M "$pkgdir"/usr/bin/dosbox
```
fabled: Use paxmark here. It does not take the '-c' argument. But should be otherwise be flag compatible. paxmark adds xattr markings which are now the standard.

testing/dosbox/APKBUILD
```diff
@@ -0,0 +1,61 @@
...
+_builddir="$srcdir"/dosbox-code-0-${_cvstag}/dosbox/trunk/ 
+prepare() {
+	local i
+	cd "$_builddir"
+	update_config_sub
+}
+
+build() {
+	cd "$_builddir"
+	./autogen.sh
```
fabled: Normally, calling autogen.sh or autoreconf is done in prepare().

nmeum:
@skrzyp Could you please update this PR?

clandmeter:
Closed due to inactivity.

PR #700, opened 2017-01-07
main/rxvt-unicode: add rxvt-unicode-terminfo subpackage

skrzyp:
Using urxvt to connect with remote Alpine servers causes them to act
very weird due to missing terminfo data.

This package contains only terminfo data to use with headless instances,
but main rxvt-unicode package adds is as dependency because it cant't
work properly without it.

### DIFF
--- a/main/rxvt-unicode/APKBUILD
+++ b/main/rxvt-unicode/APKBUILD
@@ -1,5 +1,6 @@
 # Contributor: Moritz Wilhelmy
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
+# Contributor: Jakub Skrzypnik <j.skrzypnik@openmailbox.org>
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=rxvt-unicode
 pkgver=9.22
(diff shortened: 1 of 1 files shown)


fabled:
I'm confused. Why do you need to split the package, if the subpackage is mandatory dependency of the main package? Why not just keep everything in the same one package?

kaniini:
@fabled it is for when you SSH into a remote machine so it has proper terminfo.  that way you do not have to install all of rxvt-unicode + X11 just to get the terminfo files.

main/rxvt-unicode/APKBUILD
```diff
@@ -57,13 +58,23 @@ build() {
...
+	# TERMINFO env var is used by rxvt-unicode makefile
+	# leaving it as is ~skrzyp
 	export TERMINFO="$pkgdir/usr/share/terminfo"
 	install -d "$TERMINFO"
 
 	make DESTDIR="$pkgdir" \
 		-C "$_builddir" install || return 1
 }
 
+terminfo() {
```
jirutka: Please use only tabs for indentation (and alignment) in APKBUILDs and other shell scripts, not spaces.

PR #702, opened 2017-01-07
testing/gradle: new aport

asiekierka:
A Java-based build system

### DIFF
--- a/testing/gradle/APKBUILD
+++ b/testing/gradle/APKBUILD
@@ -1,0 +1,56 @@
+# Contributor: Adrian Siekierka <kontakt@asie.pl>
+# Maintainer: Adrian Siekierka <kontakt@asie.pl>
+pkgname=gradle
+pkgver=3.3
+pkgrel=0
+pkgdesc="A Java-based build system"
+url="http://gradle.org/"
+arch="noarch"
+license="ASL-2.0"
+depends="openjdk8-jre-base"
+makedepends=""
+subpackages="$pkgname-doc"
+source="http://services.gradle.org/distributions/$pkgname-$pkgver-all.zip
+        gradle"
+builddir="$srcdir/$pkgname-$pkgver"
+_pkghome="/usr/share/java/$pkgname"
+
+prepare() {
+	default_prepare || return 1
+
+	cd "$builddir"
+	rm -r samples/native-binaries
+}
+
+build() {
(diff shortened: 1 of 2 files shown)


skrzyp:
well...

testing/gradle/APKBUILD
```diff
@@ -0,0 +1,54 @@
...
+license="ASL-2.0"
+depends="openjdk8-jre-base"
+makedepends=""
+subpackages="$pkgname-doc"
+source="http://services.gradle.org/distributions/$pkgname-$pkgver-all.zip
+        gradle"
+builddir="$srcdir/$pkgname-$pkgver"
+pkghome="/usr/share/java/$pkgname"
+
+prepare() {
```
jirutka: When you define custom `prepare` function, please always invoke `default_prepare || return 1`, usually as the first command in the function.

testing/gradle/gradle
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+APP_NAME="Gradle"
+APP_BASE_NAME=`basename "$0"`
```
jirutka: Backticks syntax for subshell is deprecated for decades, please use `$(…)`.

testing/gradle/APKBUILD
```diff
@@ -0,0 +1,54 @@
...
+url="http://gradle.org/"
+arch="noarch"
+license="ASL-2.0"
+depends="openjdk8-jre-base"
+makedepends=""
+subpackages="$pkgname-doc"
+source="http://services.gradle.org/distributions/$pkgname-$pkgver-all.zip
+        gradle"
+builddir="$srcdir/$pkgname-$pkgver"
+pkghome="/usr/share/java/$pkgname"
```
jirutka: Custom global variables should be prefixed with underscore (i.e. `_pkghome`).

testing/gradle/gradle
asiekierka: Sorry for that one - this was derived from Gradle's own /usr/bin script, just trimmed down heavily as we don't need things like Darwin compatibility.

jirutka:
The build fails because of bad checksum:

```
>>> gradle: Checking sha512sums...
gradle-3.3-all.zip: OK
gradle: FAILED
```

clandmeter:
Closed due to inactivity.

PR #707, opened 2017-01-07
community/xpra: update to 1.0.1 / create xpra-webclient subpkg

itoffshore:
* https://xpra.org/trac/wiki/News

* html5 `websockets` client split into `xpra-webclient` subpkg

if the `xxhash` PR  referenced below can be applied I can update this `xpra` PR to include `xxhash` support.


itoffshore:
continuous integration is not working (pull requests with both `xpra` & `ffmpeg` commits also failed)

`ffmpeg` needs to be pulled before `xpra` will build successfully

PR #716, opened 2017-01-11
community/php7: upgrade to 7.1.0, fixes and improvements

vakartel:
* for common binaries added unsuffixed symlinks
  (php7 -> php etc.)

* -embed subpackage renamed to -libs

* 'peardev' script moved from -dev to -pear subpackage

* PEAR data directory setted to /usr/share/pear for compatibility with existing packages

* 'phar.phar' script moved from -phar extension subpackage
  to -phar-phar subpackage

* added 'xmlwriter' shared extension.

* added 'recode' shared extension and resolve its building
  and loading conflict with 'imap' extenstion.

* build 'gd' extension without jis-conv japanse language support
  because it suggests any non-latin character as JIS, not UTF-8
  by default (php bug #39543 from 2006-11-17 still unresolved).
  Thanks Alex Isakov <bobrocoder@gmail.com> for investigation.

* php7-fpm.initd: cosmetic fixes

* remade extensions build:
  - extension dependencies needs to be set strictly for counting loading index.
  - a fake '--' dependency added to just increment loading index.
  - descriptions are picked out from ext/extension/CREDITS file or can be set strictly.

* for now this APKBUILD can be implemented for building
  php5 with minimal changes.


vakartel:
@andypost I made a new PR for php7 only. It's simplier and have more fixes then the previous one. 
Any comments are welcome.
Thanks.

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
-	# phpdbg
+	# build phpdbg
 	_build --enable-phpdbg \
-		--enable-phpdbg \
 		--enable-phpdbg-webhelper \
 		--disable-cgi \
 		--disable-cli \
 		|| return 1
-	# apache2 module
+	# build apache2 module + recode extension
```
andypost: useful addition

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
 
-	install -Dm644 php.ini-production "$pkgdir"/etc/$pkgname/php.ini || return 1
+	rm -fr "$pkgdir"/.[[:alpha:]]* || return 1
 
-	local file; for file in pear peardev pecl; do
-		sed -i -e "s~/usr/bin/php~/usr/bin/php7~g" \
-			-e "s~PHP=php~PHP=php7~" \
-			"$pkgdir"/usr/bin/$file || return 1
-	done
-	paxmark -m "$pkgdir"/usr/bin/php7
```
jirutka: Why have you removed paxmark? Are you sure that it’s not needed anymore under grsecurity kernel?

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
-
-	find "$pkgdir" -name '.*' | xargs rm -rf || return 1
-	rmdir "$pkgdir"/var/run
+	ln -sf php$_suffix "$pkgdir"/usr/bin/php
```
jirutka: This will conflict with php5 package.

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
+	local file
+	for file in php-config phpize; do
+		mv "$pkgdir"/usr/bin/$file$_suffix \
+			"$subpkgdir"/usr/bin/ || return 1
+		ln -sf $file$_suffix "$subpkgdir"/usr/bin/$file
+	done
+
 	mv "$pkgdir"/usr/lib/$pkgname/build \
-		"$subpkgdir"/usr/lib/$pkgname/
+		"$subpkgdir"/usr/lib/$pkgname/ || return 1
```
jirutka: `|| return 1` is actually not needed here (it’s has zero effect, like writing `true || true`) since it’s the last statement of the function (implicit return).

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
 	default_doc
+	install_if="docs $pkgname-common=$pkgver-r$pkgrel"
 
-	mkdir -p "$subpkgdir"/usr/share/doc/$pkgname
-
+	cd "$builddir"
 	local file
-	for file in CODING_STANDARDS CREDITS EXTENSIONS INSTALL LICENSE NEWS README* UPGRADING*; do
-		cp "$builddir"/$file "$subpkgdir"/usr/share/doc/$pkgname/ || return 1
+	for file in $(ls CODING_STANDARDS CREDITS EXTENSIONS INSTALL LICENSE NEWS README* UPGRADING*); do
```
jirutka: Why have you changed that? o.O

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
 phpdbg() {
-	pkgdesc="Interactive PHP debugger"
+	provides="php-phpdbg"
+	pkgdesc="$pkgdesc (interactive debugger)"
 	depends="$pkgname-common"
 
 	install -Dm755 "$builddir"/sapi/phpdbg/phpdbg \
-		 "$subpkgdir"/usr/bin/phpdbg7
+		 "$subpkgdir"/usr/bin/phpdbg$_suffix || return 1
+	ln -sf phpdbg$_suffix "$subpkgdir"/usr/bin/phpdbg
```
jirutka: This is in conflict with php5-phpdbg package.

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
-	pkgdesc="PHP Common Gateway Interface"
+	provides="php-cgi"
+	pkgdesc="$pkgdesc (common gateway interface)"
 	depends="$pkgname-common"
 
 	mkdir -p "$subpkgdir"/usr/bin
-	mv "$pkgdir"/usr/bin/php-cgi7 "$subpkgdir"/usr/bin/
+	mv "$pkgdir"/usr/bin/php-cgi$_suffix \
+		 "$subpkgdir"/usr/bin/ || return 1
+	ln -sf /usr/bin/php-cgi$_suffix "$subpkgdir"/usr/bin/php-cgi
```
jirutka: This is in conflict with php5-cgi package.

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
 	mkdir -p "$subpkgdir"/usr/bin \
 		"$subpkgdir"/etc/$pkgname
 
-	local file; for file in phpize7 pecl pear; do
-		mv "$pkgdir"/usr/bin/$file "$subpkgdir"/usr/bin/
-	done
-	mv "$pkgdir"/etc/$pkgname/pear.conf "$subpkgdir"/etc/$pkgname/ || return 1
-	mv "$pkgdir"/usr/share "$subpkgdir"/usr/
+	mv "$pkgdir"/usr/bin/pecl \
+		"$pkgdir"/usr/bin/pear \
```
jirutka: This is in conflict with php5-pear package.

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
-	pdo_*) depends="$depends $pkgname-pdo"
-	esac
+_extension() {
+	local extname=${subpkgname#*-}
 
-	# extension dependencies and loading order
-	local elo=00
-	[ -z "$extdepends" ] || elo=01
+	pkgdesc=$(eval echo \$_pkgdesc_$extname)
+	[ ! "$pkgdesc" ] && pkgdesc=$(head -n1 "$builddir"/ext/$extname/CREDITS)
```
jirutka: Please replace with `[ -n "$pkgdesc" ] || pkgdesc=…`

community/php7/APKBUILD
```diff
@@ -137,231 +143,244 @@ _build() {
...
+	[ ! "$pkgdesc" ] && return 1
```
jirutka: Please replace with `[ -n "$pkgdesc" ] || return 1`

community/php7/php7-fpm.initd
```diff
@@ -52,21 +50,21 @@ start_pre() {
 	# directory exists and has correct privileges.
 	local listen="$(conf_get listen)"
 	if [ "${listen:0:1}" = "/" ]; then
-		checkpath -d -o $user:$group "$(dirname "$listen")"
+		checkpath -d -o $user:$group ${listen%/*}
 	fi
 
-	checkpath -d "$(dirname "$pidfile")"
+	checkpath -d ${$pidfile%/*}
```
jirutka: This is wrong, should be `${pidfile%/*}`.

community/php7/php7-fpm.initd
```diff
@@ -52,21 +50,21 @@ start_pre() {
 	# directory exists and has correct privileges.
 	local listen="$(conf_get listen)"
 	if [ "${listen:0:1}" = "/" ]; then
-		checkpath -d -o $user:$group "$(dirname "$listen")"
+		checkpath -d -o $user:$group ${listen%/*}
```
jirutka: Why have you removed quotes around the `listen` variable? It’s a good practice to always quote variables that contain externally provided file path.

community/php7/php7-fpm.initd
```diff
@@ -6,8 +6,6 @@
 # variable "user" and optionally "group" in conf.d file, or in the $fpm_config
 # file (the former has precedence).
 
-: ${name:="PHP FastCGI Process Manager"}
```
jirutka: Why have you removed this?

vakartel:
@andypost  7.1.1 is out :)

community/php7/APKBUILD
vakartel: @jirutka 
paxmark was included in earlier versions by mistake.
I know because I included it :)

community/php7/APKBUILD
vakartel: @jirutka 
But how else to know if this directory's name  would be changed in next versions?

community/php7/APKBUILD
vakartel: @jirutka 
Because of items with wildcards (README* UPGRADING*)

community/php7/APKBUILD
vakartel: @jirutka 
cli, cgi, phpdbg and dev-tools are in conflict with php5 ones for now.

community/php7/APKBUILD
vakartel: @jirutka 
pecl and pear are the same in all php versions, so it does not matter for now.

community/php7/APKBUILD
vakartel: clear

community/php7/APKBUILD
vakartel: clear

community/php7/php7-fpm.initd
vakartel: ok, move back

community/php7/php7-fpm.initd
vakartel: fixed and quoted

community/php7/php7-fpm.initd
jirutka: The question is why. I added this variable for purpose. This runscript is designed to be symlinked for particular applications. It makes sense for me to allow users to modify the name in log messages.

community/php7/APKBUILD
jirutka: I don’t understand your question.

community/php7/php7-fpm.initd
vakartel: ok, moved back

vakartel: But I think it's better to use name of init-script ($RC_SVCNAME) as unique identifier in logs

jirutka:
Agree with that.

community/php7/php7-fpm.initd
jirutka: Agree with that.

community/php7/APKBUILD
vakartel: @jirutka 
ok, I remove it.

vakartel:
@jirutka 
Does I fixed all you requested ?

Do I need to add 'replaces=' to conflicting subpackages or stay as is?

andypost:
FYI http://php.net/archive/2017.php#id2017-01-19-1 7.1.1 release with security fixes, I will make PR for 7.0 later today

andypost:
I mean to edit title of this PR to `... upgrade to 7.1.1 ...`

vakartel:
@andypost 
What do you think about to combine all php versions (5, 7.0, 7.1) in one apkbuild. 5 as good-old stable, 7.0 as stable and 7.1 as a current?

andypost:
@vakartel I really like to see all 3 versions available. But looks that will require to give another suffix for packages (php5, php7 and php71 - looks ugly)
Also it's a good question how to maintain 3.5 release with 7.0 security updates! I mean that may require to create PRs against 3.5-stable branch?!

On one hand I mostly use php for Drupal and next month release will be compatible with 7.1 but
on other hand it makes sense to have 7.0 for tests and some old applications that I need to support.

So if you have any ideas about how to maintain all 3 releases - I'm in and will try to help

andypost:
Also there's a good question about other packages that have dependency. For example xhprof is not compatible with 7.1 and other extensions could also fail. So I think it will require update their dependencies somehow so that should be a separate PR I guess

vakartel:
@andypost 
let's discuss it direct valery.kartel<at>gmail.com

JulienBreux:
Hi guys.

We following the comments ;)

Please resume after your private party!

Thanks for your work!
If you pass in France, I offer beer 🍺

andypost:
Btw https://github.com/php/php-src/pull/2236 accepted to PHP upstream so very probably `fix-asm-constraints-in-aarch64-multiply-macro.patch` no longer needed at least for 7.1 but for 7.0 it needs testing

andypost:
According https://bugs.php.net/bug.php?id=73898 7.0+ so please remove it

JulienBreux:
@jirutka Are-you ok with last changes?

JulienBreux:
@jirutka sorry for abusing :/ but can you review last changes? Can I help you for something?

f3l1x:
@JulienBreux @jirutka Can't wait too. Good jobs guys.

ncopa:
Sorry to say, but I really don't like the idea of building all the php versions from same APKBUILD. it is messy.

I also don't think we have resources to maintain php5.x, php7.0.x and php7.1.x for stable branches.

Can't we just upgrade the php7 from 7.0.x -> 7.1.x?

andypost:
@ncopa it makes sense to update but a lot of other packages depends on php7 (means do not work with 7.1)
So maybe better to upgrade to php7.1 in testing and test dependent packages somehow

vakartel:
@ncopa 
I don't like it too. I made this apkbuild just as an universal template to build all php versions. It must be splitted to versioned apkbuilds.

Many 3rd-party extensions are not compatible with php7.1 so it can't be in as main php7 version. And many extensions are not compatible with php7 at all. So there we have two production stable versions php5.6 and php7.0 and a testing current php7.1

fabled:
Wow. Looks like lots of things have been fixed on this. Good work.
This has been pending for quite a while -- and I'd like to get done with this. Would it be possible to get testing/php7.1 as single squashed commit? We could then use it as basis to rebuild php5 and php7 aports?

vakartel:
I want to ask about a new naming variant: main/php56 (provides php, php5), community/php70 (provides php, php7), testing/php71 (provides php)

andypost:
I'd better use `php5` (as it now the only 5.x version) cos php 5.6 will be supported next 2 years
See http://php.net/supported-versions.php

7.x branch IMO should get split because 7.0 is only production ready version and will be supported also next 2 years

PHP 7.1 should land in `testing/` so could use `php7` as package name. But that will cause a collisions with dependent packages to I'd better use `php7.1` as package name

jirutka:
What about not supporting PHP 7.1 at all, until all pkgs that depends on PHP 7.x are compatible with 7.1?
I’d strongly suggest to keep pkgname unique across all official repositories.

andypost:
The primary reason to have 7.1 available is to allow adopt it, I guess is what `testing` for.

btw I quickly checked `php*` packages and found that some of them does not declare their dependencies, that's a primary issue to have packages prefixed with `php(5|7)`
So I'm going to review dependencies and complete a table with proper deps

vakartel:
@jirutka @andypost @fabled @ncopa 
I made a new PR for testing/php71 https://github.com/alpinelinux/aports/pull/839

So lets stay community/php7 as is an close this PR ;)

ncopa:
@vakartel thank you for working on this mess. I really appreciate.

PR #724, opened 2017-01-12
main/udev-init-scripts: upgrade to 32

vakartel:
- cleanups in apkbuild
- add provides=udev for autoinstalling udev-related packages


main/udev-init-scripts/APKBUILD
```diff
@@ -1,53 +1,33 @@
...
+pkgrel=0
 pkgdesc="udev startup scripts for openrc"
 url="https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/sys-fs/udev-init-scripts/"
 arch="noarch"
 license="GPLv2"
-depends=""
-depends_dev=""
-makedepends="$depends_dev"
-replaces="udev"
+provides=udev
```
fabled: If we do this, it can't co-exist with 'udev' dummy package which is sometimes used to pull in eudev.
@ncopa Should we just nuke 'udev' transitional package, and include provides here and in 'eudev' ?

vakartel: @fabled 
eudev depends on this one. So it does not need to make eudev provides udev.

Another idea is to move udev init scripts into eudev and remove this package.

fabled: Then this is actually a bug. Because provides=name will conflict with package named 'name' and this is the cycle after this:
udev (the wrapper package): depends=eudev
eudev: depends = udev-init-scripts
udev-init-scripts: provides udev

So you can no longer install the convenience package.

I think we should have eudev provides 'udev=$pkgver', remove the udev wrapper package, and remove the provides here.

PR #753, opened 2017-01-16
proftpd: move from unmaintained to testing. Upgrade to 1.3.6

vakartel:
upgrade to 1.3.5d
- patch to build with libressl
- improve init script, add logrotate script
- prxs binary moved to -dev
- other scripts and binaries moved to -utils
- shared modules moved to subpackages
- add strict module dependencies loading
- add some modules configs


fabled:
Looks good to me. Would you be able to squash all the commits to one? I also would suggest you to take maintainership because it was in unmaintained. Thanks.

testing/proftpd/proftpd.initd
```diff
@@ -0,0 +1,28 @@
...
+depend() {
+	need net
+	use logger dns
+	after firewall
+}
+
+start_pre() {
+	ebegin
+	checkpath --directory ${pidfile%/*}
+	$command -t &>/dev/null
```
jirutka: `&>` is a bashism, not POSIX-sh syntax. Please replace it with `>/dev/null 2>&1`.

kaniini:
can you flatten this down?

testing/proftpd/APKBUILD
```diff
@@ -0,0 +1,147 @@
...
+		fi
+		mkdir -p ${file%/*}
+		echo "<IfModule !$name.c>" > $file
+		local dep; for dep in $(echo $load | tr "," " "); do
+			depends="$depends $pkgname-mod_$dep"
+			echo "  <IfModule !mod_$dep.c>" >> $file
+			echo "    LoadModule mod_$dep.c" >> $file
+			echo "  </IfModule>" >> $file
+		done
+		echo -e "  LoadModule $name.c\n</IfModule>" >> $file
```
jirutka: Please don’t use `echo -e`, its behaviour is not well defined. Use `printf` instead.

testing/proftpd/APKBUILD
```diff
@@ -0,0 +1,147 @@
...
+	mv "$pkgdir"/usr/sbin/ftpscrub \
+		"$pkgdir"/usr/sbin/ftpshut \
+		"$subpkgdir"/usr/sbin || return 1
+}
+
+_module() {
+	local name=${subpkgname#$pkgname-}
+	local load=$(eval echo \$_load_$name)
+	pkgdesc="ProFTPd module: $name"
+	depends="$pkgname"
```
jirutka: Empty line between various parts of code doesn’t hurt, it increases readability a lot…

testing/proftpd/APKBUILD
```diff
@@ -0,0 +1,147 @@
...
+	install -d "$pkgdir"/etc/$pkgname/modules.d
+	install -dm750 -o ftp -g ftp "$pkgdir"/var/lib/ftp
+	install -Dm755 "$srcdir"/$pkgname.initd "$pkgdir"/etc/init.d/$pkgname
+	install -Dm644 "$srcdir"/$pkgname.logrotate "$pkgdir"/etc/logrotate.d/$pkgname
+	rm -fr "$pkgdir"/run "$pkgdir"/usr/share/locale \
+		"$pkgdir"/usr/sbin/in.$pkgname
+}
+
+dev() {
+	default_dev
```
jirutka: Unfortunately APKBUILD does not run with `set -e` (yet, it’s on our TODO list to change it), so every statement that may fail, but should not fail, and is not the last statement of the function, should end with `|| return 1`.

testing/proftpd/libressl.patch
```diff
@@ -0,0 +1,3017 @@
+--- proftpd-1.3.6rc4.orig/contrib/mod_auth_otp/crypto.c
```
jirutka: Where did you get this huge patch from? Why it contains even non-significant changes like copyright year? Is it backported from upstream or have you wrote it? This should be stated in the patch file, especially in huge patches like this one.

testing/proftpd/APKBUILD
vakartel: added || return 1

testing/proftpd/APKBUILD
vakartel: it's not too complicated to hard to read, I think

testing/proftpd/APKBUILD
vakartel: replaced with two echoes

vakartel:
@fabled @kaniini @jirutka 

done to be approved.

PR #763, opened 2017-01-18
Remove duplicate packages

vakartel:
move community/perl-timedate -> main/perl-timedate
move testing/perl-mailtools -> main/perl-mailtools
fix dependencies in existing packages (perl-mail-tools -> perl-mailtools)
delete wrong-named packages main/perl-time-date & main/perl-mail-tools


vakartel:
community/zoneminder does not rebuilds because perl-mailtools does not found for it

I think, it's not a problem.

PR #764, opened 2017-01-18
Add entry for powerpc platforms

leitao:
This patch simply adds an entry to build linux-header for powerpc
platforms.
The main target is ppc64le architecture, which uses Linux ARCH=powerpc
on the Makefiles.

### DIFF
--- a/main/linux-headers/APKBUILD
+++ b/main/linux-headers/APKBUILD
@@ -2,7 +2,7 @@
 pkgname=linux-headers
 pkgver=4.4.6
 _kernver=${pkgver%.*}
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


jirutka:
Please bump `pkgrel` (=> `pkgrel=2`).

leitao:
Hello @jirutka,

Just bumped pkgrel=2 now. 

I understand that I should change the pkgrel for every new change, if I do not change the major package version. Is my understanding correct?

jirutka:
> I understand that I should change the pkgrel for every new change, if I do not change the major package version. Is my understanding correct?

Basically, yes. If there already exist pkg with the same pkgver and pkgrel, then build infra does not build it.

PR #765, opened 2017-01-18
 Add ppc64le support

leitao:
This is a patch that adds supports for ppc64le platform.
The config file was generated with "make powernv_defconfig".

On top of it, I just changed CONFIG_POWER8=y, since I do not expect to have any
PPC64LE/ABIv2 distro running on machines older than POWER8.


main/musl/APKBUILD
```diff
@@ -53,6 +54,11 @@ build() {
 	${CROSS_COMPILE}gcc $CPPFLAGS $CFLAGS -c "$srcdir"/__stack_chk_fail_local.c -o __stack_chk_fail_local.o || return 1
 	${CROSS_COMPILE}ar r libssp_nonshared.a __stack_chk_fail_local.o || return 1
 
+	# Consider "long double" 64-bits wide on Power
+	if [ $CARCH == "ppc64le" ]; then
```
jirutka: `==` is not standard syntax per POSIX sh, please use `=` instead.

main/spl/APKBUILD
```diff
@@ -19,6 +19,7 @@ builddir="$srcdir"/spl-$pkgver
 
 build() {
 	cd "$builddir"
+	autoreconf -fi
```
jirutka: Please move `autoreconf -fi` to `prepare()` function:

```sh
prepare() {
 	default_prepare || return 1
 	autoreconf -fi
}
```

and add comment why it’s needed to run `autoreconf` despite the `configure` script exists.

leitao:
Hello @jirutka,
Thanks for the review. I corrected these points now on this PR.

main/spl/APKBUILD
```diff
@@ -17,6 +17,13 @@ source="https://github.com/zfsonlinux/zfs/releases/download/zfs-$pkgver/spl-$pkg
 
 builddir="$srcdir"/spl-$pkgver
 
+prepare() {
+	default_prepare || return 1
+	# Current package version does not contain information about ppc64le on
+	# config.sub and config.guess. Calling autoreconf to update these files.
```
jirutka: Aha, then it’s not needed to call `autoreconf`, we have two helper functions for this: `update_config_sub && update_config_guess`.

leitao: Cool. Much easier then. Should I call it inside prepare() ?

jirutka: Yes.

leitao:
@jirutka, I just call update_config_{guess,sub} now.

It was not working properly on ppc64le, thus, I fixed it with 
https://github.com/alpinelinux/abuild/pull/13/commits/2cf38f68067ce381cb25eac29b10fda34135466c

jirutka:
@ncopa, could you please review update of main/linux-vanilla? I don’t know how are the other configs generated, whether it’s okay to add config for power8 with defaults.

leitao:
It is important to say that ABI v2 (that use little endianess) requires POWER8. POWER7 is not able to boot on LE mode, thus, the requirement for POWER8.

jirutka:
When I read my pervious comment once again, I see that it's misleading. I don't have any problem with `CONFIG_POWER8=y`. 

If I understand it correctly, you've generated config with default options. My concern is that **if** our configs for other arch are not generated in similar way (except arch-specific flags), i.e. we use custom configuration, then the ppc64le kernel may have quite different settings and feature set than our kernels for other archs.

Sorry for unclear message, I don't know how to express it correctly. Anyway, know we should just wait for reply from @ncopa, he created the existing configs.

leitao:
Ok. By the way, if you wish to have access to a ppc64le system, I can some VMs for you. Unfortunately we still do not have Alpine as the OS yet. You can choose among the other distros that ppc64le support, as Debian, Fedora, Ubuntu, CentOS, etc.

leitao:
Just updated this PR to be based on top of already pushed William Pitcock changes for powerpc.

kaniini:
This is incomplete -- ppc64le support needs to be added to main/gcc.  Note the use of `--enable-decimal-float=no`, that is the magic switch to use.

leitao:
Hi  @kaniini,

We are not targeting GCC patches at this Pull Request. We will wait until we have a final chroot (A chroot that is generated with native compiled packages) in order to make sure that GCC has all the patches, so, I am not targeting GCC patches at this moment. This PR is not supposed to be complete, but more a "release early, release often" approach.

Regarding `--enable-decimal-float=no`, what do you mean by magical switch? Are you talking about the C flag required by to compile musl?

leitao:
Hi @kaniini,

Why do you think that disabling decimal float point is required to build musl in ppc64le? I am not able to configure it with decimal float point disabled, take a look: http://paste.ubuntu.com/23870027/

Anyway, if you think that the correct flag is `--enable-decimal-float=no`, go ahead and add it to gcc for ppc64le.

kaniini:
@leitao 

musl author says explicitly that the IBM 128-bit double is not supported on ppc64le.  I will go ahead and add the configure flags for GCC.

leitao:
hi @kaniini,
Right,  it makes sense now. Thanks!

By the way, is this PR OK for you? Can it be merged?

kaniini:
@leitao i have no objections to carrying the patches, but full powerpc enablement still requires grub packaging (which I am working on for ppc32, presumably the same work will work for you guys as well).  there is also the infrastructure for the port that needs to be sorted before the work is merged as the builder needs to be running off a pristine git tree once it is actually building and uploading packages into the archive

PR #769, opened 2017-01-19
main/libvirt: upgrade to 3.0.0

misery:


misery:
ping? :-)

jirutka:
It’s bump in major version, so I’m afraid what breaking changes does it contain and what it may affect. Can you please verify this?

PR #772, opened 2017-01-19
community/php7: upgrade to 7.0.15 (security fixes)

andypost:
Security release http://php.net/archive/2017.php#id2017-01-19-2


andypost:
This needs to go 3.5 as well, probably before https://github.com/alpinelinux/aports/pull/716

PR #773, opened 2017-01-19
community/dnscrypt-proxy: update to 1.9.3

itoffshore:
* This version can be compiled on Linux distributions using the **musl C** library.

* Version `1.9.3` also restores compatibility with ancient Linux kernels that didn't support
`SO_REUSEPORT`, without having to explicitly compile the package with `NO_REUSEPORT`.

* On Linux, the service now prints when the system doesn't have enough entropy
to initialize the `PRNG`.

https://github.com/jedisct1/dnscrypt-proxy/blob/master/ChangeLog


jirutka:
This abuild has been already updated to 1.9.4 in 4dc5a6dd83264357ebb8de971c2935c5369ad549.

PR #774, opened 2017-01-19
main/php5: upgrade to 5.6.30 (security fixes)

andypost:
Security release http://php.net/archive/2017.php#id2017-01-19-3


andypost:
3.5 also needs this upgrade

PR #783, opened 2017-01-23
testing/postgresql-xl: new aport

GJRTimmer:
http://www.postgres-xl.org/
Scalable Open Source PostgreSQL-based Database Cluster

### DIFF
--- a/testing/postgresql-xl/APKBUILD
+++ b/testing/postgresql-xl/APKBUILD
@@ -1,0 +1,248 @@
+# Maintainer: G.J.R. Timmer <gjr.timmer@gmail.com>
+pkgname=postgresql-xl
+_pkgname=${pkgname/ql/}
+pkgver=9.5r1.4
+pkgrel=0
+pkgdesc="Scalable Open Source PostgreSQL-based Database Cluster"
+url="http://www.postgres-xl.org/"
+arch="all"
+license="BSD"
+# Depends are deliberate, reason: avoid conflict with normal PostgreSQL
+depends="postgresql-xl-client postgresql-xl-libpq"
+install=""
+pkgusers="postgres"
+pkggroups="postgres"
+depends_dev="libressl-dev"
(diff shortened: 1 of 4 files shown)


GJRTimmer:
@jirutka request review

GJRTimmer:
updated postgresql-xl to the same solution you wrote for ```main/postgresql```

***Note:
initdb.patch is a little different from the ```main/postgresql``` because postgresql-xl version is slightly different.***

***Note:
It's correct some files are missing in relation to ```main/postgresql``` file the restore / dump files, and the init files for running it as a service; I have to look at every one of them because postgresql-xl is meant to be run as a scalable cluster and has more components then a single postgresql. Therefor this is only to be added to the alpine/testing repository; its not ready for complete deployment***

testing/postgresql-xl/APKBUILD
```diff
@@ -0,0 +1,248 @@
...
+pkgname=postgresql-xl
+_pkgname=${pkgname/ql/}
+pkgver=9.5r1.4
+pkgrel=0
+pkgdesc="Scalable Open Source PostgreSQL-based Database Cluster"
+url="http://www.postgres-xl.org/"
+arch="all"
+license="BSD"
+# Depends are deliberate, reason: avoid conflict with normal PostgreSQL
+depends="postgresql-xl-client postgresql-xl-libpq"
```
fabled: What is the intention here? To pull packages from this aport, and it's to override the automatic soname dependencies?
It might be better to use `sonameprefix` and conflict with the regular postgresql package.

GJRTimmer: @fabled Working on it.
About the naming, I will look in to it, however the product is actually named ```postgresql-xl``` 

Using the depends you override the automatic detection of libpq of the normal libpq package.

I working on gettings details about libpq from the postgres-xl team. to find out if there are differences.

clandmeter:
Closed due to inactivity.

PR #785, opened 2017-01-24
main/strace: fix pt_regs collision on ppc64le

gromero:
strace build requires <linux/ptrace.h> and it includes <asm/ptrace.h>
that, by its turn, defines the pt_regs struct. However the same struct
is also define in <bits/user.h> from musl-dev, creating therefore a
conflict. A simple solution is to add the __ASSEMBLY__ guard so pt_regs
struct from <asm/ptrace.h> is not include twice, avoiding the collision
in question.

### DIFF
--- a/main/strace/APKBUILD
+++ b/main/strace/APKBUILD
@@ -9,7 +9,8 @@ license="BSD"
 depends=
 makedepends="linux-headers"
 subpackages="$pkgname-doc"
-source="http://downloads.sourceforge.net/sourceforge/$pkgname/$pkgname-$pkgver.tar.xz"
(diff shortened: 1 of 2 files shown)


jirutka:
It fails on armhf now (see the attached log), could you please fix it?


[strace-4.15-r1.txt](https://github.com/alpinelinux/aports/files/728459/strace-4.15-r1.txt)

fabled:
I reverted this. The fixing patch is incorrect, it should not try to depend on internal definitions such as `__ASSEMBLY__`.

You could possibly use arch specific define instead, see: http://nadeausoftware.com/articles/2012/02/c_c_tip_how_detect_processor_type_using_compiler_predefined_macros

Perhaps it could be `#if !defined(__powerpc__) || !defined(__powerpc64__)` similar

PR #788, opened 2017-01-25
testing/etckeeper: new aport

HRio:
For limitations see README.alpine

### DIFF
--- a/testing/etckeeper/APKBUILD
+++ b/testing/etckeeper/APKBUILD
@@ -1,0 +1,43 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+
+pkgname=etckeeper
+pkgver=1.18.6
+pkgrel=1
+pkgdesc="Store /etc in git."
+url="https://github.com/joeyh/etckeeper"
+arch="noarch"
+license="GPL2"
+depends="findutils git perl"
+subpackages="$pkgname-doc"
+install="$pkgname.post-install $pkgname.pre-deinstall"
+source="$pkgname-$pkgver.tar.gz::https://github.com/joeyh/etckeeper/$_pkgname/archive/$pkgver.tar.gz
+	apk-commit_hook
+	$pkgname.post-install
+	$pkgname.pre-deinstall
+	"
+builddir="$srcdir/$pkgname-$pkgver"
+
+prepare() {
+	cd "$builddir"
+	sed -i 's|^PYTHON=python$|PYTHON=/bin/false|' Makefile || return 1
+	sed -i 's/_PACKAGE_MANAGER=.*/_PACKAGE_MANAGER=apk/' "$builddir"/etckeeper.conf || return 1
+}
+
+package() {
+	cd "$builddir"
+	make install DESTDIR="$pkgdir" || return 1
+	rm -rf "$pkgdir"/lib/systemd || return 1
+	rmdir "$pkgdir"/lib/ || return 1
+	rm -rf "$pkgdir"/etc/apt || return 1
+
+	install -m755 -D "$builddir"/debian/cron.daily \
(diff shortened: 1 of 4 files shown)


HRio:
@ncopa, closes http://bugs.alpinelinux.org/issues/6587 ?

testing/etckeeper/etckeeper.trigger
```diff
@@ -0,0 +1,11 @@
...
+
+# we are in trigger context so the apk database
+# contains info before the change takes effect
+/etc/etckeeper/pre-install.d/10packagelist
+
+LOCK="/lib/apk/db/lock"
+exec 200>"$LOCK"
+
+# wait for $LOCK before commit
+(flock --timeout 180 200 && etckeeper post-install) &
```
fabled: Is this trigger script hack something we get rid of after https://github.com/alpinelinux/apk-tools/pull/2 is merged?

testing/etckeeper/README.alpine
```diff
@@ -0,0 +1,4 @@
+Known Limitations
+-----------------
+* Can not block package installation
+* Can not handle commit before install
```
fabled: Both are fixable after https://github.com/alpinelinux/apk-tools/pull/2 ?

testing/etckeeper/APKBUILD
```diff
@@ -0,0 +1,50 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+
+pkgname=etckeeper
+pkgver=1.18.6
+pkgrel=1
+pkgdesc="Store /etc in git."
```
fabled: I generally dislike this, because git tracks history, and in tmpfs installs it means the whole history is also stored in the apkovl. Which is part of the reason why we wrote `aaudit`. Then again, it's much less known, less feature complete etc. And I do understand that people are possibly familiar with `etckeeper` so I think we should include it as an option.

testing/etckeeper/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+	$pkgname.post-install
+	$pkgname.pre-deinstall
+	$pkgname.trigger
+	"
+builddir="$srcdir/$pkgname-$pkgver"
+
+package() {
+	cd "$builddir"
+	sed -i 's|^PYTHON=python$|PYTHON=/bin/false|' Makefile || return 1
+	sed -i 's/_PACKAGE_MANAGER=.*/_PACKAGE_MANAGER=apk/' "$builddir"/etckeeper.conf || return 1
```
fabled: There sed scripts should probably be in prepare step.

testing/etckeeper/etckeeper.trigger
HRio: yes ;-) was not happy with that trigger at all, but it was the only way to do it with current apk-tools. Hence this trigger (removal of it) is the motivation for the work in alpinelinux/apk-tools#2 i.

testing/etckeeper/README.alpine
HRio: yes thats right.

HRio:
**Note**  2720907e05 testing/etckeeper: use apk commit_hook requires a new apk-tools package for the commit hook to function

kaniini:
Can you flatten this down?

HRio:
@kaniini sure I can push a squash.

kaniini:
It seems documentation on how to use apk commit-hook is missing.  Can you bring back the README.Alpine file for that?

HRio:
The apk commit hook will not work before a new version of apk-tools has been tagged:
https://github.com/alpinelinux/apk-tools/commits/master
 before that this PR can be kept on hold.

kaniini:
Okay.  Lets hold on this one then.

PR #795, opened 2017-01-27
main/libpciaccess: include limits.h for ppc64le

ferseiti:
When limits.h is not included for ppc64le, the macro PATH_MAX is undefined 
which causes libpciaccess to fail to build.

Signed-off-by: Fernando Seiti Furusato <ferseiti@linux.vnet.ibm.com>

### DIFF
--- a/main/libpciaccess/ppc64le-limits.patch
+++ b/main/libpciaccess/ppc64le-limits.patch
@@ -1,0 +1,13 @@
+--- a/src/linux_sysfs.c
++++ b/src/linux_sysfs.c
+@@ -66,6 +66,10 @@
+ #include <sys/ioctl.h>
+ #endif
+ 
++#ifdef __PPC64__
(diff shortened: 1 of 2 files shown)


main/libpciaccess/ppc64le-limits.patch
```diff
@@ -0,0 +1,13 @@
+--- a/src/linux_sysfs.c
++++ b/src/linux_sysfs.c
+@@ -66,6 +66,10 @@
+ #include <sys/ioctl.h>
+ #endif
+ 
++#ifdef __PPC64__
++#include <limits.h>
```
fabled: Could this be included unconditionally without the #ifdef conditionals?

ferseiti: I think so. I just assumed it was not needed for other archs. Do you want me to test it here?

PR #808, opened 2017-01-30
testing/xtrabackup: move from unmaintained and claim maitainership

Orhideous:


Orhideous:
Just tested on simple InnoDB database and MariaDB 10.1 from Alpine 3.4, works fine

testing/xtrabackup/APKBUILD
```diff
@@ -0,0 +1,80 @@
...
+
+pkgname=xtrabackup
+pkgver=2.4.5
+pkgrel=0
+pkgdesc='MySQL non-blocking backup solution for InnoDB and XtraDB data'
+arch=all
+url='http://www.percona.com/software/percona-xtrabackup/'
+license='GPL'
+depends='
+  libaio
```
fabled: We use tabs for indenting.

testing/xtrabackup/APKBUILD
```diff
@@ -0,0 +1,80 @@
...
+  libaio-dev
+  libgcrypt-dev
+  zlib-dev
+'
+source="
+  https://www.percona.com/downloads/XtraBackup/Percona-XtraBackup-$pkgver/source/tarball/percona-xtrabackup-$pkgver.tar.gz
+  sigevent_musl.patch
+"
+
+prepare() {
```
fabled: Prepare that applies patches only is no longer needed, the 'default_prepare' does that. This can be removed. You just need to set `builddir="$srcdir"/percona-xtrabackup-$pkgver`.

testing/xtrabackup/APKBUILD
```diff
@@ -0,0 +1,80 @@
...
+  for i in $source; do
+    case $i in
+      *.patch)
+        msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+    esac
+  done
+}
+
+build() {
+  cd "$srcdir"/percona-xtrabackup-$pkgver
```
fabled: Use `$builddir` here after you've defined it.

testing/xtrabackup/sigevent_musl.patch
```diff
@@ -0,0 +1,39 @@
+--- ./mysys/posix_timers.c.orig
++++ ./mysys/posix_timers.c
+@@ -22,6 +22,17 @@
+ #include <string.h>         /* memset */
+ #include <signal.h>
+
++struct linux_sigevent {
```
fabled: This looks ugly hack. It probably also breaks horribly because timer_create() is libc call, and expects one to use libc definition of the struct sigevent.

The correct solution is likely to change in the same file something like:
```
-#ifndef sigev_notify_thread_id
+#if !defined(sigev_notify_thread_id) && defined(__GLIBC__)
 #define sigev_notify_thread_id   _sigev_un._tid
 #endif
```

testing/xtrabackup/sigevent_musl.patch
```diff
@@ -0,0 +1,39 @@
...
++  struct linux_sigevent sigev;
+ 
+   memset(&sigev, 0, sizeof(sigev));
+ 
+@@ -200,7 +211,7 @@
+   sigev.sigev_notify= SIGEV_SIGNAL | SIGEV_THREAD_ID;
+   sigev.sigev_notify_thread_id= timer_notify_thread_id;
+ 
+-  return timer_create(CLOCK_MONOTONIC, &sigev, &timer->id);
++  return timer_create(CLOCK_MONOTONIC, (struct sigevent *)&sigev, &timer->id);
```
fabled: Yeah, this will break horribly if you run the code with the above struct linux_sigevent definition.

TBK:
@Orhideous Would it be possible to update the PR?
I really would like to see this merged :)

clandmeter:
Im closing this due to inactivity. @TBK if you like you could open a new one.

TBK:
@clandmeter That is alright, we have the mariadb-backup now so this is no longer nessecary. Mariadb-backup is based on the source of xtrabackup (https://mariadb.com/kb/en/library/mariadb-backup-overview/).

PR #809, opened 2017-01-30
testing/virt-what: new aport

HRio:


testing/virt-what/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+# Maintainer:
+pkgname=virt-what
+pkgver=1.15
+pkgrel=0
+pkgdesc="Detect if we are running in a virtual machine"
+url="https://people.redhat.com/~rjones/virt-what/"
+arch="x86 x86_64"
```
HRio: Note: should most probably work on arm as well, but have no way to verify it.

PR #813, opened 2017-01-31
community/gogs: upgrade to 0.9.113

7heo:


7heo:
Do not merge yet, there are two things left to do:

1. Ensure that upgrade happens correctly from previous versions (I had a problem with the home being set to `/home/gogs`in `/etc/passwd`)
2. Solve the 500 error problem I am having while testing the package.

PR #817, opened 2017-01-31
main/gsl: don't disable deprecated symbols

errm:
gsl disables "deprecated" symbols by default.

This for example breaks compatibility with the ruby bindings for gsl https://github.com/SciRuby/rb-gsl/issues/40

In debian they solved this issue with a similar approach https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=841611


errm:
@ncopa it would be good to get your thoughts on this as you are the Maintainer of this package?

PR #818, opened 2017-02-01
testing/check_postgres: new aport

GJRTimmer:
https://bucardo.org/wiki/check_postgres


testing/check_postgres/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+depends="postgresql perl"
+makedepends="perl-dev"
+install=""
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/bucardo/check_postgres/archive/${pkgver}.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+prepare() {
+	default_prepare || return 1
+}
```
jirutka: This is not needed, just omit whole `prepare()` function and default_prepare will be used by default.

PR #835, opened 2017-02-06
main/gcc: force gcc build not to use lib64 for ppc

ferseiti:
This patch uses the same approach used in gcc-pure64.patch, but within
ppc architecture configs, which were not being taken into account.

Signed-off-by: Fernando Seiti Furusato <ferseiti@linux.vnet.ibm.com>


main/gcc/APKBUILD
```diff
@@ -194,6 +194,7 @@ source="ftp://gcc.gnu.org/pub/gcc/releases/gcc-${_pkgbase:-$pkgver}/gcc-${_pkgba
 	300-main-gcc-add-musl-s390x-dynamic-linker.patch
 
 	fix-rs6000-pie.patch
+	gcc-ppc-pure64.patch
```
fabled: Could you just append these changes to 'gcc-pure64.patch' instead of adding a new patch. We did the same for s390x and aarch64. See commits f41e7533051039a224a302fafe8356548af0c543 and db1d2c069e66b3bf9962d76b0b48bfdcfc8d0384.

main/gcc/APKBUILD
```diff
@@ -245,7 +245,7 @@ build() {
 	armv7-*-*-*eabihf)	_arch_configure="--with-arch=armv7-a --with-tune=generic-armv7-a --with-fpu=vfpv3-d16 --with-float=hard --with-abi=aapcs-linux --with-mode=thumb";;
 	mipsel-*-*-*)		_arch_configure="--with-arch-32=mips2 --with-tune-32=mips32 --with-fp-32=32 --with-mips-plt --with-float=hard --with-abi=32";;
 	powerpc-*-*-*)		_arch_configure="--enable-secureplt --enable-decimal-float=no";;
-	powerpc64-*-*-*)	_arch_configure="--with-abi=elfv2 --enable-secureplt --enable-decimal-float=no";;
+	powerpc64le-*-*-*)	_arch_configure="--with-abi=elfv2 --enable-secureplt --enable-decimal-float=no";;
```
ferseiti: I have changed this part because the previous way had no effect whatsoever when building in a ppc64le machine.
The way it is now works.

main/gcc/APKBUILD
ferseiti: Hi @fabled. I have updated it in my branch and consequently in this PR.
Can you please take a look?
I have also changed powerpc64-* to powerpc64le-* because although it said it was targeting ppc64le in the original commit, it did not work for that arch. Now it does.

main/gcc/APKBUILD
fabled: Could be also powerpc64*-*-*-* ?

ferseiti: Yes, that worked. Updated.

PR #839, opened 2017-02-07
testing/php71: new aport

vakartel:
The PHP7.1 language runtime engine
http://php.net/

This APKBUILD is a template to build all three php versions (main/php5, community/php7, testing/php71)

### DIFF
--- a/testing/php7.1/APKBUILD
+++ b/testing/php7.1/APKBUILD
@@ -1,0 +1,419 @@
+# Contributor: Valery Kartel <valery.kartel@gmail.com>
+# Maintainer: Valery Kartel <valery.kartel@gmail.com>
+_suffix=7.1
+_pkgreal=php
+provides=$_pkgreal
+replaces=$provides
+pkgname=$_pkgreal$_suffix
+pkgver=7.1.1
+pkgrel=0
+_apiver="20160303"
+pkgdesc="The PHP${pkgver%.*} language runtime engine"
+url="http://www.php.net/"
+arch="all"
+license="PHP-3"
+options=""
+depends="$pkgname-config"
+depends_dev="$pkgname"
+makedepends="autoconf bison re2c apache2-dev libxml2-dev libxslt-dev libzip-dev bzip2-dev zlib-dev
+	aspell-dev enchant-dev expat-dev pcre-dev curl-dev gmp-dev icu-dev imap-dev
+	libical-dev libressl-dev openldap-dev net-snmp-dev db-dev krb5-dev gdbm-dev sqlite-dev
+	freetds-dev mariadb-dev postgresql-dev unixodbc-dev freetype-dev tidyhtml-dev libxpm-dev
+	libpng-dev libwebp-dev libjpeg-turbo-dev libmcrypt-dev gsoap-dev recode-dev
+	libedit-dev gettext-dev
+	"
+source="http://php.net/distributions/$_pkgreal-$pkgver.tar.bz2
+	$_pkgreal-module.conf.in
+	$_pkgreal-fpm.logrotate.in
+	$_pkgreal-fpm.initd.in
+	$_pkgreal-fpm.patch.in
+	includedir.patch.in
+	install-pear.patch
+	"
+
+# unimplemented extensions: com_dotnet interbase oci8 pdo_firebird pdo_oci
+_extensions="bcmath bz2 calendar ctype curl dba dom enchant exif ftp gd gettext gmp iconv imap:1 intl json
+	ldap mbstring mcrypt mysqli mysqlnd odbc opcache openssl pcntl pdo pdo_dblib pdo_mysql
+	pdo_odbc pdo_pgsql pdo_sqlite pgsql phar posix pspell recode session shmop snmp soap
+	sockets sqlite3 sysvmsg sysvsem sysvshm tidy wddx xml xmlreader xmlrpc xmlwriter xsl zip zlib
+	"
+for _extension in $_extensions; do
+	[ -z "${_extension##*:*}" ] && eval _index_${_extension%:*}=${_extension#*:}
+	subpackages="$subpackages $pkgname-${_extension%:*}:_extension"
+done
+_prefix_opcache="zend_"
+_pkgdesc_opcache="Zend OPcache"
+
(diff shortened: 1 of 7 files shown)


testing/php71/tidy-buffio.patch
```diff
@@ -0,0 +1,11 @@
+--- a/ext/tidy/tidy.c
++++ b/ext/tidy/tidy.c
+@@ -31,7 +31,7 @@
+ #include "ext/standard/info.h"
+ 
+ #include "tidy.h"
+-#include "buffio.h"
++#include "tidybuffio.h"
```
andypost: is this patch needed?

testing/php71/fix-asm-constraints-in-aarch64-multiply-macro.patch
```diff
@@ -0,0 +1,31 @@
+From 6cbb9f4c247c5361b8c165fbb40b4118d5d7c0e5 Mon Sep 17 00:00:00 2001
+From: Andreas Schwab <schwab@linux-m68k.org>
+Date: Mon, 25 Apr 2016 11:59:14 +0200
+Subject: [PATCH] Fix asm constraints in aarch64 multiply macro
```
andypost: this patch already accepted

testing/php71/tidy-buffio.patch
vakartel: yes, for php7 (and for php5 if you buil it with community-repo enabled)

vakartel:
Look at APKBUILD's first 'case' statement - all needed patches for specific version are listed here

vakartel:
I repeat again, this is not the only php71 build but a template for building all other versions

andypost:
Discussed at IRC and looks everyone agree to have `testing/php71` to start adoption

> ncopa> and if its not solved til we do v3.6 then we keep it in testing til v3.7
> ncopa> jirutka: i think we can enforce not creating any php71-foobar packages if we keep the php7 name

andypost:
This way we can have a time to make sure that all dependent on `php7` packages are compatible with 7.1

ncopa:
> I repeat again, this is not the only php71 build but a template for building all other versions

I am very skeptic to that approach.

What do you think about:
- leave php5 untouched. Its in maintenance mode, we just add sec fixes updates there. This will be removed as soon we can.
- leave `community/php7` as it is. this is also in maintenance mode and will only get `7.0.x` bugfixes.
- we copy `community/php7` to `testing/php7` and upgrade it to `7.1.x`. Here we do improvements that can be tested. Then we can consider backport improvements to `community/php7`

When everything we support is tested and works with php 7.1 we do:
`git rm -r community/php7 && git mv testing/php7 community/`

People who needs `php 5` for production can use v3.5 stable. People who needs `php 7.0` for production can use v3.5 stable. People who want test or develop for `php 7.1` can use `edge/testing` til we move it to `community`.

By having `community/php7` as 7.0 for now will prevent immediate breakage for edge users. By adding `testing/php7` with 7.1 we give people an opportunity to test 7.1 and we can fix the `php-*` packages one by one (possibly by having a copy in `testing/php7-*`).

If you add edge/testing to your repositories, the testing repo from the rolling release branch, then you should get the latest version of php7. If you use community repo from the rolling release branch, edge, then you will at some point get php7 upgraded from 7.0 to 7.1.

f3l1x:
Can't wait for 7.1. 👍  Ready to test guys. Just let me know.

fabled:
> I repeat again, this is not the only php71 build but a template for building all other versions
This should not be done. abuild is not designed to handle this, please keep only php-7.1 parts, and resubmit.

fabled:
> I repeat again, this is not the only php71 build but a template for building all other versions
While we appreciate the effort, this really is against how abuild and apkbuilds are designed to work. We want to keep things explicit in each apkbuild. We've learned that trying to unify things in 'smart' way usually ends up to problems and complications. We need anyway different APKBUILDs for the build system to pick them up, so you need synchronization between them even if trying to share most of the file. Also when the code base differs enough, you can't share the patches anymore, but they are complicated with the 'sed' transformation and creating patches becomes more trouble. The checksums portion would be also different, because only files in $source get in checksums. We have also other packages with major versions, and we need to keep the APKBUILDs 'unmarried'. Please repost after deleting all the non-7.1 specific parts. After that we can consider copying that APKBUILD + patches and modifying them for 7.0.

andypost:
@vakartel please squash 2 commits into 1, no reason to have useless files in history
```
git checkout upstream # checkout master
git pull  # refresh the branch
git checkout php71   # get feature branch
git rebase master # build feature on top of current sprint
git branch -m  php71 php71-old # rename local branch
git checkout -b php71 master # create new branch to push

git merge --squash php71-old  # squash commits
git commit

git push -fu origin php71 # Push new branch forcely
git branch -D php71-old # delete Old branch
```

andypost:
Now that looks perfectly clear to be accepted

ncopa:
Why cannot we simplify by simply drop the `71` suffix?

If it is not possible to drop the `71` suffix, why are we not using a dot `.` there, like we do with `lua5.1`, `lua5.2` and `lua5.3`?

vakartel:
The suffix is mandatory for library and config path-s. There is no problem
to rename it to 7.1.



2017-02-13 16:21 GMT+02:00 Natanael Copa <notifications@github.com>:

vakartel:
The suffix is mandatory for library and config path-s. There is no problem
to rename it to 7.1.

2017-02-13 16:21 GMT+02:00 Natanael Copa <notifications@github.com>:

JulienBreux:
@vakartel Thx!
@andypost LGTY?

andypost:
@JulienBreux yep, the main question in naming and @ncopa opinion

ncopa:
I would have preferred to call it `php7` instead of `php7.1`, but i suppose the maintainers opinion counts more.

@vakartel thank you very much for working on this and for not giving up along the way.

JulienBreux:
@vakartel A big thx! A beer awaits you in France / Paris 😄

qex:
I've tried new php7.1 aport and found `php7.1-gd` raises error below ...

> PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php7.1/modules/gd.so' - Error relocating /usr/lib/libpng16.so.16: inflateValidate: symbol not found in Unknown on line 0

am I doing something wrong ..?

PR #840, opened 2017-02-07
main/git: Upgrade to 2.11.1

pbregener:


pbregener:
Thanks @fabled! (How) Will this end up in 3.5 (given this maintenance release is mostly fixing regressions)? Will someone pick the commit or should I open another PR against the `3.5-stable` branch?

fabled:
We have no formal procedure for that. Normally, just a comment like this is enough. I can pick it up from here. If you want it to be tracked, adding a separate PR against right branch is good too.

pbregener:
Alright, cool. Would appreciate it if you picked it up from here 👍

PR #846, opened 2017-02-09
testing/iprutils: Adding a new package

leitao:
Iprutils is an userspace package required to handle most of
the storage controllers on ppc64le arch.

Build tested on ppc64le and amd64.

### DIFF
--- a/testing/iprutils/APKBUILD
+++ b/testing/iprutils/APKBUILD
@@ -1,0 +1,40 @@
+# Maintainer: Breno Leitao <breno.leitao@gmail.com>
+pkgname=iprutils
+pkgver=2.4.14
+pkgrel=0
+pkgdesc="user space tools required by the ipr device driver."
+url="https://github.com/bjking1/iprutils"
+arch="ppc64le"
+license="CPL"
+depends="zlib ncurses"
+makedepends="autoconf automake libtool linux-headers m4 ncurses-dev zlib-dev"
+install=""
+subpackages="$pkgname-doc"
+_pkgver=${pkgver//./-}
+source="$pkgname-$pkgver.tar.gz::https://github.com/bjking1/$pkgname/archive/rel-$_pkgver.tar.gz"
+
+builddir="$srcdir"/iprutils-rel-$_pkgver
+
+build() {
+	cd "$builddir"
+	./bootstrap.sh
+
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--localstatedir=/var \
+		--mandir=/usr/share/man \
+		--infodir=/usr/share/info \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install || return 1
+}
+md5sums="267200f22684688f49c2b0da8d4b18d4  iprutils-2.4.14.tar.gz"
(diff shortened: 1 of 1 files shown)


testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Maintainer: Breno Leitao <breno.leitao@gmail.com>
+pkgname=iprutils
+pkgver=2.4.14
+pkgrel=0
+pkgdesc="user space tools required by the ipr device driver."
+url="https://github.com/bjking1/iprutils"
+arch="all"
+license="CPL"
+depends="automake libtool m4 autoconf ncurses-dev zlib-dev"
+makedepends=""
```
fabled: Based on the builld error:
`
iprlib.h:26:25: fatal error: linux/types.h: No such file or directory
`

makedepends should have `linux-headers`

testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+install=
+subpackages="$pkgname-doc"
+_pkgver=$(echo $pkgver | sed 's/\./-/g')
```
fabled: We try to avoid subshells in main context. This is to do with the build scripting performance.
Wouldn't something like `${pkgver//./-}` work ?

testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+source="$pkgname-$pkgver.tar.gz::https://github.com/bjking1/$pkgname/archive/rel-$_pkgver.tar.gz"
+
+_builddir="$srcdir"/iprutils-rel-$_pkgver
```
fabled: This variable should be `builddir`. This got renamed recently, when we added default_prepare to use it.

testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Maintainer: Breno Leitao <breno.leitao@gmail.com>
+pkgname=iprutils
+pkgver=2.4.14
+pkgrel=0
+pkgdesc="user space tools required by the ipr device driver."
+url="https://github.com/bjking1/iprutils"
+arch="all"
```
fabled: Even if this builds on all, is it useful? Or would it make sense to be limited to ppc?

testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+		--localstatedir=/var \
+		--mandir=/usr/share/man \
+		--infodir=/usr/share/info \
+		|| return 1
+	make || return 1
+}
+
+package() {
+	cd "$_builddir"
+	make DESTDIR="$pkgdir" install
```
fabled: Currently this should have `|| return 1` since it's not the last command.

testing/iprutils/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+	rm -rf "$pkgdir"/usr/lib
+	return 0
```
fabled: `return 0` is not really needed, by default the status of last command is returned, which should be zero when successful.

testing/iprutils/APKBUILD
leitao: No, it is not useful on !ppc64le arches. I am changing all these ppc64le packages to be ppc64le-only packages.

testing/iprutils/APKBUILD
leitao: Makes sense. I suppose I should not call  `default_prepare`  because I do not have external patches, right?

leitao:
Thanks @fabled. I just reworked on this package. I think it is ready for the second review.

PR #849, opened 2017-02-09
community/tclx: remove use of rresvport

snoe925:
Remove use of network routines in libc on Debian.  In this case rresvport is just a helper to allocate a port < 1024.  The user can do that as root.


jirutka:
> Remove use of network routines in libc on Debian.

WAT? This is not Debian… o.O

community/tclx/APKBUILD
```diff
@@ -11,11 +11,18 @@ depends="tcl"
...
+	rresvport.patch
+	"
 
 _major=${pkgver%.*}
 builddir="$srcdir"/tclx${_major}
 
+prepare() {
+	cd "$builddir"
+	default_prepare
+}
```
jirutka: Why have you added this? `default_prepare` is called by default when custom `prepare` function is not defined.

snoe925: I used main/openssh as an example.  I cut and pasted the prepare, but after removing one line I did not understand to remove the entire block.

fabled:
Applied with pkgrel bump in commit 1ab30ef138c7b4200ee7c543bafa9502869799a3. I had to merge manually using the diff because this was on top of your earlier tree, and not on top of the aports master. Please submit further changes rebased on top of the aports master branch. Thanks.

PR #856, opened 2017-02-10
main/libwebp: upgrade to 0.6.0

misery:


jirutka:
⚠️ **Note for maintainers:** soname change, all dependent packages must be rebuilt.

PR #863, opened 2017-02-11
testing/zerotier-one: set more specific initd depends

kyleparisi:
The reason for this is that zerotier-one can be started without
functioning network and will create it's net-devices as soon as it can.
So, it provides net devices which might be used in e.g. /etc/fstab


jirutka:
Please update checksums (run `abuild checksum`) and update this PR.

PR #877, opened 2017-02-15
main/py-cryptography: upgrade to 1.7.2

misery:


jirutka:
This Python module requires [asn1crypto](https://pypi.python.org/pypi/asn1crypto) now, but we don’t have package for it yet, so it must be added.

misery:
Are you sure with asn1crypto and v1.7? Changelog means v1.8

https://cryptography.io/en/latest/changelog/

fabled:
Seems this got upgraded in commit 5b669ec64c5a43478b271bac768ef49d1a5e7c66.

PR #878, opened 2017-02-15
testing/php7-apcu: new aport 5.18

andypost:
Compatibility with php7.1


rnalrd:
Hi,
pkgrel is set to 1 instead of 0.
Also "check" stage is missing. Enabling it causes the build to fail. Please either fix it (recommended) or add "options=!check" as a _temporary_ workaround.

Thanks!

andypost:
Closing as outdated

PR #880, opened 2017-02-16
main/sqlite: increase SQLITE_MAX_VARIABLE_NUMBER

andypost:
To support large amount of variables in queries it needs to be up.

Debian/Ubuntu is 250000
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=717900

### DIFF
--- a/main/sqlite/APKBUILD
+++ b/main/sqlite/APKBUILD
@@ -17,7 +17,7 @@ esac
 [ $_d -lt 10 ] && _d=0$_d
 _ver=${_a}${_b}${_c}${_d}
 
-pkgrel=0
+pkgrel=1
 pkgdesc="A C library that implements an SQL database engine"
 url="http://www.sqlite.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


andypost:
I faced with it while trying to use sqlite with php in https://www.drupal.org/node/2031261

jirutka:
Sorry, but if some software needs more than 999 placeholders in a single query, then it’s doing something very, very wrong (or I misunderstood what this parameter do). Read https://www.sqlite.org/limits.html#max_variable_number.

andypost:
That's not correct, it's not rare case when you need to do insert with a lot of values instead of subsequent tons of inserts.
It's mostly about concurrency and transactions.

The most common case is insert or reaplace 1-2 thousands rows (each row has few fields) in atomic way, see https://www.sqlite.org/lang_conflict.html

That's why mostly all distros increasing this parameter slightly.
And it's always a consensus between memory and CPU when talking about performance

vakartel:
Seems like sqlite upstream and end-point users are on different planets.
If we see this value increased in many well-known distros, so it really need in real-life projects.

andypost:
btw Sqlite widely used on embed devices that has small RAM so probably that's why default so small

kaniini:
The following distros have changed to 250k:

- Debian (and derivatives like Ubuntu)
- Arch
- Homebrew (not really a distro but work with me here)

It's not "all distros".  With that said, it seems that this limit applies to the SQL99 `IN` operator, which could very easily have more than 999 values.  So lets apply this.

jirutka:
It's only few distros, not most, and both of them are known for not very good QA... I checked Fedora and as I've expected, they did not increased this limit.

PR #885, opened 2017-02-17
community/docker: Fixing APKBUILD, adding tini.

gena01:
* Trying to fix broken APKBUILD. (Wasn't catching package
  errors)
* Adding tini (docker-init binary)

### DIFF
--- a/community/docker/APKBUILD
+++ b/community/docker/APKBUILD
@@ -1,33 +1,36 @@
 # Maintainer: Eivind Uggedal <eivind@uggedal.com>
 pkgname=docker
-pkgver=1.13.1
-_ver=${pkgver/_/-}
+pkgver=17.03.0
+_ver=${pkgver/_/-}-ce
 pkgrel=0
 _gitcommit=v$_ver
 pkgdesc="Pack, ship and run any application as a lightweight container"
 url="http://www.docker.io/"
 arch="all"
 license="ASL 2.0"
-depends="iptables git xz"
+depends="iptables"
 depends_dev=""
-makedepends="go btrfs-progs-dev bash linux-headers coreutils libseccomp-dev"
-options="!strip"
+makedepends="go btrfs-progs-dev bash linux-headers coreutils libseccomp-dev cmake lvm2-dev"
+options="!check"
 install="$pkgname.pre-install"
 
 # $ grep COMMIT hack/dockerfile/binaries-commits
 # TOMLV_COMMIT=9baf8a8a9f2ed20a8e54160840c492f937eeaf9a
-# RUNC_COMMIT=2f7393a47307a16f8cee44a37b262e8b81021e3e
-# CONTAINERD_COMMIT=03e5862ec0d8d3b3f750e19fca3ee367e13c090e
+# RUNC_COMMIT=a01dafd48bc1c7cc12bdb01206f9fea7dd6feb70
+# CONTAINERD_COMMIT=977c511eda0925a723debdc94d09459af49d082a
 # TINI_COMMIT=949e6facb77383876aeff8a6944dde66b3089574
 # LIBNETWORK_COMMIT=0f534354b813003a754606689722fe253101bc4e
 # VNDR_COMMIT=f56bd4504b4fad07a357913687fb652ee54bb3b0
+# BINDATA_COMMIT=a0ff2567cfb70903282db057e799fd826784d41d
 
-_runc_ver=2f7393a47307a16f8cee44a37b262e8b81021e3e
-_containerd_ver=03e5862ec0d8d3b3f750e19fca3ee367e13c090e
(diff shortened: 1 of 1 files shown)


community/docker/APKBUILD
```diff
@@ -36,7 +37,7 @@ source="
 	runc-$_runc_ver.tar.gz::https://github.com/opencontainers/runc/archive/$_runc_ver.tar.gz
 	containerd-$_containerd_ver.tar.gz::https://github.com/docker/containerd/archive/$_containerd_ver.tar.gz
 	libnetwork-$_libnetwork_ver.tar.gz::https://github.com/docker/libnetwork/archive/$_libnetwork_ver.tar.gz
-
+	tini-$_tini_ver.tar.gz::https://github.com/krallin/tini/archive/$_tini_ver.tar.gz
```
misery: tini is already in repo....

https://github.com/alpinelinux/aports/blob/master/community/tini/APKBUILD

gena01: It's a bit more complicated than that. It expects a specific version (commit) and a specific binary name (docker-init)

ncopa: can we get away with a symlink?

gena01: My concern is the `docker info` line:
init version: N/A (expected: 949e6facb77383876aeff8a6944dde66b3089574)

Might be tied to specific version of tini.

community/docker/APKBUILD
```diff
@@ -2,15 +2,15 @@
...
 pkgver=1.13.0
 _ver=${pkgver/_/-}
-pkgrel=0
+pkgrel=1
 _gitcommit=v$_ver
 pkgdesc="Pack, ship and run any application as a lightweight container"
 url="http://www.docker.io/"
 arch="all"
 license="ASL 2.0"
 depends="iptables git xz"
```
gena01: @ncopa any idea why git and xz are dependencies for Docker?

ncopa: I dont know. It seems to be introduced with commit 4248d1a88bc, which was the first commit.

pickfire:
Now docker is in monthly releases, maybe bump the version here as well?

community/docker/APKBUILD
gena01: @ncopa I can try adding it as a dependency. Who would own the symlink? Is there a clean way to do this? I could also try to patch in the tini path instead of the symlink name.

gena01: @ncopa I have added a shell wrapper for this in my local copy (didn't push it yet). Also fixed the git dependency (actually needs to be a make dependency). Also re-enabled aufs (any idea why it's off?). Now if you look the PR is failing because it can't find a header file when trying compile with btrfs-progs-dev (not sure what to do here. I could also disable btrfs graphdriver)

pickfire:
Looks like someone has already submitted a patch on patchwork.

http://patchwork.alpinelinux.org/patch/3200/

gena01:
@pickfire Thanks for the link. Most of the changes look very similar to what I have in the PR. I might pull in the fish piece. Also was looking to add a "-doc" package with man pages (need to fix the building part).

@ncopa btrfs graphdriver doesn't compile for me. See why Travis CI failed. I tried wrapping tini, but it doesn't quite work right. Tini compiled from "git clone" injects a git sha1 commit into "tini --version" output. The release build doesn't. So "docker info" can't detect the git commit, but should work otherwise. with my wrapper. We could patch up the tini package to include git hash.

pickfire:
@gena01 He removed the patch knowing you had it. His version built.

community/docker/APKBUILD
```diff
@@ -127,12 +136,13 @@ package() {
...
 		"$pkgdir"/usr/bin/docker-containerd-ctr || return 1
 
-	install -Dm755 bundles/$ver/dynbinary/dockerinit-$ver \
-		"$pkgdir"/usr/lib/docker/dockerinit
+	install -Dm755 "$_tini_builddir"/tini-static \
+		"$pkgdir"/usr/bin/docker-init || return 1
+
 	install -Dm755 contrib/init/openrc/docker.initd \
-		"$pkgdir"/etc/init.d/docker
+		"$pkgdir"/etc/init.d/docker || return 1
```
pickfire: Leonardo wrote:
> +build() {
> +     cd "$builddir"
> +     make || return 1
> +}

We don't need any more to use "|| return 1". Please see
http://lists.alpinelinux.org/alpine-devel/5608.html

> +
> +package() {
> +     cd "$builddir"
> +     make DESTDIR="$pkgdir" install || return 1
> +}
> +

http://lists.alpinelinux.org/alpine-aports/4391.html

ncopa: we should keep the `|| return 1` til after 3.6-stable is branched. (when v3.6.0 is released)

http://lists.alpinelinux.org/alpine-devel/5608.html

PR #888, opened 2017-02-17
testing/sbt: new aport

errm:
http://www.scala-sbt.org/ The interactive build tool, sbt is a build
tool for Scala, Java, and more.

### DIFF
--- a/testing/sbt/APKBUILD
+++ b/testing/sbt/APKBUILD
@@ -1,0 +1,40 @@
+# Contributor: Ed Robinson <ed@reevoo.com>
+# Maintainer: Ed Robinson <ed@reevoo.com>
+pkgname=sbt
+pkgver=0.13.13
+pkgrel=0
+pkgdesc="The interactive build tool, sbt is a build tool for Scala, Java, and more."
+url="http://www.scala-sbt.org/"
+arch="noarch"
+license="BSD-3"
+depends="openjdk8-jre bash"
+source="https://dl.bintray.com/$pkgname/native-packages/$pkgname/$pkgver/$pkgname-$pkgver.tgz
+global-sbtopts.patch"
+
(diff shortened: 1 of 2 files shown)


testing/sbt/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+global-sbtopts.patch"
+
+builddir="$srcdir/$pkgname-launcher-packaging-$pkgver"
+_basedir="/usr/share/java/$pkgname"
+
+prepare() {
+	default_prepare || return 1
+	cd "$builddir"
+
+	# Remove windows specific files
```
kaniini: why are we removing these files?  you say they are windows specific, but there's no need to remove them just because of that.

testing/sbt/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+	local confdir="$pkgdir/etc/$pkgname"
+	cd "$builddir"
+
+	install -dm755 "$destdir"/bin || return 1
+	install -m755 -t "$destdir"/bin bin/* || return 1
+
+	install -dm755 "$confdir" || return 1
+	install -m644 -t "$confdir" conf/* || return 1
+
+        mkdir -p "$pkgdir"/usr/bin
```
kaniini: spurious indentation (spaces vs tabs)

testing/sbt/APKBUILD
jirutka: Why to add into pkg files that are totally useless on Alpine? These files definitely should be removed. Or do you disagree just with removing them in prepare phase?

kaniini: In the prepare phase.  Obviously they should not be included in the package itself.

errm:
Changes made @kaniini PTAL

PR #889, opened 2017-02-17
python 3.6.0

scadu:


jirutka:
I’ve removed commit “main/python3: rebuild against new version of python3” – I assume that it was here by accident – and “main/boost: fix checksums” – changes in already released tarballs are suspicious (especially on SourceForge) and we should not just overlook it. I’ve notified ncopa to look at it.

clandmeter:
I just found out that many python packages were still build against 3.5 in edge.
How did you generate the list of python packages that needed to be upgraded?

jirutka:
IIRC I generated it with SQL query on aports DB. Could you please name some pkgs that were not rebuilt, so I can check how are they different and what went wrong?

jirutka:
Actually, it's PR by scadu, so more likely he somehow generated the list and I just checked it using SQL query? Don't really remember. :/

scadu:
@jirutka I used the script provided by @ncopa. I will share it here once I have an access to my machine.

clandmeter:
You can find a list of packages in aports which i bumped today.

scadu:
@clandmeter Here is the [script](https://gist.github.com/scadu/db1549a1b0afaca7b49e202ded0c9160).

PR #890, opened 2017-02-17
community/php5: backport features and fixes from testing/php7.1

vakartel:
rebuilt with php7.1 template

I know that php5 is old and can be removed in future.
But I want to stay it as bugless as possible for history.


vakartel:
@andypost , that's it

andypost:
> ERROR: Failed to create usr/lib/libpcre16.a: No space left on device
> ERROR: pcre-dev-8.40-r0: No space left on device

main/php5/php-fpm.initd
```diff
@@ -8,14 +8,13 @@
 
 : ${name:="PHP FastCGI Process Manager"}
 
-command="/usr/bin/php-fpm"
+command="/usr/sbin/php-fpm5"
```
andypost: I strongly against rename the binary
so after package update I will get 2 binaries (old in old path and new)
also that will break all php5 docker images that expect that binary with old name

vakartel: It was a mistake when original php5 apkbuild was created. php-fpm binary natively lives in /usr/sbin

I made a compatibility link to /usr/bin

PR #891, opened 2017-02-18
testing/logstash: update to 5.2.2 / use supervise-daemon

itoffshore:
adds service supervision with OpenRC's builtin `supervise-daemon`
`logstash/bin` added to `$PATH`
take ownership

### DIFF
--- a/testing/logstash/APKBUILD
+++ b/testing/logstash/APKBUILD
@@ -1,64 +1,91 @@
 # Contributor: Steeve Chailloux <steeve@chaahk.com>
-# Maintainer:
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=logstash
-pkgver=1.4.2
-pkgrel=2
+pkgver=5.2.2
+pkgrel=0
 pkgdesc="A flexible, open source, data collection, parsing and enrichment pipeline"
 url="http://logstash.net"
 arch="noarch"
 license="apache"
-depends="openjdk7-jre-base"
-depends_dev=""
+depends="openjdk8-jre-base java-jffi-native libc6-compat bash"
 makedepends="$depends_dev"
 install="$pkgname.pre-install"
 pkgusers="logstash"
-subpackages=""
-source="
-    https://download.elasticsearch.org/${pkgname}/${pkgname}/${pkgname}-${pkgver}.tar.gz
-    $pkgname.confd
-    $pkgname.initd
-    "
-
-_builddir="${srcdir}/${pkgname}-${pkgver}"
-prepare() {
-	local i
-	cd "$_builddir"
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
-}
+subpackages="$pkgname-doc"
+options="!strip !check"
+source="$pkgname-$pkgver.tar.gz::https://artifacts.elastic.co/downloads/$pkgname/$pkgname-$pkgver.tar.gz
+	$pkgname.confd
+	$pkgname.initd
+	$pkgname.conf
+	$pkgname.profile
+	"
+builddir="${srcdir}/${pkgname}-${pkgver}"
 
 build() {
-    echo "Nothing to build"
+	return 0
 }
 
 package() {
-	cd "$_builddir"
+	local datadir=/var/lib/logstash logdir=$datadir/logs confdir=/etc/logstash
+	local homedir=/usr/share/logstash heapsize=150m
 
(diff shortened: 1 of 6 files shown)


itoffshore:
`supervise-daemon` restarts `logstash` ok when the process is killed.

It's not clear [in the guide](https://github.com/OpenRC/openrc/blob/master/supervise-daemon-guide.md) but each service needs it's own pid - perhaps use a format of `$SVCNAME.sd.pid`

testing/logstash/APKBUILD
```diff
@@ -1,64 +1,88 @@
...
+pkgver=5.2.1
+pkgrel=0
 pkgdesc="A flexible, open source, data collection, parsing and enrichment pipeline"
 url="http://logstash.net"
-arch="noarch"
+arch="all"
 license="apache"
-depends="openjdk7-jre-base"
-depends_dev=""
+depends="openjdk8-jre-base libc6-compat"
```
kaniini: why is libc6-compat needed?  is there now native JNI code?

same for `$arch`

testing/logstash/logstash.conf
```diff
@@ -0,0 +1,12 @@
+input {
+        file {
+        type => "syslog"
+        path => ["/var/log/**/*.log"]
+        }
+}
+
+output {
+        elasticsearch { hosts => ["localhost:9200"] }
```
kaniini: this does not seem like a reasonable default as the package does not have a hard dependency on ES

testing/logstash/APKBUILD
itoffshore: without `libc6-compat` tailing files does not work

testing/logstash/logstash.conf
itoffshore: I just used the settings from the [ES reference docs](https://www.elastic.co/guide/en/logstash/current/configuration.html) - have commented out `elasticsearch`.

I also changed the log directory in the `confd` so logstash does not consume it's own logs

By default the `stdout` plugin prints to the `/var/lib/logstash/logs/error.log`)

jirutka:
If it contains JARs with native binaries, then these should be removed. Please read at least https://github.com/jruby/jruby/wiki/JRuby-on-Alpine-Linux

jirutka:
BTW I gave up packaging Logstash for a reason... I hope that the new version is not so horrible ball of mud as it was a year ago, otherwise it'd be just a waste of time.

itoffshore:
@jirutka - ok - I'll check which `jar` is built against `glibc` & build a native version

jirutka:
Wait a moment, IIRC Logstash runs on JRuby and there's no jruby in dependencies. Aren't you bundling it, are you!?

jirutka:
If they haven't pulled another deps with native extensions, then we already have all the needed java libs with native extensions in aports.

itoffshore:
I was using the `vendor` libraries that come with `logstash` - I'll try & work out which one is the problem

itoffshore:
it uses `jffi` which need `libc6-compat` which I have as a depends

itoffshore:
I'll add your `java-jffi-native` as a depends & remove the bundled `jffi`

jirutka:
Do not even consider bundling jruby or any JAR with included native binaries.

I've already made package for jffi:
https://pkgs.alpinelinux.org/package/edge/community/x86_64/java-jffi

Actually I did whole work with packaging JRuby just to make a package for Logstash. I haven't finished Logstash itself, b/c it was so messy, that it was nearly impossible to do it right.

testing/logstash/logstash.post-deinstall
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
 deluser logstash 2>/dev/null
+rm -f /etc/profile.d/logstash 2>/dev/null
```
jirutka: What the... Maybe you want also remove whole home directory as well? Just for fun, users love when pkgs removes their files when uninstalling...

itoffshore: it's just removing `/usr/share/logstash/bin` from `$PATH`

itoffshore: I removed the `post-deinstall` script & fixed the merge conflict.

testing/logstash/logstash.conf
itoffshore: `logstash.conf` has `elasticsearch` input commented out now  - can it be pulled ?

testing/logstash/APKBUILD
itoffshore: fixed to use Alpine's `java-jffi-native` & `libc6-compat`

PR #892, opened 2017-02-18
testing/lxd: upgrade to 2.9.1; fix dependencies.

vifino:
Hello.

To actually use LXD, you need to have a few things installed at runtime, like iptables/ip6tables, dnsmasq and LXC.


jirutka:
lxd depends on `lxc-libs` (this is not explicitly declared in APKBUILD, because this dependency is automatically resolved from so libs), isn’t that sufficient?

vifino:
It is not. 

LXD requires `/usr/lib/lxc` and `lxc-monitord` for example. I just tried it with just `lxc-libs`, it needs the full `lxc`.

vifino:
Ah, I forgot. It lacks other deps to be usable, too. iptables/ip6tables and dnsmasq for sure, otherwise the networking is broken, `lxd init` doesn't work when creating a bridge, etc..

vifino:
I also took the liberty of bumping the package to 2.9.1. Hope that's fine by you.

vifino:
Bumped to 2.9.3.

ncopa:
looks like this was merged

PR #893, opened 2017-02-18
community/php7: backport features and fixes from testing/php7.1

vakartel:
php7.1 build template used to build this.


vakartel:
@andypost - let's try to make bugless one

community/php7/APKBUILD
```diff
@@ -1,71 +1,70 @@
...
 	$pkgname-module.conf
-	install-pear.patch
-	tidy-buffio.patch
+	$pkgname-fpm.logrotate
+	$pkgname-fpm.initd
+	$pkgname-fpm.patch
 	includedir.patch
+	tidy-buffio.patch
+	install-pear.patch
 	fix-asm-constraints-in-aarch64-multiply-macro.patch
```
andypost: this probably no longer needed, at least it was picked to upstream

community/php7/APKBUILD
```diff
@@ -1,71 +1,70 @@
...
+depends="$pkgname-config"
+depends_dev="$pkgname"
+replaces="$_pkgreal $_pkgreal-cli"
+provides="$replaces $pkgname-cli"
 makedepends="autoconf bison re2c apache2-dev libxml2-dev libxslt-dev libzip-dev bzip2-dev zlib-dev
 	aspell-dev enchant-dev expat-dev pcre-dev curl-dev gmp-dev icu-dev imap-dev
 	libical-dev libressl-dev openldap-dev net-snmp-dev db-dev krb5-dev gdbm-dev sqlite-dev
 	freetds-dev mariadb-dev postgresql-dev unixodbc-dev freetype-dev tidyhtml-dev libxpm-dev
-	libpng-dev libwebp-dev libjpeg-turbo-dev libmcrypt-dev gsoap-dev recode-dev
-	libedit-dev paxmark gettext-dev
```
andypost: recode & paxmark - why they removed while recode extension new shared?

community/php7/APKBUILD
```diff
@@ -1,71 +1,70 @@
...
+# unimplemented extensions: com_dotnet interbase oci8 pdo_firebird pdo_oci
+_extensions="bcmath bz2 calendar ctype curl dba dom enchant exif ftp gd gettext gmp iconv imap:1 intl json
 	ldap mbstring mcrypt mysqli mysqlnd odbc opcache openssl pcntl pdo pdo_dblib pdo_mysql
-	pdo_odbc pdo_pgsql pdo_sqlite pgsql phar:phar posix pspell session shmop snmp soap
-	sockets sqlite3 sysvmsg sysvsem sysvshm tidy wddx xml xmlreader xmlrpc xsl zip zlib
-	"
-subpackages="$pkgname-dev $pkgname-doc $pkgname-apache2 $pkgname-phpdbg $pkgname-embed
-	$pkgname-litespeed $pkgname-cgi $pkgname-fpm $pkgname-pear::noarch
+	pdo_odbc pdo_pgsql pdo_sqlite pgsql phar posix pspell recode session shmop snmp soap
+	sockets sqlite3 sysvmsg sysvsem sysvshm tidy wddx xml xmlreader xmlrpc xmlwriter xsl zip zlib
```
andypost: xmlwriter/recode flux... makes sense to have them as shared

community/php7/APKBUILD
```diff
@@ -137,231 +135,270 @@ _build() {
...
-	esac
+	pkgdesc=$(eval echo \$_pkgdesc_$name)
+	: ${pkgdesc:=$(head -n1 "$builddir"/ext/$name/CREDITS)}
+	pkgdesc="PHP$_suffix extension: ${pkgdesc:-$name}"
 
-	# extension dependencies and loading order
-	local elo=00
-	[ -z "$extdepends" ] || elo=01
+	for dep in $(grep -o "_DEP('$name'.*" "$builddir"/ext/$name/config.w32 | sed "s/[',);]//g"); do
+		[ -z "${_extensions##*$dep*}" ] && depends="$depends $pkgname-$dep"
```
andypost: Nice trick ++

community/php7/APKBUILD
vakartel: recode-dev is on place
gsoap-dev and paxmark are removed because are not used anywhere

community/php7/APKBUILD
andypost: Actually this might wrong because *nix based build of php should have different deps because it have much more extensions bundled

community/php7/APKBUILD
andypost: Looks it still required for 7.0, because was not backported in https://github.com/php/php-src/pull/2236

vakartel:
I think, maximum number of extensions must be compiled as shared, because many of them dependent on external libraries.

At 'Bundled extensions' list i can point to: 

gd (libgd -> libpng, libjpeg, freetype etc.)
iconv (libiconv)
mbstring (libmbfl)
pcre (libpcre)
sqlite3 (libsqlite3)
zlib (libz)

All those extensions are in 'bundled` list, because have an own copy of dependent lib sources and can be built without external system ones. See php/ext sources.

In any case we make a general-purpose build. If somebody want to have some special in production, they can build it's own php with compiled-in extensions they want.

And about bcmath. official docs says: "The Windows version of PHP has built-in support for this extension. You do not need to load any additional extensions".
So, on other systems than Windows you can have it as shared, not compiled into main binary.

ncopa:
IIRC I patched php5 to be able to use system `xml`.

In general, we want use shared system libs for as much as possible because it reduces maintenance work for security fixes in future. (we avoid needing to patch same issues multiple places)

vakartel:
I recheck it. For now use system libxml & expat libs. I can't use system xmlrpc due to compile errors.

community/php7/APKBUILD
jirutka: No, it’s not in 7.1.4, see https://github.com/php/php-src/blob/php-7.1.4/Zend/zend_multiply.h#L56.

community/php7/APKBUILD
```diff
@@ -1,72 +1,71 @@
...
+	$pkgname-litespeed $pkgname-cgi $pkgname-fpm $pkgname-pear::noarch
+	$pkgname-phar-utils:_phar:noarch $subpackages $pkgname-config::noarch"
+builddir="$srcdir/$_pkgreal-$pkgver"
 
 prepare() {
 	cd "$builddir"
 
+	sed "$(($(grep -n 'Pool Definitions' sapi/fpm/php-fpm.conf.in | cut -d: -f1)+3)),\$d" \
+		-i sapi/fpm/php-fpm.conf.in
+	echo include=@php_fpm_sysconfdir@/php-fpm.d/*.conf >> sapi/fpm/php-fpm.conf.in
```
jirutka: This is quite cryptic and moreover totally unnecessary, because `include=@php_fpm_sysconfdir@/php-fpm.d/*.conf` is already there in 7.0.16 (the version specified in this exact revision) and newer.

community/php7/APKBUILD
```diff
@@ -101,11 +100,10 @@ _build() {
 		--enable-exif=shared \
 		--enable-ftp=shared \
 		--with-gd=shared --with-webp-dir=shared --with-jpeg-dir=shared --with-png-dir=shared --with-xpm-dir=shared \
-		--with-freetype-dir=shared --enable-gd-native-ttf --enable-gd-jis-conv \
+		--with-freetype-dir=shared --enable-gd-native-ttf --disable-gd-jis-conv \
```
jirutka: Why have you disabled `gd-jis-conv`?

community/php7/APKBUILD
```diff
@@ -86,9 +86,10 @@ _build() {
 		--with-config-file-path=/etc/$pkgname \
 		--with-config-file-scan-dir=/etc/$pkgname/conf.d \
 		--disable-short-tags \
+		--enable-fileinfo=shared \
 		--with-openssl=shared --with-kerberos --with-system-ciphers \
-		--with-pcre-regex --with-pcre-dir=/usr \
-		--with-zlib=shared --with-zlib-dir=/usr \
+		--with-pcre-regex=/usr --with-pcre-dir \
+		--with-zlib=shared --with-zlib-dir \
```
jirutka: Why did you change `--with-zlib-dir=/usr` to `--with-zlib-dir`? Doesn’t it mean that bundled lib will be used instead of the system-provided?

community/php7/APKBUILD
```diff
@@ -138,214 +136,275 @@ _build() {
...
 	install -Dm755 "$builddir"/sapi/phpdbg/phpdbg \
-		 "$subpkgdir"/usr/bin/phpdbg7
+		 "$subpkgdir"/usr/bin/phpdbg$_suffix || return 1
+	ln -sf /usr/bin/phpdbg$_suffix "$subpkgdir"/usr/bin/phpdbg
 }
 
-embed() {
-	pkgdesc="PHP Embedded Library"
-	depends="$pkgname-common"
+libs() {
```
jirutka: PHP calls it “embed SAPI”, VoidLinux “php-embed”, Fedora “php-embedded”, Arch Linux “php-embed”, Debian “libphp-embed”… So what reason do you have to rename it from php7-embed to php7-libs?!

community/php7/APKBUILD
```diff
@@ -138,214 +136,275 @@ _build() {
...
-	local file; for file in pear peardev pecl; do
-		sed -i -e "s~/usr/bin/php~/usr/bin/php7~g" \
-			-e "s~PHP=php~PHP=php7~" \
-			"$pkgdir"/usr/bin/$file || return 1
-	done
-	paxmark -m "$pkgdir"/usr/bin/php7
-
-	find "$pkgdir" -name '.*' | xargs rm -rf || return 1
-	rmdir "$pkgdir"/var/run
+	rm -fr "$pkgdir"/.[[:alpha:]]* || return 1
```
jirutka: Maybe you’ve overlooked that dot files are not only in root, but also in `usr/share/pear`…? I’m not sure if they are any useful and should be preserved, or not. For example VoidLinux removes them.

community/php7/APKBUILD
vakartel: I think, it must be renamed to libphp7. Those -embed or -embedded suffix do not need - it's just a shared library.

community/php7/APKBUILD
vakartel: those files are useless for system-wide install, because we use 'apk' for install new packets and pear know nothing about it. It's just for perl-user installs

community/php7/APKBUILD
vakartel: Because it broke any non-latin font rendering. This flag for japan users only any nobody knows why it enabled be default. It's an old known bug.

community/php7/APKBUILD
vakartel: As you remember, a time ago I suggested a common APKBUILD for all php5, php7.0 and php7.1. So this code is for php5 only. It does nothing for php7.0 and php7.1. I don't removed it from php7.0 and 7.1 just to have a minimum diff-s when I change something.
So now it have no sence.

community/php7/APKBUILD
vakartel: Win32 configs have more stricted deps definitions, so I used it first.

community/php7/APKBUILD
jirutka: I don't see any reason to rename it. We don't follow Debian's conventions for naming pkgs.

jirutka:
Replaced by #1339. I’ve picked useful changes from this branch and applied it in small commits.

PR #898, opened 2017-02-19
community/elasticsearch: update to 5.2.1 / use supervise-daemon 

itoffshore:
ES now uses `$ES_CONF_DIR/jvm.options` to set java options so duplicate
java options have been removed from the `initd` / `conf.d`.

service supervision with OpenRC's `supervise-daemon` added

added `README.alpine` with notes for upgrading ES `2.x` => `5.x`

### DIFF
--- a/community/elasticsearch/APKBUILD
+++ b/community/elasticsearch/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=elasticsearch
-pkgver=2.3.5
+pkgver=5.2.1
 pkgrel=0
 pkgdesc="Open Source, Distributed, RESTful Search Engine"
 url="https://www.elastic.co/products/elasticsearch"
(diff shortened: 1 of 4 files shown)


itoffshore:
upgrading `2.x => 5.x` should be as simple as stopping & restarting with the new `initd`.

I've [uploaded `64bit` pkgs](http://dev.it-offshore.co.uk/downloads/elasticsearch-5.2.1.tar.xz) so you can see they work as intended.

jirutka:
> ES now uses `$ES_CONF_DIR/jvm.options` to set java options so duplicate java options have been removed from the `initd` / `conf.d`.

IIRC it provides it since beginning and I haven’t used it for purpose. But don’t remember details and don’t wanna look into it now.

itoffshore:
It just seemed less confusing to have the java settings where they were expected to be according to the [elasticsearch reference docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/setting-system-settings.html) (& also making the `initd` easier to read).

I didn't notice any difference in the java commands in `htop`

kaniini:
This seems fine.

PR #912, opened 2017-02-20
testing/bloaty: new aport

mitchty:
Note, this entire apkbuild is a hack, there is no upstream source
tarball anywhere. So I hacked in the build function to git clone
the source and checkout at a specific known version for now.

Not sure what the best option is in cases like these.

### DIFF
--- a/testing/bloaty/APKBUILD
+++ b/testing/bloaty/APKBUILD
@@ -1,0 +1,42 @@
+#-*-mode: Shell-script; coding: utf-8;-*-
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+_commitid=2234386bcee7297dfa1b6d8a5d20f95ea4ed9bb0
+_commitid_libfuzzer=e9387f0b6d411e18ca75f08aecfec1238b8e113f
+pkgname=bloaty
+# Fake a version until upstream does one
+# format is simply 0.0.0.0.${commit_count}
+pkgver=0.0_git20170301
+pkgrel=0
+pkgdesc="A size profiler for binaries"
+url="https://github.com/google/bloaty"
+arch="all"
+license="apache"
+depends=""
+makedepends="gcc git"
+# This software has no releases yet, and has git submodules for 3 things
+# I managed to get everything but libFuzzer from a released tarball.
+# Upstream already has an issue from an arch AUR maintainer to fix this.
+source="https://github.com/google/bloaty/archive/${_commitid}.tar.gz
+	https://github.com/google/re2/archive/2017-01-01.tar.gz
+	https://github.com/google/googletest/archive/release-1.8.0.tar.gz
+	byteswap.patch
+	"
+builddir="$srcdir/${pkgname}-${_commitid}"
+
+build() {
+	rmdir third_party/re2 third_party/googletest || return 1	
+	ln -s ../../re2-2017-01-01 third_party/re2 || return 1
+	ln -s ../../googletest-release-1.8.0 third_party/googletest || return 1
+	git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer third_party/libFuzzer || return 1
+	(cd third_party/libFuzzer && git checkout ${_commitid_libfuzzer}) || return 1
+	make || return 1
+}
+
+package() {
+	install -Dm0755 "${builddir}/bloaty" "${pkgdir}/usr/bin/bloaty" || return 1
+	strip "${pkgdir}/usr/bin/bloaty" || return 1
+}
+sha512sums="2e9545623ce4e20318cf8c8a13f006993318841769ebfee62910659c7c66b6099e228975739c70dd6499cbb109b5440aa559a3d384c0f1574748565eeeda7333  2234386bcee7297dfa1b6d8a5d20f95ea4ed9bb0.tar.gz
+52ec99fe1bfb2754e11e3767a561208f14c3e5c03142c37b43df7cf5a406896396854a3776ae536ddf293a770d8ca80f9c761c2c476394693a5954c15550a731  2017-01-01.tar.gz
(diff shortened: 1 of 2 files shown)


fabled:
For github you can download tarball of specific commit. For an example, see: main/raspberrypi/APKBUILD.

mitchty:
Ah cool, I keep forgetting about that feature of github, I'll update this to use that then thanks!

mitchty:
So it turns out there were three submodules in this too, I couldn't find a release for one of the three, libFuzzer as used from chromium. So not sure how to handle that one outside of just checking it out at a certain point in time.

testing/bloaty/APKBUILD
```diff
@@ -0,0 +1,41 @@
+#-*-mode: Shell-script; coding: utf-8;-*-
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+_commitid=2234386bcee7297dfa1b6d8a5d20f95ea4ed9bb0
+pkgname=bloaty
+# Fake a version until upstream does one
+# format is simply 0.0.0.0.${commit_count}
+pkgver=0.0.0.0.79
```
jirutka: Sigh, yet another Google project that doesn’t bother to do releases…

For such cases it’s IMO better to use version number in format `0.0_gitYYYYMMDD` (i.e. `0.0_git20161228`) or if you prefer commits count, then `0.0_gitX` (i.e. `0.0_git79`). It makes it more clear that it’s a snapshot from git.

testing/bloaty/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+# Fake a version until upstream does one
+# format is simply 0.0.0.0.${commit_count}
+pkgver=0.0.0.0.79 
+pkgrel=0
+pkgdesc="A size profiler for binaries"
+url="https://github.com/google/bloaty"
+arch="all"
+license="apache"
+depends=""
+makedepends="$depends gcc git"
```
jirutka: Don’t add `$depends` into `makedepends`, this is implicit.

testing/bloaty/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+	byteswap.patch
+	"
+builddir="$srcdir/${pkgname}-${_commitid}"
+
+build() {
+	rmdir third_party/re2 third_party/googletest || return 1	
+	ln -s ../../re2-2017-01-01 third_party/re2 || return 1
+	ln -s ../../googletest-release-1.8.0 third_party/googletest || return 1
+	git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer third_party/libFuzzer || return 1
+	(cd third_party/libFuzzer && git checkout e9387f0b6d411e18ca75f08aecfec1238b8e113f) || return 1
```
jirutka: Please move this SHA into a top-level variable, so one does not simply overlook that there’s some hard-coded version.

testing/bloaty/APKBUILD
mitchty: For the date should it be the last commit or just say today? Or is it more whatever works in this case, I'm impartial really, my goal with that version was to make a silly version that was almost guaranteed to be lower than whatever was chosen by the upstream.

testing/bloaty/APKBUILD
mitchty: Will do, also would the git checkout be better done in a prepare function?

testing/bloaty/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+	https://github.com/google/googletest/archive/release-1.8.0.tar.gz
+	byteswap.patch
+	"
+builddir="$srcdir/${pkgname}-${_commitid}"
+
+build() {
+	rmdir third_party/re2 third_party/googletest || return 1	
+	ln -s ../../re2-2017-01-01 third_party/re2 || return 1
+	ln -s ../../googletest-release-1.8.0 third_party/googletest || return 1
+	git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer third_party/libFuzzer || return 1
```
kaniini: i don't like this, it should be possible to build a package without requiring a functioning internet connection.

if we have to git clone during the build process, then it makes the build process much more complex

testing/bloaty/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+	https://github.com/google/re2/archive/2017-01-01.tar.gz
+	https://github.com/google/googletest/archive/release-1.8.0.tar.gz
+	byteswap.patch
+	"
+builddir="$srcdir/${pkgname}-${_commitid}"
+
+build() {
+	rmdir third_party/re2 third_party/googletest || return 1	
+	ln -s ../../re2-2017-01-01 third_party/re2 || return 1
+	ln -s ../../googletest-release-1.8.0 third_party/googletest || return 1
```
kaniini: there is googletest here which indicates a testsuite is present; but no `check()` function.

mitchty: It looks like that causes this:

```
git submodule init && git submodule update
make: *** No rule to make target 'third_party/googletest/googlemock/gtest/libgtest.a', needed by 'tests/range_map_test'.  Stop.
make: *** Waiting for unfinished jobs....                                                                                                                        touch .gitmodules -r third_party/googletest/CMakeLists.txt
>>> ERROR: bloaty: all failed
>>> bloaty: Uninstalling dependencies...
```

I'll try a prepare and just tar.gz a full git checkout, at this point the things this depends on are fractal in missing sub dependencies.

mitchty:
Ok I gave up and created a snapshot function that just tar.xz's the entire git checkout with submodules and threw it up on my vps. I'm not sure thats the best way to do this or if it would better be on alpinelinux.org or not. Otherwise I'll just keep building this thing as a static binary and just leave this be.

Note I committed this via git commit --squash HEAD and didn't rebase --autosquash so the differences can be seen.

mitchty:
Looks like the make test needs cmake, i'll fix it.

mitchty:
ugh, I give up, I'll just keep building the binary statically.

PR #914, opened 2017-02-20
testing/ghc: multiple cleanups/improvements

mitchty:
Multiple commits/changes, consisting of:
- Add a post-install script to recache the ghc pkg db on install so we don't have a warning when running ghc-pkg list
- Split ghc into two disparate packages like in debian, ghc and ghc-prof, the latter being the profiled debugging libraries. Which means if one wants to debug ghc programs they'll need to install ghc-prof.
- Run strip --strip-unneeded on all shared libraries and archives to save some disk space, ~98MiB. There is likely more low hanging fruit here.
- Add provides for the ghc pkgs that ghc itself provides similar to how the arch linux build works.
- Remove a few redundant or unnecessary lines.

I bumped the release version as well due to the changes.

### DIFF
--- a/testing/ghc/APKBUILD
+++ b/testing/ghc/APKBUILD
@@ -32,7 +32,32 @@ license="custom:bsd3"
 # that version as it greatly simplifies the apkbuild process. The
 # apks built on 3.5 will not work on any prior version of alpine linux.
 depends="gmp-dev perl gcc>=6.2.1 llvm3.7"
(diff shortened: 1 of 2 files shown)


mitchty:
Question on the sub package name, would ghc-dev be more appropriate and in line with existing packages?

mitchty:
I just switched it to ghc-dev, when I get hackage package building apkbuild files I'll ensure profiled versions also get built for the packages as well in their own -dev packages. No need to mimic debian entirely with the name.

PR #915, opened 2017-02-21
testing/cabal: new aport

mitchty:
Add an initial version of cabal-install, aka the cabal package manager
for ghc.

Note this version uses bootstrap.sh, this is a temporary measure until
I can get autogenerated APKBUILD files via updates to cblrepo.

Note, this relies on #914 with its ghc/ghc-dev separation.

### DIFF
--- a/testing/cabal/APKBUILD
+++ b/testing/cabal/APKBUILD
@@ -1,0 +1,208 @@
+#-*-mode: Shell-script; coding: utf-8;-*-
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+pkgname=cabal
+pkgver=1.24.0.2
+pkgrel=0
+pkgdesc="The Haskell Cabal"
+url="http://haskell.org"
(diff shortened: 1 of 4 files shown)


fabled:
Ok. I'm also thinking bootstrapping new architectures. How will the bootstrap build done be in future? Or can/should cabal be also cross-built from working architecture to the new one?

mitchty:
Essentially we won't be using the bootstrap.sh stuff in the future. All of the .tar.gz/.cabal combination sources in the list will become apkbuilds in their own right.

Essentially it would make what you're proposing possible. We would technically be able to build things for other environments via a ghc cross compiler. For now though this is good enough to get a version of cabal that people can use to install things on their own. With the caveat that all this stuff needs is a working ghc on the architecture its building on.

mitchty:
Ping on this PR, ghc the compiler isn't of much use without the package manager. Similar to rust and cargo, kinda need both to be able to reasonably do things.

PR #916, opened 2017-02-21
testing/libwrap && testing/tcp-wrappers: new aport

itoffshore:
* patched ipv4 + ipv6 version used by [Red Hat](http://pkgs.fedoraproject.org/cgit/rpms/tcp_wrappers.git/commit/)

* built without linking to `glibc` libs that do not exist in `musl`

* hostname lookups are therefore disabled


ncopa:
do we really want libwrap and tcpwrappers support? What should we use it for?

I would prefer not add those unless absolutely necessary.

itoffshore:
originally this was for adding tcp wrapper support to `stunnel` - but in the end I'm not using `stunnel` - this is ok to close.

PR #920, opened 2017-02-22
main/openssh: fix patches, split to subpackages, remade init-script

vakartel:
- remove unapplied patches
- rename *.diff to *.patch because *.diff are ignored by
  default_prepare
- split to subpackages with auto pkgdesc creation and
  auto dependency tracking
- remove gentoo-specific stuff from init, simplify it.
- cleanups in APKBUILD


ncopa:
this looks scary. Have you tested that you can restart sshd remotely without losing the current login? I don't want that anyone gets locked out of their remote system on upgrade.

ncopa:
After reading it a bit closer. We cannot merge this.

If you make an error in `sshd_config` and restart sshd you will end up stopping sshd. Earlier it would not stop the service unless the config is valid.

This is not an improvement.

vakartel:
@ncopa 
Look closer to stop_post(). It terminates current sessions only if runlevel is "shutdown". It needs to to terminate ssh sessions while interfaces are still up.
And yes, I tested it. I make a typo in sshd_config and tries to reload|restart:

\# echo Portzz 22 >> /etc/ssh/sshd_config

\# sshd -t
/etc/ssh/sshd_config: line 120: Bad configuration option: Portzz
/etc/ssh/sshd_config: terminating, 1 bad configuration options

\# /etc/init.d/sshd reload
 * Reloading sshd ...
/etc/ssh/sshd_config: line 13: Bad configuration option: Portzz
/etc/ssh/sshd_config: terminating, 1 bad configuration options                                       [ !! ]

\# /etc/init.d/sshd restart
 * Stopping sshd ...                                                                                                          [ ok ]
 * ERROR: sshd failed to stop

For now I'm still logged in ...

vakartel:
I remade init.d script to to dealing with foreground sshd process and and suppress process-generated pidfile. But it closed ((

ncopa:
@vakartel `post_stop` checks if it should kill client connections on shutdown. sshd will not kill current sessions if you stop it which is why you still are connected even if you stop sshd. `post_stop` will execute *after* the process is stopped. Earlier the check was done *before* the sshd process stopped.

Also, you removed the `/etc/init.d/sshd checkconfig`. How do you know that nobody is using that?

You also change the configure options. I have not taken the time to investigate the consequences of those changes and the reason why you do it is unclear.

In general, the risk is high, the cost may be high on breakage and the benefit is not that big. I also don't feel very confident that you actually know what you are doing.

vakartel:
@ncopa 
init.d config options are the same as in original.
I just assign it to standard openrc variables.
So 'start', 'stop' 'restart' definitions are openrc-default, not redefined.
I can return checkconfig as an alias of start_pre.

But I think you don't want to change it at all. So let's forget about init.d script. 

What about *.diff patches?  it's completely ignored by default_prepare() ...
I found it when I saw old closed bug in sftp interactive mode.

vakartel:
I made compatibility fixes and rechecked it twice to be full-compatible with existing one.

If you think it's a kind idiotic to split it to (config|script|binary) submodules, I answer: there is a problem in existing one that some binaries use certain config-files and some do not. So I just find a way to split in correctly.

main/openssh/APKBUILD
```diff
@@ -39,32 +39,22 @@ source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_myver.tar
...
-
 build() {
 	cd "$builddir"
-	export LD="$CC"
 	./configure \
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
 		--sysconfdir=/etc/ssh \
-		--datadir=/usr/share/openssh \
```
ncopa: why remove `--datadir`? I mean, why rely on the default datadir?

main/openssh/APKBUILD
```diff
@@ -39,32 +39,22 @@ source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_myver.tar
...
 		--libexecdir=/usr/lib/ssh \
 		--mandir=/usr/share/man \
 		--with-pid-dir=/run \
 		--with-mantype=man \
 		--with-ldflags="${LDFLAGS}" \
-		--disable-strip \
 		--disable-lastlog \
+		--disable-strip \
 		--disable-wtmp \
-		--with-privsep-path=/var/empty \
```
ncopa: why rely on default privsep path? You later use a hardcoded path in `start_pre`. For that reason I think it might make sense to explicitly set it hard here too.

main/openssh/APKBUILD
```diff
@@ -39,32 +39,22 @@ source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_myver.tar
...
+		--with-xauth=/usr/bin/xauth \
```
ncopa: Why was this added? (might be good to include the reason why in commit message since its not evident)

main/openssh/sshd.initd
```diff
@@ -1,100 +1,45 @@
...
-: ${SSHD_PIDFILE:=/run/${SVCNAME}.pid}
-: ${SSHD_BINARY:=/usr/sbin/sshd}
+cfgfile=${SSHD_CONFIG:-${SSHD_CONFDIR:-/etc/ssh}/sshd_config}
+pidfile=${SSHD_PIDFILE:-/run/$RC_SVCNAME.pid}
+command=${SSHD_BINARY:-/usr/sbin/sshd}
+command_args="-D $SSHD_OPTS"
+command_background=yes
+extra_commands=checkconfig
+extra_started_commands=reload
+required_files=$cfgfile
```
ncopa: this change is good

main/openssh/sshd.initd
```diff
@@ -1,100 +1,45 @@
...
-	fi
-
-	[ "${SSHD_PIDFILE}" != "/run/sshd.pid" ] \
-		&& SSHD_OPTS="${SSHD_OPTS} -o PidFile=${SSHD_PIDFILE}"
-	[ "${SSHD_CONFIG}" != "/etc/ssh/sshd_config" ] \
-		&& SSHD_OPTS="${SSHD_OPTS} -f ${SSHD_CONFIG}"
-
-	"${SSHD_BINARY}" -t ${SSHD_OPTS} || return 1
+	[ "$cfgfile" = "/etc/ssh/sshd_config" ] || command_args="$command_args -f $cfgfile"
+	$command -t $command_args
```
ncopa: I would like to get rid of the the `$RC_CMD` conditional.

main/openssh/sshd.initd
```diff
@@ -1,100 +1,45 @@
...
-
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --exec "${SSHD_BINARY}" \
-	    --pidfile "${SSHD_PIDFILE}" \
-	    -- ${SSHD_OPTS}
-	eend $?
+stop_post() {
+	local pids
+	[ "$RC_RUNLEVEL" = "shutdown" ] && pids=$(pgrep ${command##*/})
+	[ -z "$pids" ] || kill -TERM $pids >/dev/null 2>&1
```
ncopa: This hunk is ok.

main/openssh/sshd.initd
```diff
@@ -1,100 +1,45 @@
...
-	if [ "${RC_CMD}" = "restart" ] ; then
-		checkconfig || return 1
-	fi
-
-	ebegin "Stopping ${SVCNAME}"
-	start-stop-daemon --stop --exec "${SSHD_BINARY}" \
-	    --pidfile "${SSHD_PIDFILE}" --quiet
+reload() {
+	ebegin "Reloading $RC_SVCNAME"
+	start_pre && start-stop-daemon --signal HUP --pidfile $pidfile
```
ncopa: I don't like this.We can make it clearer with:

````shell
    checkconfig && start-stop-daemon --signal HUP --pidfile $pidfile
````

File: main/openssh/sshd.initd [hunk 8]
```diff
@@ -1,100 +1,45 @@
...
-	eend $?
+stop_post() {
+	local pids
+	[ "$RC_RUNLEVEL" = "shutdown" ] && pids=$(pgrep ${command##*/})
+	[ -z "$pids" ] || kill -TERM $pids >/dev/null 2>&1
 }
 
-stop() {
-	if [ "${RC_CMD}" = "restart" ] ; then
-		checkconfig || return 1
```
ncopa: I'd like to keep the `checkconfig` on `restart` before stopping the service. Maybe we can do:

````sh
stop_pre() {
        if [ "${RC_CMD}" = "restart" ] ; then
            checkconfig || return 1
        fi
}
````

or:
````sh
stop_pre() {
        [ "${RC_CMD}" != "restart" ] && return 0
        checkconfig
}
````

Former is more readable.

File: main/openssh/sshd.initd [hunk 9]
```diff
@@ -1,100 +1,45 @@
...
-reload() {
-	checkconfig || return 1
-	ebegin "Reloading ${SVCNAME}"
-	start-stop-daemon --signal HUP \
-	    --exec "${SSHD_BINARY}" --pidfile "${SSHD_PIDFILE}"
-	eend $?
+checkconfig() {
+	ebegin "Checking $RC_SVCNAME config"
+	start_pre
+	eend $1
```
ncopa: I want make the opposite:
````sh
checkconfig() {
        $command -t -f "${cfgfile:-/etc/ssh/sshd_config}"
}

start_pre() {
        ...
        checkconfig || return 1
}
````

That way its more clear whats going on.

**Comment by ncopa** [comment]
> @vakartel But I think you don't want to change it at all. So let's forget about init.d script.

I don't like your changes to the init.d script because:

1) I still believe it changes the current behavior of restart. Currently it will check config *before* stopping sshd and if config is not ok it will not stop sshd at all. New behavior will check the config *after* stopping sshd but before starting it again. If config is not ok it will leave the `sshd` no longer running. (current connection still runs though but it will not accept new connections)

2) I dont like it because i think the new code is unreadable. I believe it could be simpler and clearer.

> @vakartel What about *.diff patches? it's completely ignored by default_prepare() ... I found it when I saw old closed bug in sftp interactive mode.

I think that might have been unintentional.

@fabled was the HPN patches intentionally disabled? (commit 756f181a56f8c7d4445ba4aacec5354e3bf9e051) I suspect they were not since the commit message does not say anything about why they were disabled.

> @vakartel If you think it's a kind idiotic to split it to (config|script|binary) submodules, I answer: there is a problem in existing one that some binaries use certain config-files and some do not. So I just find a way to split in correctly.

I don't think its idiotic. I think it is a bit risky since it may break things for people who does apk upgrade. So I don't want there to be any doubts on why this is an improvement that is worth the risk.

It would be good if the commit message explained what binaries needs what config files.

Thank you for working on this and thank you for your patience with me being pedantic.

File: main/openssh/APKBUILD (hunk 1 above)
Comment (vakartel) [inline main/openssh/APKBUILD]: I remove it because there is no datadir in this package. I think it appears as a copy-paste result from another APKBUILD.

File: main/openssh/APKBUILD (hunk 2 above)
Comment (vakartel) [inline main/openssh/APKBUILD]: I remove it because it's a configure-default path. But you are right. I move it back.

File: main/openssh/sshd.initd (hunk 8 above)
Comment (vakartel) [inline main/openssh/sshd.initd]: Agree. ssh often the only way to get access to remote host. So it must have.

Comment (vakartel) [inline main/openssh/sshd.initd]: Agree. ssh often the only way to get access to remote host. So it must have.

A I think we need to allow sshd with broken config to stop if the RC_RUNLEVEL=shutdown only.

Comment (ncopa) [inline main/openssh/sshd.initd]: i don't think we bother care about shutdown. If you have broken config and do shutdown, you will get an error message on screen "failed to stop sshd" and you will have a hint that something is wrong with ssh. shutdown will continue and the sshd will eventually get killed.

But you might get an idea of why you cannot reconnect with ssh after its powered on again.

If there was a clean sshd shutdown you would get disconnected and have no idea why you can not log in after its powered on again.

So i'd say we don't bother check for RC_RUNLEVEL=shutdown and let it just shutdown with error message on screen.

File: main/openssh/APKBUILD (hunk 1 above)
Comment (ncopa) [inline main/openssh/APKBUILD]: add a comment in commit message like "remove unused --datadir configure option"

File: main/openssh/APKBUILD (hunk 3 above)
Comment (ncopa) [inline main/openssh/APKBUILD]: could be mentioned in commit message: "set xauth with configure option instead of with sed"

**Comment by vakartel** [comment]
in original pack there was 5 subpackages:
* _keygen_ - keygen app
* _client_ - all client apps and utils (/usr/bin/*) + configs (ssh_config, moduli) + /var/empty dir needed for server only
* _keysign_ - helper app for host-based auth
* _sftp-server_ - helper app for sftp subsystem
* _server_ - ssh server + init scripts depends on keygen.

I split it to:
* _moduli_ - DH config. used by: keyscan, ssh, server
* subpackages for clients: ssh (depend on moduli), sftp (depend on ssh), scp (depended on ssh)
* subpackages for utils: add, agent, keygen, keyscan (depend on moduli), copy-id (depend on ssh)
* subpackages for helpers: keysign (depend on ssh), sftp-server, pkcs11-helper (depend on agent)
* _clients_ pack: ssh, sftp, scp
* _utils_ pack: add, agent, keygen, keyscan, copy-id
* _client_ compatibility pack: clients + utils
* _server_ sshd server (depends on moduli) + init scripts (depends on keygen) + /var/empty dir.

_openssh_ main pack depends on all of it: clients + utils + server + keysign + sftp-server + pkcs11-helper

Remade init script: check config before restart to not allow to stop ssh server with broken config.

File: main/openssh/APKBUILD [hunk 10]
```diff
@@ -29,16 +28,6 @@ source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_pkgver.ta
...
-# secfixes:
-#   7.4_p1:
-#     - CVE-2016-10009
-#     - CVE-2016-10010
-#     - CVE-2016-10011
-#     - CVE-2016-10012
-#       https://bugs.alpinelinux.org/issues/6583
-
-# HPN patches are from: http://www.psc.edu/index.php/hpn-ssh
-
```
ncopa: We need to keep this comment. It is used to create a whitelist for a CVE scanner.

vakartel: ok, move it back

vakartel: but ... this url is invalid: http://www.psc.edu/index.php/hpn-ssh

ncopa:
I'm looking at this again and trying to merge/group the patches and revert "meaningless" changes (like renaming `_myver` to `_pkgver`) so it becomes more visible what is actually changing.

While doing this i noticed that you removed `makedepend_build` and `makedepends_host` which was introduced with 756f181a56f8c7d4445ba4aacec5354e3bf9e051. Can you confirm that you have verified that you are not breaking cross compile?

ncopa:
you also remove `export LD="$CC"` which also was introduced with commit 756f181a56f8c7d4445ba4aacec5354e3bf9e051. Have you verified that this is no longer needed for cross compile?

ncopa:
I would like to apply your work but I want rebase it into 3 commits:

1) fix patches
2) rearrange package split
3) rewrite of init.d

I think that will make it easier to follow the changes and make it more visible that we haven't introduced any unintentional bugs. Do you think this make sense?

ncopa:
I rebased the patches: https://github.com/ncopa/aports/commits/openssh and made minor modifications:

- keep the `_myver` instead of replacing it with `_pkgrel` to make the new rebased pach more readable.
- increase the `pkgrel` a couple of more times.
- keep the `makedepends_build` and `makedepends_host` to avoid breaking crosscompile.
- keep the order of the patches to make rebased patch more readable.
- fix the URL comment for HPN patches (thanks for spotting that)
- keep the `export LD="$CC"`

If you look at the 3 new patches you should see that it is much easier to follow the changes. We can also more easily backport the HPN patch fix to stable branch if we need that, and its easy to revert either the package split or init-script refactor without affecting the other if things breaks badly.

@vakartel I need your :+1: for this since I made you the author to the commits. I also wonder if I should close this pull request and create a new with the 3 rebased patches so @fabled and @jirutka also can review your work. What do you think?

The diff between the rebased patches and your current pull request:
````diff
diff --git a/main/openssh/APKBUILD b/main/openssh/APKBUILD
index aa6278b419..6527e32002 100644
--- a/main/openssh/APKBUILD
+++ b/main/openssh/APKBUILD
@@ -2,8 +2,8 @@
 pkgname=openssh
 pkgver=7.4_p1
-pkgrel=2
-_pkgver=${pkgver/_/}
+_myver=${pkgver%_*}${pkgver#*_}
+pkgrel=4
 pkgdesc="Port of OpenBSD's free SSH release"
 url="http://www.openssh.org/portable.html"
 arch="all"
@@ -12,24 +12,24 @@ options="suid"
 depends="$pkgname-clients $pkgname-utils $pkgname-server
        $pkgname-keysign $pkgname-pkcs11-helper $pkgname-sftp-server
        "
-makedepends="libressl-dev linux-headers zlib-dev"
+makedepends_build=""
+makedepends_host="libressl-dev zlib-dev linux-headers"
+makedepends="$makedepends_build $makedepends_host"
 subpackages="$pkgname-doc $pkgname-add:_sub $pkgname-agent:_sub $pkgname-ssh
        $pkgname-scp:_sub $pkgname-sftp:_sub $pkgname-keygen:_sub $pkgname-keyscan:_sub
        $pkgname-keysign $pkgname-sftp-server:_sub $pkgname-pkcs11-helper:_pkcs
        $pkgname-copy-id:_copy:noarch $pkgname-moduli:_sub:noarch $pkgname-server
        $pkgname-clients::noarch $pkgname-utils::noarch $pkgname-client::noarch
        "
-source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_pkgver.tar.gz
+source="http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/$pkgname-$_myver.tar.gz
+       openssh7.4-peaktput.patch
+       openssh7.4-dynwindows.patch
+       fix-utmp.patch
+       bsd-compatible-realpath.patch
        sshd.initd
        sshd.confd
-       fix-utmp.patch
        sftp-interactive.patch
-       bsd-compatible-realpath.patch
-       openssh7.4-peaktput.patch
-       openssh7.4-dynwindows.patch
        "
-builddir="$srcdir/$pkgname-$_pkgver"
-
 # secfixes:
 #   7.4_p1:
 #     - CVE-2016-10009
@@ -38,10 +38,13 @@ builddir="$srcdir/$pkgname-$_pkgver"
 #     - CVE-2016-10012
 #       https://bugs.alpinelinux.org/issues/6583

-# HPN patches are from: http://www.psc.edu/index.php/hpn-ssh
+# HPN patches are from: http://hpnssh.sourceforge.net/
+
+builddir="$srcdir"/$pkgname-$_myver

 build() {
        cd "$builddir"
+       export LD="$CC"
        ./configure \
                --build=$CBUILD \
                --host=$CHOST \
@@ -141,10 +144,10 @@ server() {
 }

 sha512sums="4f3256f461f01366c5d5e0e45285eec65016e2643b3284b407f48f53d81087bf2c1caf7d5f7530d307a15c91c64de91446e1cba948e8fc68f82098290fe3b292  openssh-7.4p1.tar.gz
-60fc8bcd675bec9a80dadfd9717a04e5e19870b32b16fbe2935480245ca7c982964506b3ed8db00c66c2a108421b8724bfdab1147679565ddcfb80323c7e8588  sshd.initd
-67c03760ff3f483eb929e9e34919d4a0a02acfbef541bad4b66cb6dfa25c7adf25cfd9ff30a3d15f0aff9e7c336e588835518713bb49f5c5cace8fddc65df97c  sshd.confd
+398096a89aa104abeff31aa043ac406a6348e0fdd4d313b7888ee0b931d38fd71fc21bceee46145e88f03bc27e00890e068442faee2d33f86cfbc04d58ffa4b6  openssh7.4-peaktput.patch
+b9d736eae9b43de91fa3eb277ba8abc6290a8436b0fb00ae3b0f1b2eabba9983e4d2a1e3c68f5514247d0a3f120037f0795fd88fbf302aabd2d1b54a325a04ee  openssh7.4-dynwindows.patch
 f35fffcd26635249ce5d820e7b3e406e586f2d2d7f6a045f221e2f9fb53aebc1ab1dd1e603b3389462296ed77921a1d08456e7aaa3825cbed08f405b381a58e1  fix-utmp.patch
-c1d09c65dbc347f0904edc30f91aa9a24b0baee50309536182455b544f1e3f85a8cecfa959e32be8b101d8282ef06dde3febbbc3f315489339dcf04155c859a9  sftp-interactive.patch
 f2b8daa537ea3f32754a4485492cc6eb3f40133ed46c0a5a29a89e4bcf8583d82d891d94bf2e5eb1c916fa68ec094abf4e6cd641e9737a6c05053808012b3a73  bsd-compatible-realpath.patch
-398096a89aa104abeff31aa043ac406a6348e0fdd4d313b7888ee0b931d38fd71fc21bceee46145e88f03bc27e00890e068442faee2d33f86cfbc04d58ffa4b6  openssh7.4-peaktput.patch
-b9d736eae9b43de91fa3eb277ba8abc6290a8436b0fb00ae3b0f1b2eabba9983e4d2a1e3c68f5514247d0a3f120037f0795fd88fbf302aabd2d1b54a325a04ee  openssh7.4-dynwindows.patch"
+60fc8bcd675bec9a80dadfd9717a04e5e19870b32b16fbe2935480245ca7c982964506b3ed8db00c66c2a108421b8724bfdab1147679565ddcfb80323c7e8588  sshd.initd
+67c03760ff3f483eb929e9e34919d4a0a02acfbef541bad4b66cb6dfa25c7adf25cfd9ff30a3d15f0aff9e7c336e588835518713bb49f5c5cace8fddc65df97c  sshd.confd
+c1d09c65dbc347f0904edc30f91aa9a24b0baee50309536182455b544f1e3f85a8cecfa959e32be8b101d8282ef06dde3febbbc3f315489339dcf04155c859a9  sftp-interactive.patch"
````

jirutka:
@ncopa Do you know that you can push into the contributor’s branch in his “fork” once he opened a pull request from it? So you can just push your changes into it instead of adding patches into comment.

ncopa:
@jirutka I didn't know that. But the purpose was to rearrange the work @vakartel did in commits that makes it much more visible what was actually changed. I can probably push force to his pull request but I dont want to purge @vakartel's work without he give me permission to do so.

The diff i pasted here was to show that there were no significant changes in the rebase.

It would be nice if the rebased patches could be reviewed (once @vakartel gives me ok that we take it from there). I think we want those changes, but they are pretty intrusive so I'd like a review from someone more than me. The rebase I did was to make that review much easier.

vakartel:
It's no problem if you push it.
one note: As I viewed in abuild sources, you don't need to define makedepends="$makedepends_build $makedepends_host", because abuild made it automatically if makedepends not defined or empty.

vakartel:
@ncopa 
There is openssh7.5 is out. What to do with this PR?

andypost:
http://pkgs.alpinelinux.org/package/edge/main/x86_64/openssh already **7.5_p1-r4**

So basically it needs rebase & fix merge conflict

vakartel:
@andypost 
I think it's easier to create new PR from scratch rather then resolve conflicts.

andypost:
I mean you can push to that branch new code to keep context & already done reviews. This changes looks great but needs polishing

vakartel:
finally I did it

andypost:
Another upgrade will require rebase https://github.com/alpinelinux/aports/pull/2908

andypost:
Sadly it is stolen again...

clandmeter:
looks like the initd script has been updated by @jirutka in 11f6721ffc08a0cf067bc71aec635a9c1544ea3d

I wonder if its better to close this and if needed create a new PR(s) with smaller changes which are easier to review.

PR #921, opened 2017-02-22
testing/hamlib: new aport

bradfordboyle:
Library to control radio transceivers and receivers

http://www.hamlib.org/

### DIFF
--- a/testing/hamlib/APKBUILD
+++ b/testing/hamlib/APKBUILD
@@ -1,0 +1,63 @@
+# Contributor: Bradford D. Boyle <bradford.d.boyle@gmail.com>
+# Maintainer: Bradford D. Boyle <bradford.d.boyle@gmail.com>
+pkgname=hamlib
+pkgver=3.0
+pkgrel=0
+pkgdesc="Library to control radio transceivers and receivers"
+url="http://www.hamlib.org/"
+arch="all"
+license="LGPLv2+"
+depends=""
+depends_dev=""
+makedepends="$depends_dev linux-headers perl-dev python-dev swig tcl-dev"
+install=""
+subpackages="$pkgname-dev $pkgname-doc $pkgname-perl $pkgname-python $pkgname-tcl"
+source="http://downloads.sourceforge.net/project/hamlib/hamlib/3.0/hamlib-3.0.tar.gz"
+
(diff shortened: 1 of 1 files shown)


testing/hamlib/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+        # fix perl module location
+        cd "$pkgdir"/usr/lib/perl5/site_perl/
+        mkdir current
+        mv auto current
+        mv Hamlib.pm current
+        rm perltest.pl
+
+        /usr/bin/find "$pkgdir" -name '.packlist' -delete
+        /usr/bin/find "$pkgdir" -name '*.pod' -delete
```
kaniini: Why delete documentation?

testing/hamlib/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgname=hamlib
+pkgver=3.0
+pkgrel=0
+pkgdesc="Library to control radio transceivers and receivers"
+url="http://www.hamlib.org/"
+arch="all"
+license="LGPLv2+"
+depends="perl python"
+depends_dev=""
+makedepends="$depends_dev linux-headers perl-dev python-dev swig tcl-dev"
```
kaniini: I see there are perl, python and tcl bindings.  These should be separated into subpackages so that installing hamlib does not pull in perl, python and TCL from the start.

testing/hamlib/APKBUILD
bradfordboyle: I was copying the [PKGBUILD from Arch](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hamlib), but now that I've split the package into subpackages, I've left them in with the perl subpackage.

testing/hamlib/APKBUILD
bradfordboyle: Good suggestion. I think I've gotten it split into subpackages now.

kaniini:
While building this, I noted that the Perl binding had a NULL rpath entry, which is not allowed.  I added a chrpath(1) call to remove the erroneous rpath entry.

I also converted the spaces in the APKBUILD to tabs to match established style guidelines.

PR #925, opened 2017-02-23
testing/grub: fixup grub-mkconfig

HRio:
Minor fixup of grub-mkconfig so the generated config works with alpine.


HRio:
Message from Travis: 
```

The job has been terminated
```

I have no problem building the package on other hosts, but the log gets 6.3M large. So it seems this package is not build-able by Travis with its current configuration.

jirutka:
That's not a problem, we use Travis just to easily check pull requests. Packages are built on our infrastructure that doesn't have such limitations. 

So this only means that some from us should try to build this pull request manually before merging.

kaniini:
I am working on this.

PR #943, opened 2017-03-01
community/entr: update to 3.7

itoffshore:
I've contacted the author about the one failing test

### DIFF
--- a/community/entr/APKBUILD
+++ b/community/entr/APKBUILD
@@ -1,16 +1,26 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=entr
-pkgver=3.6
-_hgrev=c15b0be493fc
+pkgver=3.7
 pkgrel=0
 pkgdesc="Event Notify Test Runner: Run arbitrary commands when files change"
 url="http://entrproject.org"
 arch="all"
 license="ISC BSD"
 subpackages="$pkgname-doc"
-source="http://entrproject.org/code/$pkgname-$pkgver.tar.gz"
-builddir="$srcdir/eradman-$pkgname-$_hgrev"
+makedepends="vim mercurial file util-linux"
+source="http://entrproject.org/code/$pkgname-$pkgver.tar.gz
+	fix-tests.patch
+	"
+builddir="$srcdir/eradman-$pkgname"
+
+prepare() {
+	# Top-level directory inside the tar archive contains SHA of a commit...
+	cd "$srcdir"
+	mv eradman-$pkgname-* "$builddir"
+
+	default_prepare
+}
 
 build() {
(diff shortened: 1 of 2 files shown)


eradman:
The problem with the failing test appears to be the use of a variable without brackets

    $ ash -c 'echo ${0}; exit 2' xyz
    xyz

Does this change fix it?

    diff -r 330e3104784e system_test.sh
    --- a/system_test.sh    Mon Feb 27 04:35:29 2017 -0500
    +++ b/system_test.sh    Wed Mar 01 15:37:56 2017 -0500
    @@ -298,7 +298,7 @@
    
     try "exec a command using shell option"
            setup
    -       ls $tmp/file* | ./entr -ps 'file $0; exit 2' >$tmp/exec.out 2>$tmp/exec.err &
    +       ls $tmp/file* | ./entr -ps 'file ${0}; exit 2' >$tmp/exec.out 2>$tmp/exec.err &
            bgpid=$! ; zz
            echo 456 >> $tmp/file2 ; zz
            kill -INT $bgpid

jirutka:
> The problem with the failing test appears to be the use of a variable without brackets
>
>    $ ash -c 'echo ${0}; exit 2' xyz
>    xyz

Brackets are definitely not required in this case.

eradman:
@jirutka, you are right. I ran a faulty test

community/entr/APKBUILD
```diff
@@ -1,15 +1,18 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=entr
-pkgver=3.6
-_hgrev=c15b0be493fc
+pkgver=3.7
+_hgrev=c5b62bde107d
```
jirutka: Please avoid this, it complicates bumping of the pkg and it’s actually unnecessary. You can workaround this problem with hash in the directory with something like:

```sh
builddir="$srcdir/eradman-$pkgname"

prepare() {
	# Top-level directory inside the tar archive contains SHA of a commit...
	cd "$srcdir"
	mv $pkgname-v$pkgver-* "$builddir"

	default_prepare
}
```

eradman:
I was able to reproduce this on Alpine and made a small change in the system test as a workaround

https://bitbucket.org/eradman/entr/commits/d3fcdd8b6d2edad3a4eece2eb5292a271beb3752

On Alpine two streams of STDOUT (one from `entr`, one from the shell) are merged (as I would expect); on other platforms the output from `entr` is not visible.

itoffshore:
* added `prepare()` to solve the hash issue
* amended the patch to reflect the [upstream commit](https://bitbucket.org/eradman/entr/commits/d3fcdd8b6d2edad3a4eece2eb5292a271beb3752)

kaniini:
I added bash to `checkdepends` as the testsuite requires bash.  Thanks.

PR #948, opened 2017-03-02
testing/wireguard-grsec / wireguard-vanilla / wireguard-tools

itoffshore:
WireGuard is a novel VPN that runs inside the Linux Kernel and utilizes
**state-of-the-art cryptography**.
    
It aims to be faster, simpler, leaner, and more useful than IPSec, while
 avoiding the massive headache. It intends to be considerably more performant
than OpenVPN.
    
WireGuard is designed as a general purpose VPN for running on embedded
interfaces and super computers alike, fit for many different circumstances.
It runs over UDP.
    
https://www.wireguard.io

### DIFF
--- a/testing/wireguard-grsec/APKBUILD
+++ b/testing/wireguard-grsec/APKBUILD
@@ -1,0 +1,67 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+
+_flavor=${FLAVOR:-grsec}
+_kpkg=linux-$_flavor
+_kver=4.9.13
+_kpkgrel=0
+
+# when changing _ver we *must* bump _mypkgrel
+_ver=0.0.20170223
+_mypkgrel=0
+_name=wireguard
+
+# verify the kernel version before entering chroot
+_kapkbuild=../../linux-${_flavor}/APKBUILD
+if [ -f $_kapkbuild ]; then
+	. $_kapkbuild
+	pkgname=$_name-$_flavor
+	[ "$_kver" != "$pkgver" ] && die "please update _kver to $pkgver"
+	[ "$_kpkgrel" != "$pkgrel" ] && die "please update _kpkgrel to $pkgrel"
+fi
+
+_kpkgver="$_kver-r$_kpkgrel"
+_abi_release=${_kver}-${_kpkgrel}-${_flavor}
+
+pkgname=${_name}-${_flavor}
+pkgver=$_kver
+pkgrel=$(($_kpkgrel + $_mypkgrel))
+pkgdesc="Next generation secure network tunnel: kernel modules for $_flavor"
+arch='all'
+url='https://www.wireguard.io'
+license="GPLv2"
+makedepends="linux-grsec-dev=$_kpkgver libmnl-dev sparse"
+source="https://git.zx2c4.com/WireGuard/snapshot/WireGuard-$_ver.tar.xz"
+builddir="$srcdir"/WireGuard-$_ver
+
+build() {
+	cd "$builddir"
+	# only building module: see wireguard-tools for userspace
+	make -C src/ \
+		KERNELDIR=/lib/modules/${_abi_release}/build \
+		module \
+		|| return 1
+}
+
+package() {
+	cd "$builddir/src"
+	install_if="linux-${_flavor}=${_kver} $pkgname"
+
+	local module=
+	for module in *.ko; do
+		install -v -D -m644 ${module} \
+			"$pkgdir/lib/modules/$_abi_release/extra/${module}"
+	done
+}
+
+check() {
+	return 0
+	# currently failing: attribute 'nocapture': unknown attribute
+	# not part of musl libc ?
+	make -C src/ \
+		KERNELDIR=/lib/modules/${_abi_release}/build \
+		check \
+		|| return 1
+}
+
+sha512sums="273ef6463d447cb04b608a0379cce5c0ed4065f988b3f449995593592b42f2fc269fc249a8e3c22c28bfa682430ee20b5b7a46a96803c9c67d1b6fed7b800455  WireGuard-0.0.20170223.tar.xz"
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -1,0 +1,65 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+
+pkgname=wireguard-tools
+pkgver=0.0.20170223
+pkgrel=0
+pkgdesc="Next generation secure network tunnel: userspace tools"
+arch='all'
+url='https://www.wireguard.io'
+license="GPLv2"
+makedepends="libmnl-dev sparse"
+subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch \
+		$pkgname-tests:tests:noarch"
+source="https://git.zx2c4.com/WireGuard/snapshot/WireGuard-$pkgver.tar.xz"
+builddir="$srcdir"/WireGuard-$pkgver
+
+build() {
+	cd "$builddir"
+	make -C src/tools || return 1
+}
+
+package() {
+	cd "$builddir"
+
+	install_if="linux-${_flavor}=${_kver} $pkgname"
+	mkdir -p "$pkgdir"/usr/share/doc/$_name
+
+	make -C src/tools \
+		DESTDIR="$pkgdir" \
+		WITH_BASHCOMPLETION=yes \
+		WITH_WGQUICK=yes \
+		WITH_SYSTEMDUNITS=no \
+		install || return 1
+
+	find "$builddir"/contrib/examples -name '.gitignore' -delete
+	cp -rf "$builddir"/contrib/examples "$pkgdir"/usr/share/doc/$_name/
+}
+
+bashcomp() {
+	depends="bash"
+	pkgdesc="Wireguard BASH completions"
+
+	mkdir -p "$subpkgdir"/usr
+	mv "$pkgdir"/usr/share "$subpkgdir"/usr
+}
+
+tests() {
+	pkgdesc="Wireguard test suites"
+
+	mkdir -p "$subpkgdir"/usr/share/wireguard
+	mv "$builddir"/src/tests "$subpkgdir"/usr/share/wireguard
+	cp -rf "$builddir"/contrib/external-tests "$subpkgdir"/usr/share/wireguard
+}
+
+check() {
+	return 0
+	# currently failing: attribute 'nocapture': unknown attribute
+	# not part of musl libc ?
+	make -C src/ \
+		KERNELDIR=/lib/modules/${_abi_release}/build \
+		check \
+		|| return 1
+}
+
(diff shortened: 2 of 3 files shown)


testing/wireguard-grsec/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+_abi_release=${_kver}-${_kpkgrel}-${_flavor}
+
+pkgname=${_name}-${_flavor}
+pkgver=$_kver
+pkgrel=$(($_kpkgrel + $_mypkgrel))
+pkgdesc="Next generation secure network tunnel: kernel modules for $_flavor"
+arch='all'
+url='https://www.wireguard.io'
+license="GPLv2"
+makedepends="linux-grsec-dev=$_kpkgver libmnl-dev sparse"
```
zx2c4: You probably do not need to depend on sparse, unless you're actually doing development. It's not the "test suite" that you think it is.

testing/wireguard-grsec/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+		|| return 1
+}
+
+package() {
+	cd "$builddir/src"
+	install_if="linux-${_flavor}=${_kver} $pkgname"
+
+	local module=
+	for module in *.ko; do
+		install -v -D -m644 ${module} \
```
zx2c4: Why not instead run `make -C src/ .... module-install`?

testing/wireguard-grsec/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+			"$pkgdir/lib/modules/$_abi_release/extra/${module}"
+	done
```
zx2c4: Do you need to run depmod after installing kernel modules, or does some higher level alpine layer already do that for you?

testing/wireguard-grsec/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+}
+
+check() {
+	return 0
+	# currently failing: attribute 'nocapture': unknown attribute
+	# not part of musl libc ?
+	make -C src/ \
```
zx2c4: Just remove this section. It doesn't do what you want in the first place. It's for development, not for usage by packagers. If this confusion becomes widespread, I'll just remove it from the package I distribute and keep the shortcut locally.

testing/wireguard-tools/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+
+pkgname=wireguard-tools
+pkgver=0.0.20170223
+pkgrel=0
+pkgdesc="Next generation secure network tunnel: userspace tools"
+arch='all'
+url='https://www.wireguard.io'
+license="GPLv2"
+makedepends="libmnl-dev sparse"
```
zx2c4: Sparse has nothing to do with the tools and is not needed. Remove this dependency.

testing/wireguard-tools/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+	find "$builddir"/contrib/examples -name '.gitignore' -delete
+	cp -rf "$builddir"/contrib/examples "$pkgdir"/usr/share/doc/$_name/
+}
+
+bashcomp() {
+	depends="bash"
+	pkgdesc="Wireguard BASH completions"
+
+	mkdir -p "$subpkgdir"/usr
+	mv "$pkgdir"/usr/share "$subpkgdir"/usr
```
zx2c4: Slightly confused about the purpose of this moving gymnastics, but I assume it's some clever alpine package management thing for making subpackages?

testing/wireguard-tools/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+}
+
+tests() {
+	pkgdesc="Wireguard test suites"
+
+	mkdir -p "$subpkgdir"/usr/share/wireguard
+	mv "$builddir"/src/tests "$subpkgdir"/usr/share/wireguard
+	cp -rf "$builddir"/contrib/external-tests "$subpkgdir"/usr/share/wireguard
```
zx2c4: These too are also not what you think they are. They're little tests to assist other developers with their implementations, but not something a user or a package manager should ever see. Again, maybe I should remove these from the repo. Please don't distribute them.

testing/wireguard-tools/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+}
+
+check() {
+	return 0
```
zx2c4: This has no business at all of being in the -tools package, and doesn't do what you'd hope it does anyway. Please remove this section.

testing/wireguard-vanilla/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+_abi_release=${_kver}
+
+pkgname=${_name}-${_flavor}
+pkgver=$_kver
+pkgrel=$(($_kpkgrel + $_mypkgrel))
+pkgdesc="Next generation secure network tunnel: kernel modules for $_flavor"
+arch='all'
+url='https://www.wireguard.io'
+license="GPLv2"
+makedepends="linux-vanilla-dev=$_kpkgver libmnl-dev sparse"
```
zx2c4: Sparse is not a requirement. Please remove it.

testing/wireguard-vanilla/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+		|| return 1
+}
+
+package() {
+	cd "$builddir/src"
+	install_if="linux-${_flavor}=${_kver} $pkgname"
+
+	local module=
+	for module in *.ko; do
+		install -v -D -m644 ${module} \
```
zx2c4: Why not instead run make -C src/ .... module-install?

testing/wireguard-vanilla/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+			"$pkgdir/lib/modules/$_abi_release/extra/${module}"
+	done
```
zx2c4: Do you need to run depmod after installing kernel modules, or does some higher level alpine layer already do that for you?

testing/wireguard-vanilla/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+}
+
+check() {
```
zx2c4: Just remove this section. It doesn't do what you want in the first place. It's for development, not for usage by packagers. If this confusion becomes widespread, I'll just remove it from the package I distribute and keep the shortcut locally.

testing/wireguard-grsec/APKBUILD
itoffshore: have removed `make check` & changed the `makedepends`

testing/wireguard-grsec/APKBUILD
itoffshore: `module-install` tries to install into `/lib` & does not obey `DESTDIR="$pkgdir"` passed to `make`

testing/wireguard-grsec/APKBUILD
itoffshore: the kernel modules should install themselves automatically when `wireguard-tools` is installed - fixing in https://github.com/alpinelinux/aports/pull/954

testing/wireguard-grsec/APKBUILD
itoffshore: removed

testing/wireguard-tools/APKBUILD
itoffshore: it was never a hard dependency to the userspace tools (only a build time depends)

removed in https://github.com/alpinelinux/aports/pull/954

testing/wireguard-tools/APKBUILD
itoffshore: it is necessary - you can [view the contents of the various wireguard packages here](https://pkgs.alpinelinux.org/packages?name=wireguard*&branch=&repo=&arch=&maintainer=)

testing/wireguard-tools/APKBUILD
itoffshore: the tests are not distributed with the userspace tools - I've renamed the tests subpkg to `wireguard-tests` & made the pkg description explicit that they are developer only tests in https://github.com/alpinelinux/aports/pull/954

testing/wireguard-tools/APKBUILD
itoffshore: Alpine Linux's `apkbuild` runs `check()` on every single package in the repo now. If available it should run the software's built in unit tests. 

Have removed `make check` as requested in https://github.com/alpinelinux/aports/pull/954

testing/wireguard-vanilla/APKBUILD
itoffshore: see https://github.com/alpinelinux/aports/pull/954

testing/wireguard-vanilla/APKBUILD
itoffshore: `module-install` tries to install into `/lib` & does not obey `DESTDIR="$pkgdir"` passed to `make`

testing/wireguard-vanilla/APKBUILD
itoffshore: the kernel modules should install themselves automatically when `wireguard-tools` is installed - fixing in https://github.com/alpinelinux/aports/pull/954

testing/wireguard-vanilla/APKBUILD
itoffshore: Alpine Linux's `apkbuild` runs `check()` on every single package in the repo now. If available it should run the software's built in unit tests. 

Have removed `make check` as requested in https://github.com/alpinelinux/aports/pull/954

itoffshore:
@zx2c4 - in summary:

* you can view the [various wireguard packages + their contents & dependencies here](https://pkgs.alpinelinux.org/packages?name=wireguard*&branch=&repo=&arch=&maintainer=).
* `sparse` was never a hard dependency (only build time) & is moved in https://github.com/alpinelinux/aports/pull/954
* developer tools are contained in `wireguards-tests` not `wireguard-tools`
* running `apk add wireguard-tools` will also automatically install the appropriate kernel modules:

zx2c4:
> developer tools are contained in wireguards-tests not wireguard-tools

That tests package is bogus and will never work. It should *not ever* be distributed as a package by distributions. As written elsewhere, if other people make this same confusion, then I'll just remove that code from the tree.

> running apk add wireguard-tools will also automatically install the appropriate kernel modules.

Awesome! That's super easy. I like the reverse dependency logic of apk; very nice.

testing/wireguard-tools/APKBUILD
zx2c4: Those tests do not belong in any distro package ever. They don't work the way you wish they do. Please remove this package all together.

testing/wireguard-tools/APKBUILD
zx2c4: I'm not sure how this would work for you, but the WireGuard module actually has built-in selftests when built in DEBUG mode. When enabled, either the module inserts and everything is fine, or it fails to insert, which implies the tests have failed (and you get something in dmesg). This probably is too difficult to do in a package though...

testing/wireguard-vanilla/APKBUILD
zx2c4: module-install will respond to `INSTALL_MOD_PATH` being set, actually. So this is the right way to do it. In the next WireGuard release, I'll investigate wiring DESTDIR up to `INSTALL_MOD_PATH`, so that these can be unified. Thanks for pointing out the shortcomings to me.

PR #950, opened 2017-03-02
testing/xombrero: A minimalists web browser, vi-like but with traditi…

dlintw:
testing/xombrero: A minimalists web browser, vi-like but with traditional web browser behavior.


stwa:
From the github project page: "xombrero has been retired and is no longer under development and supported."

dlintw:
@stwa, It runs well on Arch Linux, and it use minimal memory and works on most web sites.  So I think it is good alternative than big firefox/chromium.

dlintw:
faced segmentation fault at runtime

PR #952, opened 2017-03-02
testing/lxterminal-gtk: new package for support Chinese input by gtk3

dlintw:
To support gcin or other im modules.


dlintw:
Why vte-dev not automatically installed in CI?

pickfire:
@lintw If we `--enable-gtk3`, supposingly we need to add extra dependencies into it. Isn't it better for us to put that in `lxterminal-gtk3`?

dlintw:
@pickfire, I think it is OK, anyway, I require a terminal which support Chinese and work correctly. But, in this CI, I don't know why CI can not work. In my box, it works.

pickfire:
@dlintw Dependencies issue, it didn't even get to prepare the package.

dlintw:
@pickfire I've already add vte3 & vte3-dev on APKBUILD, I guess this is problem of CI builder.

jirutka:
No, it’s not CI problem. vte3 is in the community repository, but lxterminal is in the main repository. Packages in the main must not depend on packages in community or testing.

pickfire:
Nice, now this works.

dlintw:
@pickfire Do you know what's the next process to let it really put into testing repository?

pickfire:
Just move to testing directory?

dlintw:
I still can not found the pre-build binary in testing repository.  I guess there is some process should do.

testing/lxterminal-gtk3/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+_pkgname=lxterminal
+pkgver=0.3.0
+pkgrel=0
+pkgdesc="Desktop-independent VTE-based terminal emulator support gtk3 IME input"
+url="http://lxde.org/"
+arch="all"
+license="GPL2+"
+depends="vte3 gtk+3.0"
+makedepends="vte3-dev intltool gtk+3.0-dev"
+conflict="lxterminal"
```
jirutka: There’s no such variable in abuild. The correct approach is to add `!lxterminal` into `depends`.

testing/lxterminal-gtk3/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+replace="lxterminal"
```
jirutka: There should be `replaces`, not `replace`.

testing/lxterminal-gtk3/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+# Maintainer: Daniel YC Lin <dlin.tw@gmail.com>
+pkgname=lxterminal-gtk3
+_pkgname=lxterminal
+pkgver=0.3.0
+pkgrel=0
+pkgdesc="Desktop-independent VTE-based terminal emulator support gtk3 IME input"
+url="http://lxde.org/"
+arch="all"
+license="GPL2+"
+depends="vte3 gtk+3.0"
```
jirutka: These dependencies are automatically detected via dynamic links, so it’s not needed (nor desirable) to specify it here.

```
>>> lxterminal-gtk3*: Tracing dependencies...
	vte3
	gtk+3.0
	so:libc.musl-x86_64.so.1
	so:libgdk-3.so.0
	so:libgdk_pixbuf-2.0.so.0
	so:libglib-2.0.so.0
	so:libgobject-2.0.so.0
-->	so:libgtk-3.so.0
	so:libintl.so.8
	so:libpango-1.0.so.0
-->	so:libvte-2.91.so.0
```

dlintw: removed

testing/lxterminal-gtk3/APKBUILD
dlintw: removed

testing/lxterminal-gtk3/APKBUILD
dlintw: changed

dlintw:
@jirutka is there anything I should modify?

tmpfile:
You have 2 commits, could you squash them into one?
Also remove all `|| return 1` instances as that's done by default now.

jirutka:
> You have 2 commits, could you squash them into one?

@tmpfile, that’s not needed, we can trivially squash it during merging.

tmpfile:
I think I remembered someone asked me to squash in the past.
Well then I must apologize for the many comments I made recently asking for squashes.

But I agree not demanding squashes can be good.
The less work for contributors -> the higher chances to get contributions.

clandmeter:
Closed due to inactivity.

PR #954, opened 2017-03-02
testing/wireguard: various small fixes

itoffshore:
* fixes `$install_if` so the kernel modules install automatically when adding `wireguard-tools`
* test subpkg removed at the [request of the author](https://github.com/alpinelinux/aports/pull/948)
* `make check` removed & `sparse` removed from `makedepends` at the [request of the author](https://github.com/alpinelinux/aports/pull/948)

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,14 +3,14 @@
 
 pkgname=wireguard-tools
 pkgver=0.0.20170223
-pkgrel=0
+pkgrel=1
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
 url='https://www.wireguard.io'
 license="GPLv2"
-makedepends="libmnl-dev sparse"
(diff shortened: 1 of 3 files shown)


itoffshore:
note to @zx2c4

testing/wireguard-tools/APKBUILD
```diff
@@ -45,7 +43,7 @@ bashcomp() {
 }
 
 tests() {
```
zx2c4: This "tests package" simply _will not work_ out of context like this. It does not belong in a distribution package, no matter what it's called. If misunderstandings like this become more widespread than just you, then I wouldn't hesitate to just remove the code from the repository. It really doesn't belong.

itoffshore: test subpkg completely removed

zx2c4: Great, thanks!

zx2c4:
Latest commit looks good to me. Please merge!

PR #961, opened 2017-03-03
main/lzo: upgrade to 2.10 and modernize abuild

misery:


kaniini:
this was already bumped by someone else

PR #980, opened 2017-03-05
testing/ghc: Remove perf-llvm and -O3

mitchty:
While it makes the resulting ghc compile faster, apparently its controversial.

So remove it, it isn't strictly necessary. Bumping release.

### DIFF
--- a/testing/ghc/APKBUILD
+++ b/testing/ghc/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
 pkgname=ghc
 pkgver=8.0.2
-pkgrel=2
+pkgrel=3
 pkgdesc="The Glasgow Haskell Compiler"
 url="http://haskell.org"
 subpackages="$pkgname-doc $pkgname-dev"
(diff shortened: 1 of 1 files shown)


jirutka:
Since LLVM backend produces smaller binaries, I’d prefer to keep it. Have you compared the resulting size between `-O3` and `-O2`?

mitchty:
Yep, in my opinion the increase with -O3 is worth the boost in speed. Ghc doesn't seem to really produce meaningfully larger binaries or shared objects with -O3 over -O2. At least not to any significant degree after stripping. 

```
$ diff -u current current-O2
--- current     2017-03-13 21:34:28.000000000 -0500
+++ current-O2  2017-03-13 21:33:10.000000000 -0500
@@ -5,7 +5,7 @@
 http://haskell.org

 ghc-8.0.2-r3 installed size:
-610201600
+610177024

 ghc-8.0.2-r3 description:
 The Glasgow Haskell Compiler
@@ -14,7 +14,7 @@
 http://haskell.org

 ghc-8.0.2-r3 installed size:
-610222080
+610205696

 ghc-dev-8.0.2-r3 description:
 The Glasgow Haskell Compiler
@@ -23,7 +23,7 @@
 http://haskell.org

 ghc-dev-8.0.2-r3 installed size:
-555851776
+555839488

 ghc-dev-8.0.2-r3 description:
 The Glasgow Haskell Compiler
zsh: exit 1     diff -u current current-O2
```

jirutka:
Okay.

PR #983, opened 2017-03-06
community/go: upgrade to 1.8

errm:
default-buildmode-pie.patch has been removed since is addressed upstream
by golang/go@53aec79

set-external-linker.patch adresses golang/go#18243

### DIFF
--- a/community/go/set-external-linker.patch
+++ b/community/go/set-external-linker.patch
@@ -1,0 +1,11 @@
+--- go/src/cmd/link/internal/amd64/obj.go.orig	2017-03-06 23:55:57.000000000 +0000
++++ go/src/cmd/link/internal/amd64/obj.go	2017-03-06 23:56:16.000000000 +0000
+@@ -67,7 +67,7 @@ func Init() {
+ 	ld.Thearch.Append64 = ld.Append64l
+ 	ld.Thearch.TLSIEtoLE = tlsIEtoLE
+ 
+-	ld.Thearch.Linuxdynld = "/lib64/ld-linux-x86-64.so.2"
++	ld.Thearch.Linuxdynld = "/lib/ld-musl-x86_64.so.1"
+ 	ld.Thearch.Freebsddynld = "/libexec/ld-elf.so.1"
+ 	ld.Thearch.Openbsddynld = "/usr/libexec/ld.so"
+ 	ld.Thearch.Netbsddynld = "/libexec/ld.elf_so"
--- a/community/go/APKBUILD
+++ b/community/go/APKBUILD
@@ -1,10 +1,10 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Eivind Uggedal <eivind@uggedal.com>
 pkgname=go
-pkgver=1.7.4
+pkgver=1.8.1
 # This should be the latest commit on the corresponding release branch
-_toolsver="26c35b4dcf6dfcb924e26828ed9f4d028c5ce05a"
-pkgrel=2
+_toolsver="5682db0e919ed9cfc6f52ac32e170511a106eb3b"
+pkgrel=0
 pkgdesc="Go programming language compiler"
 url="http://www.golang.org/"
 arch="all"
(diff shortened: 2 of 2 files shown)


errm:
yuck this is failing and I am looking into why

I think it is somehow related to https://github.com/golang/go/issues/18243

ncopa:
The golang/go@53aec79 commit will make default non-pie. I think we want default pie, just like we have gcc defaulting to pie.

errm:
Ok this should do it...

errm:
https://github.com/golang/go/commit/53aec79 doesn't change the default, it just passes -no-pie to the linker when buildmode is `exe`.

I can try and get this to build with the patch to change the default buildmode to pie if you think that is better... however I am not sure it is best to deviate from upstream like this unless we have to?  If we want pie in stuff we are building its simple enough to set --buildmode=pie.

As I understood this patch was here since go didn't have a way to set -no-pie on the linker, not because we were changing the default behaviour of an upstream tool...

errm:
Humm so when trying to build with the `default-buildmode-pie.patch`

the build fails with

```
2017/03/07 13:19:05 cannot handle R_TLS_IE (sym sync/atomic.(*Value).Store) when linking internally
go tool dist: FAILED: /home/builder/package/src/go/pkg/tool/linux_amd64/link -o /home/builder/package/src/go/pkg/tool/linux_amd64/go_bootstrap /tmp/go-tool-dist-219041411/cmd/go/_go_.a: exit status 1
```

so I tried setting `GO_EXTLINK_ENABLED=1` to force go to use the external linker for cgo

and then I run into 

```
warning: unable to find runtime/cgo.a
/home/builder/package/src/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/lib/gcc/x86_64-alpine-linux-musl/6.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: unrecognized option '-pagezero_size'
/usr/lib/gcc/x86_64-alpine-linux-musl/6.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: use the --help option for usage information
```

community/go/set-external-linker.patch
```diff
@@ -0,0 +1,11 @@
+--- go/src/cmd/link/internal/amd64/obj.go.orig	2017-03-06 23:55:57.000000000 +0000
++++ go/src/cmd/link/internal/amd64/obj.go	2017-03-06 23:56:16.000000000 +0000
+@@ -67,7 +67,7 @@ func Init() {
+ 	ld.Thearch.Append64 = ld.Append64l
+ 	ld.Thearch.TLSIEtoLE = tlsIEtoLE
+ 
+-	ld.Thearch.Linuxdynld = "/lib64/ld-linux-x86-64.so.2"
++	ld.Thearch.Linuxdynld = "/lib/ld-musl-x86_64.so.1"
```
fabled: If you need to patch the ld name here, you likely need to patch it for all supported architectures.

errm: This is the path to the dynamic linker right? so I would imagine that we don't need to patch it for other GOOSes ?

kaniini: yes, just for the other architectures (x86, arm, etc).

cblecker:
Go 1.8.1 has been released. Edge should probably target it, as it contains a number of bug fixes:
- https://github.com/golang/go/issues?q=milestone%3AGo1.8.1
- https://golang.org/doc/devel/release.html#go1.8.minor

errm:
Updated to target go 1.8.1

fabled:
I posted to golang-nuts asking how to make the default buildmode PIE. Seems the build still fails with the pie patch applied.

fabled:
golang-nuts thread started: https://groups.google.com/forum/#!topic/golang-nuts/gQ6ArCTlPGU

cblecker:
It looks like the golang team is working on an alpine builder. I don't know if any of their work could be leveraged here:
- https://github.com/golang/go/issues/17891
- https://github.com/golang/go/issues/19938

fabled:
The build can be fixed with:
```
@@ -72,10 +72,11 @@ build() {
                fi
        done
 
-       ./make.bash --no-clean || return 1
+       ./make.bash
        for os in $_gocross; do
+               local hostarch="$GOARCH"
                for arch in "386" "amd64"; do
-                       GOARCH=$arch GOOS=$os ./make.bash --no-clean || return 1
+                       GOHOSTARCH=$hostarch GOARCH=$arch GOOS=$os ./make.bash
                done
        done
```

I also asked about this in golang-nuts, see if they can give suggestions how to fix this without removing --no-clean as it slows downs things considerably.

Another things I'd like to fix while at this is splitting the -source package.

errm:
@fabled I have applied your suggested changes...

PR #985, opened 2017-03-07
testing/ghc: Add libffi to depends

mitchty:
While not strictly necessary generally, when passing in linker args to ghc
to compile the c rts invokes some ffi calls.

Not sure if I need to bump the release in cases like this or not. For now just bumped it if I don't have to I can just amend the commit.

### DIFF
--- a/testing/ghc/APKBUILD
+++ b/testing/ghc/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
 pkgname=ghc
 pkgver=8.0.2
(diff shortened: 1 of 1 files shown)


fabled:
I think this is not necessary. abuild automatically finds the libffi so dependency:
```
ghc-8.0.2-r2 depends on:
gmp-dev
perl
gcc>=6.2.1
llvm3.7
busybox
so:libc.musl-x86_64.so.1
so:libffi.so.6
so:libgmp.so.10
so:libncursesw.so.6
```

Thus libffi gets pulled in automatically.

PR #986, opened 2017-03-07
testing/cabal: Add armhf to architectures

mitchty:
Since we have a ghc for armhf, also build cabal for that arch.

Do I need to bump the release to add architectures? I presume not.


ncopa:
correct. you dont need bump petrel to add architecture.

PR #993, opened 2017-03-09
main/uwsgi : enable jansson support

tomdottom:
Adds build dependency jansson-dev and install dependency jansson.

Compalation with jansson-dev enables the --json option.
Please see uwsgi docs:
http://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#json-files

### DIFF
--- a/main/uwsgi/APKBUILD
+++ b/main/uwsgi/APKBUILD
@@ -17,6 +17,7 @@ makedepends="
 	attr-dev
 	curl-dev
 	geoip-dev
+	jansson-dev
 	libcap-dev
 	linux-headers
 	linux-pam-dev


ncopa:
Could you please change first line of commit message to something like

    main/uwsg: enable jason support

or similar. Otherwise there is no way to know what package that got json support when looking at the commit log. See yourself: https://github.com/tomdottom/aports/commits/enable_uwsgi_json_option

tomdottom:
Good point.

Commit message updated accordingly.

main/uwsgi/APKBUILD
```diff
@@ -9,14 +9,18 @@ pkgdesc="uWSGI application container server"
...
+        jansson
+        "
 install="uwsgi.pre-install"
 pkgusers="uwsgi"
 pkggroups="uwsgi"
 makedepends="
 	attr-dev
 	curl-dev
 	geoip-dev
+        jansson-dev
```
jirutka: As you can see, this file is indented with tabs, but you’ve used spaces here. Please fix it.

main/uwsgi/APKBUILD
```diff
@@ -9,14 +9,18 @@ pkgdesc="uWSGI application container server"
 url="http://projects.unbit.it/uwsgi/"
 arch="all"
 license="GPL2"
-depends="mailcap"
+depends="
+        mailcap
+        jansson
```
jirutka: I think that abuild automatically detects this dependency on base of dynamic linking, so it’s unnecessary (nor desirable) to add it here. However, I haven’t verified it (yet).

ncopa: You are right, its not needed:

````
>>> uwsgi*: Tracing dependencies...
        mailcap
        busybox
        so:libc.musl-x86_64.so.1
        so:libcap.so.2
        so:libcrypto.so.38
        so:libjansson.so.4
        so:libpcre.so.1
        so:libssl.so.39
        so:libxml2.so.2
        so:libz.so.1
````

note the automatic dependency of `so:libjansson.so.4`

tomdottom:
Indentation corrected.

Let me know wrt to the need to declare jansson as a dependency or not.

tomdottom:
jansson dependency removed

PR #1001, opened 2017-03-10
main/py-cryptography: upgrade to 1.8.1

misery:


jirutka:
This version requires asn1crypto instead of asn1, see https://github.com/pyca/cryptography/blob/1.8.1/setup.py#L38.

misery:
Ah, yes.... I forgot. Thats why I added it as new aport. I remember.

misery:
Looks like you merged it. :-)

https://github.com/alpinelinux/aports/commit/0c03086d554663c646dc0b64d96b070a2a4faff1

PR #1010, opened 2017-03-12
Using hub

pickfire:


jirutka:
Could you please write some meaningful subject for your PR…?

kaniini:
this commit is wrong way to do it, because `/run` is a tmpfs.  meaning if you reboot you will lose that directory.

the init script should create the directory instead

jirutka:
The init script already handles /run directory.

pickfire:
But it's ugly that everytime on boot:
```
tor              | * /run/tor: creating directory
tor              | * /run/tor: correcting owner
tor              | * Starting tor ...                                                                          [ ok ]
```
Sorry about that pull request, it's the first time I use `hub`.

PR #1015, opened 2017-03-13
main/liblogging: upgrade to 1.0.6 and build from tarball again

misery:


jirutka:
> build from tarball again

Why? Please update the commit message with brief explanation of this change for others.

misery:
Changed to:

           main/liblogging: upgrade to 1.0.6
           
           Build from official release tarball again
           as we do not need to generate "configure"
           script with automake/autoconf by ourself.

PR #1018, opened 2017-03-13
community/py-lz4: update to 0.9.0

itoffshore:
community/py-lz4: update to `0.9.0`

### DIFF
--- a/community/py-lz4/APKBUILD
+++ b/community/py-lz4/APKBUILD
@@ -1,16 +1,19 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=py-lz4
-pkgver=0.8.2
+pkgver=0.9.0
 _pkgname=${pkgname#py-}
-pkgrel=1
+pkgrel=0
 pkgdesc="LZ4 Bindings for Python"
 url="https://pypi.python.org/pypi/lz4"
 arch="all"
 license="BSD"
 makedepends="python2-dev python3-dev py-setuptools lz4-dev"
-subpackages="py3-${pkgname/py-/}:_py3 py2-${pkgname/py-/}:_py2 $pkgname-tests:_tests"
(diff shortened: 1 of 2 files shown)


community/py-lz4/APKBUILD
```diff
@@ -44,12 +46,17 @@ _py3() {
...
 
 	mkdir -p "$subpkgdir"/usr/share/$pkgname
 	mv "$builddir"/tests "$subpkgdir"/usr/share/$pkgname
 }
 
-md5sums="50dc6c2cb99c43ce6ed95cf6ae965f0e  lz4-0.8.2.tar.gz"
-sha256sums="6bf49061d73d69c453e892ace4586b99ccffc7de558f921d18b9418235692ac7  lz4-0.8.2.tar.gz"
-sha512sums="e516d5c3309f214fa422b040a0aacdda079395624dd4f5548ed7af637b1594012aac75b309cef260b41c10a623410f815fc089b54ca5d80955d5f2ead12841b1  lz4-0.8.2.tar.gz"
+check() {
+	echo "*** running python2 tests ***"
```
jirutka: Please use `msg` (function provided by abuild) instead of `echo`.

itoffshore: fixed to `msg` & rebased

itoffshore:
fixed & rebased

jirutka:
This package requires `nose` for tests. Now it’s automatically installed when running `setup.py test`, that’s not ideal. Please add `checkdepends="py-nose"`.

itoffshore:
I've added the `checkdepends="py-nose"`  - however the test suite still downloads py-nose

I'll open another issue with the author

jirutka:
I don’t think that it’s author’s fault, but the way how `setup.py test` works. :/

jirutka:
Maybe run `nose` directly instead of via `setup.py test`?

itoffshore:
the `python2` tests are run will the repo version of `nose` - I need to create a `python3` version of `nose` so the `python3` tests are also run with the Alpine repo version of `nose`

PR #1019, opened 2017-03-14
testing/audit: initial import

tych0:
An initial import of the audit daemon, libraries, and utilities.

Signed-off-by: Tycho Andersen <tycho@docker.com>

PR #1021, opened 2017-03-14
testing/audit: fix indentation

tych0:
Here's one of the fixes requested in #1019. I need to do some more work to understand exactly what auditd needs to write an init script for it, but I should have that some time this week.


kaniini:
great!

PR #1023, opened 2017-03-15
testing/apt-dater-host: add check

HRio:


HRio:
Could this package be moved from testing to community? or what is needed?

testing/apt-dater-host/APKBUILD
```diff
@@ -3,23 +3,32 @@
...
 source="apt-dater-host-$pkgver-g$_pkgver.tar.gz::https://github.com/DE-IBH/apt-dater-host/archive/$_pkgver.tar.gz
 	0001-apk-try-to-use-virt-what.patch
+	0001-apk-add-some-unit-tests-called-via-make-check.patch
 	"
 builddir="$srcdir/"
 
 prepare() {
 	cd "$builddir"/"$pkgname"-"$_pkgver"*/ || return 1
 	patch -p1 -i "$srcdir"/0001-apk-try-to-use-virt-what.patch
+	patch -p1 -i "$srcdir"/0001-apk-add-some-unit-tests-called-via-make-check.patch
```
jirutka: Do not manually apply patches, just call `default_prepare`.

testing/apt-dater-host/APKBUILD
```diff
@@ -3,23 +3,32 @@
 pkgname=apt-dater-host
 pkgver=1.0.0
```
jirutka: The source is not version 1.0.0, please don’t lie to Alpine users. If you really need to use unreleased revision, then add appropriate suffix to the version number, e.g. `1.0.0_git20170216`.

jirutka: Aha, c3d10e0 is only one commit after 1.0.0 and it adds apk support. Then it’s better to just add it as a patch instead of fetching unreleased source.

jirutka:
Fixed in f8f30b412d2048942f459d92616169357d02858d.

testing/apt-dater-host/APKBUILD
HRio: OK, thanks, was a bit unsure how to best handle this

PR #1026, opened 2017-03-15
libsemanage, minor fixes to libse*

tych0:
Hi,

Towards the end of eventually getting policycoreutils into alpine, here's libsemanage, and some fixes that make the symlinks in /usr/lib/* valid in some other packages.

Thoughts welcome.


ncopa:
for the `ln` fixes, we could depend on `coreutils` (~5.4MB) instead of `python` (~39.7MB).

jirutka:
IMHO `ln --relative` is not needed at all, just use relative path as the first argument to `ln`.

ncopa:
That is what we currently do but apparently it does not work.

ncopa:
i just pulled in coreutils as build time dep. Thanks!

tych0:
Thanks @ncopa!

PR #1028, opened 2017-03-16
community/maven: remove openjdk depends

itoffshore:
forcing a depends on `openjdk` breaks using `maven` with `oracle jdk`
in `docker`.

### DIFF
--- a/community/maven/APKBUILD
+++ b/community/maven/APKBUILD
@@ -4,12 +4,12 @@
 pkgname=maven
 pkgver=3.3.9
 _pkgname="$pkgname-${pkgver%%.*}"
-pkgrel=1
+pkgrel=2
 pkgdesc="A Java project management and project comprehension tool."
 url="http://maven.apache.org"
 arch="noarch"
 license="ASL-2.0"
-depends="openjdk8-jre"
+options="!check"
 source="http://mirror.hosting90.cz/apache/$pkgname/$_pkgname/$pkgver/binaries/apache-$pkgname-$pkgver-bin.tar.gz"
 builddir="$srcdir/apache-$pkgname-$pkgver"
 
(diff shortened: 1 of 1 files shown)


jirutka:
> forcing a depends on `openjdk` breaks using `maven` with `oracle jdk`
in `docker`.

Oracle JDK does not work on Alpine (without very nasty hacks), because it’s provided only as a binary compiled against glibc.

apolzin:
Hi,
so is openjdk the only supported jdk on alpine?

jirutka:
Yes. Please note that Oracle JDK is basically just a branded, proprietary distribution of OpenJDK. The source code-base is the same.

This means that basically all Java applications that run on Oracle JDK 8 run also on OpenJDK 8. That said, exceptions exist, but they are quite rare (more common in graphic apps). I run many, even enterprise (e.g. Alfresco), server applications on OpenJDK without any problem.

Oracle JDK just adds some proprietary, licensed components like Java Mission Control (you'd know about them if you need them) and closed-source reimplementation of few components. You can read more e.g. at http://stackoverflow.com/a/29706592.

Oracle JDK is a proprietary product, they don’t provide source-code of all their modifications to OpenJDK and build system, just binaries. These binaries are built against glibc (GNU C Library). Alpine Linux doesn’t use glibc, it’s based on (much better) musl libc.

Glibc and musl libc are (partially) incompatible (mainly because glibc doesn’t strictly comply with POSIX standards). Therefore software built against glibc may not work on musl libc without recompiling. This is the case of Oracle JDK. Because it’s a proprietary product, we can’t recompile it and there’s nothing we can do about it.

That Docker image with Oracle JDK on Alpine uses ugly hacks to add glibc into Alpine. It's not (and can't be) supported by us, it may not work correctly in some cases and definitely isn't a production quality solution.

apolzin:
Hi,

thanks for the info! then having openjdk as a dependency makes perfect sense.
I know about about the problems with the oracle JDK on Alpine. And i would prefer not to use it.
Sadly our client insists.

Thanks again for the info!

jirutka:
If your clients insist, then tell them to pay Oracle for musl support or use a different Linux distribution (e.g. Oracle Linux with paid support). Let's make money talk, maybe they will eventually change their requirements...

PR #1031, opened 2017-03-16
community/php7: upgrade to 7.0.17

andypost:
Bug fix release http://php.net/archive/2017.php#id2017-03-16-1


andypost:
> configure: error: OpenSSL version >= 1.1 is not supported.
>>> ERROR: php7: all failed

andypost:
Looks it needs patch for libressl
Ref https://github.com/remicollet/remirepo/blob/master/scl-php70/php/php-7.0.17-openssl11.patch

vakartel:
@andypost 
I made a little fix in patch you point to (changes in aclocal.m4 instead of acinclude.m4)
Now it builds fine.
Place it to $startdir and add "$_pkgreal-$pkgver-openssl11.patch" to the end of 'sources' list.
Ref http://alpinelinux.org.ua/vaka/php-7.0.17-openssl11.patch

andypost:
we are building with libressl and libssl version (1.0.2 now) should not prevent configuring

PR #1039, opened 2017-03-17
testing/nginx-naxsi: => 1.11.10 / naxsi 0.55.3 / use supervise-daemon

itoffshore:
* `initd` now uses `supervise-daemon` for service supervision
* removes duplicate `--with-http_realip_module`
* removes `scgi_params` & `uwsgi_params` as the servers are disabled

### DIFF
--- a/testing/nginx-naxsi/nginx.initd
+++ b/testing/nginx-naxsi/nginx.initd
@@ -1,12 +1,14 @@
 #!/sbin/openrc-run
+supervisor=supervise-daemon
 
 description="Nginx http and reverse proxy server"
 extra_started_commands="reload reopen upgrade"
 
 cfgfile=${cfgfile:-/etc/nginx/nginx.conf}
-pidfile=/run/nginx/nginx.pid
+pidfile=/run/nginx/$RC_SVCNAME.sd.pid
 command=/usr/sbin/nginx
 command_args="-c $cfgfile"
+command_args_foreground='-g "daemon off;"'
 required_files="$cfgfile"
 
 depend() {
@@ -23,45 +25,24 @@ start_pre() {
 
 reload() {
 	ebegin "Reloading ${SVCNAME} configuration"
-	start_pre && start-stop-daemon --signal HUP --pidfile $pidfile
+	start_pre && $command -s reload
 	eend $?
 }
 
 reopen() {
 	ebegin "Reopening ${SVCNAME} log files"
-	start-stop-daemon --signal USR1 --pidfile $pidfile
+	$command -s reopen
 	eend $?
 }
 
 upgrade() {
-	start_pre || return 1
+	restart
+}
 
-	ebegin "Upgrading ${SVCNAME} binary"
-
-	einfo "Sending USR2 to old binary"
-	start-stop-daemon --signal USR2 --pidfile $pidfile
-
-	einfo "Sleeping 3 seconds before pid-files checking"
-	sleep 3
-
-	if [ ! -f $pidfile.oldbin ]; then
-		eerror "File with old pid ($pidfile.oldbin) not found"
-		return 1
-	fi
-
-	if [ ! -f $pidfile ]; then
-		eerror "New binary failed to start"
-		return 1
-	fi
-
-	einfo "Sleeping 3 seconds before WINCH"
-	sleep 3 ; start-stop-daemon --signal 28 --pidfile $pidfile.oldbin
-
-	einfo "Sending QUIT to old binary"
-	start-stop-daemon --signal QUIT --pidfile $pidfile.oldbin
-
-	einfo "Upgrade completed"
-
-	eend $? "Upgrade failed"
+restart() {
+	stop
+	# prevents bind() failed (98: Address in use) error msg
+	sleep 0.05
+	start
 }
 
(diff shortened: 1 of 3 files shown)


testing/nginx-naxsi/nginx.initd
```diff
@@ -1,67 +1,40 @@
 #!/sbin/openrc-run
+supervisor=supervise-daemon
 
 description="Nginx http and reverse proxy server"
 extra_started_commands="reload reopen upgrade"
 
 cfgfile=${cfgfile:-/etc/nginx/nginx.conf}
-pidfile=/run/nginx/nginx.pid
+pidfile=/run/nginx/$SVCNAME.sd.pid
```
jirutka: This variable should be named `$RC_SVCNAME`, please read [openrc-run(8)](http://manpages.org/openrc-run/8#environment).

testing/nginx-naxsi/nginx.initd
```diff
@@ -1,67 +1,40 @@
...
 depend() {
-	need net
 	use dns logger netmount
 }
 
 start_pre() {
 	ebegin
 	checkpath --directory --owner nginx:nginx ${pidfile%/*}
 	$command $command_args -t -q
+	sleep 0.012
```
jirutka: Why?

testing/nginx-naxsi/nginx.initd
```diff
@@ -1,67 +1,40 @@
...
 cfgfile=${cfgfile:-/etc/nginx/nginx.conf}
-pidfile=/run/nginx/nginx.pid
+pidfile=/run/nginx/$SVCNAME.sd.pid
 command=/usr/sbin/nginx
 command_args="-c $cfgfile"
+command_args_foreground='-g "daemon off;"'
 required_files="$cfgfile"
 
 depend() {
-	need net
```
jirutka: nginx is a web server and web servers usually needs network…

itoffshore: `need net` breaks `lxc` containers if they set an ip address in their `config`

If you set an ip address in an `lxc` config the container does not need to start networking

testing/nginx-naxsi/nginx.initd
itoffshore: without a small `sleep`:
```
musl64 [~/aports]$ sudo rc-service nginx restart
 * Caching service dependencies ...                                                                                                                         [ ok ]
 * Stopping nginx ...                                                                                                                                       [ ok ]
 * Starting nginx ...                                                                                                                                       [ ok ]
musl64 [~/aports]$ nginx: [emerg] bind() to 0.0.0.0:80 failed (98: Address in use)
nginx: [emerg] bind() to [::]:80 failed (98: Address in use)

musl64 [~/aports]$
```
it stops the error message from `nginx` starting so quickly (the server starts ok with the error message - so it's just confusing)

Also for an `upgrade` the sleep is for `0.012` instead of `3` seconds

testing/nginx-naxsi/nginx.initd
itoffshore: corrected & rebased

testing/nginx-naxsi/nginx.initd
jirutka: Okay, then please open an issue on [bugs.alpinelinux.org](https://bugs.alpinelinux.org/projects/alpine/issues). It’s not a problem of this runscript, but of our OpenRC setup for LXC. OpenRC can deal with different requirements of various platforms, I think that this can be somehow solved using keywords.

testing/nginx-naxsi/nginx.initd
jirutka: Then it should sleep only when you invoke restart, not on every start.

testing/nginx-naxsi/nginx.initd
itoffshore: ok will open a bug 

I've left `depends()` alone & rebased

testing/nginx-naxsi/nginx.initd
itoffshore: ok fixed so it will only `sleep` during `restart()` & increased to `0.05`

jirutka:
BTW You run nginx with modules contained in this abuild, right? Have you encountered some problems? Is it stable?

itoffshore:
I monitor my logs with `logcheck` & have not seen any unexpected `daemon` restarts.

I've just been using `http_sysguard` on my servers so far & have had no problems. I'll add some other modules & see if there are any problems.

testing/nginx-naxsi/nginx.initd
```diff
@@ -23,45 +25,24 @@ start_pre() {
...
-	einfo "Upgrade completed"
+	restart
+}
 
-	eend $? "Upgrade failed"
+restart() {
+	rc-service $RC_SVCNAME stop
+	# prevents bind() failed (98: Address in use) error msg
+	sleep 0.05
+	rc-service $RC_SVCNAME start
```
jirutka: Can this be replaced just with `stop && sleep 0.05 && start`? I’m not sure if it can be used with supervisor, I haven’t tried it myself yet.

itoffshore: yes that works - rebased with changes

PR #1040, opened 2017-03-18
main/openssh: enable support for Kerberos and GSSAPI

efcasado:


jirutka:
This adds many new dependencies that most of the users really don’t need. It must be separated into a subpackage, if possible.

clandmeter:
Closed due to inactivity.

PR #1046, opened 2017-03-19
testing/xbps: new aport

misery:
Add package management of VoidLinux like
pacman or dpkg to bootstrap VoidLinux.

### DIFF
--- a/testing/xbps/APKBUILD
+++ b/testing/xbps/APKBUILD
@@ -1,0 +1,61 @@
+# Contributor: André Klitzing <aklitzing@gmail.com>
+# Maintainer: André Klitzing <aklitzing@gmail.com>
+
+pkgname=xbps
+pkgver=0.51
+pkgrel=0
+pkgdesc="The X Binary Package System"
+arch="all"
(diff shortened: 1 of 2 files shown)


testing/xbps/APKBUILD
```diff
@@ -0,0 +1,61 @@
+# Contributor: André Klitzing <aklitzing@gmail.com>
+# Maintainer: André Klitzing <aklitzing@gmail.com>
+
+pkgname=xbps
+pkgver=0.51
+pkgrel=0
+pkgdesc="XBPS to bootstrap VoidLinux"
```
kaniini: the package description should be more accurate, while it is true that it is mainly meant for bootstrapping voidlinux, that doesn't necessarily have anything to do with the package itself

misery:
Done :-)

testing/xbps/APKBUILD
misery: done

jirutka:
@misery Is this package complete and fully working? I’d like to move it into community.

misery:
@jirutka Yes, it works without problems for me.

PR #1047, opened 2017-03-19
testing/stargazer: new aport

galeksandrp:


testing/stargazer/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+license="GPL2+"
+depends=""
+makedepends="expat-dev linux-headers"
+install=""
+subpackages=""
+source="http://www.$pkgname2.dp.ua/download/server/$pkgver/$pkgname2-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname2-$pkgver/projects/$pkgname"
+options="!check"
+
+prepare() {
```
jirutka: Please call `default_prepare || return 1` at the beginning of custom `prepare` function.

galeksandrp: Added.

testing/stargazer/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Contributor: Alexander Georgievskiy <galeksandrp@gmail.com>
+# Maintainer:
+pkgname=stargazer
+pkgname2=stg
```
jirutka: When you need to define custom global variable, please prefix it with underscore (to avoid potential conflicts with future abuild variables). Common name for this variable is `_pkgname`.

testing/stargazer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+build() {
+	cd "$builddir"
+
+	./build || return 1
+}
+
+package() {
+	cd "$builddir"
+	
+	mkdir -p "$pkgdir/usr/sbin"
```
jirutka: Is this really needed? Makefile is usually able to create directories itself.

testing/stargazer/APKBUILD
galeksandrp: Prefix added.

testing/stargazer/APKBUILD
galeksandrp: Yes.

testing/stargazer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+pkgver=2.408
+pkgrel=0
+pkgdesc="A billing system for small home and office networks"
+url="http://www.$_pkgname.dp.ua"
+arch="all"
+license="GPL2+"
+depends=""
+makedepends="expat-dev linux-headers"
+install=""
+subpackages=""
```
jirutka: Please add subpackage `$pkgname-dev`, there are many header files.

testing/stargazer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+	./build || return 1
+}
+
+package() {
+	cd "$builddir"
+	
+	mkdir -p "$pkgdir/usr/sbin"
+
+	make PREFIX="$pkgdir" install
```
jirutka: What are these files and directories?

```
/var/stargazer/admins/admin.adm
/var/stargazer/tariffs/tariff.tf
/var/stargazer/users/test/conf
/var/stargazer/users/test/stat
```

testing/stargazer/APKBUILD
galeksandrp: Added.

PR #1051, opened 2017-03-20
main/{git,imagemagick,jemalloc,nss,redis}: update

pickfire:


pickfire:
For `6efe253` I am not quite sure why it fails.

vakartel:
@pickfire 
It fails on packaging. I reviewed those horror and made some changes. Now it builds fine.
[nss.patch.txt](https://github.com/alpinelinux/aports/files/857404/nss.patch.txt)

vakartel:
@pickfire 

ImageMagick fails because there is no 7.0.5.2. Change it to 7.0.5.3 :)

pickfire:
> ImageMagick fails because there is no 7.0.5.2. Change it to 7.0.5.3 :)

@vakartel I know, I had changed that hours ago but looks like I forgot to push it.

Maybe you might want to add nss.patch.txt as a commit here? I don't want to take the blame (`git blame`) for it.

By the way, thanks a lot for helping out with the nss thing, I didn't know that without check, it will fail.

pickfire:
@vakartel Thanks a lot for helping out. I forgot what I just typed so I just make it short.

vakartel:
@pickfire 
I can't add commits to your PR. So I added my proposed changes as attached file.

PS there is an unresolved conflict.

pickfire:
@vakartel Yes, nss is still broken after your patch. I wonder why can't you add commit to my PR.

fabled:
Git was already updated to newer version than in this. Perhaps drop that?

vakartel:
I have no push access to alpinelinux/aports, So I can't made changes in
others PR-s.

Just now I successfuly rebuild nss again. I don't see you applied patch I
sent.

there is attached main/nss/APKBUILD I used to build.

2017-03-22 5:55 GMT+02:00 Ivan Tham <notifications@github.com>:

pickfire:
@fabled Oh, looks like I am the one that updated git. Sorry http://git.alpinelinux.org/cgit/aports/commit/main/git?id=c6b849ea0ea4952815b911e89a943ac2e5db7b1c

@vakartel Just commit to your repo, I can help to add that or do you prefer that I remove that so that you can push it yourself?

ncopa:
imagemagick breaks abi. everything using the imagemagic's libs needs to be rebuilt. Alternatively we rename current imagemagic to imagemagic6

jirutka:
@pickfire Could you please rebase this PR, remove git and apply @vakartel’s patch for NSS?

ncopa:
i already pushed it. sorry. what was the patch for NSS?

ncopa:
ah, yes, @vakartel's patch should be applied.

jirutka:
It seems that are relevant commits from this PR are merged, so closing it.

@pickfire Next time please really open a separate PR for each package (when the changes does not depend on each other). I know that I’ve suggested that it’s not necessary to create single PR for each pkg, to make it easier for you, but as we see here, it’s not good.

pickfire:
Okay, I think next time if I had successfully build the package, I will just send to ML because it is a lot easier unless there is some parts that I am not sure.

ncopa:
@vakartel's patch broke things. The problem is that the `libssl3.so` symlink needs to be in main package, while the `default_dev` function will move it to the `$pkgname-dev` subpackage. I will have to revert the last hunk.

jirutka:
@ncopa Yes, that’s why no one, including you, should not skip reviews, as you did here. Especially in the case of packages like NSS.

ncopa:
@jirutka Depends on what you mean with "reviews". I did review this pull request, i did test build it, i did fix things i didn't like with it and I did rebuild the needed packages for imagemagick.

After pushing I did notice the nss patch, which I did review before I decided to give it a test compile. Since all looked good I pushed. Travis would not have caught this.

Or do you mean that I should not push anything without making a pull request on github first?

jirutka:
Aha, I didn’t know that you’ve tested all of it locally and fixed issues. I’m sorry for that bad assumption. So you just haven’t noticed that files were unintentionally moved between (sub)packages due to Valery’s patch?

> Travis would not have caught this.

No, but checkapk (that is run after the pkg is built on Travis) generates diff between old and new build of the package, so you can notice here when some files are moved.

> Or do you mean that I should not push anything without making a pull request on github first?

No, not at all. I just didn’t like that CI revealed breakage, Valery proposed a patch, we asked the contributor to apply it and rebase, and then see if the build will succeed or not. And before he did it you’ve just merged it without a word and now there’s a massive breakage. I admit that I didn’t know what’s behind (i.e. that you’ve actually reviewed and verified it) and overreacted a bit.

pickfire:
@jirutka Yes, notice that http://git.alpinelinux.org/cgit/aports/commit/?id=968d2718ea097e63b3d532bb6e8b84d1fa20a282 shows `Natanael Copa <ncopa@alpinelinux.org>` so he did review it.

> Or do you mean that I should not push anything without making a pull request on github first?

Oh, I hope no.

PR #1052, opened 2017-03-20
*/py-*: update

pickfire:


main/py-cryptography/APKBUILD
```diff
@@ -3,7 +3,7 @@
 pkgname=py-cryptography
 _pkgname=${pkgname#py-}
 pkgver=1.7.2
```
misery: upgrade ???

by the way: https://github.com/alpinelinux/aports/pull/1001

jirutka:
The only updated pkg in this PR is main/py-requests-oauthlib, all others have just set pkgrel to 0.

pickfire:
@jirutka Did you set the pkgrel to 0? I thought abump automatically sets that?

jirutka:
I've just rebased it against master, to use more up-to-date mirror. You actually didn't bump pkgver, except one package...

pickfire:
What do I need to change now?

jirutka:
Actually bump packages... Please look at diff here in the PR to see what you've really committed.

main/py-cryptography/APKBUILD
pickfire: Ah, I didn't notice this.

pickfire:
Ah, looks like there's a lot packages that aren't upgraded.

PR #1057, opened 2017-03-21
testing/percona-xtradb-cluster-galera: new aport

Flowman:
https://www.percona.com/software/mysql-database/percona-xtradb-cluster
Codership Galera library for Percona XtraDB Cluster

### DIFF
--- a/testing/percona-xtradb-cluster-galera/APKBUILD
+++ b/testing/percona-xtradb-cluster-galera/APKBUILD
@@ -1,0 +1,51 @@
+# Contributor: Peter Szalatnay <theotherland@gmail.com>
+# Maintainer: Peter Szalatnay <theotherland@gmail.com>
+pkgname=percona-xtradb-cluster-galera
+pkgver=5.7.16
+_pkgver="$pkgver-27.19"
+pkgrel=0
+pkgdesc="Codership Galera library for Percona XtraDB Cluster"
+url="https://www.percona.com/software/mysql-database/percona-xtradb-cluster"
+arch="all"
+license="GPL2"
+depends=""
+depends_dev=""
+makedepends="scons asio-dev check-dev"
+source="https://github.com/percona/galera/archive/pxc_$_pkgver.tar.gz
+        fix-gu_arch.patch
+        fix-gu_alloc.patch
+        "
+
+builddir="$srcdir/galera-pxc_$_pkgver"
+
+build() {
+        cd "$builddir"
+
+        local GALERA_REVISION="$(test -r GALERA-REVISION && cat GALERA-REVISION)"
+ 
+        CFLAGS="$CFLAGS -DSIGEV_THREAD_ID=4 -U_FORTIFY_SOURCE"
+        CXXFLAGS="$CXXFLAGS -U_FORTIFY_SOURCE -Wno-error=cpp -Wno-error=missing-field-initializers"
+
+        scons psi=1 --config=force revno="$GALERA_REVISION" libgalera_smm.so || return 1
+        scons --config=force revno="$GALERA_REVISION" garb/garbd || return 1
+}
+
+package() {
+        cd "$builddir"
+
+        install -Dm 755 "$builddir"/libgalera_smm.so \
+            "$pkgdir"/usr/lib/galera3/libgalera_smm.so || return 1
+
+        install -Dm 755 "$builddir"/garb/garbd \
+            "$pkgdir"/usr/bin/garbd || return 1
+}
+
+md5sums="a9f774cefeecec27271c7f5dfd30eeeb  pxc_5.7.16-27.19.tar.gz
+43a717bf490f0a2782d2ade4b7bbc300  fix-gu_arch.patch
+c84b8405a1efbafd97da777bd9821d04  fix-gu_alloc.patch"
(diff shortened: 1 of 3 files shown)


Flowman:
So this fails as it requires asio-dev that is in testing. How can that be fixed?

Flowman:
The checksum works fine on local build but Travis does not like them.

testing/percona-xtradb-cluster-galera/APKBUILD
```diff
@@ -0,0 +1,60 @@
+# Contributor: Peter Szalatnay <theotherland@gmail.com>
+# Maintainer: Peter Szalatnay <theotherland@gmail.com>
+pkgname=percona-xtradb-cluster-galera
+pkgver=5.7.16
+_pkgver=5.7.16-27.19
```
ncopa: i dont like this. we have a script called `abump` that we use for simple version bumps. if we run `abump percona-xtradb-cluster-galera-5.7.17` to upgrade, then would it modify the `pkgver` only, build would succeed but the actual downloaded and built version would be 5.7.16 since that was unmodified.

better do:
````sh
_pkgver="$pkgver-27.19"
````

The would alteast the download fail and we'd not end up with a package with wrong version number.

File: testing/percona-xtradb-cluster-galera/APKBUILD [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+_builddir="$srcdir/galera-pxc_$_pkgver"
+prepare() {
+        local i
+        cd "$_builddir"
+        for i in $source; do
+                case $i in
+                *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+                esac
+        done
+}
```
ncopa: please replace `_builddir` with `builddir` and remove the `prepare()` function to let abuild apply the patches.

testing/percona-xtradb-cluster-galera/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+
+build() {
+        cd "$_builddir"
+
+        local GALERA_REVISION="$(test -r GALERA-REVISION && cat GALERA-REVISION)"
+ 
+        local CFLAGS="-DSIGEV_THREAD_ID=4 -U_FORTIFY_SOURCE"
```
ncopa: we should keep the system CFLAGS. do `CFLAGS="$CFLAGS ..."`

testing/percona-xtradb-cluster-galera/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+        local CXXFLAGS="-U_FORTIFY_SOURCE -Wno-error=cpp -Wno-error=missing-field-initializers"
```
ncopa: we should keep the system CPPFLAGS. do `CPPFLAGS="$CPPFLAGS ..."`

ncopa:
can you please also merge the checksum fixes into the first patch?

testing/percona-xtradb-cluster-galera/APKBUILD
Flowman: The problem here is that perconas version number works like this, mysql.version-wsrep.version. Sometimes they bump the wsrep version but not the mysql version, and sometimes the other way around. I wanted to add the whole version number to pkgver but abuild does not like dashes. What would be the best solution for this?

ncopa:
This fails now. Looks like libressl related:
````
galerautils/src/gu_asio.cpp: In function 'void gu::ssl_init_options(gu::Config&)':
galerautils/src/gu_asio.cpp:85:74: error: no matching function for call to 'asio::ssl::context::context(asio::io_service&, asio::ssl::context_base::method)'
             asio::ssl::context ctx(io_service, asio::ssl::context::sslv23);
                                                                          ^
In file included from /usr/include/asio/ssl/context.hpp:755:0,
                 from /usr/include/asio/ssl.hpp:18,
                 from galerautils/src/gu_asio.hpp:33,
                 from galerautils/src/gu_asio.cpp:6:
/usr/include/asio/ssl/impl/context.ipp:62:1: note: candidate: asio::ssl::context::context(asio::ssl::context_base::method)
 context::context(context::method m)
 ^~~~~~~
/usr/include/asio/ssl/impl/context.ipp:62:1: note:   candidate expects 1 argument, 2 provided
In file included from /usr/include/asio/ssl.hpp:18:0,
                 from galerautils/src/gu_asio.hpp:33,
                 from galerautils/src/gu_asio.cpp:6:
/usr/include/asio/ssl/context.hpp:35:7: note: candidate: asio::ssl::context::context(const asio::ssl::context&)
 class context
       ^~~~~~~
/usr/include/asio/ssl/context.hpp:35:7: note:   candidate expects 1 argument, 2 provided
galerautils/src/gu_asio.cpp: In function 'void gu::ssl_prepare_context(const gu::Config&, asio::ssl::context&, bool)':
galerautils/src/gu_asio.cpp:145:37: error: 'class asio::ssl::context' has no member named 'impl'
         SSL_CTX_set_cipher_list(ctx.impl(), conf.get(param).c_str());
                                     ^~~~
scons: *** [galerautils/src/gu_asio.os] Error 1
scons: building terminated because of errors.
````

clandmeter:
Closed due to inactivity.

PR #1056, opened 2017-03-21
testing/percona-xtradb-cluster: new aport

Flowman:
https://www.percona.com/software/mysql-database/percona-xtradb-cluster
Percona XtraDB Cluster is an active/active high availability and high
scalability open source solution for MySQL clustering

PR #1058, opened 2017-03-21
testing/percona-xtrabackup: new aport

Flowman:
http://www.percona.com/software/percona-xtrabackup/
Percona XtraBackup is a MySQL non-blocking backup solution for InnoDB
and XtraDB data

### DIFF
--- a/testing/percona-xtrabackup/APKBUILD
+++ b/testing/percona-xtrabackup/APKBUILD
@@ -1,0 +1,54 @@
+# Contributor: Peter Szalatnay <theotherland@gmail.com>
+# Maintainer: Peter Szalatnay <theotherland@gmail.com>
+pkgname=percona-xtrabackup
+pkgver=2.4.6
+pkgrel=0
+pkgdesc='Percona XtraBackup is a MySQL non-blocking backup solution for InnoDB and XtraDB data'
+arch=all
+url='http://www.percona.com/software/percona-xtrabackup/'
+license='GPL'
+makedepends='bash bison cmake ncurses-dev libaio-dev libgcrypt-dev zlib-dev curl-dev libev-dev vim'
+source="https://github.com/percona/percona-xtrabackup/archive/percona-xtrabackup-$pkgver.tar.gz
+        fix-posix_timers.patch
+        "
+
+subpackages="$pkgname-doc $pkgname-test:mytest"
+
+builddir="$srcdir/percona-xtrabackup-percona-xtrabackup-$pkgver"
+
+build() {
+        cd "$builddir"
+
+        CFLAGS="$CFLAGS -DSIGEV_THREAD_ID=4"
+
+        cmake . -DBUILD_CONFIG=xtrabackup_release \
+                -DCMAKE_INSTALL_PREFIX=/usr \
+                -DWITH_MAN_PAGES=OFF \
+                -DDOWNLOAD_BOOST=1 \
+                -DWITH_BOOST="$builddir"/libboost \
+                || return 1
+
+        make || return 1
+}
+
+package() {
+        cd "$builddir"
+        make DESTDIR="$pkgdir/" install || return 1
+
+        install -Dm644 COPYING "$pkgdir"/usr/share/licenses/$pkgname/COPYING || return 1
+}
+
+mytest() {
+    pkgdesc="The test suite distributed with Percona XtraBackup"
+    mkdir -p "$subpkgdir"/usr/bin || return 1
+    mv "$pkgdir"/usr/xtrabackup-test \
+        "$subpkgdir"/usr/bin/ \
+        || return 1
+}
+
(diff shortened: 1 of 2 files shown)


testing/percona-xtrabackup/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+license='GPL'
+makedepends='bash bison cmake ncurses-dev libaio-dev libgcrypt-dev zlib-dev curl-dev libev-dev vim'
+source="https://github.com/percona/percona-xtrabackup/archive/percona-xtrabackup-$pkgver.tar.gz
+        fix-posix_timers.patch
+        "
+
+subpackages="$pkgname-doc $pkgname-test:mytest
+        "
+
+_builddir="$srcdir/percona-xtrabackup-percona-xtrabackup-$pkgver"
```
ncopa: please use `builddir` instead of `_builddir` and remove the `prepare()` function. If builddir is set then will abuild apply the patches itself.

testing/percona-xtrabackup/APKBUILD
```diff
@@ -0,0 +1,64 @@
+# Contributor: Peter Szalatnay <theotherland@gmail.com>
+# Maintainer: Peter Szalatnay <theotherland@gmail.com>
+pkgname=percona-xtrabackup
+pkgver=2.4.6
+pkgrel=0
+pkgdesc='Percona XtraBackup is a MySQL non-blocking backup solution for InnoDB and XtraDB data'
+arch=all
+url='http://www.percona.com/software/percona-xtrabackup/'
+license='GPL'
+makedepends='bash bison cmake ncurses-dev libaio-dev libgcrypt-dev zlib-dev curl-dev libev-dev vim'
```
ncopa: why is vim needed during build?

testing/percona-xtrabackup/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+                case $i in
+                *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
+                esac
+        done
+}
+
+build() {
+        cd "$_builddir"
+
+        local CFLAGS="-DSIGEV_THREAD_ID=4"
```
ncopa: CFLAGS is a global variable that we set systemwide. To append cflags do:

````sh
CFLAGS="$CFLAGS -DSIGEV_THREAD_ID=4"
````

That way you dont unset the systemwide `-Os` (or what ever default flags is in use)

File: testing/percona-xtrabackup/APKBUILD [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+
+        cmake . -DBUILD_CONFIG=xtrabackup_release \
+                -DCMAKE_INSTALL_PREFIX=/usr \
+                -DWITH_MAN_PAGES=OFF \
+                -DDOWNLOAD_BOOST=1 \
```
ncopa: will it download boost? Can't you use system boost by adding boost-dev to makedepends?

Flowman:
It needs boost 1.59 and that is not in the latest alpine 3.5 as 1.62 has replaced it.

testing/percona-xtrabackup/APKBUILD
Flowman: It is a dependency it needs vim-common on debian and it does not build without it.

testing/percona-xtrabackup/APKBUILD
Flowman: It needs boost version 1.59 and latest alpine comes with 1.62

fabled: We would rather patch the package than carry bundled old boost. Is there upstream bug and or commits that would fix build against current boost?

testing/percona-xtrabackup/APKBUILD
```diff
@@ -0,0 +1,54 @@
...
+        cd "$builddir"
+
+        CFLAGS="$CFLAGS -DSIGEV_THREAD_ID=4"
+
+        cmake . -DBUILD_CONFIG=xtrabackup_release \
+                -DCMAKE_INSTALL_PREFIX=/usr \
+                -DWITH_MAN_PAGES=OFF \
+                -DDOWNLOAD_BOOST=1 \
+                -DWITH_BOOST="$builddir"/libboost \
+                || return 1
```
fabled: `|| return 1` are deprecated with latest abuild as the functions are now executed with `set -e`. Please remove them all.

testing/percona-xtrabackup/fix-posix_timers.patch
```diff
@@ -0,0 +1,50 @@
+--- a/mysys/posix_timers.c
++++ b/mysys/posix_timers.c
+@@ -188,10 +188,20 @@
+           On error, -1 is returned, and errno is set to indicate the error.
+ */
+
++struct linux_sigevent {
```
fabled: This is wrong. See PR #808 and #1056 review comments.

clandmeter:
Closed due to inactivity.

PR #1060, opened 2017-03-21
rename: testing/php7.1 -> testing/php7

vakartel:

PR #1061, opened 2017-03-21
Remade php7 extensions for testing/ (php 7.1.x) and for community/ (php 7.0.x)

vakartel:
@andypost 
I'll review all testing/php7-* packages (do cleanups, upgrade if need, etc.) and will move it all to community

So, there would be a pack of patches.


vakartel:
all those fails - It's a bug in community/php7-dev. I just send a patch to fix it...

vakartel:
in progress ...

jirutka:
Some of these abuilds are duplicated for php5 and php7. When the source is the same and it just compiles against different PHP version, there should be just single abuild with two subpackages. It simplifies maintenance and make it easier to keep them consistent. That’s what we do for Lua, Python and some others. Please take a look at [php-xdebug](https://github.com/alpinelinux/aports/blob/master/community/php-xdebug/APKBUILD) for example of php package with subpackages for php5 and php7.

vakartel:
@jirutka 
I learned how python and lua modules builds and I like it. And I thinked will made something simmilar.
I agree with you and don't like a double-work, but for now it's impossible. We have two similar-named php7 in community and testing. I really don't know the way how to point building process to make  two packages in community and testing in one way. When it named differently php7 and php7.1 there was a sence.
I had many building variants and I played with php-apkbuild many times. I had php-imagick and php-gmagick builded for all 5, 7, 7.1 at once, but I didn't push it to git. After dozens remakes I thinked I have an universal template for 3rd-party php extensions, But when I started to migrate extensions from testing to community I made 4  improvements of my template at one day.
Some extensions have different versions for php5.x and php7.x (php-apcu - 4.x for php5, 5.x - for php7), some extensions has a -dev part, so all it would made multiversion APKBUILD really complex.
And about php5: It's not updated anymore, so any new php5 package would be one-timed APKBUILD.
So the only benefit - ease maintenance.
And it's not worth this complication

andypost:
PHP5 will security updates til Dec 2017

vakartel:
@jirutka , @andypost 

I moved all php7-* from testing. So now it can be reviewed

Next step: I copy it to testing again with php7.1 specific changes.

vakartel:
@jirutka , @andypost 

It's done.

vakartel:
That all good, but now we have totally broken php7 3rd party extensions for those who use edge/testing. any docker image based on alpine:edge is affected. php-xdebug widely used by developers is unusable now.
When I provided #1060 I thniked there would be some debates for weeks, disaggreements ... etc.
So when I saw #1060 suddenly merged, I started to fix all broken extensions as quick as possible.
And what I see? #1061 in hold ... nice.
IMHO For now it would be better to revoke #1060 at the time while #1061 would be in hold/review and then merege it together.

jirutka:
> And what I see? #1061 in hold ... nice.

I didn’t have time to reply to your comment, so I’ve marked it as Hold just to ensure that someone will not merge it meanwhile.

> We have two similar-named php7 in community and testing.

This is only a temporary state. I hope that you will upgrade all php 7.0 packages to 7.1 soon and finally get rid of this insanity. ;)

> I really don't know the way how to point building process to make two packages in community and testing in one way.

There’s no clean way for this and that’s not what I asked you for. I’d like to just unify php[57] abuilds in main/community, avoid having single upstream source in two abuilds in non-testing repositories.

> Some extensions have different versions for php5.x and php7.x (php-apcu - 4.x for php5, 5.x - for php7)…

If php-apcu 5.x is not compatible with php5, then we need to keep two separate abuilds – php5-apcu and php7-apcu.

> …some extensions has a -dev part…

There’s no need to create -dev subpackage for each supported version. Example: let there be an PHP extension named _foo_ in version 1.2.3. This extension supports both php5 and php7. It also installs some header files and documentation. Then you should have package php-foo with subpackages php5-foo, php7-foo, php-foo-dev and php-foo-doc.

> And about php5: It's not updated anymore, so any new php5 package would be one-timed APKBUILD.

Once we have them in repository, we must maintain them and at least patch security issues. Also we will probably have the same situation once php8 (or even yet another incompatible php7.x) is released, so it’s better to be prepared. I consider it as a mistake that we ever allowed to create this mess of duplicated abuilds for the same source package.

> …now we have totally broken php7 3rd party extensions for those who use edge/testing. any docker image based on alpine:edge is affected.

edge is unstable branch and testing is über unstable. If someone rely on packages in testing, then it’s only his/her problem.

community/php5-xdebug/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+	./configure --prefix=/usr || return 1
+	make
+}
+
+check() {
+	make -C "$builddir" NO_INTERACTION=1 test
+}
+
+package() {
+	local confdir=$(php-config|grep -o "scan-dir=[[:graph:]]*"|cut -d= -f2)
```
jirutka: Please put one space around `|`. This is quite unreadable.

testing/php7-zmq/APKBUILD
```diff
@@ -1,37 +1,38 @@
...
+	make -C "$builddir" NO_INTERACTION=1 test
 }
 
 package() {
-	cd "$builddir"
-	make INSTALL_ROOT="$pkgdir"/ install || return 1
-	install -d "$pkgdir"/etc/php7/conf.d || return 1
-	echo "extension=$_pkgreal.so" > "$pkgdir"/etc/php7/conf.d/50_$_pkgreal.ini
+	local confdir=$(php-config|grep -o "scan-dir=[[:graph:]]*"|cut -d= -f2)
+	local apiver=$(($(phpize -v|awk -F: '/PHP Api/{print $2}')))
```
jirutka: What’s this? Why you use arithmetic expression with a command substitution inside? o.O

testing/php-couchbase/APKBUILD
```diff
@@ -1,62 +0,0 @@
...
-
-	make -C "$builddir-$ver" INSTALL_ROOT="$subpkgdir" install || return 1
-
-	install -d "$subpkgdir"/etc/php$ver/conf.d || return 1
-	echo "extension=$_pkgname.so" > "$subpkgdir"/etc/php$ver/conf.d/$_pkgname.ini
-}
-
-md5sums="f58ee1596ade978832f033f2e1662579  couchbase-2.2.3.tgz"
-sha256sums="5c998e9c98e51a783a5d74ec36aa6ac740a4a958dd8c206ed9e9d1580234907f  couchbase-2.2.3.tgz"
-sha512sums="d1feeb33a8ce450919731d17f84b170eaa68da104cedfe8ffa56f4e2e6062e250077d00babaca4872cc5bf271509daee2ca83e35cfb4b8ceef4b402cc20c1bf3  couchbase-2.2.3.tgz"
```
jirutka: Why the heck did you divided this abuild back to two duplicated aports?!

jirutka:
> So when I saw #1060 suddenly merged,

Suddenly? If it wasn’t ready / you didn’t want to get it merged, then you should explicitly state this (be adding WIP to the PR subject and maybe an explanation to the description).

andypost:
Sorry late here, used to discus it with Valery in person

I used to create https://docs.google.com/spreadsheets/d/10nGKgwym2j7rle85pHj72I_2hEQ_VpLCd2CVKLGL_BE/edit#gid=0
here's compatibility tested, there's no reason to maintain *5 packages for ages when last LTS is done
that's why some extensions no longer makes sense for php7 and later...
btw Is there deprecation process?

Another concerns about php dependencies for things like php*admin (here we need set of extensions) so kinda grouping will be needed and php7 vs php5 is a question of year

jirutka:
I don't have any problem with removing php5 extensions and keeping just php7. My concern is only duplicated abuilds (multiple APKBUILDs for single upstream source).

community/php5-xdebug/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+pkgver=2.5.1
+pkgrel=0
+_pkgver=${pkgver/_rc/RC}
+_phpver=${pkgname%%-*}
+_pkgreal=${pkgname#*-}
+pkgdesc="$_phpver extension: provides functions for function traces and profiling"
+url="http://pecl.php.net/package/$_pkgname"
+arch="all"
+license="PHP"
+pecldepends="$_phpver-dev autoconf"
```
vakartel: I don't like it too, but I puted this variable because old phpext-template has it.
So I'll just replace it in community/php7-* to "php7-dev automake".
For testing/php7-* I'll use php7-pecl instead of php7-dev, because php7-pecl contains all dependencies to build extension.

testing/php7-zmq/APKBUILD
vakartel: It's a trick to make two things at once:
 - trim spaces around
 - verify that we have numeric value

community/php5-xdebug/APKBUILD
vakartel: agree...
for php7 (7.1) I want to make a helper that do all this grep/cut and just shows confdir or apiver values. And put this helper to php7-pecl subpackage.

testing/php-couchbase/APKBUILD
vakartel: because it fails to build in travis-ci
apk add php5-dev php7-dev raise an error about conflict in /usr/bin

`$ apk add php5-dev php7-dev ; echo $?

(1/7) Installing php5-config (5.6.30-r1)
(2/7) Installing libxml2 (2.9.4-r1)
(3/7) Installing php5 (5.6.30-r1)
(4/7) Installing php5-dev (5.6.30-r1)
(5/7) Installing php7-config (7.1.3-r0)
(6/7) Installing php7 (7.1.3-r0)
ERROR: php7-7.1.3-r0: trying to overwrite usr/bin/php owned by php5-5.6.30-r1.
(7/7) Installing php7-dev (7.1.3-r0)
ERROR: php7-dev-7.1.3-r0: trying to overwrite usr/bin/phpize owned by php5-dev-5.6.30-r1.
ERROR: php7-dev-7.1.3-r0: trying to overwrite usr/bin/php-config owned by php5-dev-5.6.30-r1.
Executing busybox-1.26.2-r0.trigger
2 errors; 414 MiB in 164 packages
2
$`

vakartel:
> I don't have any problem with removing php5 extensions

for example main/phpldapadmin (last release in 2012) depends on php5 only.
at sourceforge its package name is phpldapadmin-php5.

Or you mean 3rd-party php5-extensions only?

vakartel:
For now it's impossible to make multiversion apkbuild because of conflict while installing both php5-dev and php7-dev.

I tried to resolve it by playing with provides, replaces
 - php5-dev provides=php-dev, replaces=php-dev
 - php7-dev provides=php-dev, replaces=php-dev

but have no luck.

vakartel:
@jirutka 
And what do you think about to move all php5 related packages (and phpldapadmin too) from main to community?

jirutka:
> For now it's impossible to make multiversion apkbuild because of conflict while installing both php5-dev and php7-dev.

This is totally unrelated to “multiversion” APKBUILD. Just don’t use `install_if`.

> And what do you think about to move all php5 related packages (and phpldapadmin too) from main to community?

Great idea! 👍

vakartel:
> This is totally unrelated to “multiversion” APKBUILD. Just don’t use install_if.

Sometime in the past it was possible, not now. I understand about install_if ... etc. But we have two identical-named packages for now. So while it exists - let's maintain it as is, without any complications.
When we have all php-related packs would be remaded to one template, then I start to combine it where it possible.  

I really want to know your and others oppinion about this. Because I want to remade all php5-* extensions too.

jirutka:
Aha, I understand the problem now, your example in [previous comment](https://github.com/alpinelinux/aports/pull/1061/files/def062c3b2be3db635d9564872451985f1c3cfef#r107851199) helped. My bad, I forgot about this consequence. You can’t merge php5-foo and php7-foo into single APKBUILD, because it requires installing both php5 and php7 simultaneously for building.

Well, I remember that I’ve warned you about this some time ago… This must be solved before branching v3.6, php5 and php7 must not be in conflict.

I suggest to add version suffix to both php5 and php7 binaries (e.g. php5, php7, phpize5, phpize7, …) and create symlink without suffix for the default PHP version (php7). It’s future proof approach. We do the same with Python (python2, python3, python -> python2 for now) and Lua.

vakartel:
@jirutka 

I can resolve php[57]-dev dependencies via removing unsuffixed links to scripts and binaries (phpize[57], php-config[57, /usr/bin/php[57]) So, the combined package would built succesfuly except we need to switch off all package's testing phase because all tests uses unsuffixed /usr/bin/php for it.

But still I insist not to do this and leave the packages as is. It seems to me more correct. Maybe I do not understand your global idea for 3.6.

I want to discuss this (add @andypost to read this)

And anyway I agree to make php[57]- combined packs in future. But before it we need to provide all it to single template, move all it to single repo (community), then we need to made php5 and php7 cores single templated (It's done in my backports and I provided those PR's)

vakartel:
@jirutka 
Ok, I started to remade community/php5 and my backported php5 PR to not collide with php7-dev ones.
At ideal I want my backported php5 PR be accepted. So the php7 (testing/php7) and community/php5 would have identical behavior.

vakartel:
And about php5-pear-* source-based extensions. @andypost sayed me thats all deprecated, not used and be in pre-death state. Have a sense to review it have used as dependency, and if not - move it to cemetery (unmaintained)

rnalrd:
I was reviewing http://patchwork.alpinelinux.org/patch/3242/
I don't think it's worth the hassle to maintain a single APKBUILD for php-xdebug and php-couchbase (currently broken). Just choose to use a single PHP version like other APKBUILDs, then make php5 and php7 "conflict=" each other.
Yeah, proper fix is to add a versioning suffix to the binaries for php5, but I won't do it now with AL3.6 release approaching.

jirutka:
> Yeah, proper fix is to add a versioning suffix to the binaries for php5, but I won't do it now with AL3.6 release approaching.

I disagree, IMO this **must** be before v3.6, because the current state is just **broken** – you can’t install both php5 and php7 now.

rnalrd:
> I disagree, IMO this must be before v3.6, because the current state is just broken – you can’t install both php5 and php7 now.

Yes, it's broken when assuming that php5 and php7 should be installed side by side, but I'm not sure this was the original idea.
Looks like that other distros (at least ubuntu/debian and archlinux) don't allow this by default. I'm not saying we should do the same, I'm just looking at options and costs. Given that php5 will be without security support in the last 4-5 months of Alpine 3.6 lifecycle, I'd just update all APKBUILDs to use only php7, if supported from the upstream project, or php5 and leave php5 as it is.
Cheers

jirutka:
> Yes, it's broken when assuming that php5 and php7 should be installed side by side, but I'm not sure this was the original idea.

We support this for all other language runtimes. Keeping php5/php7 in conflict has multiple negative consequences.

> Given that php5 will be without security support in the last 4-5 months of Alpine 3.6 lifecycle, I'd just update all APKBUILDs to use only php7, if supported from the upstream project, or php5 and leave php5 as it is.

Agree. @kaniini suggested to remove php5 completely, so it seems that this idea has wider support.

@andypost @vakartel Do we have any packages in main or community that does not support php7 at all and the upstream project is not already dead?

jirutka:
Update: @ncopa now supports keeping php5, but only for PHP applications that does not support php7, removing all php5-* packages that are not needed for the php5-only applications, and declaring php5 and php7 as conflicting (php7: `depends="!php5"`, php5: `depends="!php7"`).

To sum it up, the “action point” for now is to create a list of packages in main/community that do not support php5 (and maybe subset of them that can be easily patched for php7).

vakartel:
webapps like phpldapadmin do not use 3rd party php extensions. So I think it's ok to remove it all. Just stay pure community/php5 with it's own bundled extensions.

A time ago I made a subpackage for cacti - cacti-php5 - just a bundle of requied extensions. It does not need any 3rd party ones.

But php5 / php7 are complex package-sets to declare a simple depends=!php[57]. It conflicts is several binaries and scripts stored in several subpackages (php5 and php7 are only binary cli packages), And have no conflict in extensions, php-fpm service, apache module etc.

vakartel:
I want also to ask to review/apply #890, #893  to make all php5/php7 builds have similar behavior in 3.6

kaniini:
This has been implemented.

vakartel:
@kaniini 
what exactly?

vakartel:
I just received a message from the first victim of packet removal.
Someone still use testing/php5-* (php5-redis) on their docker images ...

f3l1x:
@vakartel 

I was using:

```
        php5-amqp@testing \
        php5-apcu@testing \
        php5-bcmath \
        php5-calendar \
        php5-cgi \
        php5-cli \
        php5-common \
        php5-ctype \
        php5-curl \
        php5-gettext \
        php5-gd \
        php5-iconv \
        php5-imagick \
        php5-imap \
        php5-intl \
        php5-json \
        php5-ldap \
        php5-mcrypt \
        php5-memcached@testing \
        php5-mysqli \
        php5-openssl \
        php5-pdo \
        php5-pdo_mysql \
        php5-pdo_pgsql \
        php5-pdo_sqlite \
        php5-phar \
        php5-pgsql \
        php5-redis@testing \
        php5-sqlite3 \
        php5-ssh2@testing \
        php5-xdebug \
        php5-xmlrpc \
        php5-xsl \
        php5-zip && \
```

ncopa:
@f3l1x yes we want drop support for php5-* packages. Are there any chance that you can use php7 instead?

If not, exactly which are the missing packages if you use `alpine:3.5` docker image?

You should try get the things you need into a stable alpine branch. Things like this can happen and will happn in `edge`.

ncopa:
> @vakartel Someone still use testing/php5-* (php5-redis) on their docker images ...

This can happen if you use things from testing. Things that is needed for production needs to  be moved to either community or main.

This is the reason we do stable branches.

kaniini:
In this case, #1061 was blocking QA efforts for weeks, so an immediate solution was needed.  This situation had been stalled for several weeks, so I went ahead and killed off the php5-* packages.

f3l1x:
@ncopa I use 5.6 and also php 7.1 packages. There's no problem with 7.1 at all and neither with 5.6. I've used alpine/edge based distro. 

I've already fixed it by using alpine/3.5 and dropped php5-redis.

https://github.com/dockette/ci

Thank you.

f3l1x:
@vakartel @jirutka I'm not sure if I'm doing something bad, but there is a new name for php binary in v3.5 and edge? I mean `php7`?

vakartel:
@f3l1x 
Both 3.5/community/php7 and edge/testing/php7 cli binaries are php7.
Also edge/testing/php7 provides a symlink /usr/bin/php7 -> /usr/bin/php

f3l1x:
@vakartel I'm using v3.5/community/php7 but there is no php symlink. Is it right? I'm little bit confused, all my docker images have suddenly stopped working.

vakartel:
@f3l1x 
yea, 3.5 php7 cli binary is php7 and have no php link.

PR #1066, opened 2017-03-22
main/ruby: Just remove files if they exist

leitao:
With the new abuild version, ruby is not able to build because it tries
to remove some files that might not exist, returning error.

this patch checks if the files to be removed before removing them.

### DIFF
--- a/main/ruby/APKBUILD
+++ b/main/ruby/APKBUILD
@@ -94,7 +94,7 @@ package() {
 	rm -R "$pkgdir"$_gemdir/cache/*
 
 	# Remove bundled CA certificates.
-	rm -R "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem
+	rm -R "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem || true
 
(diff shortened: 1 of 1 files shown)


main/ruby/APKBUILD
```diff
@@ -94,7 +94,9 @@ package() {
 	rm -R "$pkgdir"$_gemdir/cache/*
 
 	# Remove bundled CA certificates.
-	rm -R "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem
+	if [ -e "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem ]; then
+		rm -R "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem
+	fi
```
jirutka: This condition is unnecessary and also wrong. It’s enough to just add `|| true` after the command that may fail, i.e. `rm -R "$pkgdir"$_rubydir/rubygems/ssl_certs/*.pem || true`.

leitao:
hi @jirutka, thanks for the review. I just changed the patch according to your suggestion.

PR #1069, opened 2017-03-23
Fixes for the packages I maintain

leitao:


leitao:
Jenkins failed because the package does not build for !ppc64le platform.

ncopa:
I'd like to wait with those changes til after 3.6-stable is branched, mostly because it makes it more time consuming to backport security fixes.

http://lists.alpinelinux.org/alpine-devel/5608.html

tmpfile:
@ncopa 3.6-stable is branched, good to merge now?

fabled:
Rebased and applied. Thanks. Interestingly `testing/ndctl` tests are mostly SKIPPED instead of really ran (at least on x86_64).

PR #1070, opened 2017-03-23
community/elasticsearch: update to 5.3.0 / add bash depends

itoffshore:
* also fixes plugin path: 

`ES` expects to find plugins in the same path as the libs. Looking at the `java` command line in `htop` `ES` sets `path.home` a 2nd time for the libs & expects the plugins to also be in this path (which is why there is no configuration for the plugin path in `elasticsearch.yml`).

* `mkdir -p $builddir/modules` in `APKBUILD` also required now or `abuild checksum` fails:

```
musl64 [~/aports/community/elasticsearch]$ abuild checksum
find: ‘/home/stuart/aports/community/elasticsearch/src/elasticsearch-5.2.1/modules’: No such file or directory
```

### DIFF
--- a/community/elasticsearch/APKBUILD
+++ b/community/elasticsearch/APKBUILD
@@ -1,14 +1,15 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=elasticsearch
-pkgver=5.2.1
+pkgver=5.3.0
 pkgrel=0
 pkgdesc="Open Source, Distributed, RESTful Search Engine"
 url="https://www.elastic.co/products/elasticsearch"
 arch="x86 x86_64"
 license="ASL-2.0"
-depends="java-jna-native>=4.1 openjdk8-jre"
+depends="java-jna-native>=4.1 openjdk8-jre bash"
 makedepends=""
+options="!check"
 install="$pkgname.pre-install"
 subpackages="$pkgname-doc"
 source="https://artifacts.elastic.co/downloads/$pkgname/$pkgname-$pkgver.tar.gz
@@ -17,14 +18,6 @@ source="https://artifacts.elastic.co/downloads/$pkgname/$pkgname-$pkgver.tar.gz
 	README.alpine
 	"
 builddir="$srcdir/$pkgname-$pkgver"
-
-default_module=transport-netty4
-_modules=$(find $builddir/modules -mindepth 1 -maxdepth 1 -type d ! -name $default_module -exec basename {} \;)
-for _mod in $_modules; do
-	subpackages="$subpackages $pkgname-$_mod:_${_mod//-/_}"
-	eval "_${_mod//-/_}() { _builtin_module $_mod; }"
-done
-
 _basedir="/usr/share/java/$pkgname"
 
 build() {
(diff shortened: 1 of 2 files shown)


jirutka:
Why bash?

itoffshore:
the following scripts have `bash` headers:
```
root@elastic [~]# head -1 /usr/share/java/elasticsearch/bin/*
==> /usr/share/java/elasticsearch/bin/elasticsearch <==
#!/bin/bash

==> /usr/share/java/elasticsearch/bin/elasticsearch-plugin <==
#!/bin/bash

==> /usr/share/java/elasticsearch/bin/elasticsearch-systemd-pre-exec <==
#!/bin/bash

==> /usr/share/java/elasticsearch/bin/elasticsearch-translog <==
#!/bin/bash

==> /usr/share/java/elasticsearch/bin/elasticsearch.in.sh <==
#!/bin/bash
```
at the moment they will probably work patched to use `/bin/sh` headers - but I thought it would be less chance of breakage in the future if `bash` was a depends.

Or perhaps raise an issue with `ES` to change to `/bin/sh` ?

community/elasticsearch/APKBUILD
```diff
@@ -19,6 +20,7 @@ source="https://artifacts.elastic.co/downloads/$pkgname/$pkgname-$pkgver.tar.gz
 builddir="$srcdir/$pkgname-$pkgver"
 
 default_module=transport-netty4
+mkdir -p $builddir/modules
 _modules=$(find $builddir/modules -mindepth 1 -maxdepth 1 -type d ! -name $default_module -exec basename {} \;)
```
ncopa: does this even work? It does not look good at all.

Remember that other scripts will parse/execute the APKBUILD to figure out buildtime dependencies. I don't think we want create and directories at that point.

Then as second step the `_modules=$(find $builddir/modules .. )` will always run on an empty, newly created dir, so we could simplify with:
````diff
diff --git a/community/elasticsearch/APKBUILD b/community/elasticsearch/APKBUILD
index ecf27f1f8a..47bbb97805 100644
--- a/community/elasticsearch/APKBUILD
+++ b/community/elasticsearch/APKBUILD
@@ -20,13 +20,6 @@ source="https://artifacts.elastic.co/downloads/$pkgname/$pkgname-$pkgver.tar.gz

 default_module=transport-netty4
-mkdir -p $builddir/modules
-_modules=$(find $builddir/modules -mindepth 1 -maxdepth 1 -type d ! -name $default_module -exec basename {} \;)
-for _mod in $_modules; do
-       subpackages="$subpackages $pkgname-$_mod:_${_mod//-/_}"
-       eval "_${_mod//-/_}() { _builtin_module $_mod; }"
-done
-
 _basedir="/usr/share/java/$pkgname"

 build() {
````

and it should still give the exact same result.

Comment (jirutka) [inline community/elasticsearch/APKBUILD]: Yeah, it’s horrible. I didn’t want this to be merged. Actually I’m still thinking about reverting it. :/ @itoffshore basically threw the cleanup around init script I did in previous version away…

Comment (itoffshore) [inline community/elasticsearch/APKBUILD]: the `find` worked & created the subpkgs correctly but I've removed it. Now the module names are added manually as they were before. When the package is updated new modules will need to be checked for manually.

the `initd` was cleaned up because in `ES` now options for `java` are set in `jvm.options` (previously they were not). I thought it was more intuitve to set options for `java` in the one place they are expected to be (& not also have them set in the `initd`)

I've been running `ES` for over a week with `supervise-daemon` with no problems.

**Comment by itoffshore** [comment]
the `find` works without `mkdir` & only runs once with:
```

default_module=transport-netty4
if [ -d "$builddir/modules" ]; then
        _modules=$(find $builddir/modules -mindepth 1 -maxdepth 1 -type d \
                ! -name $default_module -exec basename {} \;)

        for _mod in $_modules; do
                subpackages="$subpackages $pkgname-$_mod:_${_mod//-/_}"
                eval "_${_mod//-/_}() { _builtin_module $_mod; }"
        done
fi

_basedir="/usr/share/java/$pkgname"
```

rebase with this method - or set the module names manually ?

**Comment by itoffshore** [comment]
I rebooted for a kernel update but have had no problems with `elasticsearch 5.2.2`
```
root@elastic [~]# uptime
 17:24:16 up 6 days, 21:08,  load average: 0.38, 0.19, 0.14
```

itoffshore:
module subpkgs removed

community/elasticsearch/APKBUILD
itoffshore: module subpkgs removed

itoffshore:
this `PR` has been outstanding so long I need to redo to resolve the conflict

PR #1071, opened 2017-03-23
Fixing extra packages for ppc64le

leitao:
Enabling new packages to be built on ppc64le.

### DIFF
--- a/community/webkitgtk/APKBUILD
+++ b/community/webkitgtk/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=webkitgtk
 pkgver=2.4.11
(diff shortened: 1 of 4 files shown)


community/webkitgtk/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=webkitgtk
 pkgver=2.4.11
-pkgrel=3
+pkgrel=4
```
ncopa: I dont think we need bump pkgrel in this case as it only affects ppc64le which is not yet built. There are no changes that should affect any previously built packages.

ncopa:
i rebased the patches to not bump the pkgrel since none of the changes affects already built packages

PR #1079, opened 2017-03-23
community/influxdb: move from testing

stwa:
Stable for ~4 months, move to community.


jirutka:
Maybe stable, but the abuild is bad. Why do you git clone the repository instead of fetching tarball?

stwa:
This has already been discussed in https://github.com/alpinelinux/aports/pull/557, especially https://github.com/alpinelinux/aports/pull/557#issuecomment-264798286

7heo:
I would argue that one core developer merging one PR once doesn't really mean that a consensus was reached. As @jirutka noted, we should not be using git repositories as sources for our packages. Even more so when those repositories do not have a strict correlation between releases and the master branch.

stwa:
My point was not that there is a consensus because a core developer earlier merged my PR, the problem is upstream which uses build scripts depending hardly on information from git branches and tags. I totally agree that git clone is a bad thing for package sources, so I wonder how I could improve this APKBUILD. Sticking with a script provided by upstream in this case seems more stable to me than reimplementing the whole build process. My other (dirty) solution was to patch the build script, which also wasn't accepted.

7heo:
Ideally, we should correct upstream. In practice, we need a "good reason"™.

stwa:
Yes, I would like to see this fixed upstream, too! There is already an issue, see https://github.com/influxdata/influxdb/issues/7696
I thought that "testing" in alpine means that software in this repo might be unstable, not that the APKBUILD isn't ideal.

kaniini:
testing means the package is unsupported and will not be included in a release distribution, for whatever reason.

in order to leave testing, a package needs to be known good and the package itself needs to appropriately follow policy.  we intend for reproducible builds (where the output is the same every time) to become a policy requirement after 3.6.  we also wish to avoid network accesses by builders, and may block network access on builders in the future, too.

stwa:
Thank you for the clarification!

So, the question was:
> Why do you git clone the repository instead of fetching tarball?
The answer was in the first link I provided including an alternative but dirty solution.

What do you want me to do? If this package doesn't comply to the policy and you see no way this can be fixed just say it straight and close this PR.

andypost:
Looks it stuck on update in https://github.com/alpinelinux/aports/pull/3034

clandmeter:
@stwa with the recent changes we could probably move it to community.
Can you verify its working correctly?

stwa:
@clandmeter not yet, I haven't upgraded to 1.5.2 - just updated my local branches and pushed them to my repo, hence the update.

stwa:
@clandmeter looks good, I already was on 1.5.0 on my custom repo, 1.5.2 also works correctly.

clandmeter:
applied in 11ae2ed9b20cef67fb2437001585d1e901921d32

PR #1085, opened 2017-03-24
main/iptables: Does not install libiptc.a if it does not exists

leitao:
Currently iptables is not being built on ppc64le because it tries to
install libiptc.a, that was not generated.

This is breaking only on abuild version 3. On version 2, it is
still working fine.

This patch simply checks if the file exists before installing it.

### DIFF
--- a/main/iptables/APKBUILD
+++ b/main/iptables/APKBUILD
@@ -66,7 +66,9 @@ package() {
 	install -m644 include/iptables.h include/ip6tables.h \
 		"$pkgdir"/usr/include/ || return 1
 	install include/libiptc/*.h "$pkgdir"/usr/include/libiptc/
-	install -m644 libiptc/libiptc.a "$pkgdir"/usr/lib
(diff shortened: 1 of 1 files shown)


ncopa:
i think we should simple just remove the line:
````diff
diff --git a/main/iptables/APKBUILD b/main/iptables/APKBUILD
index 12a8a95a61..4fd5168674 100644
--- a/main/iptables/APKBUILD
+++ b/main/iptables/APKBUILD
@@ -66,7 +66,6 @@ package() {
        install -m644 include/iptables.h include/ip6tables.h \
                "$pkgdir"/usr/include/ || return 1
        install include/libiptc/*.h "$pkgdir"/usr/include/libiptc/
-       install -m644 libiptc/libiptc.a "$pkgdir"/usr/lib
        install -m755 "$startdir"/iptables.initd "$pkgdir"/etc/init.d/iptables
        install -m644 "$startdir"/iptables.confd "$pkgdir"/etc/conf.d/iptables
 }
````

PR #1096, opened 2017-03-25
main/tinc: Create intermediate directories

leitao:
Currently tinc tries to create $pkgdir/etc/tinc without creating
$pkgdir/etc first, which causes a failure.

This patch allows the creation of the directory chain if it does not
exists.

### DIFF
--- a/main/tinc/APKBUILD
+++ b/main/tinc/APKBUILD
@@ -47,7 +47,7 @@ package() {
 	cd "$_builddir"
 	make DESTDIR="$pkgdir" install || return 1
 
-	mkdir "$pkgdir"/etc/tinc
+	mkdir -p "$pkgdir"/etc/tinc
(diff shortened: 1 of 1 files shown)


main/tinc/APKBUILD
```diff
@@ -47,7 +47,7 @@ package() {
 	cd "$_builddir"
 	make DESTDIR="$pkgdir" install || return 1
 
-	mkdir "$pkgdir"/etc/tinc
+	mkdir -p "$pkgdir"/etc/tinc
```
jirutka: This `mkdir` is not be needed at all, `install -D` automatically creates intermediate repositories.

jirutka:
I’ve fixed it in c38cf9bb99d8ea8c3977c2e00f87b846e9b3afb3.

leitao:
Excellent. Thanks!

PR #1098, opened 2017-03-25
main/taskd: Avoid using file owners during packaging

leitao:
Currently taskd FTBFS if user 'taskd' does not exist in the builder.
Since file ownership shouldn't be done externally of the system,
due to the fact that users and groups IDs are unique to system, this
patch removes this option.

A possible solution, if this is required, would be to change the
ownership during the install process (postinst).


main/taskd/APKBUILD
```diff
@@ -35,7 +35,7 @@ pki() {
 package() {
 	cd "$builddir"
 	make DESTDIR="$pkgdir" install || return 1
-	install -dm700 -o taskd -g taskd "$pkgdir"/var/lib/$pkgname
```
jirutka: This is correct. However, `pkgusers="taskd"` and `pkggroups="taskd"` must be defined in the abuild.

leitao:
Ok, fixing it.

leitao:
This should be fixed now.

PR #1101, opened 2017-03-25
main/perl-test-output: Fix make dependencies

leitao:
This package depends on perl-test-tester which does not exist
on Alpine anymore, thus, this package is not able to build due
to lack of dependency.

I understand that the natural replacement for perl-test-tester is
perl-test-simple, although I am not a perl expert.


jirutka:
@fcolista, could you please review this change? I don’t know Perl libraries.

fabled:
Fixed in git by earlier commit.

PR #1103, opened 2017-03-26
aports/mariadb: remove glibc dep on ppc

gromero:
Remove dependency on glibc by using gcc builtin function and no glibc
wrappers.

Currently there are no surrogates in musl for:

  __ppc_get_timebase()
  __ppc_set_ppr_low()
  __ppc_set_ppr_med()

however glibc __ppc_get_timebase() is just a wrapper for GCC builtin
__builtin_get_timebase() available since GCC 4.8 [1], so assuming that
aports on ppc64le will never be built using GCC < 4.8 we can just
switch directly to the GCC builtin function.

Also __ppc_set_ppr_{low,med}() are not available on musl but both
are simple glibc wrappers on a single asm instruction, hence there
is no harm in using asm directly instead. Actually, using asm
directly was the first solution adopted in MariaDB [2].

[1] https://goo.gl/jxLV6O
[2] https://goo.gl/9bjuVC

### DIFF
--- a/main/mariadb/APKBUILD
+++ b/main/mariadb/APKBUILD
@@ -21,6 +21,7 @@ source="https://downloads.mariadb.org/interstitial/mariadb-$pkgver/source/mariad
 	$pkgname.initd
 	fix-mysql-install-db-path.patch
 	fix-ucontext-check.patch
+	ppc_remove_glibc_dep.patch
 	"
 
 # secfixes:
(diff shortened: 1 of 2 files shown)


jirutka:
I’ve copied your description into the patch, renamed it and merged as 17f250dd159dbb6a8c5abf794532b434c0e7ca08. Thanks for your contribution!

PR #1105, opened 2017-03-26
main/{iptables,libnetfilter_acct}: upgrade

misery:


jirutka:
There’s a change in version of `libxtables.so` provided by the iptables package. The following packages depends on `so:libxtables.so.11` and so needs to be rebuilt:

* connman
* iproute2
* xtables-addons

PR #1111, opened 2017-03-27
main/ncftp: Adjust package checksums

rgdoliveira:
Adjust ncftp checksum to match the last available version and
remove a patch that is already applied in the new ncftp version.


jirutka:
@ncopa, there’s yet another suspicious change of upstream tarball, could you please verify it?

ncopa:
seems like there have been significant changes. We should ask upstream what happened here.
https://gist.github.com/ncopa/83233e841957e9f9324960f65fcc8a38

jirutka:
7k lines long diff!? 🙀 I’m really very curious what has happened here.

ncopa:
There are two different dates on the version string so it seems to be intentional:
https://gist.github.com/ncopa/83233e841957e9f9324960f65fcc8a38#file-ncftp-diff-L5202

The proper fix is that upstream makes a `3.2.7` release of the `LibNcFTP 3.2.6 (November 12, 2016)` version.

jirutka:
> There are two different dates on the version string so it seems to be intentional.

Then the author is total idiot and I suggest to remove this aport and ban all software from this author. 😠

rgdoliveira:
fix for this change is already applied. Closing it...

PR #1112, opened 2017-03-27
testing/etcd: subpackage etcdctl

errm:
etcdctl can be a useful utility to have
without etcd itself, when connecting to
a remote cluster, or from within a container.

Equaly it might not be needed when installing
the server.


jirutka:
Please bump pkgrel to trigger package rebuild.

errm:
@jirutka pkgrel bumped ...

PR #1113, opened 2017-03-27
main/smokeping: Add perl in makedepends

leitao:
smokeping fails to build on a clean build machine due to the lack
of perl-dev package, which provides pod2man. Adding this package
as a make-dependency.

### DIFF
--- a/main/smokeping/APKBUILD
+++ b/main/smokeping/APKBUILD
@@ -29,6 +29,7 @@ depends="perl fping rrdtool perl-rrd
 	perl-cgi
 	perl-cgi-fast
 	"
+makedepends="perl-dev perl-try-tiny"
 arch="noarch"
 license="GPL"
 subpackages="$pkgname-doc"


jirutka:
Okay, but why perl-try-tiny?

leitao:
@jirutka, we don't need perl-try-tiny, just updated the commit.

leitao:
In Fact, we need it, and now I remember why. If we don't add perl-try-tiny, we get the following error:

> checking checking for perl module 'LWP'... Can't locate Try/Tiny.pm in @INC (you may need to install the Try::Tiny module) (@INC contains: /usr/thirdparty/lib/perl5 /usr/local/lib/perl5/site_perl /usr/local/share/perl5/site_perl /usr/lib/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl .) at /usr/share/perl5/vendor_perl/LWP/Protocol.pm line 11.

I will update this PR

PR #1114, opened 2017-03-27
main/libgsasl: update config guess before build

rgdoliveira:
updates configure guess file before build to fix build break


main/libgsasl/APKBUILD
```diff
@@ -17,6 +17,7 @@ _builddir="$srcdir"/$pkgname-$pkgver
 
 build() {
 	cd "$_builddir"
+	update_config_guess || return 1
```
jirutka: This does not belong to the build phase, it should be called from `prepare()` function:

```sh

prepare() {
	default_prepare || return 1
	update_config_guess
}
```

PR #1115, opened 2017-03-27
main/libquvi-scripts: update config guess before build

rgdoliveira:
updates configure guess file before build to fix build break


kaniini:
This was already included.

PR #1119, opened 2017-03-27
main/lua-xml: Several changes

leitao:
Several changes to the lua-xml package, that now is named luaxml.
These are the changes:
  * Upgrade to the latest version
  * Change project URL (now on github)
  * Rebased the patch
  * Changed $_builddir to $builddir

### DIFF
--- a/main/lua-xml/APKBUILD
+++ b/main/lua-xml/APKBUILD
@@ -2,10 +2,10 @@
 _luaversions="5.1 5.2"
 pkgname=lua-xml
 _name=LuaXML
-pkgver=130610
-pkgrel=3
+pkgver=1.7.5
+pkgrel=0
 pkgdesc="a minimal set of functions for the processing of XML data in Lua"
-url="http://viremo.eludi.net/LuaXML/index.html"
+url="https://github.com/LuaDist/luaxml"
 arch="all"
 license="MIT"
 depends=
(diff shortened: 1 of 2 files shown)


main/lua-xml/APKBUILD
```diff
@@ -1,11 +1,11 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 _luaversions="5.1 5.2"
-pkgname=lua-xml
+pkgname=luaxml
```
jirutka: All Lua libraries are named as `lua-<name>` or lua<lver>-<name>`, where `<lver>` is Lua version (e.g. lua-xml, lua5.1-xml, …).

main/lua-xml/APKBUILD
```diff
@@ -17,28 +17,28 @@ for _v in $_luaversions; do
...
 done
 
-source="http://viremo.eludi.net/LuaXML/LuaXML_$pkgver.zip
-0001-Fix-error-where-TAG-should-be-xml.TAG.patch"
-
-_builddir="$srcdir"
+source="$pkgname-$pkgver.tar.gz::https://github.com/LuaDist/luaxml/archive/$pkgver.tar.gz
+	0001-Fix-error-where-TAG-should-be-xml.TAG.patch"
 
+builddir="$srcdir"
```
jirutka: This should point to the root of the unpacked package, so  `$srcdir/$pkgname-$pkgver`.

main/lua-xml/APKBUILD
```diff
@@ -17,28 +17,28 @@ for _v in $_luaversions; do
...
 prepare() {
-	cd "$_builddir"
+	cd "$builddir"
 	for i in $source; do
 		case $i in
 		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
 		esac
 	done
```
jirutka: This is unnecessary. Applying of patches is already handled by the `default_prepare` function that is called when no custom `prepare` function is declared. When custom function is declared, just call `default_prepare`, usually as the first command of the function.

main/lua-xml/APKBUILD
```diff
@@ -17,28 +17,28 @@ for _v in $_luaversions; do
...
 
+	dest="$pkgname-$pkgver"
```
jirutka: Custom variables inside functions should be always declared as `local`, if not needed in more functions. Moreover, this is totally unnecessary, just use `buildddir` properly.

leitao:
hi @jirutka. Thanks for your review. I just updated the commit to adjust to your recommendations.

leitao:
hi @jirutka, any feedback about this new version?

jirutka:
Sorry, I forgot to this patch, I'll look at it tomorrow.

jirutka:
The current state of LuaXML is really a mess, there are few forks and it’s not clear what is actually the original LuaXML, since the upstream source is down. However, it seems that https://github.com/LuaDist2/luaxml corresponds to the version we currently have and also what is on LuaRocks. There are many changes between this and https://github.com/LuaDist/luaxml, even different versioning scheme.

So I prefer to use https://github.com/LuaDist2/luaxml as new source. I’ve already updated lua-xml in 5a461824f433db3d66099e34f371c4608a93fed0, 6a1e7a93bbc27a044874af176b4cd32f42c69e08 and 6a672cdda3dc2c560540ccf164b1d18ea8105906.

PR #1122, opened 2017-03-28
Move php5 and all depenent packages from main to community

jirutka:
There two main reasons for this change:

* Active support of PHP 5.x ended on January 2017, security support will
  end on December 2018. Packages in the main repository should be
  supported for at least 2 years, this means until first quarter of 2019
  for the upcoming v3.6.

* php7 and its extensions are currently in the community repository, so
  we can't use single abuild for both php5-* and php7-* packages (as we
  do for Python and Lua packages).

This change was suggested by @vakartel (https://github.com/alpinelinux/aports/pull/1061#issuecomment-288981165), approved by @ncopa, @kaniini, and
@jirutka.

/cc @andypost, @clandmeter, @fabled, @rnalrd


jirutka:
Hm, zabbix has failed:

```
checking for SSH2 support... no
configure: error: SSH2 library not found
>>> ERROR: zabbix: all failed
```

rnalrd:
On 28/03/2017 16:08, Jakub Jirutka wrote:
> Hm, zabbix has failed:
> 
> |checking for SSH2 support... no configure: error: SSH2 library not found
>>>> ERROR: zabbix: all failed |
> 

It needs libssh2-dev not libssh2

|_eo

PR #1123, opened 2017-03-28
community/docker: Fixing packaging issues, adding docker-doc pkg

gena01:
* Needed to tweak tini to return not just version but also commit.
  This is needed so docker doesn't complain it can't get version. This
 should also duplicate behavior building from inside a git clone.
* Added docker-doc man pages (sub package)

### DIFF
--- a/community/docker/APKBUILD
+++ b/community/docker/APKBUILD
@@ -1,6 +1,6 @@
 # Maintainer: Eivind Uggedal <eivind@uggedal.com>
 pkgname=docker
-pkgver=17.03.0
+pkgver=17.03.1
 _ver=${pkgver/_/-}-ce
 pkgrel=0
 _gitcommit=v$_ver
@@ -10,7 +10,7 @@ arch="all"
 license="ASL 2.0"
 depends="iptables"
 depends_dev=""
-makedepends="go btrfs-progs-dev bash linux-headers coreutils libseccomp-dev cmake lvm2-dev"
+makedepends="go btrfs-progs-dev bash linux-headers coreutils libseccomp-dev cmake lvm2-dev glide"
 options="!check"
 install="$pkgname.pre-install"
 
@@ -33,6 +33,7 @@ subpackages="
 	$pkgname-fish-completion:fishcomp:noarch
 	$pkgname-zsh-completion:zshcomp:noarch
 	$pkgname-vim:vim:noarch
+	$pkgname-doc
 	"
 source="
 	$pkgname-$pkgver.tar.gz::https://github.com/docker/docker/archive/v$_ver.tar.gz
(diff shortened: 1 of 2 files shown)


community/docker/APKBUILD
```diff
@@ -27,12 +27,14 @@ _runc_ver=a01dafd48bc1c7cc12bdb01206f9fea7dd6feb70
 _containerd_ver=977c511eda0925a723debdc94d09459af49d082a
 _libnetwork_ver=0f534354b813003a754606689722fe253101bc4e
 _tini_ver=949e6facb77383876aeff8a6944dde66b3089574
+_tini_gitver=949e6fa
```
ncopa: I dont like this. we should be able to extract this from `_tini_ver` so we don't need specify same info twice. We also want avoid the situation where `_tini_ver` is updated but `_tini_gitver` is forgotten, which leads to to a tini build which lies about which version it really is.

community/docker/APKBUILD
```diff
@@ -105,12 +107,33 @@ build() {
 	msg "building tini"
 	cd "$_tini_builddir" || return 1
 	cmake . || return 1
+	sed -i "s/TINI\_GIT *\"[^\"]*/TINI_GIT     \" - git.$_tini_gitver/"  tiniConfig.h
```
ncopa: I think we should report this upstream to tini. As I understand the problem here is that docker will check for a specific git revision, the tini build script will execute `git log -n 1  --pretty=format:%h` to get the git revision of tini, but since we are using a tarball generated from the given git commit, the tini build script fails to set the correct version string.

I think a better solution would be to patch tini to allow override the `tini_VERSION_GIT` from the cmake run. something like:
`cmake -Dtini_VERSION_GIT="$(echo $_tini_gitver | cut -b1-7)"`

community/docker/APKBUILD
```diff
@@ -105,12 +107,33 @@ build() {
...
+	rmdir src && mv vendor src
+	mkdir -p src/github.com/docker
+	ln -s `pwd`/../../ src/github.com/docker/docker
+	cd src/github.com/docker/docker
+	GOPATH="$builddir/man/genman" go build -o man/gen-manpages ./man
+
+	msg "Generating man pages"
+	cd "$builddir/man"
+	mkdir -p man1
+	./gen-manpages ./man1
```
ncopa: I get the following error:

````
/usr/bin/abuild: line 1: glide: not found
````

Maybe glide should be in makedepends?

community/docker/APKBUILD
gena01: @ncopa I need to look and see how to extract the first 7 characters from the version string.

jirutka: `${foo:0:7}`

ncopa:
I reported the tini issue upstream: https://github.com/krallin/tini/issues/79

gena01:
@ncopa ideally if they fix this upstream, it would simplify things a lot on our side. 

I'd like to see git version included in alpine tini package that we build. Would make simplifying docker build easier.

community/docker/APKBUILD
gena01: @ncopa updated

PR #1130, opened 2017-03-28
main/mesa: Do not fail on file move

leitao:
Currently mesa is failing to build on ppc64le due to a file that
does not exist, and APKBUILD tries to move it.

Adding some changes in the script to let it pass if the source file
does not exists.


jirutka:
No, this is definitely not a correct solution. If some file does not exist, then probably something was changed in the upstream package and the abuild must be fixed, not just modified to ignore the error. That’s exactly why we (finally) enabled `sh -e`, to avoid silent errors.

Isn’t it already fixed by 8dc250b7cf59e3019d55d963ebe589f6ca71ba79 ?

leitao:
This is fixed by 8dc250b. Closing it!

PR #1148, opened 2017-03-28
community/aha: update to version 0.4.10.5

rgdoliveira:
update to last available version and update checksum


jirutka:
This package has been already updated in #1147.

PR #1153, opened 2017-03-28
comunity/geany-plugins: check if .so exist before moving it

rgdoliveira:
Check if .so exist before trying to move it

### DIFF
--- a/community/geany-plugins/APKBUILD
+++ b/community/geany-plugins/APKBUILD
@@ -86,7 +86,9 @@ _plugin() {
 	pkgdesc="$2"
 	local _d
 	mkdir -p "$subpkgdir"/usr/lib/geany
-	mv "$pkgdir"/usr/lib/geany/$1.so "$subpkgdir"/usr/lib/geany/ || return 1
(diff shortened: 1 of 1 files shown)


jirutka:
Why it doesn’t exist? Some plugins are not built on ppc64le or are never built on any arch, because they are not in the upstream source anymore…?

ncopa:
Seems to happen on x86_64 as well
````
>>> geany-plugins-gproject*: Running split function gproject...
mv: cannot stat '/home/ncopa/aports/community/geany-plugins/pkg/geany-plugins/usr/lib/geany/gproject.so': No such file or directory
>>> ERROR: geany-plugins-gproject*: gproject failed
>>> ERROR: geany-plugins*: prepare_subpackages failed
>>> ERROR: geany-plugins: all failed
````

ncopa:
i think this is fixed with the geany/geany-plugins update

ncopa:
fixed with commit 9bb70c50838ddc33d379e25b41c70688b2051f5e

PR #1159, opened 2017-03-29
main/xf86-video-rendition:  package changes

leitao:


jirutka:
@ncopa @kaniini, what should we do with the following error?

```
>>> xf86-video-rendition*: Stripping binaries
strip: Unable to recognise the format of the input file `./usr/lib/xorg/modules/v20002d.uc'
```

ncopa:
In this case abuild finds ELF files that seems to be some firmware for the GPU or similar.

This is from my x86_64 machine:

````
./usr/lib/xorg/modules/v10002d.uc:               ELF 32-bit MSB executable, *unknown arch 0x3d32* version 1 (SYSV), statically linked, not stripped, with debug_info
./usr/lib/xorg/modules/drivers/rendition_drv.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped, with debug_info
./usr/lib/xorg/modules/v20002d.uc:               ELF 32-bit MSB executable, *unknown arch 0x3d32* version 1 (SYSV), statically linked, not stripped, with debug_info
````
I think we have the following options:
- mark the package options="!strip" and strip manually in APKBUILD
- make the automatic stripping ignore strip errors
- try fix the automatic stripping to search for ELF files that corresponds with current arch

I'm leaning to the first option.

ncopa:
we have similar problem in `valgrind`

leitao:
Hi @ncopa, I just implemented your first option. How does it sound?

PR #1164, opened 2017-03-29
main/cmocka: Fix FTBFS on ppc64le

leitao:
Currently cmocka fails to build on ppc64le due to a redefinition
for uintptr_t. This is declared at Musl and redeclared by cmocka.

This patch check if uintptr_t was not defined by MUSL before
re-defining it.

This patch also skips the tests for ppc64le, mainly because there
is a single test that fails (segfault). It should be analyzed later.

### DIFF
--- a/main/cmocka/APKBUILD
+++ b/main/cmocka/APKBUILD
@@ -10,12 +10,18 @@ license="ASL-2.0"
 depends=""
 makedepends="cmake"
 subpackages="$pkgname-dev"
-source="https://cmocka.org/files/${pkgver%.*}/$pkgname-$pkgver.tar.xz"
+source="https://cmocka.org/files/${pkgver%.*}/$pkgname-$pkgver.tar.xz
+	musl_uintptr.patch"
 builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
 	mkdir -p "$builddir"/build || return 1
 	cd "$builddir"/build
+
+	# Just one single test (customer_database_test) is breaking on ppc64le
+	if [ "$CARCH" = "ppc64le" ] ; then
+		sed -i '/customer_database_test/d' "$builddir"/example/CMakeLists.txt
+	fi
 
 	cmake .. \
 		-DCMAKE_INSTALL_PREFIX=/usr \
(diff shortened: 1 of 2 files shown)


main/cmocka/APKBUILD
```diff
@@ -28,13 +29,15 @@ build() {
 }
 
 check() {
-	make -C "$builddir"/build test
+	# Just one single test (customer_database_test) is breaking on ppc64le
+	if [ "$CARCH" != "ppc64le" ] ; then
```
jirutka: If just a single test fails, then it’d be better to just remove this test on ppc64le then skipping all tests.

leitao: Yes, I agree, but, in order to be able to disable this test, I will need to patch examples/CMakeLists.txt. On abuild, I don't think I can apply a patch just for one platform, thus, I would need to hack heavily change CMakeList.txt to detect the platform, and remove one test if the platform is ppc64le.

jirutka: Hm, what about `sed -i '/customer_database_test/d' example/CMakeLists.txt`?

leitao:
You are smart @jirutka, and your suggestion worked. How does this new patch sound for you?

jirutka:
Thanks 😊 It looks good, gonna merge it.

jirutka:
Just one thing, this sed should be in `prepare`, not `build`.

clandmeter:
This also fixed the issue on aarch64 but also fails on multiple tests.
Did this get reported upstream?

leitao:
I didn't report it upstream, since I didn't dig further. I should be able to do so after the freeze. On ppc64le the test is segfaulting, but not on amd64.

clandmeter:
Do you mean no test fail on aarch64?

I am getting

```
91% tests passed, 4 tests failed out of 45

Total Test time (real) =   0.11 sec

The following tests FAILED:
         38 - calculator_test (Failed)
         42 - customer_database_test (Failed)
         44 - product_database_test (Failed)
         45 - waiter_test_wrap (Failed)
Errors while running CTest
```

PR #1169, opened 2017-03-30
community/openjdk7: fix build on PPC when musl is used instead of glibc

gromero:
Musl on Power does not define regs member as a pt_regs pointer type,
hence it's necessary to use member gp_regs instead.


ncopa:
this only affects ppc64le and the job is too much for travis.

PR #1171, opened 2017-03-30
main/freeradius: Bump version to 3.0.13

leitao:
Current version (3.0.12) it not available in the $source URL anymore,
thus, cauing a FTBFS. This patch bumpts version to 3.0.13.

This new version is crashing to be built with LibreSSL, thus fixing
it also. This patch was already sent to upstream and accepted on version
4. I also needed to rebase freeradius-305-default-config.patch.


leitao:
Just found that problem was already fixed by 6b68ea0dba2ea365ca77a7d647349f8aaafad18c

PR #1172, opened 2017-03-30
main/less: upgrade to 487

misery:


rnalrd:
@jirutka there was earlier patch than this PR in patchwork, which I marked superseded @vakartel

PR #1174, opened 2017-03-30
main/apg: Getting package from a mirror

leitao:
Official apg website is not online, grabbing the package from a
mirror temporarily. This patch should be reverted when the address
www.adel.nursat.kz is being resolved.


rgdoliveira:
btw. I sent an email to Adel, who is the project maintainer, but I got no answer.

leitao:
Hi @jirutka, any feedback about this PR?

PR #1175, opened 2017-03-30
main/luajit: enable support for ppc64le

gromero:
Enable support for ppc64le and change _builddir to
builddir so default_prepare can apply the patch by
itself.


gromero:
@jirutka That patch is not yet upstream, but already in debian/experimental.

PR #1177, opened 2017-03-31
main/qemu: remove strip action from APKBUILD

rgdoliveira:
APKBUILD was stripping qemu binaries and also a .img file and this was making
the build fail. As qemu Makefile already strip the generated binaries, APKBUILD
does not need to strip them again, so this patch removes the strip from APKBUILD.


jirutka:
This doesn't affect only PPC, but all platforms, right? Then pkgrel should be bumped to trigger rebuild.

PR #1180, opened 2017-03-31
main/valgrind: fix when stripping binaries

rgdoliveira:
APKBUILD was trying to strip a binary (no_op_client_for_valgrind)
that was deprecated and no longer exists. This patch removes this binary
from strip list and add two valgrind binaries (valgrind-di-server and
vgdb) that were not being stripped.


ncopa:
we need bump `pkgrel` in this case because the built package will be different. 2 more stripped binaries.

PR #1181, opened 2017-03-31
main/linux-vanilla: improve config file for ppc64le

leitao:


leitao:
Any feedback about this PR?

ncopa:
It was me who asked @leitao to bring the config closer to alpine's other kernel configs. I think we can pull this and tweak it as we go.

PR #1182, opened 2017-03-31
main/aiccu: remove from alpine build

rgdoliveira:
removes aiccu from alpine build because of the project sunset.
On 2017-06-06 they will shut down services, and close the sunsetting
project on 2017-07-01.

For more information: https://www.sixxs.net/sunset/


kaniini:
I think in reality we should just drop the package.

rgdoliveira:
@kaniini as you suggested, I sent a new version of the patch dropping the package.

PR #1183, opened 2017-04-01
testing/cockroach: new aport

benesch:
https://www.cockroachlabs.com/
An open source, survivable, strongly-consistent SQL database.

This is a rebase of #358 that uses the latest CockroachDB beta, which shipped with numerous packageability improvements. Even with these improvements, the recipe still requires a few patches to convince Cockroach's C dependencies to build on Alpine Linux. We'll hopefully have these patches integrated upstream by next week's beta release.

Please let me know if any of these files doesn't conform to Alpine standards. I'm unfamiliar with Alpine, so I've left most of @nathanejohnson's original work changed.

@nathanejohnson, thanks again for all your work! I've left you listed as the contributor and the maintainer. Let me know if you'd rather I take responsibility.

Also worth noting that this PR is blocked on having Go 1.8 available (#983). I wanted to put this up for review in the meantime, though.

+cc @tamird @knz

---
This change is [Reviewable](https://reviewable.io/reviews/alpinelinux/aports/1183)

### DIFF
--- a/testing/cockroach/APKBUILD
+++ b/testing/cockroach/APKBUILD
@@ -1,0 +1,59 @@
+# Contributor: Nathan Johnson <nathan@nathanjohnson.info>
+# Maintainer: Nathan Johnson <nathan@nathanjohnson.info>
+pkgname=cockroach
+pkgver=1.0
+pkgrel=0
+pkgdesc="Distributed SQL database"
+url="https://www.cockroachlabs.com"
+arch="all"
+license="ASL 2.0"
+depends=""
+depends_dev=""
+makedepends="$depends_dev go go-tools cmake xz bash linux-headers"
+install="$pkgname.pre-install"
+subpackages="$pkgname-doc"
+pkgusers="cockroach"
+pkgroups="cockroach"
+source="$pkgname-$pkgver.tar.gz::https://binaries.cockroachdb.com/$pkgname-v$pkgver.src.tgz
+	$pkgname.initd
+	$pkgname.confd"
+builddir="$srcdir/$pkgname-v$pkgver"
+options="!strip"
+
+build() {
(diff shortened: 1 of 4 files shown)


kaniini:
hold until go 1.8 as per comment

nathanejohnson:
@benesch Sorry for taking so long to get back to you.  This looks good to me!  Great work, and thanks for picking up the torch!

tamird:
I believe all the patches can be removed now, although 1.8 is not merged yet anyway https://github.com/alpinelinux/aports/pull/983.

benesch:
1.8 has landed! I've updated this aport to install CockroachDB 1.0, which requires no patches. I've also removed the logrotate config, because Cockroach will rotate logs automatically.

tmpfile:
@benesch abuild now runs with `set -e` as default, so we don't need all those ` || return 1` statements any more. They should be removed.
And now it also only checks for `sha512sum` so the other hashes should be removed too.
LGTM otherwise.

TBK:
@benesch Would it be possible to update the PR?
I really would like to see this merged :)

benesch:
@clandmeter I assume it's still very close to working. There was a CI failure back in May of last year that I never got around to fixing, but it must have been something minor. I was able to build and install cockroach manually on a local test machine.

I'm afraid I haven't had the time to bring this back up-to-date and I've mostly paged out the details of building for Alpine. Perhaps I'll be able to find some time this week, though.

> After scanning the makefile it seems this fetches go sources somehow? are all sources locked to a specific version/tag/hash?

We "vendor" (snapshot and commit) all of our Go dependencies (github.com/cockroachdb/vendored) and include them in the source tarball, so the build system doesn't need to dynamically fetch any code at compile time.

PR #1189, opened 2017-04-02
main/nano: fix regex syntax 'Invalid character class name'

itoffshore:
add `--with-wordbounds` to `configure` to fix bad regexes:

https://lists.gnu.org/archive/html/help-nano/2013-01/msg00001.html

See https://savannah.gnu.org/bugs/?50705

`TL;DR:` seems like a bug in configure which ends up wrongly assuming it
needs to use an alternative word-boundary regex syntax `[[:<:]]` - which
the resulting binary doesn't actually support, and so it spawns a lot of
errors while trying to load and parse color syntax files.

As a workaround until it's resolved upstream, adding `--with-wordbounds`
forces it to use a regex syntax which the binary does support, and color
syntax files are now loaded and used correctly and without errors.

### DIFF
--- a/main/nano/APKBUILD
+++ b/main/nano/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=nano
 pkgver=2.8.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Text Editor. GNU nano is designed to be a free replacement for the Pico text editor."
 url="http://www.nano-editor.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


avih:
Oh, you beat me at #1190 by 10 minutes :)

I also filed a bug at https://savannah.gnu.org/bugs/?50705 describing the issue and its causes.

It might be worth adding reference to the bug at the commit message, and remove `--with-wordbounds` when (if?) the issue is resolved upstream. The reference to the mailing list at this commit message is 4 years old and doesn't state that `--with-wordbounds` is the way to override it.

Alternatively, the commit message at #1190 already does that ;)

itoffshore:
@avih - updated the commit message ;o)

avih:
> @avih - updated the commit message ;o)

It does appear so :)

PR #1190, opened 2017-04-02
main/nano: fix errors while parsing color syntax files

avih:
See https://savannah.gnu.org/bugs/?50705

TL;DR: seems like a bug in configure which ends up wrongly assuming it
needs to use an alternative word-boundary regex syntax [[:<:]] - which
the resulting binary doesn't actually support, and so it spawns a lot of
errors while trying to load and parse color syntax files.

As a workaround until it's resolved upstream, adding --with-wordbounds
forces it to use a regex syntax which the binary does support, and color
syntax files are now loaded and used correctly and without errors.


ncopa:
duplicate of #1189

PR #1193, opened 2017-04-02
main/postgresql: build with GSSAPI support

qfire:


jirutka:
Unfortunately enabling GSSAPI does not produce additional shared library that can be split into a subpackage. Enabling this flag would add dependency on libgssapi_krb5.so (krb5-libs) into postgresql and libpq packages for all users. Most of the users don’t need any dependency for Kerberos, so it would just add additional unnecessary bloat into their system.

The only thing we can do is to basically build PostgreSQL twice and create packages with and without GSSAPI. However, this would create another problems and I don’t think that it’s worth in this case.

Where do you need GSSAPI support? In the PostgreSQL server or just client library (libpq)?

qfire:
I'm currently using it for both server and client purposes.

I understand the desire to keep install size down, but we're talking about a fairly heavy database here. Another megabyte seems like a small thing when compared to the amount of data in the database itself, especially considering the convenience it provides to environments that use kerberos/AD extensively.

Also, note that the postgresql packages are already compiled with LDAP authentication which is less secure than kerberos where the password is actually transmitted to the database server vs kerberos where only a temporary ticket is used instead.  Using kerberos provides one less place and two networks (from client to database server and from database server to LDAP server) where an account's password has to pass through.

ncopa:
> @qfire Another megabyte seems like a small thing when compared to the amount of data in the database itself, 

Problem is when you have a tiny client which accesses the huge database remotely, without needing the huge data storage locally.

| package | size (kb) |
| ----------- | ------ |
| libpq     | 176 |
| libldap | 612 |
| krb5-libs |  1748 |

adding the krb5 dependency will more than double the required size for libpq.

However, the security benefits may be worth it.

I wonder if we could reduce the needed dependencies:

````shell
$ lddtree -l /usr/lib/libgssapi_krb5.so.2 | xargs du
0       /usr/lib/libgssapi_krb5.so.2
264     /usr/lib/libgssapi_krb5.so.2.2
0       /usr/lib/libkrb5.so.3
784     /usr/lib/libkrb5.so.3.3
0       /lib/libkeyutils.so.1
16      /lib/libkeyutils.so.1.5
0       /usr/lib/libk5crypto.so.3
92      /usr/lib/libk5crypto.so.3.1
0       /lib/libcrypto.so.38
1692    /lib/libcrypto.so.38.0.0
0       /lib/libcom_err.so.2
16      /lib/libcom_err.so.2.1
0       /usr/lib/libkrb5support.so.0
44      /usr/lib/libkrb5support.so.0.1
0       /lib/libc.musl-x86_64.so.1
552     /lib/ld-musl-x86_64.so.1
````

If we had libgssapi_krb5.so.2, libkrb5.so.3.3, libk5crypto.so.3 and libkrb5support.so.0.1 in separate package(s), we'd reduce the extra needed size to ~1.1MB?

jirutka:
I really doubt that it's worth it. Kerberos is a complex beast used only in enterprise, I bet that 99 % of users will not ever need it in PostgreSQL. I'd be more favour of creating another libpq variant then adding this dependency to the current package.

PR #1200, opened 2017-04-04
main/mosh: fix build with muscl on ppc64le

rgdoliveira:
mosh was breaking when building in ppc64le using musl, because ioctl() is defined
as ioctl(int, int) in musl and mosh is using TIOCSWINSZ macro as parameter. This was
triggering a gcc warning and make the build fail.

This patch does an explicit integer conversion in TIOCSWINSZ, as no bits get
lost.


jirutka:
Please next time add the description also into the patch itself (as I did in 17ee07c9a8e698ee9c06a435eddcccae27ae1bbd), so one can read what is the patch doing and why without need to query git log.

PR #1204, opened 2017-04-04
community/pdns: init crashed process tracking

brimston3:
Start via start-stop-daemon to allow init to catch crashed procs. Without this change a killed process does not show as crashed in rc-status.

### DIFF
--- a/community/pdns/pdns.initd
+++ b/community/pdns/pdns.initd
@@ -16,8 +16,10 @@ PDNS_CONFIG=""
 if [ -n "${PDNS_INSTANCE}" ] && [ "${PDNS_INSTANCE}" != "pdns" ]
 then
 	PDNS_CONFIG="--config-name=${PDNS_INSTANCE}"
+	PDNS_PIDFILE="/run/pdns-${PDNS_INSTANCE}.pid"
 else
 	PDNS_INSTANCE="default"
+	PDNS_PIDFILE="/run/pdns.pid"
 fi
 
 depend() {
@@ -27,7 +29,8 @@ depend() {
 
 start() {
 	ebegin "Starting PowerDNS (${PDNS_INSTANCE})"
-	${daemon} \
+	start-stop-daemon --start --exec ${daemon} \
+		--pidfile ${PDNS_PIDFILE} -- \
 		${PDNS_CONFIG} \
 		--daemon=yes \
 		--guardian=yes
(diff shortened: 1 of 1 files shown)


Shizmob:
IMO, this script should be moved towards the standard OpenRC method of using `command=`, `command_args=` and `pidfile=`. (Ideally, a flag to foreground the daemon as well (`--daemon=no`?) and setting `command_background=yes`). That way the `start()`, `stop()` and `monitor()` functions can be removed entirely, and the script would be more generic for a future without `start-stop-daemon` (e.g. proper supervision, which is at least partially implemented in OpenRC).

brimston3:
I'm not able to figure out how to implement the graceful shutdown command `/usr/bin/pdns_control [--config-name=XYZ] quit` with the requested change.

What is the equivalent to Service.ExecStop from systemd?

Shizmob:
As far as I know, there is no direct equivalent except for simply overriding `stop()`, of course.

However, looking at their [shipped systemd service](https://github.com/PowerDNS/pdns/blob/b7eeb5059d756f93a8858f5234800e3c7459b382/pdns/pdns.service.in) it seems it is not used there either for shutting the service down. This leads me to believe that `pdns_control` is not actually needed as long as the service manager simply sends a correct signal to the process. I'd suggest doing something like this:

```sh
command=/usr/sbin/pdns_server
command_args="--guardian=no --daemon=no $PDNS_OPTS"
command_background=yes
```

brimston3:
Modified to use `command=`, etc. variables, removing `start-stop-daemon`.

Shizmob:
You can set set pidfile to `"/run/$RC_SVCNAME.pid"` unconditionally, as the instance is embedded into the service name. The rest looks good to me, good work!

brimston3:
pdns will always format the pid filename as `pdns.pid` for the default instance or `pdns-${PDNS_INSTANCE}.pid` for subsequent instances. If I understand correctly, the script currently uses '.' and '-' to split the instance name from the script name. If only '-' is allowed, then your suggestion works correctly, otherwise it should probably be conditional like this.

Shizmob:
The pidfile specified is actually created by OpenRC this case, not `pdns`. If you just want to deduplicate pidfiles (since you'd have both `/run/pdns-foo.pid` and `/run/pdns.foo.pid`), I'd suggest something like this then: `pidfile="/run/${RC_SVCNAME/-/.}.pid"`.

jirutka:
Merged in 3a679fc9e1d34deefe4f4d2d380b552f368b2a8b, thanks for your contribution!

PR #1206, opened 2017-04-04
testing/opensmtpd-extras: new aport

Shizmob:
This aport packages the standard OpenSMTPD extras, providing it
with various filters, queues, schedulers, and tables.

### DIFF
--- a/testing/opensmtpd-extras/APKBUILD
+++ b/testing/opensmtpd-extras/APKBUILD
@@ -1,0 +1,92 @@
+# Contributor: Shiz <hi@shiz.me>
+# Maintainer: Shiz <hi@shiz.me>
+pkgname=opensmtpd-extras
+pkgver=201703132115
+pkgrel=0
+pkgdesc="OpenSMTPD addons"
+url="https://opensmtpd.org/"
+arch="all"
+license="ISC"
+depends=""
+_depends_filter_monkey="libressl"
+_depends_queue_python="python2"
+_depends_scheduler_python="python2"
+_depends_table_mysql="mariadb-client-libs"
+_depends_table_postgres="libpq"
+_depends_table_python="python2"
+_depends_table_redis="hiredis"
+_depends_table_sqlite="sqlite-libs"
+makedepends="libressl-dev libevent-dev python2-dev mariadb-dev postgresql-dev hiredis-dev sqlite-dev"
+subpackages="$pkgname-doc"
+source="https://www.opensmtpd.org/archives/opensmtpd-extras-$pkgver.tar.gz
+	remove-decls.patch"
+builddir="$srcdir/opensmtpd-extras-$pkgver"
+
+_extras="
+	filter-monkey
+	filter-stub
+	filter-trace
+	filter-void
+	queue-null
+	queue-python
+	queue-ram
+	queue-stub
+	scheduler-python
+	scheduler-ram
+	scheduler-stub
+	table-ldap
+	table-mysql
+	table-passwd
+	table-postgres
+	table-python
+	table-redis
+	table-socketmap
+	table-sqlite
+	table-stub
+"
+
+for _extra in $_extras; do
+	subpackages="$subpackages opensmtpd-$_extra:_package_extra"
+	depends="$depends opensmtpd-$_extra"
+done
+
+build() {
+	cd "$builddir"
+	local with_extras
+	local extra; for extra in $_extras; do
+		with_extras="$with_extras --with-$extra"
+	done
+
+	./configure \
+		--prefix=/usr \
+		--sysconfdir=/etc/smtpd \
+		--mandir=/usr/share/man \
+		--libexecdir=/usr/lib \
+		--with-pie \
+		--with-mantype=man \
+		$with_extras
+	# the Redis table expects <hiredis.h> without subdirectory, so indulge it
+	make REDIS_CPPFLAGS='-I /usr/include/hiredis'
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install || return 1
+}
+
+_package_extra() {
+	local extraname="${subpkgname#opensmtpd-}"
+
+	pkgdesc="OpenSMTPD $extraname addon"
+	depends="$(eval "echo \$_depends_${extraname//-/_}")"
+
+	mkdir -p "$subpkgdir"/usr/lib/opensmtpd
+	mv "$pkgdir"/usr/lib/opensmtpd/$extraname "$subpkgdir"/usr/lib/opensmtpd
+}
+
+md5sums="a6fd3238ca61d6f48dfd640e5ba9f816  opensmtpd-extras-201703132115.tar.gz
+d7be474d22719470a5e013710d8f32e1  remove-decls.patch"
+sha256sums="4672afdd756ef279c0adcfcbf8151d6d8a370d5a628a705967373a1c83ee0ef5  opensmtpd-extras-201703132115.tar.gz
+5305d2c2a2e26b7e104cece736e0834109743cc13e1d504adfa7865afdd576d5  remove-decls.patch"
+sha512sums="d841c63445ca674b368a5aa6012ae90e46b0c31b650067bdcea5badca3d818b1b732be880b2a18421fb39d07291dd413455944b178597bc13ae6ff5c75ac9aed  opensmtpd-extras-201703132115.tar.gz
+36efd3b6cf75728cc8b75b3d9d6bf464d1e949ccfbe6151ed53dbba0f9ee1e50eb61afcca05af302ab359bc9ea1136e7750a15e5f5b824899141298d3060782a  remove-decls.patch"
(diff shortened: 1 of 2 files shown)


testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+# Maintainer: Shiz <hi@shiz.me>
+pkgname=opensmtpd-extras
+pkgver=201703132115
+pkgrel=0
+pkgdesc="OpenSMTPD addons"
+url="https://opensmtpd.org/"
+arch="all"
+license="ISC"
+depends="libevent"
+depends_filter_monkey="libressl"
```
jirutka: All custom global variables (i.e. variables not specified by abuild) should be prefixed with an underscore (i.e. `_depends_filter_monkey`).

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+depends_table_redis="hiredis"
+depends_table_sqlite="sqlite-libs"
+makedepends="libressl-dev libevent-dev"
+makedepends_queue_python="python2-dev"
+makedepends_scheduler_python="python2-dev"
+makedepends_table_mysql="mariadb-dev"
+makedepends_table_postgres="postgresql-dev"
+makedepends_table_python="python2"
+makedepends_table_redis="hiredis-dev"
+makedepends_table_sqlite="sqlite-dev"
```
jirutka: These `makedepends_*` are quite an overkill. `makedepends` specifies dependencies that should be installed for building the abuild as whole. You can’t build just a single subackage and these dependencies are not written in APKINDEX.

Just put all build dependencies into `makedepends` and don’t overcomplicate it.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+	filter-void
+	table-sqlite
+	table-redis
+	table-socketmap
+	table-passwd
+"
+
+for _extra in $_extras; do
+	_extraident="${_extra/-/_}"
+	_extramakedeps="$(eval "echo \$makedepends_$_extraident")"
```
jirutka: It’s unnecessary to declare yet another global variable `_extraident` just to use it only one line below. You can do:

```diff
- _extramakedeps="$(eval "echo \$makedepends_$_extraident")"
+ _extramakedeps="$(eval "echo \$makedepends_${_extra/-/_}")"
```

Also you probably want `${_extra//-/_}` (replace all occurrences of `-`, not just the first one).

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+	subpackages="$subpackages opensmtpd-$_extra:_package_extra"
+	makedepends="$makedepends $_extramakedeps"
+	case "$_extradocs" in *$_extra*) subpackages="$subpackages opensmtpd-$_extra-doc:_package_extra_doc" ;; esac
```
jirutka: We don’t usually provide separate -doc subpackage for every subpackage, that’s quite an overkill. I’d just declare single -doc subpackage for all docs.

Anyway, please don’t write `case` like this on single line.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+	cd "$builddir"
+	local with_extras
+	local extra; for extra in $_extras; do
+		with_extras="$with_extras --with-$extra"
+	done
+
+	./configure \
+		--prefix=/usr \
+		--sysconfdir=/etc/smtpd \
+		--with-libssl=/usr/include/openssl \
```
jirutka: Is it really needed to explicitly declare location of openssl headers? This is usually autodetected by autohells.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+		--mandir=/usr/share/man \
+		--libexecdir=/usr/lib \
+		--with-pie \
+		--with-mantype=man \
+		$with_extras \
+		|| return 1
```
jirutka: It’s not needed to use `|| return 1` anymore for new aports.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+	make REDIS_CPPFLAGS='-I /usr/include/hiredis' || return 1
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install || return 1
+}
+
+_package_extra() {
+	local extraname; extraname="${subpkgname#opensmtpd-}"
```
jirutka: This is unnecessary verbose, just write `local extraname="${subpkgname#opensmtpd-}"`.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+	local extraident; extraident="${extraname/-/_}"
+	local extradeps; extradeps="$(eval "echo \$depends_$extraident")"
```
jirutka: Same as above, `extraident` is unnecessary:

```diff
-	local extraident; extraident="${extraname/-/_}"
-	local extradeps; extradeps="$(eval "echo \$depends_$extraident")"
+	local extradeps="$(eval "echo \$depends_${extra//-/_}")"
```

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,134 @@
...
+
+	pkgdesc="OpenSMTPD $extraname addon"
+	depends="$depends $extradeps"
+	install_if="$pkgname=$pkgver-r$pkgrel"
```
jirutka: This doesn’t make much sense for me, what behaviour do you expect here?

testing/opensmtpd-extras/remove-decls.patch
```diff
@@ -0,0 +1,15 @@
+--- opensmtpd-extras-201703132115/extras/tables/table-ldap/ber.h
++++ opensmtpd-extras-201703132115-patched/extras/tables/table-ldap/ber.h
```
jirutka: Please replace `opensmtpd-extras-201703132115` with `a` and `opensmtpd-extras-201703132115-patched` with `b` and write a description why is this patch here.

testing/opensmtpd-extras/APKBUILD
Shizmob: I copied it from the OpenSMTPD abuild -- I'll check to be sure.

testing/opensmtpd-extras/APKBUILD
Shizmob: Gotcha!

testing/opensmtpd-extras/APKBUILD
Shizmob: This doesn't work for all shells, though.

Shizmob:
Thanks for the review. Updated!

testing/opensmtpd-extras/APKBUILD
jirutka: `local` itself doesn’t work in all shells, it’s not defined by POSIX. However, all common shells (ash, dash, ZSH, Bash) support it and none of them has problem with `local foo=bar`. I don’t know about any shell that supports `local`, but not with assignment.

testing/opensmtpd-extras/APKBUILD
```diff
@@ -0,0 +1,92 @@
...
+license="ISC"
+depends=""
+_depends_filter_monkey="libressl"
+_depends_queue_python="python2"
+_depends_scheduler_python="python2"
+_depends_table_mysql="mariadb-client-libs"
+_depends_table_postgres="libpq"
+_depends_table_python="python2"
+_depends_table_redis="hiredis"
+_depends_table_sqlite="sqlite-libs"
```
jirutka: According to the [build log](https://travis-ci.org/alpinelinux/aports/builds/218695592#L1268) all these dependencies are auto-detected via dynamic links, so you can remove all these `_depends` and so simplify the abuild even more. :)

```
>>> opensmtpd-filter-monkey*: Tracing dependencies...
    libressl
    so:libc.musl-x86_64.so.1
    so:libcrypto.so.38
    so:libevent-2.1.so.6

>>> opensmtpd-queue-python*: Tracing dependencies...
    python2
    so:libc.musl-x86_64.so.1
    so:libcrypto.so.38
    so:libpython2.7.so.1.0

>>> opensmtpd-scheduler-python*: Tracing dependencies...
    python2
    so:libc.musl-x86_64.so.1
    so:libpython2.7.so.1.0

>>> opensmtpd-table-mysql*: Tracing dependencies...
    mariadb-client-libs
    so:libc.musl-x86_64.so.1
    so:libmysqlclient.so.18

>>> opensmtpd-table-postgres*: Tracing dependencies...
    libpq
    so:libc.musl-x86_64.so.1
    so:libpq.so.5

>>> opensmtpd-table-python*: Tracing dependencies...
    python2
    so:libc.musl-x86_64.so.1
    so:libpython2.7.so.1.0

>>> opensmtpd-table-redis*: Tracing dependencies...
    hiredis
    so:libc.musl-x86_64.so.1
    so:libhiredis.so.0.13

>>> opensmtpd-table-sqlite*: Tracing dependencies...
    sqlite-libs
    so:libc.musl-x86_64.so.1
    so:libsqlite3.so.0
```

PR #1207, opened 2017-04-04
testing/exim: add crypteq support / fix initd

itoffshore:
* `checkpath` added to `initd` so `exim` starts
* `crypteq` added for hashed password support in flat files


jirutka:
I’m quite nervous from all these:

```
expr: syntax error
sh: 0: unknown operand
```
in the build log… Are they okay?

itoffshore:
the error existed in building version `4.88` too - I've been using `exim` for over a year as a smarthost with no problems

PR #1209, opened 2017-04-05
testing/refpolicy: inital import

tych0:
This is the SELinux policy library, which contains a large number of
interface (and other) definitions for use in SELinux policies.

Signed-off-by: Tycho Andersen <tycho@docker.com>

### DIFF
--- a/testing/refpolicy/APKBUILD
+++ b/testing/refpolicy/APKBUILD
@@ -11,48 +11,52 @@ depends_dev=""
 makedepends="$depends_dev checkpolicy python gawk"
 install=""
 subpackages="$pkgname-doc"
-source="https://raw.githubusercontent.com/wiki/TresysTechnology/refpolicy/files/refpolicy-2.$pkgver.tar.bz2"
+source="https://raw.githubusercontent.com/wiki/TresysTechnology/refpolicy/files/refpolicy-2.$pkgver.tar.bz2
+	Makefile.devel"
 builddir="$srcdir/refpolicy"
 
 # refpolicy config
 monolithic=n
-# This causes various distro-specific policies to be built. Since most SELinux
-# policies are written against redhat-based distros, we set that here.
-distro=redhat
+distro=gentoo
+# unknown perms here means what to do with perms that are unknown to the
+# current userspace, because the kernel version is newer. By default, we deny.
 unknown_perms=deny
 
 # These are somewhat related to what is in the CentOS spec file, although they
 # are slightly differnet in what they install.
-makeCmds() {
+#
+# https://selinuxproject.org/page/NB_RefPolicy#Reference_Policy_Build_Options_-_build.conf
+# are the build options: M{L,C}S_CATS are the number of categories for m{l,c}s policies.
+make_cmds() {
 	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=1024 bare || return 1
 	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=1024 conf || return 1
 }
 
-installCmds() {
-	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=0124 SEMOD_EXP="/usr/bin/semodule_expand -a" base.pp
+install_cmds() {
+	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=1024 SEMOD_EXP="/usr/bin/semodule_expand -a" base.pp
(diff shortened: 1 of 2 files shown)


testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+install=""
+subpackages="$pkgname-doc"
+source="https://raw.githubusercontent.com/wiki/TresysTechnology/refpolicy/files/refpolicy-2.$pkgver.tar.bz2"
+builddir="$srcdir/refpolicy"
+
+# refpolicy config
+monolithic=n
+# This causes various distro-specific policies to be built. Since most SELinux
+# policies are written against redhat-based distros, we set that here.
+distro=redhat
```
kaniini: This seems suspect to me.
There is a lot of things different in Alpine that would not mesh well with Redhat's policy definitions.
I think either `debian` or `gentoo` would be a much closer fit.

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+unknown_perms=deny
```
kaniini: It would probably be good to document what `unknown_perms` is for.

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+
+# These are somewhat related to what is in the CentOS spec file, although they
+# are slightly differnet in what they install.
+makeCmds() {
```
kaniini: nitpick: use lowercase function names please, also below with `installCmds`

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=1024 bare || return 1
```
kaniini: As above, it would probably be good to document what `MLS_CATS` and `MCS_CATS` are.

It would be *nice* to have a knob to allow rebuilding the package with UBAC support.

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=1024 conf || return 1
+}
+
+installCmds() {
+	make UNK_PERMS=$3 NAME=$1 TYPE=$2 DISTRO=$distro UBAC=n DIRECT_INITRC=y MONOLITHIC=$monolithic MLS_CATS=1024 MCS_CATS=0124 SEMOD_EXP="/usr/bin/semodule_expand -a" base.pp
```
kaniini: Here `MCS_CATS` is incorrectly defined as `0124` in many of the commands.

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+}
+
+build() {
+	cd "$builddir"
+	makeCmds targeted mcs $unknown_perms || return 1
+}
+
+package() {
+	cd "$builddir"
+	echo packaging now
```
kaniini: `msg`, not `echo`

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+	installCmds targeted mcs $unknown_perms || return 1
+	mkdir -p $pkgdir/usr/share/selinux/devel || return 1
+	mv "$pkgdir/usr/share/selinux/targeted/include" "$pkgdir/usr/share/selinux/devel/include"
+	cp $startdir/Makefile.devel "$pkgdir/usr/share/selinux/devel/Makefile" || return 1
```
kaniini: Add this to `$sources`, and refer to it from `$srcdir`.  The files in `$startdir` may not be available in all cases.

testing/refpolicy/APKBUILD
```diff
@@ -0,0 +1,58 @@
...
+	cd "$builddir"
+	makeCmds targeted mcs $unknown_perms || return 1
+}
+
+package() {
+	cd "$builddir"
+	echo packaging now
+	installCmds targeted mcs $unknown_perms || return 1
+	mkdir -p $pkgdir/usr/share/selinux/devel || return 1
+	mv "$pkgdir/usr/share/selinux/targeted/include" "$pkgdir/usr/share/selinux/devel/include"
```
kaniini: I'm a little suspect on this.  I think we might want to keep the includes where they are, and just copy them instead.

testing/refpolicy/APKBUILD
tych0: I don't know what they are :)

tych0: But I can find a link to some document, probably.

testing/refpolicy/APKBUILD
tych0: Sure, I can make the switch.

testing/refpolicy/APKBUILD
tych0: Oh, this is leftover from debugging, I can just delete it.

testing/refpolicy/APKBUILD
tych0: It sort of depends, in ubuntu/debian/rhel, there are several packages all built with various policies enabled and disabled, but I think they all install to mostly the same target directories. I only built one set (the devel set) in this package for now, so that's why I did a `mv`. I can make it a `cp -r`, though I suspect this will need to change back when someone starts using this in anger.

PR #1210, opened 2017-04-05
testing/ed: New aport

leitao:
GNU ed is a line-oriented text editor. It is used to create, display,
modify and otherwise manipulate text files, both interactively and via
shell scripts. A restricted version of ed, red, can only edit files in
the current directory and cannot execute shell commands. Ed is the
"standard" text editor in the sense that it is the original editor for
Unix, and thus widely available. For most purposes, however, it is
superseded by full-screen editors such as GNU Emacs or GNU Moe.

This aport depends on an abuild change to understands .tar.lz files.

### DIFF
--- a/testing/ed/APKBUILD
+++ b/testing/ed/APKBUILD
@@ -1,0 +1,32 @@
+# Maintainer: Breno Leitao <breno.leitao@gmail.com>
+pkgname=ed
+pkgver=1.14.2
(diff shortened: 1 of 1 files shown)


tmpfile:
abuild now supports .tar.lz files since https://github.com/alpinelinux/abuild/pull/15 was merged

PR #1212, opened 2017-04-05
testing/ipset-blacklist: new aport

itoffshore:
this `initd` is designed for blacklisting ip addresses / ranges & has
some additional options:

`info =>` List active IP sets / raw iptables rules
`load =>` Load inactive /etc/ipset.blacklist files
`unload =>` Unload / remove specific Blacklist IP set

script options can be reviewed with:

`rc-service ipset.blacklist describe`


kunkku:
In my comment on issue #341, I suggested moving the iptables-related blacklisting functions to a separate package. I did not mean creating an alternative init script for ipset. I would prefer a single ipset init script implementation, which would satisfy the requirements of all stakeholders. This is just my opinion, though.

kunkku:
I do not see my input having been taken into account. Do you think we should merge this anyway?

PR #1218, opened 2017-04-06
main/guile: upgrade to 2.2.0

vakartel:


tmpfile:
This can be closed in favor of https://github.com/alpinelinux/aports/pull/1714 (although we still have to figure out why strip fails there)

PR #1219, opened 2017-04-06
community/psmisc: Fix FTBFS on ppc64le

leitao:
Currently psmisc fails to build on ppc64le because there are some
issues with the package:

 * Usage of macros not included in the header
 * Using lintl (gettext) functions without passing -lintl
 * Not explicilty adding gettext dependencies.

### DIFF
--- a/community/psmisc/APKBUILD
+++ b/community/psmisc/APKBUILD
@@ -6,10 +6,13 @@ pkgdesc="A set of some small useful utilities that use the proc filesystem"
 url="http://psmisc.sourceforge.net/"
 arch="all"
 license="BSD GPL"
-depends=""
-makedepends="ncurses-dev"
+depends="libintl gettext"
+makedepends="gettext-dev ncurses-dev"
 subpackages="$pkgname-doc"
-source="http://downloads.sourceforge.net/psmisc/$pkgname-$pkgver.tar.gz"
+source="http://downloads.sourceforge.net/psmisc/$pkgname-$pkgver.tar.gz
+	peekfd_lintl.patch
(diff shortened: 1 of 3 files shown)


leitao:
Just sent this upstream: 
https://gitlab.com/psmisc/psmisc/merge_requests/12

community/psmisc/peekfd_lintl.patch
```diff
@@ -0,0 +1,30 @@
...
+    not have a explicit link against gettext, causing the following error:
+    
+      peekfd.o: In function `print_version':
+      peekfd.c:(.text+0x160): undefined reference to `libintl_gettext'
+      peekfd.c:(.text+0x18c): undefined reference to `libintl_gettext'
+      peekfd.c:(.text+0x1ac): undefined reference to `libintl_gettext'
+    
+    peekfd depends on libint because of the following macro:
+    
+      #define _(String) gettext (String)
```
kaniini: Why not just make _() macro a noop?

We generally do not want to internationalize CLI tools in Alpine, only GUI software.

community/psmisc/APKBUILD
```diff
@@ -6,10 +6,13 @@ pkgdesc="A set of some small useful utilities that use the proc filesystem"
 url="http://psmisc.sourceforge.net/"
 arch="all"
 license="BSD GPL"
-depends=""
-makedepends="ncurses-dev"
+depends="libintl gettext"
+makedepends="gettext-dev ncurses-dev"
```
kaniini: depends isn't necessary, and see other comment instead of depending on gettext-dev

PR #1225, opened 2017-04-09
Gearman

nickandrew:
Add new aports for gearman perl client and worker modules.


tmpfile:
perl-string-crc32 was added with https://github.com/alpinelinux/aports/commit/ca8a98d14ef7222a4006baa304b0e20972a1c5da
So that specific commit should be removed.
The other 3 commits LGTM.
I would just remove all ` || return 1` instances. That's implicit now because abuild runs with `set -e`.

nickandrew:
Give it a try now.

PR #1233, opened 2017-04-11
main/newsbeuter: Fixing regression

leitao:
My last patch for this package caused a regression on amd64. Fixing it
now that properly builds on all arches.


ncopa:
fixed with a2507a8e53902915fa0fa1bcbf9335620911a776

PR #1234, opened 2017-04-11
unmaintained/ipmitool: update to version 1.8.18

mmlb:


mmlb:
I'll be looking to move back to testing or preferably community.

ncopa:
does it build? if it does then move it to testing. nothing will build anything in unmaintained.

mmlb:
@ncopa yup it does build, thanks for the note about moving to testing. travis actually did build it, probably not what you meant, but I figured that would probably get the ball rolling.

mmlb:
Also I have a question about an `abuild` warning: 
```
ipmitool: ipmievd.initd is not an openrc #!/sbin/openrc-run
```

Looks like I just need to update the shebang?

ncopa:
Yes, you you need update the shebang.

Can you do that and merge your commits into one, with a commit message: `testing/ipmitool: move from unmaintained and upgrade to 1.8.18` or similar.

 Thanks!

mmlb:
@ncopa done

mmlb:
Hey @ncopa any thoughts on when I can expect the package to be out on the mirrors?

mmlb:
whoops and thanks for fixing the missing checksum update

PR #1249, opened 2017-04-13
main/nano: upgrade to 2.8.1

misery:


misery:
Maybe revert c66acd7b780be1346aed7066f302a94587d59ca3 now?

avih:
> Maybe revert c66acd7 now?

Yeah, that got fixed upstream at 2.8.1 .`--with-wordbounds` can now be removed from `configure`.

However, there's now another minor bug which seemingly manifests on Alpine but not elsewhere (not exhaustive), where non default `titlecolor` config is not respected until the screen is refreshed fully.

That's the bug: https://savannah.gnu.org/bugs/?50787

And that's a workaround patch (from that bug): https://savannah.gnu.org/bugs/download.php?file_id=40373

PR #1257, opened 2017-04-14
Use GeoIP legacy when building GoAccess

allinurl:
Ensure GoAccess builds with GeoIP legacy support, removed unnecessary dependency and updated license.

### DIFF
--- a/main/goaccess/APKBUILD
+++ b/main/goaccess/APKBUILD
@@ -28,7 +28,7 @@ build() {
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
 		--localstatedir=/var \
-		--enable-geoip \
+		--enable-geoip=legacy \
 		|| return 1
 	make || return 1
 }


main/goaccess/APKBUILD
```diff
@@ -28,7 +28,7 @@ build() {
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
 		--localstatedir=/var \
-		--enable-geoip \
+		--enable-geoip=legacy \
```
allinurl: @jirutka Right now the options are `legacy` which will utilize the original GeoIP databases and `mmdb` which will utilize the enhanced GeoIP2 databases.

kaniini: Would we not want to support the enhanced databases then?

allinurl: @kaniini Certainly, but I'm not entirely sure if Alpine has a GeoIP2 package?

allinurl: Any updates on this? Thanks.

jirutka: It seems that we don’t have geoip 2.x yet.

jirutka:
**Note for myself:** The source repository is gone, thus https://github.com/alpinelinux/aports/pull/1257.patch does not work. We must figure out how to get these patches or recreate them.

PR #1259, opened 2017-04-15
main/cmake: Update to 3.8.1 and clean up APKBUILD

pbregener:
Remove obsolete/outdated/wrong comment from APKBUILD, remove explicit `--system-...` flags, as `--system-libs` should do the trick.

### DIFF
--- a/main/cmake/APKBUILD
+++ b/main/cmake/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=cmake
-pkgver=3.8.0
-pkgrel=1
+pkgver=3.8.1
+pkgrel=0
 pkgdesc="CMake is a cross-platform open-source make system"
 url="http://www.cmake.org"
 arch="all"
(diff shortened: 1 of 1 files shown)


pbregener:
I know that you added the explicit `--system` flags as part of bd86165d1a5e2da18463ef50479741c1464241c7 @ncopa, but I guess it was done to point out the (previous) difference regarding `jsoncpp` more clearly. Are you fine with these changes?
The Alpine `jsoncpp` package does not depend on `cmake` anymore (but uses `scons` instead).

BTW: As per https://github.com/open-source-parsers/jsoncpp#building-and-testing-with-scons using scons to build jsoncpp is actually deprecated.

ncopa:
> BTW: As per https://github.com/open-source-parsers/jsoncpp#building-and-testing-with-scons using scons to build jsoncpp is actually deprecated.

Then we should probably build jsoncpp with cmake and cmake with bundled jsoncpp so we avoid bootstrap issue?

pbregener:
Updated accordingly to use bundled jsoncpp again. Now someone should probably update the jsoncpp package to actually use cmake for building.

pbregener:
Pushed an additional commit updating to cmake 3.8.1.

cc @ncopa @jirutka

pbregener:
ping

pbregener:
ping @ncopa

pbregener:
This also needs to be picked into `3.6-stable`. Thanks! :)

pbregener:
Just a quick reminder that this should also be backported to 3.6 @ncopa @jirutka

ncopa:
why is this needed for `3.6-stable`?

pbregener:
Because it's a patch release with regression fixes for cmake 3.8.0 which is included in Alpine 3.6.

PR #1261, opened 2017-04-16
main/{llvm,clang}: upgrade to 4.0.0

jirutka:
**TODO:**

- [ ] Verify that clang really works (on all architectures), mainly that the updated patches are correct (make tests working?).
- ~~[ ] Add -doc subpackage to llvm4.~~ (documentation is kinda broken, it fails to build)
- [x] Update packages that depends on main/llvm to use llvm4, llvm3.9, or llvm3.7:
    - [x] main/compiler-rt
    - [x] main/mesa
    - [x] testing/afl
    - [x] testing/creduce
    - [x] testing/lldb

* * *
Note that upstream has changed versioning scheme:

> Starting with this release, LLVM is using a new versioning scheme, increasing the major version number with each major release. Stable updates to this release will be versioned 4.0.x, and the next major release, six months from now, will be version 5.0.0.

That’s why new package is named llvm4 and not llvm4.0.


kaniini:
I will do an experimental rebuild with this change and see what happens tonight.

kaniini:
Regressions :(

```$ clang
: CommandLine Error: Option 'x86-use-base-pointer' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options```

**Comment by jirutka** [comment]
I’ve tried to remove `-DLLVM_LINK_LLVM_DYLIB=ON`, i.e. link clang with LLVM statically (as before), and this error has disappeared.

**Comment by tmh1999** [comment]
Build good for me on x86_64 and s390x. Tested hello world with clang.
It would be great if you add s390x next to ppc64le in compiler-rt in : 

https://github.com/jirutka/alpine-aports/blob/adf6533755198bfe35d03c47a9ee85e3ed4dc7d0/main/compiler-rt/APKBUILD#L38

Great work !

File: main/llvm4/APKBUILD [hunk 1]
```diff
@@ -0,0 +1,202 @@
...
+		-DLLVM_ENABLE_LIBCXX=OFF \
+		-DLLVM_ENABLE_PIC=ON \
+		-DLLVM_ENABLE_RTTI=ON \
+		-DLLVM_ENABLE_SPHINX=OFF \
+		-DLLVM_ENABLE_TERMINFO=ON \
+		-DLLVM_ENABLE_ZLIB=ON \
+		-DLLVM_HOST_TRIPLE="$CHOST" \
+		-DLLVM_INCLUDE_EXAMPLES=OFF \
+		-DLLVM_LINK_LLVM_DYLIB=ON \
+		-DLLVM_TARGETS_TO_BUILD='X86;ARM;AArch64;PowerPC;AMDGPU;NVPTX;Mips;BPF' \
```
tmh1999: SystemZ is the target to build for s390x architecture

jirutka: Aha, didn’t know that. I’ll add it.

jirutka:
Merged 🎉

jirutka:
@tmh1999 Could you please look at the failing tests on s390x? http://build.alpinelinux.org/buildlogs/build-edge-s390x/main/llvm4/llvm4-4.0.0-r0.log

tmh1999:
@jirutka It's probably a problem with the builder. On my test s390x machine, the build is good. I am seeing where things went wrong.

PR #1262, opened 2017-04-16
Update Rust and Cargo to fully pass the test suite

Shizmob:
This series of commits reorganizes and adds the appropriate patches for Rust to work properly on Alpine as far as I can see it, passing the full Cargo test suite and producing correct dynamic, static PIE, and fully dynamic (dynamically linking to Rust libraries) binaries.


testing/rust/alpine-target.patch
```diff
@@ -0,0 +1,146 @@
...
++        target_endian: "little".to_string(),
++        target_pointer_width: "32".to_string(),
++        data_layout: "e-m:e-p:32:32-f64:32:64-f80:32-n8:16:32-S128".to_string(),
++        arch: "x86".to_string(),
++        target_os: "linux".to_string(),
++        target_env: "musl".to_string(),
++        target_vendor: "alpine".to_string(),
++        options: base,
++    })
++}
```
jirutka: Why do you define new targets from scratch (hence copypasting some stuff) instead of extending some existing base target?

testing/rust/minimize-rpath.patch
```diff
@@ -0,0 +1,93 @@
+--- a/src/librustc_trans/back/rpath.rs
```
jirutka: Please add patch header and description – what this patch actually do and why it’s needed.

testing/rust/need-rpath.patch
```diff
@@ -0,0 +1,49 @@
+--- a/src/librustc_back/target/mod.rs
```
jirutka: Please add patch header and description – what this patch actually do and why it’s needed. You can probably just copy description from `force-rpath-on-prefer-dynamic.patch`…?

testing/rust/alpine-target.patch
```diff
@@ -0,0 +1,146 @@
...
++        data_layout: "e-m:e-i64:64-f80:128-n8:16:32:64-S128".to_string(),
++        arch: "x86_64".to_string(),
++        target_os: "linux".to_string(),
++        target_env: "musl".to_string(),
++        target_vendor: "alpine".to_string(),
++        options: base,
++    })
++}
+--- /dev/null
++++ b/mk/cfg/i586-alpine-linux-musl.mk
```
jirutka: I’m almost sure that this is used only by (deprecated) Makefile-based build system, so it’s useless here.

Shizmob: Nope, the bootstrap process checks for its existence to see if it's a correct target, and while it's here we may as well make it correct.

testing/rust/alpine-target.patch
Shizmob: Following in line with how all other targets are defined. It also makes us more robust against if the default `$arch-unknown-linux-musl` targets are changed.

jirutka: But we must copypaste it for every architecture…

Shizmob: Only needs to happen once, right? Different architectures have different options anyhow...

testing/rust/minimize-rpath.patch
Shizmob: Done!

testing/rust/need-rpath.patch
Shizmob: Done!

testing/rust/musl-fix-static-linking.patch
```diff
@@ -0,0 +1,148 @@
+--- a/src/librustc/session/mod.rs 
++++ b/src/librustc/session/mod.rs
+@@ -409,6 +409,11 @@
+         return crt_static;
+     }
+ 
++    pub fn fully_static(&self) -> bool {
```
jirutka: What exactly does this mean? Static without PIE? This patch also miss description. :)

jirutka:
Merged. 🎉 Thanks for excellent work!

PR #1267, opened 2017-04-18
testing/snapcast: create home directory

xsteadfastx:


clandmeter:
few comments.

1. just a tip from my side, always use a [feature branch](https://guides.github.com/introduction/flow/) when submitting pr's
2. not sure you did this deliberately but normally you need to bump pkgrel on changes (this case is kind of special as a user who installed this would never get the home directory as the user already existed)
3. and dont forget to build it local and update checksums (this case its not needed as we dont add install files to $source.

Thanks for noticing this bug! if this works as expected we can move it to community.

PR #1271, opened 2017-04-18
testing/xterm: Enable SIXEL support

7heo:


7heo:
In case anyone wonders, the binary size increases by 12kB (12608 Bytes) with this new feature, and the package size increases by 8kB (8903 Bytes).

PR #1272, opened 2017-04-18
main/binutils: fix URL in cross-compilation

xentec:
Bug introduced in bae49db117dd2ef7aabc23c1bfef571f3d8c349f

Changing pkgname to cross-compile version before setting source URLs breaks them. This PR moves that part after all paths are set. No side effects were found.


xentec:
Fixed in f741c9ab82d6eaf9bf2d86dc12ec97309328335b.

PR #1274, opened 2017-04-18
testing/virt-what: modernize APKBUILD

HRio:


testing/virt-what/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Maintainer:
 pkgname=virt-what
 pkgver=1.15
-pkgrel=0
+pkgrel=1
```
Shizmob: There don't seem to be functional changes in the resulting package here, so there is no need to bump `$pkgrel`.

PR #1275, opened 2017-04-18
main/libbsd: enable build on ppc64le

rgdoliveira:
Libbsd includes a.out.h header to get some definitions, like struct nlist,
struct exec.
In Alpine x86_64 this header is provided by linux-headers package (linux/a.out.h),
in Ubuntu ppc64le it is provided by glibc (a.out.h), but in Alpine ppc64le it is not
available.

A workaround for this problem is to define the types that are being used
by libbsd instead of including the invalid header for ppc64le.

### DIFF
--- a/main/libbsd/fix-build-with-musl-on-ppc64le.patch
+++ b/main/libbsd/fix-build-with-musl-on-ppc64le.patch
@@ -1,0 +1,60 @@
+Libbsd includes a.out.h header to get some definitions like struct nlist, struct exec.
+
+In Alpine x86_64 this header is provided by linux-headers package (linux/a.out.h),
+in Ubuntu ppc64le it is provided by glibc (a.out.h), but in Alpine ppc64le it is not
+available.
+
+A workaround for this problem is to define the types that are being used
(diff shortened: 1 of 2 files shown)


main/libbsd/fix-build-with-musl-on-ppc64le.patch
```diff
@@ -0,0 +1,60 @@
...
++
++#define N_UNDF  0
++#define N_ABS   2
++#define N_TEXT  4
++#define N_DATA  6
++#define N_BSS   8
++#define N_FN    15
++#define N_EXT   1
++#define N_STAB  0340
++
```
tmh1999: Also, I have to make this https://github.com/tmh1999/aports/commit/338d442f8d29236497a07fc84b02fdbe2b6d2fac in order to have a good build of netcat-openbsd and libvirt on s390x. Looks like ppc64le is also having a problem with that?

rgdoliveira:
@tmh1999 Actually I did not define N_INDR, N_SETA (and so on) because the libbsd does not seem to be using them

PR #1279, opened 2017-04-19
testing/lshell: fix pre-install / add tests subpkg

itoffshore:
testing/lshell: fix pre-install / add tests subpkg

### DIFF
--- a/testing/lshell/APKBUILD
+++ b/testing/lshell/APKBUILD
@@ -2,39 +2,35 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=lshell
 pkgver=0.9.18
-pkgrel=1
+pkgrel=2
 pkgdesc="Python-based limited Shell"
 url="http://ghantoos.org/limited-shell-lshell"
 arch="noarch"
 license="GPL3+"
 depends="python2"
-depends_dev=""
+options="!check"
 makedepends="python2-dev py-setuptools"
-install=""
-subpackages="$pkgname-doc"
+install="$pkgname.pre-install"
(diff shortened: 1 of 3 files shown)


testing/lshell/APKBUILD
```diff
@@ -2,39 +2,35 @@
...
 }
 
 package() {
-	cd "$_builddir"
-	python2 setup.py install --prefix=/usr --root="$pkgdir" || return 1
+	cd "$builddir"
+	python2 setup.py install --prefix=/usr --root="$pkgdir"
+}
+
+_tests() {
```
Shizmob: The subpackage function doesn't need a leading underscore.

testing/lshell/APKBUILD
```diff
@@ -2,39 +2,35 @@
...
 pkgver=0.9.18
-pkgrel=1
+pkgrel=2
 pkgdesc="Python-based limited Shell"
 url="http://ghantoos.org/limited-shell-lshell"
 arch="noarch"
 license="GPL3+"
 depends="python2"
-depends_dev=""
+options="!check"
```
Shizmob: Does the package testsuite not work? Also, do the tests need to be packaged for a specific reason?

itoffshore: the testsuite requires `lshell` to be installed - so I bundled the tests rather than run the testsuite against the previous version of `lshell` (I was going to run the testsuite on my servers - but since making this `PR` I have used `rshell` instead)

if you think it's better to run the testsuite against the previous version of `lshell` I will revise & rebase.

testing/lshell/APKBUILD
itoffshore: fixed & rebased

PR #1280, opened 2017-04-19
testing/py-bandit: new aport

itoffshore:
https://wiki.openstack.org/wiki/Security/Projects/Bandit

### DIFF
--- a/testing/py-bandit/APKBUILD
+++ b/testing/py-bandit/APKBUILD
@@ -1,0 +1,55 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-bandit
+_pkgname=${pkgname#*-}
+pkgver=1.4.0
+pkgrel=0
+pkgdesc="Security oriented static analyser for python code"
+url="https://wiki.openstack.org/wiki/Security/Projects/Bandit"
+arch="noarch"
+license="ASL-2.0"
+depends="python"
+# temporarily disable as unit tests require py-bandit
+options="!check"
+makedepends="python-dev python3-dev py-setuptools"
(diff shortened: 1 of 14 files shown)


testing/py-bandit/APKBUILD
```diff
@@ -0,0 +1,47 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-bandit
+_pkgname=${pkgname#*-}
+pkgver=1.4.0
+pkgrel=0
+pkgdesc="Security oriented static analyser for python code"
+url="https://wiki.openstack.org/wiki/Security/Projects/Bandit"
+arch="noarch"
+license="Apache2"
```
Shizmob: I believe `ASL-2.0` is preferred.

testing/py-bandit/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+depends="python"
+makedepends="python-dev python3-dev py-setuptools"
+options="!check"
```
Shizmob: Could you add a comment why there is no check function, e.g. because of the lack of upstream test suite? Is there a simple quick smoke test you could add (e.g. invocation of the resulting binary/code) to make sure that at least something didn't go terribly wrong?

itoffshore:
added a note why the tests are temporarily disabled (they require the python `bandit.core` module) + some packages not yet in the repos.

I also need `py-bandit` pulled to finish the `lshell` PR.

itoffshore:
can this be pulled ?

no "smoke test" is possible on the binaries in a brand new package that requires it's own python libs

itoffshore:
this needs a few python depends

itoffshore:
still needs 2 more new aports

itoffshore:
still needs a few more

itoffshore:
I'm going to split this up into a few - `PR's`

it's too many commit's to `rebase` with all the dependencies involved.

PR #1284, opened 2017-04-19
testing/php7: upgrade to 7.1.4

andypost:
Bug fix release http://php.net/archive/2017.php#id2017-04-13-2

### DIFF
--- a/testing/php7/APKBUILD
+++ b/testing/php7/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Maintainer: Valery Kartel <valery.kartel@gmail.com>
 pkgname=php7
-pkgver=7.1.3
-pkgrel=1
+pkgver=7.1.4
+pkgrel=0
 _apiver=20160303
 _compat=7.1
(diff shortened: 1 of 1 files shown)


andypost:
@vakartel Here's just upgrade on top of http://patchwork.alpinelinux.org/patch/3338/

vakartel:
@andypost 
In patchwork3338 I also fixed depends_dev, depends_doc, pecl()-depends, added correct replaces* for coexist with php5, removed compatibility with php7.1 naming, and added disabling pcre-jit on s390x.

So It's more complex than just version bump.

jirutka:
@vakartel And more importantly broken it even more with incorrect usage of `replaces`…

vakartel:
@jirutka 
You asked to move php7 (7.1) from testing to community. So I made a PR #1305

fabled:
@vakartel I know of https://patchwork.alpinelinux.org/patch/3338/ but it did not apply as is, can you please follow up on top of this, as this applied and looked a step forward. Thanks.

vakartel:
@fabled Now it does not matter. Because testing/php7 would be removed soon #1305

PR #1306, opened 2017-04-24
main/alpine-base: add layouts for mail and web servers

vakartel:
I want to introduse a kind of common layout for mail and web servers.

Beacuse:
- every mail server uses 'aliases' file and /var/spool/mail as a default storage for local mail.
- every web server creates /var/www/localhost/htdocs directory and create www-data group. Many of webapps does the same.

I think it's a good idea to unify these things for all of them.

### DIFF
--- a/main/alpine-base/APKBUILD
+++ b/main/alpine-base/APKBUILD
@@ -2,18 +2,21 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=alpine-base
 pkgver=3.5.0
-pkgrel=0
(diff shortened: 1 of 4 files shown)


vakartel:
@ncopa 
I do not know if this will be useful. I want to know your opinion.

vakartel:
about 'mail' user and group creation: For now mail user and group exists in baselayout, but if this would pass, it's better to remove it, because no busybox-related stuff use it.

kaniini:
personally, i would prefer to sit on this until after 3.6 release engineering is done

PR #1308, opened 2017-04-25
testing/p0f - new aport

archey:
Updated and fixed p0f build for alpine


testing/p0f/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+pkgname=p0f
+pkgver=3.09b
+pkgrel=0
+pkgdesc="Passive traffic fingerprinting tool"
+url="http://lcamtuf.coredump.cx/p0f3/"
+arch="all"
+license="GPL2"
+makedepends="bash libpcap-dev"
+subpackages="$pkgname-doc"
+source="http://lcamtuf.coredump.cx/p0f3/releases/$pkgname-$pkgver.tgz p0f-fixes.patch"
```
Shizmob: Please put all sources on their own line.

testing/p0f/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+
+build() {
+	cd "$srcdir/$pkgname-$pkgver"
```
Shizmob: Can you set `$builddir` and use that?

testing/p0f/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	make
+	cd tools
+	make p0f-client
+	make p0f-sendsyn
+	make p0f-sendsyn6
+}
+
+package() {
+	cd "$srcdir/$pkgname-$pkgver"
+        install -Dm 755 p0f "$pkgdir/usr/bin/p0f"
```
Shizmob: This seems to have some inconsistent indentation. Also, you can use `cp` if the existing permissions are correct.

testing/p0f/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+        install -Dm 644 p0f.fp "$pkgdir/etc/p0f/p0f.fp"
+        install -Dm 644 docs/COPYING "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
```
Shizmob: I'm not sure this is needed unless the license is a custom one.

testing/p0f/p0f-fixes.patch
```diff
@@ -0,0 +1,42 @@
+--- p0f-3.09b.orig/build.sh
```
Shizmob: Could you add a comment at the start of the file explaining what it fixes and why it's needed? If it's too many different changes, it may be useful splitting up as well.

PR #1316, opened 2017-04-25
testing/grub: fix build on ppc64le

rgdoliveira:
grub was not building on ppc64le. This patch adds the correct 'platform'
value in configure and make necessary adjusts to make grub build.

### DIFF
--- a/testing/grub/APKBUILD
+++ b/testing/grub/APKBUILD
@@ -11,30 +11,33 @@ license="GPLv3+"
 depends=""
 depends_dev=""
 makedepends="$depends_dev bison flex linux-headers xz-dev lvm2-dev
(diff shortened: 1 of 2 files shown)


ncopa:
i think i'd want replace those `if [ "$CARCH" = ...]` with a `case "$CARCH" in`

ncopa:
it fails to build on ppc64le. it needs `freetype-dev` in makedepends
````
checking for freetype-config... no
configure: error: qemu, powerpc-ieee1275, coreboot and loongson ports needs build-time grub-mkfont (need freetype2 library)
>>> ERROR: grub: all failed
````

rgdoliveira:
@ncopa Thanks, just added the freetype-dev in makedepends

PR #1318, opened 2017-04-26
main/mosh: Skip one test on ppc64le

leitao:
Currently test unicode-later-combining.test is failing on ppc64le.
Let's skip it for 3.6, and analyze it later.


leitao:
It seems that this test is not only failing on ppc64le, but on all platform. Should I remove this test from all arches?

PR #1320, opened 2017-04-26
testing/nginx-naxsi: update to 1.13.8

itoffshore:
* add `sysguard.conf` to the sysguard module
* fixes `${SVCNAME}` => `${RC_SVCNAME}` in `initd`
* add `server_tokens off;` to `default.conf`

https://nginx.org/en/CHANGES

### DIFF
--- a/testing/nginx-naxsi/APKBUILD
+++ b/testing/nginx-naxsi/APKBUILD
@@ -5,8 +5,8 @@
 
 pkgname=nginx-naxsi
 _pkgname=nginx
-pkgver=1.11.10
-pkgrel=2
+pkgver=1.13.8
+pkgrel=0
 pkgdesc="Lightweight HTTP and reverse proxy server with Naxsi WAF support, see also 'nxapi'"
 url="http://www.nginx.org | https://github.com/nbs-system/naxsi"
 arch="all"
@@ -22,7 +22,7 @@ _ngx_cache_purge_ver=2.3.0.1
 _ngx_cache_purge_dir="$srcdir/$_ngx_cache_purge_name-$_ngx_cache_purge_ver"
 
 _ngx_upstream_fair_name=nginx-upstream-fair
-_ngx_upstream_fair_ver=0.1.1
+_ngx_upstream_fair_ver=0.1.2
 _ngx_upstream_fair_dir="$srcdir/$_ngx_upstream_fair_name-$_ngx_upstream_fair_ver"
 
 _ngx_http_sysguard_name=tengine-http-sysguard
@@ -53,6 +53,7 @@ source="http://nginx.org/download/$_pkgname-$pkgver.tar.gz
 	nginx.logrotate
 	nginx.conf
 	default.conf
+	sysguard.conf
 	"
 builddir="$srcdir"/$_pkgname-$pkgver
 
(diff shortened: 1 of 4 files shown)


itoffshore:
please pull - `nginx-naxsi` has `PaX` protections enabled

jirutka:
> nginx-naxsi has PaX protections enabled

I'm confused now, why did you mention this?

itoffshore:
only because the previous  `PR` was waiting for 24 days - so long a rebase for the latest version included a massive merge to fix a merge conflict on github (hence the new clean `PR`)

Am just highlighting there is a major difference between `nginx-naxsi` & the version in main (I also have anonymization recommendations from caromel.org & some servers disabled)

itoffshore:
ok - I understand the holdup now https://github.com/libressl-portable/portable/issues/307

Shizmob:
Could you maybe bring this in line with the [current nginx APKBUILD](https://github.com/alpinelinux/aports/blob/master/main/nginx/APKBUILD), with the tests and the like?

jirutka:
@Shizmob, I’d like to add the modules that @itoffshore actually needs to the main/nginx and eventually remove this package.

itoffshore:
I'd like `PaX` **not** to be disabled on `main/nginx` - but I know it breaks the `lua` module

if `PaX` cannot be enabled I would like to continue with `nginx-naxsi` - or perhaps rename it to `nginx-pax` & track the latest version of `nginx` (not stable)

Shizmob:
Can't we build the Lua module as a separate `.so` and apply the PaX mark to that specific file?

jirutka:
It’s trivial to re-enable PaX MPROTECT for nginx binary after installation (just `paxmark M /usr/sbin/nginx`). Isn’t that sufficient for you? I’m not sure if it’s really worth it to overcomplicate this nginx package or maintain separate package just because of this.

> Can't we build the Lua module as a separate .so and apply the PaX mark to that specific file?

We build Lua module as a dynamic module, so it is a separate `.so` file. I’m not sure if it’d be sufficient to disable MPROTECT just for `ngx_http_lua_module.so`, but don’t remember if I’ve tried it or not.

jirutka:
Hm, what about paxmarking nginx module after installation, using post-install script associated with the Lua module?

Shizmob:
That would likely be better in my opinion, if just PaX-marking the Lua module does not suffice -- but I would like to try that approach first.

jirutka:
@itoffshore Do I understand correctly that you don’t use any extra module from this package anymore, so we don’t need to add them into main/nginx? Not saying that I don’t wanna add them, but IIRC some of these modules were not in very good condition, some needs patching nginx etc.

itoffshore:
I use `naxsi` & `sysguard` - both of these should be in good shape (I ported `sysguard` from the current `tengine` sources) - `ngx_cache_purge` is still maintained & the author of `upstream-fair` [sent a `PR` a few weeks ago with an `openssl` fix](https://github.com/itoffshore/nginx-upstream-fair/releases).

itoffshore:
you could also just pull this - I am quite happy maintaining a lighter `nginx` with `PaX` enabled + anonymization patch

PR #1325, opened 2017-04-27
main/dovecot: upgrade to 2.2.29.1 and pigeonhole plugin to 0.4.18

ashb:


main/dovecot/APKBUILD
```diff
@@ -3,11 +3,11 @@
...
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=dovecot
-pkgver=2.2.27
+pkgver=2.2.29.1
 pkgrel=2
-_pigeonholever=0.4.16
+_pigeonholever=0.4.18
 _pluginextdataver=39
-_pkgvermajor=${pkgver%.*}
+_pkgvermajor=$(echo $pkgver | cut -d. -f1-2)
```
jirutka: Please don’t use subshell here, use parameter expansion instead.

jirutka:
libressl patch must be updated:

```
>>> dovecot: libressl.patch
patching file src/lib-ssl-iostream/dovecot-openssl-common.c
Hunk #1 FAILED at 101.
1 out of 1 hunk FAILED -- saving rejects to file src/lib-ssl-iostream/dovecot-openssl-common.c.rej
```

main/dovecot/APKBUILD
ashb: I'll see if I can get a param expansion version working. (The previous version didn't cope with a changing number of dots in the version, i.e. it went from 2.2.28 to 2.2.29.1)

ashb: I can't think of any easy way of dealing with changing number of dots (i.e. that would cope with 2.2.29.1 and 2.2.30) without a subshell . Either we have a subshell here so we don't have to change with it in the future, or I change the pattern for this current case only.

Which would you rather?

ashb:
Have removed the Libressl patch - it was fixed upstream in https://github.com/dovecot/core/commit/b91659702c83b8f144e369ad5fa82724242c210c

main/dovecot/APKBUILD
Shizmob: I personally think it's fine to just use `cut` to deal with the varying number of dots.

Shizmob:
You also need to remove `libressl.patch` from `source=` -- please test your APKBUILD. :)
Also, please squash all your commits at the end into a single one.

ashb:
@Shizmob doh thanks! (I'm out and about so didn't test it properly which is sloppy of me)

vakartel:
Just played a little with ash to grab major version (first two numbers) w/o subshell. Need to cut minor tail first. Then you get major part.
```
pkgver=2.29.30.1
_pkgverminor=${pkgver##[[:digit:]].[[:digit:]]?.}
_pkgvermajor=${pkgver%.$_pkgverminor}
```

PS: subshell variant is more readable :)

ashb:
@vakartel Ah cool didn't know you could use `$var` in the pattern side. Have just gone with hard coding `_pkgvermajor` right next to the `pkgver` after chatting on IRC

tmpfile:
LGTM, than

tmpfile:
@ashb LGTM, you should just squash the 2 commits into a single commit IMO.

ashb:
@tmpfile have done that.

jirutka:
@tmpfile Please don’t tell ppl to squash commits. In this case, it’d be cleaner to upgrade pkg in one commit and add pigeonhole plugin in another one. In case of fixup commits it’s easier to review changes when they are not squashed. It’s trivial to squash commits when needed during merging.

PR #1329, opened 2017-04-28
scripts: add support for ppc64le in vanilla configuration

rgdoliveira:
To boot ppc64le we need to use grub with ieee1275 platform. Said that,
a new section to support grub ieee1275 was created.
Also needed to change xorrisofs command by grub-mkrescue, that is the one
used to create a bootable image for power and it is also a wrapper to xorriso.

### DIFF
--- a/scripts/mkimg.base.sh
+++ b/scripts/mkimg.base.sh
@@ -190,6 +190,13 @@ build_grubefi_img() {
 	mcopy -s -i ${DESTDIR}/boot/grub/efiboot.img $_tmpdir/efi ::
 }
 
+section_grubieee1275() {
+	[ "$ARCH" = ppc64le ] || return 0
+	[ "$output_format" = "iso" ] || return 0
+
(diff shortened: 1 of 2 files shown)


scripts/mkimg.base.sh
```diff
@@ -249,18 +256,23 @@ create_image_iso() {
...
-		-joliet \
-		-rock \
-		-volid "alpine-$PROFILE $RELEASE $ARCH" \
-		$_isolinux \
-		$_efiboot \
-		-follow-links \
-		${iso_opts} \
-		${DESTDIR}
+
+	if [ "$ARCH" = ppc64le ]; then
```
ncopa: can grub-mkrecue be used for all archs?

rgdoliveira:
@ncopa I am not sure if we should use grub-mkrescue for all archs now. I mean, we are still trying to fix some issues with grub-mkrescue and I do not want to take the risk to break the image generation in x86 (or other archs)

PR #1339, opened 2017-05-01
community/php7: upgrade to 7.1.4 and improve abuild, remove testing/php7

jirutka:


jirutka:
Merged into master, but comments are still welcome!

community/php7/APKBUILD
```diff
@@ -157,203 +322,285 @@ build() {
...
+}
+
+check() {
+	cd "$builddir"
+
+	# PHP is so stupid that it's not able to resolve dependencies
+	# between extensions and load them in correct order, so we must
+	# help it...
+	# * opcache is Zend extension, it's handled specially in Makefile
+	# * imap can't be loaded together with recode, so we must omit it
```
vakartel: imap can be loaded together with recode, but it must loaded after recode. It fixed in my php7 variant.

community/php7/APKBUILD
```diff
@@ -157,203 +322,285 @@ build() {
...
 	rmdir "$pkgdir"/var/run
+
+	[ "$_default_php" = yes ] \
+		&& ln -s php$_suffix "$pkgdir"/usr/bin/php
 }
 
 dev() {
 	default_dev
 	replaces="php-dev"
-	depends="$depends $pkgname $pkgname-pear"
```
vakartel: Why you not use global depends_dev instead?
And dev does not depend on pear.

For all 3rd party php-extensions to build need php (cli), autoconf and most of them pcre-dev, so look in my variant:
depends_dev="$pkgname autoconf pcre-dev"

Also I split pear on two:  pear (just runtime) and pecl (same as pear but with ability to build extension from source)

pecl() {
        replaces=
        pkgdesc="$pkgdesc (php extension community library)"
        depends="$pkgname-pear $pkgname-dev gcc make musl-dev"
        mkdir -p "$subpkgdir"
}

community/php7/APKBUILD
```diff
@@ -157,203 +322,285 @@ build() {
...
 }
 
-_package_ext() {
+_extension() {
 	local extname="${subpkgname#$pkgname-}"
-	local extdepends="$(eval "echo \$depends_$extname")"
-	pkgdesc="PHP7 extension: $extname"
-	depends="$depends $extdepends"
+	local extdepends="$(eval "echo \$_depends_$extname")"
+	pkgdesc="PHP$_suffix extension: $extname"
```
vakartel: I used this code to grab description for every extension:

pkgdesc=$(eval echo \$_pkgdesc_$name)
: ${pkgdesc:=$(head -n1 "$builddir"/ext/$name/CREDITS)}
pkgdesc="PHP$_suffix extension: ${pkgdesc:-$name}"

only opcache has no valid description, so I set it explictly in _pkgdesc_opcache

community/php7/APKBUILD
```diff
@@ -157,203 +322,285 @@ build() {
...
-	mv "$pkgdir"/etc "$subpkgdir"/ || return 1
+	cd "$pkgdir"
 
-	mkdir "$subpkgdir"/etc/$pkgname/conf.d
+	_mv usr/lib "$subpkgdir"/usr/
+	_mv etc "$subpkgdir"/
+	mkdir -p "$subpkgdir"/$_extension_confd
 }
 
 phar() {
```
vakartel: In my variant I split this on two: phar (php extension only) and phar-utils - those /usr/bin/phar and phar.phar scripts

community/php7/APKBUILD
```diff
@@ -1,152 +1,317 @@
...
-		--program-suffix=7 \
-		--libdir=/usr/lib/$pkgname \
+		--program-suffix=$_suffix \
+		--libdir=$_libdir \
 		--datadir=/usr/share/$pkgname \
 		--sysconfdir=/etc/$pkgname \
 		--localstatedir=/var \
 		--with-layout=GNU \
 		--with-pic \
 		--with-pear=/usr/share/$pkgname \
```
vakartel: There is no sence to store source-based extensions in versionized subdirectory, because it fits any php version.

This how it done in my variant: --with-pear=/usr/share/pear

And also all php[57]-pear-* packages must be remade to be unversionized.

community/php7/APKBUILD
jirutka: > Why you not use global depends_dev instead?

Good question! Because it was already here and I’ve overlooked it.

> Also I split pear on two: pear (just runtime) and pecl (same as pear but with ability to build extension from source)

But you left pecl tool in the pear subpackage and pecl subpackage let empty, just with additional dependencies. This doesn’t make much sense for me. But I don’t know if pear for some reason needs pecl command, but not all build dependencies are needed, so I just left it as-is for now.

So how it works, can I move pecl from -pear to a separate subpackge -pecl?

community/php7/APKBUILD
jirutka: But why? To create even more confusion among users?

community/php7/APKBUILD
jirutka: Ah, right, I wanted to use this approach, but then forgot to apply it.

community/php7/APKBUILD
jirutka: It seems that you’re right, but then I totally don’t understand why php refuses to even build them together… There’s actually no problem with building them together, so I thought that they cannot be loaded together (and it even failed when I tried, but just b/c of order as you’ve said).

jirutka: Yeah, even tests are okay, many of them are skipped because of “could not connect to mailbox {localhost/norsh}” (WTF?!), but this is a different issue. But it means that we need to explicitly define order not based on dependencies, omfg…

community/php7/APKBUILD
andypost: pecl require tools to build code but pear may not

community/php7/APKBUILD
andypost: order matters http://php.net/manual/en/recode.installation.php would be great to link here

community/php7/APKBUILD
andypost: discussed briefly at IRC
IMO better to have [bundled](http://php.net/manual/en/extensions.membership.php#extensions.membership.bundled) in some meta package like `php-bundled` but most of distros using `php-common` with a broader set of extensions

community/php7/APKBUILD
andypost: ++ to have pear common

community/php7/APKBUILD
andypost: btw It's really confusing to have no phar support out of box!
`composer` is the most used package manager require it

community/php7/APKBUILD
vakartel: Yes, pecl command need to be moved to -pecl subpackage.

community/php7/APKBUILD
vakartel: phar scripts used rarely then phar extension. So I assumed not to pull unneeded stuff with it. And the extension subpackage must be clean enough: only two files - binary module and config that enable it.

community/php7/APKBUILD
vakartel: php-bundled is in discuss. To not generate 100500 variants what ext-s must be in list, my propose - to include to -bundled all extensions enabled by ./configure without arguments.

community/php7/APKBUILD
vakartel: It's a quirk from php5. imap and recode was use same namespace and couldn't be builded together. In my variant I just split building process of this extensions to different phases. And I increased imap's loading index to load it after recode. In php5-mysql extension the same problem.

PR #1341, opened 2017-05-02
Upgrade Hugo package to 0.20.6

blobor:
This PR upgrades Hugo package to `0.20.6` version for 3.5 branch.


jirutka:
We don’t upgrade packages in stable branches, only fixes and security updates/patches should be backported.

blobor:
Whoops, sorry for that 🙃

PR #1345, opened 2017-05-04
community/php7: various fixes and improvements

jirutka:
TODO:

- [ ] Move pecl into separate subpackage -pecl.
- [ ] Change `--with-pear-dir` to `/usr/share/pear` and update all pear pkgs.

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -540,7 +540,8 @@ phar() {
 _extension() {
 	local extname="${subpkgname#$pkgname-}"
 	local extdepends="$(eval "echo \$_depends_$extname")"
-	pkgdesc="PHP$_suffix extension: $extname"
+	local extdesc="$(head -n1 "$builddir"/ext/$extname/CREDITS 2>/dev/null ||:)"
+	pkgdesc="PHP$_suffix extension: ${extdesc:-$extname}"
 
 	: ${extdepends:=$(_resolve_extension_deps "$extname")}
 	depends="$depends $extdepends"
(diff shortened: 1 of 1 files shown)


community/php7/APKBUILD
```diff
@@ -535,6 +539,14 @@ phar() {
 	fi
 }
 
+default_exts() {
+	pkgdesc="PHP$_suffix metapackage for all Core Extensions and Bundled Extensions"
+	depends=$(printf '%s\n' $_extensions \
+		| sed -En "s/(\w+)\*/$pkgname-\1/p")
+
+	mkdir -p "$subpkgdir"
+}
```
jirutka: @andypost @vakartel What do you think?

community/php7/APKBUILD
```diff
@@ -537,18 +539,24 @@ phar() {
 	fi
 }
 
+default_exts() {
+	pkgdesc="PHP$_suffix metapackage for all Core Extensions and Bundled Extensions"
+	depends=$(printf '%s\n' $_extensions \
+		| sed -En "s/(\w+)\*/$pkgname-\1/p")
```
jirutka: Note for myself: Fix regexp, there should be `\+`, not `\*`.

community/php7/APKBUILD
vakartel: I think it's pretty good.
But all those plussed ones are 'bundled' not 'core'
(http://php.net/manual/en/extensions.membership.php)
So IMHO this subpackage need to be named php7-bundled
and description like "PHP$_suffix metapackage for all Bundled Extensions"

jirutka: I’m afraid that _php7-bundled_ is very misleading name. _php7-bundled-exts_ would be a bit better, but still misleading, because they are in fact not bundled. I don’t like _php7-default-exts_ either, but can’t figure out anything better.

jirutka: Sorry for confusion, I’ve merged all commits from this PR _except_ the last one adding php7-default-exts. I’ll merge this separately after we resolve how to name it.

PR #1347, opened 2017-05-04
community/php7: Change php7-module.conf to use SetHandler

andypost:
Related php5 PR is https://github.com/alpinelinux/aports/pull/206


andypost:
@vakartel please check it

vakartel:
Yes, it's must have.
SetHandler is more secure than AddHandler because "AddHandler php5-script .php" treats any file with .php in it's name as php-script.
So it can try to run something like file.php.jpg etc...

PR #1356, opened 2017-05-06
testing/mailutils: new aport

Shizmob:
Also add the new aport as a requirement for `gpg-remailer`.

### DIFF
--- a/testing/mailutils/APKBUILD
+++ b/testing/mailutils/APKBUILD
@@ -1,0 +1,75 @@
+# Contributor: Shiz <hi@shiz.me>
+# Maintainer: Shiz <hi@shiz.me>
+pkgname=mailutils
+pkgver=3.2
+pkgrel=0
+pkgdesc="GNU swiss army knife of electronic mail handling"
+url="https://mailutils.org/"
+arch="all"
+license="GPL3+"
+provides="mailx=$pkgver-r$pkgrel"
+replace="mailx"
+depends_dev="$pkgname-libs=$pkgver-r$pkgrel"
+makedepends="readline-dev libtool"
+checkdepends="autoconf"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-libs $pkgname-servers $pkgname-mh"
+source="https://ftp.gnu.org/gnu/mailutils/mailutils-$pkgver.tar.gz
+	disable-koi8-r-test.patch"
+builddir="$srcdir/mailutils-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr
+	make
+}
+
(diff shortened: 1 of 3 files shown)


testing/mailutils/APKBUILD
```diff
@@ -0,0 +1,74 @@
...
+	disable-koi8-r-test.patch"
+builddir="$srcdir/mailutils-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr
+	make -j4
```
jirutka: Abuild exposes environment variable `JOBS` with user preferred number of jobs for `make` and similar. I’m not sure if `make` implicitly uses that variable and so whether it’s needed to be set explicitly. Anyway, if you want to set number of jobs, please use `-j $JOBS`.

testing/mailutils/APKBUILD
```diff
@@ -0,0 +1,74 @@
...
+	make DESTDIR="$pkgdir" install
+
+	rm "$pkgdir"/usr/lib/charset.alias
+	# no need for these to be suid/sgid root.
+	chmod u-s "$pkgdir"/usr/sbin/maidag
+	chmod g-s "$pkgdir"/usr/bin/dotlock
+}
+
+servers() {
+	pkgdesc="GNU swiss army knife of electronic mail handling (servers)"
```
jirutka: Please don’t repeat yourself. ;) `pkgdesc="$pkgdesc (servers)"`

testing/mailutils/APKBUILD
```diff
@@ -0,0 +1,74 @@
...
+	make check
+}
+
+package() {
+	cd "$builddir"
+	# re-enable imap4d subdirectory for install
+	sed -i 's/# $(IMAP4D_DIR)/$(IMAP4D_DIR)/g' Makefile
+	make DESTDIR="$pkgdir" install
+
+	rm "$pkgdir"/usr/lib/charset.alias
```
jirutka: protip: If you `cd "$pkgdir"` above these, you don’t need to repeat `"$pkgdir"` three times ;)

testing/mailutils/APKBUILD
Shizmob: Oops, I only added this locally for testing and forgot to remove it!

testing/mailutils/APKBUILD
Shizmob: Will this actually work for multiple subpackages? I did it like this because I feared some subpackage may end up like "... (libraries) (servers)". :P

jirutka:
Merged, thanks for your contribution!

testing/mailutils/APKBUILD
jirutka: Yes, it works.

PR #1360, opened 2017-05-08
main/busybox: add -Z option to syslogd and make it the default

Shizmob:


kaniini:
This looks fine.

A concern I have is that it may be turned on for people who do not expect it though, who have not modified their conf.d for syslog, so I am not sure if we should enable it by default.

@ncopa @fabled what do you think?

Shizmob:
To keep in mind: this will only affect busybox syslogd if `/etc/localtime` (or a very early `$TZ`) exists and is set to non-UTC. In that case, the syslogd would already have messages in that timezone in `/var/log/messages` (the messages from its own initialisation/shutdown, which are always logged in the local time), regardless of whether `-Z` is passed or not.

As such, if we don't make it the default, `/var/log/messages` will be full with messages of conflicting timezones, like this:
```
May  9 06:38:57 4cefaff33a69 syslog.info syslogd started: BusyBox v1.26.2
May  8 10:39:03 4cefaff33a69 user.notice builder: nice
May  9 06:45:04 4cefaff33a69 syslog.info syslogd started: BusyBox v1.26.2
```

kaniini:
okay.  lets merge this, then.

PR #1362, opened 2017-05-09
testing/py-unoconv: new aport

TBK0000:


ncopa:
I get this:
````
>>> unoconv: Checking sha512sums...
unoconv-0.7.tar.gz: OK
setup.py: FAILED
````

jirutka:
I haven't checked it yet, but since it's a *tool*, it probably installs some executables into /usr/bin, so py2/py3 would be in conflict. Since it's a new aport, I would just omit py2.

ncopa:
> @jirutka Since it's a new aport, I would just omit py2.

I agree.

PR #1363, opened 2017-05-09
main/flex - updated flex to 2.6.4

archey:
Added check function and updated to 2.6.4
This will fix build issues with unbound


Shizmob:
LGTM.

ncopa:
from the NEWS file in flex:

> *** A number of portability fixes were introduced so building flex is
> more reliable on more platforms. Additionally, outdated function
> calls were removed.

I don't know if any of the currently built packages which uses flex depends on any of those outdated functions. I am test building `grep flex */APKBUILD` and if it passes i'll push it. This is because we already built everything for the 3.6 release.

Thanks!

Shizmob:
I think this refers to function calls to deprecated APIs within Flex, not with regards to its external API.

PR #1373, opened 2017-05-11
main/kbd: generate (binary) keymaps from X.org keyboard-config

Shizmob:


Shizmob:
This seems to lead to a severe size increase for the replacement package (~780kB -> ~17MB), due to having way more keymaps that are also more featureful...

kaniini:
I think we should compress the keymaps.  Can you handle the setup-keymap changes too?

ncopa:
I don't think it is an issue that it is a couple of megs. setup-keymap should make a copy of the selected keymap and then uninstall the package with all alternatives.

jirutka:
> setup-keymap should make a copy of the selected keymap

This is IMO quite bad approach. Wouldn't it be better to split each keymap into a subpackage?

Shizmob:
Copying is what we already do, for what it's worth.

jirutka:
Uh, there are so many keymaps?!

Shizmob:
As in my first reply, 475 of them. :)

jirutka:
Okay, then I agree that splitting all of them to separate subpackages is very bad idea. XD

Shizmob:
Relevant `setup-keymap` changes handled here: https://github.com/alpinelinux/alpine-conf/pull/5

PR #1375, opened 2017-05-11
community/php7: add meta-subpackage for “bundled” extensions

jirutka:
This PR is follow up for #1345, see discussion https://github.com/alpinelinux/aports/pull/1345#discussion_r114918627.

/cc @vakartel @andypost

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -80,7 +80,7 @@ makedepends="
 provides="$pkgname-cli"  # for backward compatibility
 subpackages="$pkgname-dev $pkgname-doc $pkgname-apache2 $pkgname-phpdbg
 	$pkgname-embed $pkgname-litespeed $pkgname-cgi $pkgname-fpm
-	$pkgname-pear::noarch
+	$pkgname-pear::noarch $pkgname-default-exts:default_exts:noarch
 	"
 source="http://php.net/distributions/$_pkgreal-$pkgver.tar.bz2
 	$pkgname-fpm.initd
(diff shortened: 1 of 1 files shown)


andypost:
Looks great, I just wondered why not improve `php7-common`... that carry just ini file

jirutka:
What do you mean by "improve php7-common"?

andypost:
I mean that there's http://pkgs.alpinelinux.org/package/edge/community/x86_64/php7-common which contains onli php.ini file but its subpackages contains much more that needed

jirutka:
php7-common contains common files needed for all SAPIs (CLI that is provided by the base package php7, FPM provided by php7-fpm subpackage and few others I don't remember). Currently it's just one ini file (should there be more files?). php7-common is a subpackage of the php7 package, subpackages don't have their own subpackages.

vakartel:
In my ABUILD this subpackage named "-config". I think it's more correct name for it. It's the only subpackage that have no dependencies. All other subpackages depends on it.

renoirb:
/me subscribing to this PR

I just worked on two extensions and I'd like to follow conventions this PR is setting in place.

I'm squashing my work WIP in https://github.com/renoirb/aports/tree/php7-mustache?files=1

jirutka:
> In my ABUILD this subpackage named "-config". I think it's more correct name for it.

I know about it. I'm not sure if it's more correct name, but I'm quite sure that it's a change that would cause many troubles to existing users and is unnecessary.

andypost:
In https://github.com/alpinelinux/aports/pull/2951 we bundled zlib so this still makes sense!
IMO it needs sort of append to php7/ABUILD about that other "bundled" expected extensions could be found in php7-bundled package

andypost:
Here's workaround to run tests https://github.com/alpinelinux/aports/blob/master/community/php7-event/APKBUILD#L27

andypost:
Meanwhile this package still makes sense

PR #1379, opened 2017-05-11
testing/zerotier-one: update to 1.2.4, take over maintainers…

adamierymenko:
testing/$pkgname: update zerotier-one to 1.2.4, take over maintainership if that's okay

$pkgurl
$pkgdesc

### DIFF
--- a/testing/zerotier-one/APKBUILD
+++ b/testing/zerotier-one/APKBUILD
@@ -1,38 +1,48 @@
-# Contributor: Kyle Parisi <kyleparisi@gmail.com>
-# Maintainer: Kyle Parisi <kyleparisi@gmail.com>
+# Contributor: ZeroTier, Inc. <contact@zerotier.com>
+# Maintainer: Adam Ierymenko <adam.ierymenko@zerotier.com>
+
+## Thanks to Kyle Parisi <kyleparisi@gmail.com> for doing the original package and getting it into testing
+
 pkgname=zerotier-one
-pkgver=1.1.14
-pkgrel=1
-pkgdesc="ZeroTier One allows systems to join and participate in ZeroTier virtual networks."
+pkgver=1.2.4
+pkgrel=0
+pkgdesc="Join ZeroTier virtual neteworks from your Alpine Linux system"
 url="https://www.zerotier.com/"
 arch="all"
 license="GPL3"
-depends=""
-makedepends="nodejs linux-headers"
-install=""
-subpackages="$pkgname-doc"
+depends="libstdc++"
+makedepends="linux-headers"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages=""
 source="
-	$pkgname-$pkgver.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/${pkgver}.tar.gz
 	zerotier-one.initd
+	zerotier-one.confd
+	${pkgname}-${pkgver}.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/${pkgver}.tar.gz
 	"
+builddir="$srcdir/ZeroTierOne-${pkgver}"
 
-builddir="$srcdir"/ZeroTierOne-$pkgver
 build() {
 	cd "$builddir"
-	sed -i.bak s:/bash:/sh:g ./doc/build.sh
-	make || return 1
+	make distclean
+	make -j4 one
 }
 
 package() {
 	cd "$builddir"
-	make DESTDIR="$pkgdir" install || return 1
+	make DESTDIR="${pkgdir}" -j1 install
+
 	install -m755 -D "$srcdir"/$pkgname.initd \
-			"$pkgdir"/etc/init.d/$pkgname || return 1
+		"$pkgdir"/etc/init.d/$pkgname || return 1
+	install -m644 -D "$srcdir"/$pkgname.confd \
+		"$pkgdir"/etc/conf.d/$pkgname || return 1
 }
 
-md5sums="5e381f0864797886b3b3bf20beb49bba  zerotier-one-1.1.14.tar.gz
-c6bc90e905d8a66c6dbffdd1d72290bf  zerotier-one.initd"
-sha256sums="d206069ad21c665159cdececb0a20a21758849ad73d91234d709962b26f634af  zerotier-one-1.1.14.tar.gz
-680d01fbac0aff5ecf7238ddc8ce9dcb91c0d19ca14fd83ec5f0949983d256dd  zerotier-one.initd"
-sha512sums="96806a5eab8204d01585c88dba426fb7d47337562074488f87eb81b7ece6643216e2f60adcf1677f9a61fa5ab5187b7569fc1b9565d7055ba106139b6363e7a5  zerotier-one-1.1.14.tar.gz
-a63f8e649d63a3de58a556b3adca440cd0c0d4b36239ea547d555b97852d89d0a1446f348d35e98f77faabe1fe4ffb76868b8290ad9f2b4cd8b6c599945a176c  zerotier-one.initd"
+md5sums="69718d792bb4187781c4bef3ad771f6c  zerotier-one.initd
+44272c03cd44d0a37d6332c598a628dd  zerotier-one.confd
+f0b127bd0919e8af841336cd70a954a6  zerotier-one-1.2.4.tar.gz"
+sha256sums="fb177e15eb1e256f0950325a2ad5de6f59974b5976df1c262c8a38682f25185e  zerotier-one.initd
+5dda94fdaadc624b84e73dfb18b58fbfb88bd00eb5731c248b2f4f2fc6a6a3c7  zerotier-one.confd
+9f275b3732b721f02cc4b8df75b51e6a8fb56dfe1a542a7bd681538d852b0358  zerotier-one-1.2.4.tar.gz"
+sha512sums="57349331f40d3c619974299fc2a6ee3bc4efbc76c5287bbc096bca7d8b21c927145adad95de441a4ca2afc69716172335fcddad880a11a285ffaada310ecd8ff  zerotier-one.initd
+4c9a356ffdd4cd25276e71c1fd252e5dd934a747f46e414d56e4941e088a818967befd6b432619fe0680b41b1649d888a3af2be3a4cd34aa9d09ca3ec9fcec80  zerotier-one.confd
+82adb110208d24ae2745e3839810afcac87955de050ebfe0517a7dc2a875881dafd40c1b16a041742d8c4d0f6513abcc71d6ea3e06c2fb89b47be2630a500363  zerotier-one-1.2.4.tar.gz"
--- a/testing/zerotier-one/zerotier-one.confd
+++ b/testing/zerotier-one/zerotier-one.confd
@@ -1,0 +1,1 @@
+zerotier_opts=""
--- a/testing/zerotier-one/zerotier-one.initd
+++ b/testing/zerotier-one/zerotier-one.initd
@@ -1,11 +1,14 @@
 #!/sbin/openrc-run
 
-depend() {
-    provide net
-    use network
-    before firewall netmount
-}
-
+name=zerotier-one
 command="/usr/sbin/zerotier-one"
+command_args="$zerotier_opts"
 command_background="yes"
-pidfile="/run/$RC_SVCNAME.pid"
+
+start_stop_daemon_args=""
+pidfile="/var/lib/zerotier-one/zerotier-one.pid"
+
+depend() {
+	after net
+	before firewall netmount
+}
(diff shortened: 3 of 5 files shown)


testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
-# Contributor: Kyle Parisi <kyleparisi@gmail.com>
-# Maintainer: Kyle Parisi <kyleparisi@gmail.com>
+# Contributor: ZeroTier, Inc. <contact@zerotier.com>
+# Maintainer: Adam Ierymenko <adam.ierymenko@zerotier.com>
+
+## Thanks to Kyle Parisi <kyleparisi@gmail.com> for doing the original package and getting it into testing
```
Shizmob: There can be multiple contributors, so you can simply keep his entry together with the ZeroTier, Inc. one. Also, did ZeroTier, Inc. themselves actually contribute to this APKBUILD?

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+
 pkgname=zerotier-one
-pkgver=1.1.14
-pkgrel=1
-pkgdesc="ZeroTier One allows systems to join and participate in ZeroTier virtual networks."
+pkgver=1.2.4
+pkgrel=0
+pkgdesc="Join ZeroTier virtual neteworks from your Alpine Linux system"
```
Shizmob: I'm personally more a fan of the old `$pkgdesc`.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
 url="https://www.zerotier.com/"
 arch="all"
 license="GPL3"
-depends=""
-makedepends="nodejs linux-headers"
-install=""
-subpackages="$pkgname-doc"
+depends="libstdc++"
```
Shizmob: This is automatically inferred by abuild (see the Travis logs) and doesn't need to be explicitly specified.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+makedepends="linux-headers"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages=""
```
Shizmob: As there are manpages, please keep the `$pkgname-doc` subpackage. They will be automatically moved to it by abuild.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
 	zerotier-one.initd
+	zerotier-one.confd
+	${pkgname}-${pkgver}.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/${pkgver}.tar.gz
 	"
+builddir="$srcdir/ZeroTierOne-${pkgver}"
 
-builddir="$srcdir"/ZeroTierOne-$pkgver
 build() {
 	cd "$builddir"
-	sed -i.bak s:/bash:/sh:g ./doc/build.sh
```
Shizmob: This change still seems needed, so please don't remove it.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
-	make || return 1
+	make distclean
```
Shizmob: You shouldn't need to do `make distclean` on a clean unpack - is there a specific reason you do this?

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+	make -j4 one
```
Shizmob: abuild automatically sets the JOBS variable, so specifying `-j` manually is both unnecessary and potentially slowing down (the ppc64le build server has far more than 4 cores, for instance :).

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
 }
 
 package() {
 	cd "$builddir"
-	make DESTDIR="$pkgdir" install || return 1
+	make DESTDIR="${pkgdir}" -j1 install
```
Shizmob: Likewise as above, specifying `-j` should not be needed.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+
 	install -m755 -D "$srcdir"/$pkgname.initd \
-			"$pkgdir"/etc/init.d/$pkgname || return 1
+		"$pkgdir"/etc/init.d/$pkgname || return 1
```
Shizmob: Can you remove the `|| return 1` stanzas here? With modern (edge) abuild, `set -e` is implicit.

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+		"$pkgdir"/etc/conf.d/$pkgname || return 1
 }
 
-md5sums="5e381f0864797886b3b3bf20beb49bba  zerotier-one-1.1.14.tar.gz
-c6bc90e905d8a66c6dbffdd1d72290bf  zerotier-one.initd"
-sha256sums="d206069ad21c665159cdececb0a20a21758849ad73d91234d709962b26f634af  zerotier-one-1.1.14.tar.gz
-680d01fbac0aff5ecf7238ddc8ce9dcb91c0d19ca14fd83ec5f0949983d256dd  zerotier-one.initd"
-sha512sums="96806a5eab8204d01585c88dba426fb7d47337562074488f87eb81b7ece6643216e2f60adcf1677f9a61fa5ab5187b7569fc1b9565d7055ba106139b6363e7a5  zerotier-one-1.1.14.tar.gz
-a63f8e649d63a3de58a556b3adca440cd0c0d4b36239ea547d555b97852d89d0a1446f348d35e98f77faabe1fe4ffb76868b8290ad9f2b4cd8b6c599945a176c  zerotier-one.initd"
+md5sums="69718d792bb4187781c4bef3ad771f6c  zerotier-one.initd
```
Shizmob: `md5sums=` and `sha256sums=` is unused and thus not needed anymore.

testing/zerotier-one/zerotier-one.initd
```diff
@@ -1,11 +1,14 @@
 #!/sbin/openrc-run
 
-depend() {
-    provide net
-    use network
-    before firewall netmount
-}
-
+name=zerotier-one
```
Shizmob: Setting `$name` shouldn't be explicitly needed if it's equivalent to the filename.

testing/zerotier-one/zerotier-one.initd
```diff
@@ -1,11 +1,14 @@
...
 command="/usr/sbin/zerotier-one"
+command_args="$zerotier_opts"
```
Shizmob: Could this be capitalised, in line with common practice for `conf.d` settings?

testing/zerotier-one/zerotier-one.initd
```diff
@@ -1,11 +1,14 @@
...
 command_background="yes"
-pidfile="/run/$RC_SVCNAME.pid"
+
+start_stop_daemon_args=""
```
Shizmob: Please omit this if it's empty.

testing/zerotier-one/zerotier-one.post-deinstall
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+/usr/sbin/deluser zerotier-one
```
Shizmob: No need to use full paths here.

testing/zerotier-one/zerotier-one.post-deinstall
```diff
@@ -0,0 +1,7 @@
...
+
+cd /var/lib/zerotier-one
```
Shizmob: Isn't it better to just `rm -r /var/lib/zerotier-one`?

testing/zerotier-one/zerotier-one.pre-install
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+/usr/sbin/addgroup -S zerotier-one
```
Shizmob: Likewise, no need to use full paths here.

testing/zerotier-one/zerotier-one.pre-install
```diff
@@ -0,0 +1,4 @@
...
+/usr/sbin/adduser -D -H -S -h /var/lib/zerotier-one -s /sbin/nologin -g 'https://www.zerotier.com/' -G zerotier-one zerotier-one
```
Shizmob: The GECOS is probably not needed, and `-s /sbin/nologin` and `-D` are implicit for system (`-S`) accounts, they can be omitted. Finally, if you omit `-G zerotier-one` you can omit the `addgroup` call above entirely. :)

Shizmob:
I think you may have forgotten to fill in some variables in your commit message. :)

Thanks for contributing, if you fix up these things I'll gladly merge it!

testing/zerotier-one/zerotier-one.pre-install
jirutka: This actually follows common pattern we currently use in most of the packages. Except this GECOS with URL, this should be removed.

testing/zerotier-one/zerotier-one.post-deinstall
jirutka: This post-deinstall script is evil and must be removed entirely.

testing/zerotier-one/zerotier-one.initd
jirutka: No, capitalized variables are just legacy practice. However, this variable is unnecessary, it’d be better to just remove this line and let users define `command_args` in conf.d if needed.

Shizmob: I'm not sure it's a good idea to have `conf.d` set OpenRC options automatically. It may get wonky if we want to add our own arguments.

testing/zerotier-one/zerotier-one.pre-install
Shizmob: Is there any reason for this pattern?

testing/zerotier-one/zerotier-one.initd
```diff
@@ -1,11 +1,14 @@
...
+pidfile="/var/lib/zerotier-one/zerotier-one.pid"
```
jirutka: PID files should be in `/run` (typically `/run/$RC_SVCNAME.pid`), definitely not in `/var/lib`!

testing/zerotier-one/zerotier-one.confd
```diff
@@ -0,0 +1 @@
+zerotier_opts=""
```
jirutka: This file is unnecessary.

testing/zerotier-one/APKBUILD
jirutka: Why have you added unnecessary braces around `pkgdir`?

testing/zerotier-one/APKBUILD
```diff
@@ -1,38 +1,48 @@
...
+makedepends="linux-headers"
+install="$pkgname.pre-install $pkgname.post-deinstall"
+subpackages=""
 source="
-	$pkgname-$pkgver.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/${pkgver}.tar.gz
 	zerotier-one.initd
+	zerotier-one.confd
+	${pkgname}-${pkgver}.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/${pkgver}.tar.gz
 	"
+builddir="$srcdir/ZeroTierOne-${pkgver}"
```
jirutka: Why have you add unnecessary braces around `pkgver`?

testing/zerotier-one/zerotier-one.pre-install
jirutka: Does zerotier-one actually use this user? It’s started as root…

testing/zerotier-one/zerotier-one.confd
Shizmob: I don't believe so, it's useful for documentation. Most files we have in `/etc/conf.d` do nothing functional, but document the possible variables. Even if we were to go with `command_args`, we can't expect end-users to know about OpenRC internals. :)

testing/zerotier-one/zerotier-one.pre-install
Shizmob: The daemon seems to drop privileges to this hardcoded user: https://github.com/zerotier/ZeroTierOne/blob/fe5257df81c4ec4b5d48f707eb794de0748b7ac0/one.cpp#L958

testing/zerotier-one/zerotier-one.confd
jirutka: Well, that’s true, but only when there are some extra variables not defined by OpenRC. However, I’m not strongly against adding this file, so okay.

testing/zerotier-one/zerotier-one.initd
jirutka: It’s fine, we can always add `command_args="--some-args --more-args $command_args"`.

testing/zerotier-one/zerotier-one.pre-install
jirutka: I don’t remember. Please remind me this topic tomorrow. :)

PR #1380, opened 2017-05-11
community/elasticsearch: update to 5.4.2 / add bash depends

itoffshore:
* also fixes plugin path:

`ES` expects to find plugins in the same path as the libs. Looking at
the java command line in htop `ES` sets `path.home` a 2nd time for the
libs & expects the plugins to also be in this path (which is why
there is no configuration for the plugin path in `elasticsearch.yml`).

### DIFF
--- a/community/elasticsearch/APKBUILD
+++ b/community/elasticsearch/APKBUILD
@@ -1,14 +1,15 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=elasticsearch
-pkgver=5.2.1
+pkgver=5.4.2
 pkgrel=0
 pkgdesc="Open Source, Distributed, RESTful Search Engine"
 url="https://www.elastic.co/products/elasticsearch"
 arch="x86 x86_64"
 license="ASL-2.0"
-depends="java-jna-native>=4.1 openjdk8-jre"
+depends="java-jna-native>=4.1 openjdk8-jre bash"
 makedepends=""
+options="!check"
 install="$pkgname.pre-install"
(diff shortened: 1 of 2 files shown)


community/elasticsearch/APKBUILD
```diff
@@ -47,23 +32,22 @@ package() {
...
 
 	# ES bin script parses the new jvm.options file
-	install -m755 -t "$destdir"/bin bin/* || return 1
+	install -m755 -t "$destdir"/bin bin/*
 
-	# ES does not run without a transport module
-	install -dm755 "$destdir/modules/$default_module"
-	install -m644 -t "$destdir/modules/$default_module" modules/"$default_module"/* || return 1
+	# Modules are no longer subpkgs
+	cp -rf modules $destdir/
```
jirutka: Why?!

itoffshore: removed as [requested by @ncopa  on 24th March](https://github.com/alpinelinux/aports/pull/1070)

jirutka: You have misinterpreted him. He was just against the method you’ve used, not against subpackages.

itoffshore: so wrapping both your `for` & my `find` in an `if` as I suggested on the 24th is ok ?

without doing this the `for` would run once unnecessarily.

jirutka: No, don’t use find here…

itoffshore: I also asked on March 24th if modules should be added manually & got no response

is fixed now

itoffshore:
can this be pulled now it is fixed ?

itoffshore:
`elasticsearch` is still broken until this PR is applied

PR #1381, opened 2017-05-11
community/php7: upgrade to 7.1.5

andypost:
http://php.net/archive/2017.php#id2017-05-11-2


Shizmob:
@vakartel do PHP extensions need to be recompiled for this change?

vakartel:
@Shizmob no need. the only need to rebuild if the php-api version is changed

PR #1383, opened 2017-05-12
testing/etckeeper: add default_prepare in prepare()

HRio:


HRio:
@jirutka sigh now I actually apply the patch, sorry about that.

PR #1392, opened 2017-05-12
main/unbound: add confd / modernize initd

itoffshore:
* `confd` with `$command_args` (for optional debugging) added

* `initd` format modernized & deprecated variables `$SVCNAME` changed to `$RC_SVCNAME`

### DIFF
--- a/main/unbound/unbound.initd
+++ b/main/unbound/unbound.initd
@@ -1,7 +1,5 @@
 #!/sbin/openrc-run
-# Copyright 1999-2011 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
-# $Header: /var/cvsroot/gentoo-x86/net-dns/unbound/files/unbound.initd,v 1.2 2011/09/13 15:14:49 matsuu Exp $
 
 name="unbound daemon"
 extra_commands="configtest"
@@ -10,10 +8,13 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host 
 description_configtest="Run syntax tests for configuration files only."
 description_reload="Kills all children and reloads the configuration."
 
+# Upper case variables are here only for backward compatibility.
+: ${cfgfile:=${UNBOUND_CONFFILE:-/etc/unbound/$RC_SVCNAME.conf}}
 
-UNBOUND_BINARY=${UNBOUND_BINARY:-/usr/sbin/unbound}
-UNBOUND_CHECKCONF=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}
-UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${SVCNAME}.conf}
+command=/usr/sbin/unbound
+command_args="-d $command_args"
+command_background=yes
+checkconf=/usr/sbin/unbound-checkconf
 
 depend() {
 	need net
@@ -22,36 +23,22 @@ depend() {
 	after auth-dns
 }
 
-checkconfig() {
-	checkpath --directory /var/run/unbound
-	UNBOUND_PIDFILE=$("${UNBOUND_CHECKCONF}" -o pidfile "${UNBOUND_CONFFILE}")
+configtest() {
+	einfo "Checking: $cfgfile" # don't show errors twice
+	pidfile=$($checkconf -o pidfile "$cfgfile" 2>/dev/null)
+	$checkconf -f $cfgfile 1>/dev/null
 	return $?
 }
 
-configtest() {
-	ebegin "Checking ${SVCNAME} configuration"
-	checkconfig
-	eend $?
-}
-
-start() {
-	checkconfig || return $?
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --pidfile "${UNBOUND_PIDFILE}" \
-		--exec "${UNBOUND_BINARY}" -- -c "${UNBOUND_CONFFILE}"
-	eend $?
-}
-
-stop() {
-	checkconfig || return $?
-	ebegin "Stopping ${SVCNAME}"
-	start-stop-daemon --stop --pidfile "${UNBOUND_PIDFILE}"
-	eend $?
+start_pre() {
+	configtest || return $?
+	checkpath --directory ${pidfile%/*}
+	einfo "Using pidfile: $pidfile"
 }
 
 reload() {
-	checkconfig || return $?
-	ebegin "Reloading ${SVCNAME}"
-	start-stop-daemon --signal HUP --pidfile "${UNBOUND_PIDFILE}"
+	configtest || return $?
+	ebegin "Reloading ${RC_SVCNAME}"
+	start-stop-daemon --signal HUP --pidfile "$pidfile"
 	eend $?
 }
--- a/main/unbound/APKBUILD
+++ b/main/unbound/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=unbound
 pkgver=1.6.7
-pkgrel=1
+pkgrel=2
 pkgdesc="Unbound is a validating, recursive, and caching DNS resolver"
 pkgusers="unbound"
 pkggroups="unbound"
@@ -21,7 +21,9 @@ source="http://unbound.net/downloads/unbound-$pkgver.tar.gz
 	update-unbound-root-hints
 	migrate-dnscache-to-unbound
 	root.hints
-	unbound.initd"
+	unbound.initd
+	unbound.confd
+	"
 builddir="$srcdir"/$pkgname-$pkgver
 
 build() {
@@ -63,6 +65,9 @@ package() {
 
 	install -m755 -D "$srcdir"/unbound.initd \
 		"$pkgdir"/etc/init.d/unbound
+	install -m755 -D "$srcdir"/unbound.confd \
+		"$pkgdir"/etc/conf.d/unbound
+
 	install -d -o unbound -g unbound "$pkgdir"/var/run/unbound
 	install -m644 "$srcdir"/root.hints "$pkgdir"/etc/unbound/
 
@@ -95,4 +100,5 @@ sha512sums="6e3d1a057081252183343d0d1b8ace742ab15e8f5244e61287340f49289d7449bed9
 b16b7b15392c0d560718ee543f1eebc5617085fb30d61cddc20dd948bd8b1634ee5b2de1c9cb172a6c0d1c5bbaf98b6fd39816d39c72a43ff619455449e668ac  update-unbound-root-hints
 b26a13c1c88da9611a65705dc59f7233c5e0f6aced0d7d66c18536a969a2de627ca5d4bb55eedd81f2f040fa11bde48eaaeca2850f376e72e7a531678a259131  migrate-dnscache-to-unbound
 0dca3470ed4ca9b76d6f47f5d20e92924e6648f0870d8594fe6735d8f1cdfeeee7296301066c2a8b2b94f7daed86c15efe00c301ca27e435e5dd2c85508dc9c8  root.hints
-540e7a11fa5421e2d103c42d69faf1ba005adcadfac2f65091795a2f00e5b5acd1436b4d2adfe2bb0fdfcbfb44d0967d6bce87620c618549fcd7e32019040f29  unbound.initd"
+ae02d4b6724684a0b2f52c1a64bb37fff8576407eb4c35b295025a73ec03b469c975c9e16a5d461ec4a984144524059cbd3d8a6513df390aed278e1b54aa1733  unbound.initd
+5cc3046fb0a0e41cf48d2af2ea0944e734a68c15a13aa306b7e70ac3aa3ba0edf379100375589407b24dde48c960c6b6823a82d863ee18c5bb2ab03f4adaa05c  unbound.confd"
--- a/main/unbound/unbound.confd
+++ b/main/unbound/unbound.confd
@@ -1,0 +1,3 @@
+# enable debugging (verbose -vv)
+# command_args="-v"
+


ncopa:
We should keep the quotes for paths, otherwise there will be problems with paths with spaces.

Try `mkdir -p "/run/test path/unbound"` and set UNBOUND_PIDFILE="/run/test path/unbound"

itoffshore:
quoting `${UNBOUND_OPTS}` breaks the `initd` if `${UNBOUND_OPTS}` is empty (which it is by `default`)

all other quotes reverted & rebased

Shizmob:
Ideally I'd like to see this initscript changed to simply use `command=`, `command_args` and `command_background=yes` and none of the start-stop-daemon stuff...

itoffshore:
`initd` modernized with `start-stop-daemon` completely removed

main/unbound/unbound.initd
```diff
@@ -10,10 +7,13 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
...
 
-
 UNBOUND_BINARY=${UNBOUND_BINARY:-/usr/sbin/unbound}
 UNBOUND_CHECKCONF=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}
-UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${SVCNAME}.conf}
+UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${RC_SVCNAME}.conf}
+
+command=${UNBOUND_BINARY}
+command_args=${UNBOUND_OPTS}
+command_background=yes
```
Shizmob: You need to pass -d to unbound so that it stays in the foreground and can be controlled by OpenRC:

`command_args="-d $UNBOUND_OPTS"`

Shizmob:
Thanks! Aside from the thing I just noted, our style is to not use variable braces if not necessary. If you fix those two things up it looks good to me!

main/unbound/unbound.initd
itoffshore: isn't forcing  `foreground` required only when using `supervise-daemon` ?

`unbound` starts & stops just fine as is

Shizmob: It's required when `command_background=yes`. It will start fine this way yes, but OpenRC will not (or diminishingly so) be able to detect when it has crashed. We prefer it this way (`command_background=yes` and forcing daemons on foreground) because it relies less on start-stop-daemon semantics and gives us flexibility to switch globally between supervisor backends in the future.

itoffshore:
fixed & rebased

main/unbound/unbound.initd
```diff
@@ -10,10 +10,13 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
 description_configtest="Run syntax tests for configuration files only."
 description_reload="Kills all children and reloads the configuration."
 
-
 UNBOUND_BINARY=${UNBOUND_BINARY:-/usr/sbin/unbound}
 UNBOUND_CHECKCONF=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}
-UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${SVCNAME}.conf}
+UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${RC_SVCNAME}.conf}
+
+command=$UNBOUND_BINARY
```
jirutka: Please remove this variable and add `/usr/sbin/unbound` directly. We really don’t expect, nor support, that users will move installed binaries. ;)

main/unbound/unbound.initd
```diff
@@ -10,10 +10,13 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
 description_configtest="Run syntax tests for configuration files only."
 description_reload="Kills all children and reloads the configuration."
 
-
 UNBOUND_BINARY=${UNBOUND_BINARY:-/usr/sbin/unbound}
 UNBOUND_CHECKCONF=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}
-UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${SVCNAME}.conf}
+UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${RC_SVCNAME}.conf}
```
jirutka: Please rename this to `cfgfile` (common name used for configs) and keep `UNBOUND_CONFFILE` as an alternative for backward compatibility.

```sh
# Upper case variables are here only for backward compatibility.
: ${cfgfile:=${UNBOUND_CONFFILE:-/etc/unbound/$RC_SVCNAME.conf}}
```

main/unbound/unbound.initd
```diff
@@ -10,10 +10,13 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
...
+
+command=$UNBOUND_BINARY
+command_args="-d $UNBOUND_OPTS"
```
jirutka: Please replace `UNBOUND_OPTS` with `command_args`:

```sh
command_args="-d $command_args"
```

main/unbound/unbound.initd
```diff
@@ -22,36 +25,20 @@ depend() {
...
+	configtest || return $?
 }
 
 reload() {
-	checkconfig || return $?
-	ebegin "Reloading ${SVCNAME}"
-	start-stop-daemon --signal HUP --pidfile "${UNBOUND_PIDFILE}"
+	configtest || return $?
+	ebegin "Reloading $RC_SVCNAME"
+	kill -SIGHUP $(head -n 1 $pidfile)
```
jirutka: Why have you replaced `start-stop-daemon --signal HUP --pidfile "$pidfile"` with `kill -SIGHUP $(head -n 1 "$pidfile")`?

main/unbound/unbound.initd
```diff
@@ -22,36 +25,20 @@ depend() {
...
-	checkpath --directory /var/run/unbound
-	UNBOUND_PIDFILE=$("${UNBOUND_CHECKCONF}" -o pidfile "${UNBOUND_CONFFILE}")
-	return $?
-}
-
 configtest() {
-	ebegin "Checking ${SVCNAME} configuration"
-	checkconfig
+	ebegin "Checking $RC_SVCNAME configuration"
+	pidfile=$("$UNBOUND_CHECKCONF" -o pidfile "$UNBOUND_CONFFILE")
```
jirutka: pidfile location should be defined in runscript, not in the program’s configuration file. It’s init/rc’s system responsibility to create and maintain pidfile (if it needs them).

main/unbound/unbound.initd
itoffshore: `UNBOUND_OPTS` is configured in `/etc/conf.d/unbound`

main/unbound/unbound.initd
itoffshore: it is there because of the extra command - configtest.

I will move it & remove the extra command

main/unbound/unbound.initd
jirutka: You can set `command_args` in conf.d too.;) And it's standard variable, not custom, different for each pkg.

main/unbound/unbound.initd
jirutka: That's not what I meant, checkconf is okay, but why it sets pidfile variable?

itoffshore:
It is how gentoo currently does it:

https://data.gpo.zugaina.org/gentoo/net-dns/unbound/files/unbound.initd

but I will change it

main/unbound/unbound.initd
```diff
@@ -22,36 +23,20 @@ depend() {
...
-	eend $?
-}
-
-start() {
-	checkconfig || return $?
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --pidfile "${UNBOUND_PIDFILE}" \
-		--exec "${UNBOUND_BINARY}" -- -c "${UNBOUND_CONFFILE}"
+	ebegin "Checking $RC_SVCNAME configuration"
+	"$checkcfg" "$cfgfile" 1>/dev/null || return $?
```
Shizmob: The `|| return $?` is not needed, `eend` will take care of that.

main/unbound/unbound.initd
```diff
@@ -22,36 +23,20 @@ depend() {
...
 }
 
-stop() {
-	checkconfig || return $?
-	ebegin "Stopping ${SVCNAME}"
-	start-stop-daemon --stop --pidfile "${UNBOUND_PIDFILE}"
-	eend $?
+start_pre() {
+	checkpath --directory /run/unbound
+	configtest || return $?
```
Shizmob: The `|| return $?` can be omitted, the last return code in a function is the function return code.

main/unbound/unbound.initd
```diff
@@ -22,36 +23,20 @@ depend() {
...
 }
 
 reload() {
-	checkconfig || return $?
-	ebegin "Reloading ${SVCNAME}"
-	start-stop-daemon --signal HUP --pidfile "${UNBOUND_PIDFILE}"
+	configtest || return $?
+	ebegin "Reloading $RC_SVCNAME"
+	start-stop-daemon --signal HUP --pidfile $pidfile
```
Shizmob: Can `$pidfile` be quoted?

main/unbound/unbound.initd
itoffshore: the `|| return $?` after `configtest` should be included - I tested the `reload` command with a config file with errors:

```
musl64 [~/aports/main/unbound]$ sudo rc-service unbound reload
 * Checking unbound configuration ...
/etc/unbound/unbound.conf:11: error: unknown keyword 'include'
/etc/unbound/unbound.conf:11: error: stray '"'
/etc/unbound/unbound.conf:11: error: unknown keyword '/etc/unbound/local.d/*.conf'
/etc/unbound/unbound.conf:11: error: stray '"'
read /etc/unbound/unbound.conf failed: 4 errors in configuration file
 * Reloading unbound ...
start-stop-daemon: option requires an argument: pidfile
Usage: start-stop-daemon [options] 

Options: [ I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:1:2:ChqVv ]
  -I, --ionice <arg>                Set an ionice class:data when starting
  -K, --stop                        Stop daemon
  -N, --nicelevel <arg>             Set a nicelevel when starting
  -R, --retry <arg>                 Retry schedule to use when stopping
  -S, --start                       Start daemon
  -a, --startas <arg>               deprecated, use --exec or --name
  -b, --background                  Force daemon to background
  -c, --chuid <arg>                 deprecated, use --user
  -d, --chdir <arg>                 Change the PWD
  -e, --env <arg>                   Set an environment string
  -k, --umask <arg>                 Set the umask for the daemon
  -g, --group <arg>                 Change the process group
  -i, --interpreted                 Match process name by interpreter
  -m, --make-pidfile                Create a pidfile
  -n, --name <arg>                  Match process name
  -o, --oknodo                      deprecated
  -p, --pidfile <arg>               Match pid found in this file
  -s, --signal <arg>                Send a different signal
  -t, --test                        Test actions, don't do them
  -u, --user <arg>                  Change the process user
  -r, --chroot <arg>                Chroot to this directory
  -w, --wait <arg>                  Milliseconds to wait for daemon start
  -x, --exec <arg>                  Binary to start/stop
  -1, --stdout <arg>                Redirect stdout to file
  -2, --stderr <arg>                Redirect stderr to file
  -P, --progress                    Print dots each second while waiting
  -h, --help                        Display this help output
  -C, --nocolor                     Disable color output
  -V, --version                     Display software version
  -v, --verbose                     Run verbosely
  -q, --quiet                       Run quietly (repeat to suppress errors)     
```

Shizmob: Huh okay, odd. Fine by me then.

jirutka: This is obviously non-sense. `|| return $?` after `configtest` is needed only in `reload()`, not in `start_pre`, for the exact reason @Shizmob said.

main/unbound/unbound.initd
```diff
@@ -10,10 +7,14 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
 description_configtest="Run syntax tests for configuration files only."
 description_reload="Kills all children and reloads the configuration."
 
+# Upper case variables are here only for backward compatibility.
+: ${cfgfile:=${UNBOUND_CONFFILE:-/etc/unbound/$RC_SVCNAME.conf}}
+: ${checkcfg:=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}}
```
jirutka: Once again, we do not expect, nor support, moving installed binaries to different locations. It doesn’t make any sense to make location of `unbound-checkconf` configurable.

main/unbound/unbound.initd
```diff
@@ -10,10 +7,14 @@ description="unbound is a Domain Name Server (DNS) that is used to resolve host
...
 
-UNBOUND_BINARY=${UNBOUND_BINARY:-/usr/sbin/unbound}
-UNBOUND_CHECKCONF=${UNBOUND_CHECKCONF:-/usr/sbin/unbound-checkconf}
-UNBOUND_CONFFILE=${UNBOUND_CONFFILE:-/etc/unbound/${SVCNAME}.conf}
+command=/usr/sbin/unbound
+command_args="-d $command_args"
+pidfile=$("$checkcfg" -o pidfile "$cfgfile" 2>/dev/null)
```
jirutka: Once again, pidfile location should be hard-coded in runscript (as `/run/$RC_SVCNAME.pid`), not read from program’s configuration file.

main/unbound/unbound.initd
itoffshore: changed

main/unbound/unbound.initd
itoffshore: changed

main/unbound/unbound.initd
itoffshore: changed

main/unbound/unbound.initd
itoffshore: quoted

itoffshore:
@jirutka - some extra changes are required if you want the `pidfile` set in the `initd`.

to make it work I've added a `sed` command to alter the `cfgfile`.

nmeum:
I would personally prefer to keep using the Gentoo OpenRC script, simply because it is easier to copy it from Gentoo instead of writing and maintaining our own. At least try to integrate your changes to the init.d script into the Gentoo ebuild.

jirutka:
Alpine is not distribution that prefer easier to correct.

If you wanna contribute our changes to Gentoo, go ahead. But I'm strongly against copying wrong runscripts from Gentoo just because it's easier.

nmeum:
> But I'm strongly against copying wrong runscripts from Gentoo just because it's easier.

Well…if it is **wrong** a bug report should be sent to the Gentoo folks. Fixing bugs in their runscripts without telling them about it seems stupid.

jirutka:
> Fixing bugs in their runscripts without telling them about it seems stupid.

I’m usually not fixing bugs in their runscripts which someone blindly copied into aports, I just remove them and rewrite from scratch, to follow up-to-date rules how to write runscripts (e.g. not redefine `start` and `stop` functions).

Also Gentoo is not our upstream. (Please note that I’m not against Gentoo, I used Gentoo for 5 years.)

> …working together with the Gentoo folks on runscripts would be a good idea.

I agree, but I’ve already tried it without success. I believe there are still some opened PRs in gentoo with small fixes in runscripts, it will be more than 2 years… Speaking about time, fixing runscripts from Gentoo instead of rewriting them in clean way for Alpine is _not_ easier (for me), it’s more time consuming.

main/unbound/unbound.initd
```diff
@@ -22,36 +22,21 @@ depend() {
...
-	eend $?
-}
-
-start() {
-	checkconfig || return $?
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --pidfile "${UNBOUND_PIDFILE}" \
-		--exec "${UNBOUND_BINARY}" -- -c "${UNBOUND_CONFFILE}"
+	ebegin "Checking $RC_SVCNAME configuration"
+	sed -i "s|.*pidfile:.*|pidfile:  \"$pidfile\"|" "$cfgfile"
```
jirutka: No, don’t do that. Start unbound in foreground (`-d`) and let OpenRC daemonize it and manage pidfile. That’s the job of rc system, not every program…

kaniini:
I prefer to rewrite runscripts where possible to use the newer declarative syntax, as it will allow us to more easily transition away from OpenRC in the future.

main/unbound/unbound.initd
itoffshore: it already does - see line 12

`command_args="-d $command_args"`

itoffshore: `unbound` uses the pid according to it's config file so the `sed` is necessary if you want the pid set in the initd

itoffshore: I added a #comment to the initd to explain the `sed`

itoffshore: checksums fixed

main/unbound/unbound.initd
```diff
@@ -22,36 +22,22 @@ depend() {
...
-}
-
-start() {
-	checkconfig || return $?
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --pidfile "${UNBOUND_PIDFILE}" \
-		--exec "${UNBOUND_BINARY}" -- -c "${UNBOUND_CONFFILE}"
+	ebegin "Checking $RC_SVCNAME configuration"
+	# unbound uses the pid in $cfgfile & we set it via initd
+	sed -i "s|.*pidfile:.*|pidfile:  \"$pidfile\"|" "$cfgfile"
```
ncopa: We cannot have init.d script modify users configuration files.

We should extract the pidfile from unbounds config (like current init.d script does) and use that instead of hardcode pidfile in init.d script and rewrite users config.

main/unbound/unbound.initd
```diff
@@ -22,36 +22,22 @@ depend() {
...
 	eend $?
 }
 
-stop() {
-	checkconfig || return $?
-	ebegin "Stopping ${SVCNAME}"
-	start-stop-daemon --stop --pidfile "${UNBOUND_PIDFILE}"
-	eend $?
+start_pre() {
+	checkpath ${pidfile%/*} 2>/dev/null
```
ncopa: Shouldn't it be `checkpath --directory` here?

I don't like that we hide stderr, how would user otherwise know if something went wrong?

main/unbound/unbound.initd
itoffshore: yes I agree - the original form of my PR which simply added support to turn on `debug` logging is all that should be applied - not hacks like this so the `pid` is set in the `initd`

main/unbound/unbound.initd
itoffshore: whole PR redone - should be all ok now:
```
root@dns [/etc/init.d]# rc-service unbound start
 * Checking: /etc/unbound/unbound.conf
 * Using pidfile: /run/unbound.pid
 * Starting unbound daemon ... [ ok ]
```

main/unbound/unbound.initd
jirutka: The best option would be to start unbound in foreground, don't let it create pidfile and let init system do its job.

itoffshore: any reason why this is still not pulled ???

jirutka: It’s a wrong approach.

> The best option would be to start unbound in foreground, don't let it create pidfile and let init system do its job.

itoffshore: if you start it in the foreground the `initd` never exits.

if anything the `-d` needs to be removed as running in the foreground breaks the `initd`

itoffshore: `-d` removed

jirutka: `command_background="yes"`. @Shizmob has already mentioned it few comments above.

itoffshore: yes it has `command_background="yes"` - so setting `-d` has zero effect so I removed the `-d`

itoffshore: broken-unbound

broken unbound `initd`

itoffshore: working-pr-unbound

working unbound `initd` running in background as in the `PR`

jirutka: This is a misunderstanding.

> The best option would be to **start unbound in foreground**…

This means start the unbound program in foreground, i.e. do not let it daemonize itself, i.e. start it without `-d` option.

>> if you start it in the foreground the initd never exits.
> command_background="yes". Shizmob has already mentioned it few comments above.

I thought that you’ve started the unbound program in foreground **without** `command_background="yes"` – that’s exactly what would happen in that case.

* * *
I admit that I’ve overlooked your last commit and assumed that it’s still in the same state as before, based on your comments.

main/unbound/unbound.initd
```diff
@@ -22,36 +23,22 @@ depend() {
...
-}
-
-start() {
-	checkconfig || return $?
-	ebegin "Starting ${SVCNAME}"
-	start-stop-daemon --start --pidfile "${UNBOUND_PIDFILE}" \
-		--exec "${UNBOUND_BINARY}" -- -c "${UNBOUND_CONFFILE}"
-	eend $?
+	einfo "Checking: $cfgfile" # don't show errors twice
+	pidfile=$($checkconf -o pidfile "$cfgfile" 2>/dev/null)
```
jirutka: Why do you do that? Unbound spreads its pidfiles over FS even when you start it without `-d`? :(

main/unbound/unbound.initd
itoffshore: `-d` in unbound means foreground

```
root@dns2 [~]# unbound -h | grep "\-d"
-d	do not fork into the background.
```

main/unbound/unbound.initd
itoffshore: to prevent:
```
root@dns2 [/etc/init.d]# rc-service unbound start
 * Checking: /etc/unbound/unbound.conf
Usage: checkpath [options] path1 [path2] [...]

Options: [ dDfFpm:o:WChqVv ]
  -d, --directory                   Create a directory if not exists
  -D, --directory-truncate          Create/empty directory
  -f, --file                        Create a file if not exists
  -F, --file-truncate               Truncate file
  -p, --pipe                        Create a named pipe (FIFO) if not exists
  -m, --mode <arg>                  Mode to check
  -o, --owner <arg>                 Owner to check (user:group)
  -W, --writable                    Check whether the path is writable or not
  -h, --help                        Display this help output
  -C, --nocolor                     Disable color output
  -V, --version                     Display software version
  -v, --verbose                     Run verbosely
  -q, --quiet                       Run quietly (repeat to suppress errors)
 * Using pidfile: 
 * Starting unbound daemon ... * command_background option used but no pidfile specified
                                                                                               [ !! ]
 * ERROR: unbound failed to start
```

itoffshore: from:
```
root@dns2 [/etc/init.d]# grep -A5 ^configtest unbound
configtest() {
	einfo "Checking: $cfgfile" # don't show errors twice
#	pidfile=$($checkconf -o pidfile "$cfgfile" 2>/dev/null)
	$checkconf -f $cfgfile 1>/dev/null
	return $?
}
```

itoffshore: this `PR` is correct & works:
```
root@dns2 [~]# rc-service unbound start
 * Caching service dependencies ... [ ok ]
 * Checking: /etc/utnbound/unbound.conf
 * Using pidfile: /run/unbound.pid
 * Starting unbound daemon ... [ ok ]
```

jirutka:
I finally verified it myself. Unbound works correctly with `-d` (run in foreground) – it does not create a pidfile, does not complain that it’s not set etc. So there’s no problem with handling it properly – configure OpenRC to daemonize it manage pidfile for it, without messing with config etc. I’m gonna fix this PR and finally merge it.

itoffshore:
I've put the `-d` back in but it still has zero effect with the command backgrounded

jirutka:
Yes, it looks like it has zero effect, because you set the same location of pidfile for both OpenRC and unbound. OpenRC starts unbound, the process forks itself and overwrites pidfile created by OpenRC with its new PID, so everything works. When you add `-d`, unbound does not create/overwrite pidfile and only OpenRC manages it.

PR #1394, opened 2017-05-13
community/php5: remaded to the same as php7

vakartel:
Remaded APKBUILD and related files to the same as in current community/php7


vakartel:
@jirutka 
I remaded it with php7/APKBUILD you made and add all php5-specific code and patches.
Somehow mostly all tests are failed so I just disable check phase.
So if it pass, php7 and php5 would have identical behavior and any future changes in APKBUILD need to be done in both php5 and php7 ones.

skwashd:
@jirutka / @vakartel any news on this one? I would like to get #1572 merged so composer can be used with either PHP 5.6 or 7.x. Should this continue to be a blocker for for the meta package PR?

andypost:
@vakartel that's still makes sense

skwashd:
@andypost what is the plan for supporting PHP 5.6? Security support ends in less than 10 months.

jirutka:
> @vakartel that's still makes sense

IMHO it’s not worth changing now. As @skwashd wrote, 5.6 is near EOL, so we will drop it after after v3.8 release.

community/php5/APKBUILD
```diff
@@ -1,514 +1,652 @@
...
-		--without-db1 \
-		--without-db2 \
-		--without-db3 \
-		--without-qdbm \
-		--with-mssql=shared \
-		--with-pdo-dblib=shared \
-		--enable-opcache \
-		"
+			--with-libzip=/usr \
+		--with-zlib=shared \
```
andypost: zlib should be bundled https://github.com/alpinelinux/aports/pull/2951

vakartel:
IMHO this PR have no sense now because EOL

andypost:
EOL is a year ahead so before 3.8 better to have php5 maintainable

jirutka:
Applying single commit with +955/-407 changes, which is basically impossible to review, would not make it better maintainable. It’s IMHO not worth the risk of possible breakages and unclear changes in git log.

PR #1396, opened 2017-05-13
community/binaryen: disable on s390x

tmh1999:
Probably bug on endianness detection

--- expected
+++ actual
@@ -3,7 +3,7 @@
  (import "env" "memory" (memory $0 1))
  (table 2 2 anyfunc)
  (elem (i32.const 0) $__wasm_nullptr $__exit)
\- (data (i32.const 16) "\d2\04\00\00\00\00\00\00)\t\00\00")
\+ (data (i32.const 16) "\00\00\04\d2\00\00\00\00\00\00\t)")


jirutka:
It fails only in few tests, so most likely it’s bug only in tests. Thus it may be safe to ignore failing tests on s390x for now instead of disabling package.

I’ve reported it to upstream https://github.com/WebAssembly/binaryen/issues/1012.

tmh1999:
I guess it is fatal enough to disable the package. In one of the previous build logs, it shows no support for big endian, which is s390x. The log is probably overridden on the builder.

jirutka:
Aha, okay.

PR #1397, opened 2017-05-13
community/ddrescue: disable test on s390x

tmh1999:
proably due to builder, test good on dev machine


jirutka:
I’d prefer to let the tests run, just ignore failure on s390x.

```sh
check() {
	cd "$builddir"

	# XXX: Some tests fail on s390x, but only on builder, so ignore it for now.
	case "$CARCH" in
		s390x) make check || true;;
		*) make check;;
	esac
}
```

tmh1999:
Fixed as requested.

PR #1399, opened 2017-05-13
community/ioping: fix bug s390x builder

tmh1999:
since s390x builder is lxc, the host kernel returns EINVAL when
fadvice() is called with POSIX_FADV_DONTNEED which gives advice to the kernel to free the test file after running ioping.
Using POSIX_FADV_NORMAL gives no advice about the test file for future
usage. Temp fix for 3.6.

### DIFF
--- a/community/ioping/APKBUILD
+++ b/community/ioping/APKBUILD
@@ -5,7 +5,7 @@ pkgver=1.0
 pkgrel=1
 pkgdesc="A simple disk i/o latency measurement tool"
 url="https://github.com/koct9i/ioping"
-arch="all !s390x"
+arch="all"
 license="GPL3+"
 depends=""
 depends_dev=""
@@ -18,6 +18,7 @@ _builddir="$srcdir"/$pkgname-$pkgver
 prepare() {
 	local i
 	cd "$_builddir"
+	[ "$CARCH" = "s390x" ] && sed -i -e 's@./$(BINARY)@./$(BINARY) -C@g' Makefile
 	for i in $source; do
 		case $i in
 		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;


HRio:
So this will not invalidate caches when running ioping on s390x, or? I.e. a bit like running ioping with option -C. If this is just to pass check() when building for s390x could we not just run the tests with -C flag instead?

tmh1999:
-C is different. -C tells ioping not to clear the cached file before executing.
POSIX_FADV_NORMAL implies no action (advice to the kernel) on the data ioping just touch after it executes.

tmh1999:
And yes of course we can do -C to pass check() on tests.

HRio:
POSIX_FADV_NORMAL give pretty much the same "happy figures" (in quick idle testing) as -C so
it would be better if this only affects check().

```
(POSIX_FADV_DONTNEED)
# ioping -c 300 .

--- . (ext4 /dev/mapper/vg_raid1-test--vol) ioping statistics ---
299 requests completed in 105.5 ms, 1.17 MiB read, 2.83 k iops, 11.1 MiB/s
generated 300 requests in 4.98 min, 1.17 MiB, 1 iops, 4.01 KiB/s
min/avg/max/mdev = 144.7 us / 352.9 us / 29.9 ms / 1.72 ms



# ioping -C -c 300 .
--- . (ext4 /dev/mapper/vg_raid1-test--vol) ioping statistics ---
299 requests completed in 3.63 ms, 1.17 MiB read, 82.5 k iops, 322.2 MiB/s
generated 300 requests in 4.98 min, 1.17 MiB, 1 iops, 4.01 KiB/s
min/avg/max/mdev = 10.4 us / 12.1 us / 64.4 us / 3.41 us


(POSIX_FADV_NORMAL)
# ioping -c 300 .
299 requests completed in 3.16 ms, 1.17 MiB read, 94.5 k iops, 369.3 MiB/s
generated 300 requests in 4.98 min, 1.17 MiB, 1 iops, 4.01 KiB/s
min/avg/max/mdev = 8.26 us / 10.6 us / 18.8 us / 1.51 us
```

HRio:
534f6b9 looks fine to me, thanks.

tmh1999:
merged in https://github.com/alpinelinux/aports/commit/c8c6c8ca8706f1f32b35cf452e25b994b76bed55.
Thank you all !

PR #1400, opened 2017-05-13
testing/apparmor: new aport

allangarret:


ncopa:
the secure_getenv could probably be implemented as inline in the header only, so we dont need the `.c` file.

allangarret:
The scandirat implementation is based on the scandir implementation from musl

PR #1401, opened 2017-05-13
community/thttpd: modernize abuild

tmpfile:


tmpfile:
@fcolista knows why this failed?
I only made cosmetic changes.

tmpfile:
It seems thttpd.pre-install is ignored.
The question is why?

Shizmob:
`.pre-install` files are ran by `apk` at install time, not by `abuild` during `make install`. It seems weird that this fails now though and presumably didn't before.

PR #1432, opened 2017-05-13
main/xen, main/binutils: Build and package Xen EFI binaries

radhus:
These PR contains two commits with the goal to package Xen with its EFI binaries, in order to boot Xen on EFI systems.

First commit addresses the need to add x86_64-pep target to binutils, which Xen requires ld to have in order to build the EFI binaries. 
Some info from other distributions regarding this change:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=675364
https://bugs.archlinux.org/task/42540

Second commit is mostly bumping the revision to rebuild and get Xen binaries, but also placing the EFI binaries in /boot with the xen-hypervisor package.

Fixes #5909.

### DIFF
--- a/main/binutils/APKBUILD
+++ b/main/binutils/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=binutils
 pkgver=2.28
-pkgrel=1
+pkgrel=2
 pkgdesc="Tools necessary to build programs"
 url="http://www.gnu.org/software/binutils/"
 depends=""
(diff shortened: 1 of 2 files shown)


radhus:
Not sure if Travis killed the job due to long output... But the new binutils must be installed before building xen, maybe I should specify a harder version dependency on binutils?

kaniini:
@radhus please split out the binutils change and then we can work on the xen one.  but we're cutting this really close to release.

radhus:
@kaniini Ok! I've created PR #1512, will close this and open a new one for Xen after that one 👍

PR #1435, opened 2017-05-13
community/openblas: fix wrong comment

tmpfile:


tmpfile:
This only failed because Travis has a limit of 4 MB

PR #1436, opened 2017-05-13
community/R: upgrade to 3.4.0

tmpfile:


tmpfile:
It thinks the checksum is wrong, but I double-checked and it's correct

PR #1438, opened 2017-05-13
community/rover: upgrade to 1.0.0

tmpfile:


tmpfile:
Upgrading this package I noticed a weird saveas- syntax.
Looks like it's implemented in abuild. https://github.com/alpinelinux/abuild/blob/master/abuild-fetch.c#L87
In my opinion it's useless becase of :: syntax.
There are only 6 more packages besides rover that use the saveas- syntax.
Could we change those 6 packages to use :: and then remove saveas- from abuild? Less is more.
The packages are mhddfs, xdesktop, xtitle, setroot, n30f and lemonbar.

Shizmob:
That seems like a good idea, yes. CCing @fabled on that.

tmpfile:
All PRs were merged so now there's no packaging using saveas.
It's safe to remove it altogether from abuild.

tmpfile:
I created a pull request to remove the saveas- syntax.
https://github.com/alpinelinux/abuild/pull/20

PR #1475, opened 2017-05-14
main/bridge-utils: upgrade to 1.6 and modernize

tmpfile:


ncopa:
i merged the second patch into the first

PR #1481, opened 2017-05-14
main/libvpx: upgrade to 1.6.1 and modernize

tmpfile:


ncopa:
this introduces abi breakage. we need rebuild various packages after this is upgraded

PR #1487, opened 2017-05-14
community/claws-mail: upgrade to 3.15.0 and modernize

tmpfile:


tmpfile:
Fails at the end because of the tnef_parse plugin, I don't know why.

mv: can't rename '/home/travis/build/alpinelinux/aports/community/claws-mail/pkg/claws-mail/usr/lib/claws-mail/plugins/tnef_parse*': No such file or directory

All other plugins work

PR #1506, opened 2017-05-14
community/marco: modernize abuild

tmpfile:


ncopa:
@tmpfile I wonder if you could take all those commits with "modernize abuild" and put them into a single PR,  and then close the individual PRs.

That is multiple commits in a single PR. That way we can git pull in one shot instead of doing it 100 times.

That is commits with only "modernize abuild", so exclude the commits which also does package upgrade.
https://github.com/alpinelinux/aports/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20label%3AA-improve%20label%3AP-low

possibly also exclude the commits which fails on travis.

tmpfile:
I don't know how I would do it...
I used github interface for most of those commits.

jirutka:
You've filled all checksums manually?! O.O

tmpfile:
No, I used apk checksum on terminal.
Then copy-pasted all using github interface.
I can't think of an easy way to join all PRs into a single PR.
Can you merge individually for now and in the future I start doing every commits on a single PR?
Sorry for the trouble...

PR #1507, opened 2017-05-14
testing/toilet: new aport

Shizmob:
http://caca.zoy.org/wiki/toilet
A free replacement for the FIGlet utility


testing/toilet/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+pkgname=toilet
+pkgver=0.3
+pkgrel=0
+pkgdesc="A free replacement for the FIGlet utility"
+url="http://caca.zoy.org/wiki/toilet"
+arch="all"
+license="WTFPL"
+makedepends="libcaca-dev"
+subpackages="$pkgname-doc"
+source="https://distfiles.alpinelinux.org/distfiles/$pkgname-$pkgver.tar.gz"
```
jirutka: What, this is not the upstream source…

Shizmob: The upstream server is extremely flaky, apparently to the extent it already existed on distfiles without me having to request it.

jirutka: Then use abuild snapshot.

Shizmob: This is not a git or svn repo, and it won't guarantee a build working if the server is down while a snapshot is being made (initially, for instance).

Shizmob: The issue is not a changing tarball, but the upstream server being offline/down a lot.

clandmeter:
Closed due to inactivity.

PR #1509, opened 2017-05-14
community/[various]: fix missing sha512 checksums

tmpfile:
I found 30 packages that were missing sha512sums and this patch fixes that.


tmpfile:
I don't understand why this failed, it was only a checksum correction, nothing invasive.
And the error is not related to checksum.

tmpfile:
@ncopa shouldn't abuild fail to build packages with missing sha512sums?
why was this working before?

PR #1510, opened 2017-05-14
main/[various]: fix missing sha512 checksums

tmpfile:
I found 203 packages in main that were missing sha512sums and this patch fixes that.


tmpfile:
This only failed because Travis has a limit of 4 MB

PR #1513, opened 2017-05-15
main/xen: build EFI binaries

radhus:
New binutils 2.28-r2 provides the x86_64-pep target which makes Xen
build its EFI bootloader.

The EFI binaries are placed in /boot and provided by xen-hypervisor:
 /boot/xen-4.8.1.efi
 /boot/xen-4.8.efi
 /boot/xen-4.efi
 /boot/xen.efi

### DIFF
--- a/main/xen/APKBUILD
+++ b/main/xen/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=xen
 pkgver=4.8.1
-pkgrel=4
+pkgrel=5
 pkgdesc="Xen hypervisor"
 url="http://www.xen.org/"
 arch="x86_64 armhf"
@@ -319,6 +319,11 @@ hypervisor() {
 	depends=
 	mkdir -p "$subpkgdir"
 	mv "$pkgdir"/boot "$subpkgdir"/
+
+	if [ "$CARCH" = "x86_64" ]; then
+		mkdir -p "$subpkgdir"/usr/lib
+		mv "$pkgdir"/usr/lib64/efi "$subpkgdir"/usr/lib/efi
+	fi
 }
 
 bridge() {


radhus:
@kaniini Here's the second one.

It seems like it makes most sense to place them in /boot and thus in the xen-hypervisor package, this is like Debian I think. But if you believe that is a more intruding change before the release, maybe I can redo with a xen-efi package instead, and perhaps in the original /usr/lib64/efi location?

kaniini:
I'm not sure it is appropriate to install them to /boot, shouldn't it be installed to a standard EFI location?

radhus:
Yes, you are right. Correct me if I'm wrong, but I don't think we have anything like that in Alpine yet?

Packages with EFI binaries that I've looked at now seem to store them in /usr somewhere, like:
* gummiboot with /usr/lib/gummiboot/gummibootx64.efi
* syslinux with /usr/share/syslinux/efi64/syslinux.efi

Do you think it would be viable to keep them in /usr/lib64/efi for now, but perhaps in the xen-hypervisor package? It would be nice to get the binaries in place for 3.6, and then maybe pursue a general "EFI boot support" feature?

radhus:
Btw, looks like Travis terminates the job due to exceeding the log limit of 4MB...

kaniini:
@radhus yes that is fine.

i am still working out the details on formal EFI support in alpine as it's kind of complicated.  so lets use /usr/lib/efi for now.

radhus:
@kaniini Ok, cool, I'll try to push that asap.

If you want help with the EFI support, I can help out with testing and development, just let me know if there's something I can work on. I will be running at least some boxes now with Xen that I will boot from EFI.

radhus:
Alright, finally, now it's in /usr/lib64/efi

If this is accepted, is there any chance this might get backported to 3.6 stable? :)
Since it's just inclusion of already built files so I feel it's quite a minor change.

jirutka:
> Alright, finally, now it's in /usr/lib64/efi

That’s wrong, we don’t use `/usr/lib64`, it should be in `/usr/lib`.

radhus:
Ah yes, sorry. And I seem to have missed committing the CARCH check for x86_64, sorry. Wait just a minute :)

radhus:
@jirutka fixed now :)

radhus:
Let me know if you want to do it any other way, I'm running one box with this now and have it verified working with Virtualbox UEFI VM's (although in both cases, you need to configure bootloader (like gummiboot) yourself of course).

radhus:
Btw, looks like /usr/lib64/efi/xen-4.8.1.efi is created in the xen packages now, both in 4.8.x and 4.9. Not really what we want I guess.

@jirutka do you have any pointers on how to proceed with this? I can of course rebase it if it's fine as it is.

PR #1517, opened 2017-05-16
gearmand: fix system user name

1ma:
As it stands now the configuration file expects the system user to be named `gearmand`, while the pre-install script names it `gearman`, therefore the daemon cannot start.

```
alpinada:~# /etc/init.d/gearmand start
 * Starting gearmand ...
 * start-stop-daemon: user `gearmand' not found
 * Failed to start gearmand                                                [ !! ]
 * ERROR: gearmand failed to start

alpinada:~# tail /var/log/messages | grep gearm
May 16 14:45:43 alpinada daemon.err /etc/init.d/gearmand[1918]: start-stop-daemon: user `gearmand' not found
May 16 14:45:43 alpinada daemon.err /etc/init.d/gearmand[1902]: ERROR: gearmand failed to start
```

### DIFF
--- a/testing/gearmand/APKBUILD
+++ b/testing/gearmand/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Aaron Hurt <ahurt@ena.com>
 pkgname=gearmand
 pkgver=1.1.15
-pkgrel=0
+pkgrel=1
 pkgdesc="A distributed job queue"
 url="http://gearman.org"
 arch="all"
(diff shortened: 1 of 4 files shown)


jirutka:
How is this user usually called? I’d personally prefer `gearman` to `gearmand`.

1ma:
Honestly, I don't know :thinking: I resolved it this way because I liked the fact that it's named the same as the program itself. But I'm no sysadmin.

Shizmob:
It seems the user is normally called `gearman`. Could you fix it in the config instead? 😅

1ma:
Sure, rebased and corrected.

Shizmob:
You should also update the checksums and pkgrel. :)

1ma:
[Aite, thanks](http://i.imgur.com/jAm9hyd.jpg) :flushed:

PR #1522, opened 2017-05-17
main/linux-vanilla: fix config file for ppc64le

rgdoliveira:
vmlinuz 4.9.28 was not booting on ppc64le because some configurations
was missing in ppc64le config file.

The most important configurations that were missing are:
CONFIG_PPC_OF_BOOT_TRAMPOLINE and CONFIG_PPC_PSERIES.

Another configurations were consequently added because of these
two configurations.


rgdoliveira:
@leitao @ncopa If you can take a look at this one, please.

PR #1535, opened 2017-05-18
main/alpine-baselayout: add hvc0 entry to inittab

rgdoliveira:
add hvc0 configuration to inittab file. Without this configuration
the boot on ppc64le was not display the login console after openrc
scripts finish executing.


ncopa:
We use hvc0 for xen domU's too, but there we set it via boot option `console=hvc0`. init in initramfs image will add it to `/etc/inittab` and `/etc/securetty`, to this change should not be needed.

rgdoliveira:
thanks for the review @ncopa 
I just made what you suggested and I opened a PR: https://github.com/alpinelinux/aports/pull/1538

Closing this one...

PR #1540, opened 2017-05-19
testing/pypy: new aport

HorridBro:
https://github.com/HorridBro/AlpinePyPy
PyPy build file

### DIFF
--- a/testing/pypy/APKBUILD
+++ b/testing/pypy/APKBUILD
@@ -1,0 +1,49 @@
+# Contributor:Mihnea Saracin <mihnea.saracin@rinftech.com>
+# Maintainer:Mihnea Saracin <mihnea.saracin@rinftech.com>
+pkgname=pypy
+pkgver=5.7.1
+pkgrel=0
+pkgdesc="pypy2"
+url="https://bitbucket.org/pypy"
+arch="all"
+license="MIT"
+depends="libssl1.0 libcrypto1.0 libffi libbz2 "
+makedepends="gcc make python linux-headers  libffi-dev pkgconf zlib-dev bzip2-dev 
+            sqlite-dev ncurses-dev readline-dev openssl-dev expat-dev openssl-dev gdbm-dev tk tk-dev libc-dev py-cffi paxmark"
+install=""
+subpackages="$pkgname-dev $pkgname-doc"
+source="https://bitbucket.org/$pkgname/$pkgname/downloads/pypy2-v$pkgver-src.tar.bz2
+        alpine.patch"
+builddir="$srcdir/pypy2-v$pkgver-src/pypy/goal"
+
+
+prepare(){
+
+        cd "$srcdir/pypy2-v$pkgver-src"
+        for i in $source; do
+                case $i in
+                *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1
+                esac
+        done
+}
+
+build() {
+        cd "$builddir"
+        /usr/bin/python ../../rpython/bin/rpython --opt=jit --nopax targetpypystandalone.py || return 1
+}
+
+package() {
+        mkdir -p "$pkgdir"
+        cd "$builddir/../tool/release"
+        ./package.py --archive-name pypy2-v$pkgver --builddir "$srcdir/pypy2-v$pkgver-alpine" || return 1
+}
+
+
+
+md5sums="73720fd65aa634d60d4633a4c4e38cbb  pypy2-v5.7.1-src.tar.bz2
+0efbcc740572b37847a5e8bade4d088b  alpine.patch"
+sha256sums="d01bee43c6df79f7bbc1149bb3e85f489491fb2358a6a1f9a7f0d6e07715832f  pypy2-v5.7.1-src.tar.bz2
+01f6e8e9ada89519543e9a84560def41fb6d8f5892bd898b80bc4fc0d42eb875  alpine.patch"
+sha512sums="1ad2dddb40c28d2d3e95a9f0730e765d981dee6e2d0664cf1274eb7c1021690a848c3485c846eac8a8b64425b44946b5b2d223058ec4699155a2122ee7d38b75  pypy2-v5.7.1-src.tar.bz2
+3df182c402ee6f37f81a87679ed7e92f7375c70beacc960548608dbff0f627037d7995c54e8e3cad544f9e0b6b08d00d8419d57b4dbfe5290bd34cbcdbb76775  alpine.patch"
+
--- a/testing/pypy/APKBUILD
+++ b/testing/pypy/APKBUILD
@@ -1,49 +1,35 @@
-# Contributor:Mihnea Saracin <mihnea.saracin@rinftech.com>
-# Maintainer:Mihnea Saracin <mihnea.saracin@rinftech.com>
+# Contributor: Mihnea Saracin <mihnea.saracin@rinftech.com>
+# Maintainer: Mihnea Saracin <mihnea.saracin@rinftech.com>
 pkgname=pypy
 pkgver=5.7.1
 pkgrel=0
-pkgdesc="pypy2"
+pkgdesc="A fast, compliant alternative implementation of the Python 2 language"
 url="https://bitbucket.org/pypy"
 arch="all"
 license="MIT"
 depends="libssl1.0 libcrypto1.0 libffi libbz2 "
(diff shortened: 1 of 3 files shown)


HorridBro:
The reason why Travis fails is because PyPy build time exceeds  easily two hours on a regular computer.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
+# Contributor:Mihnea Saracin <mihnea.saracin@rinftech.com>
```
Shizmob: Please put a space between the colon and the names.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+# Maintainer:Mihnea Saracin <mihnea.saracin@rinftech.com>
+pkgname=pypy
+pkgver=5.7.1
+pkgrel=0
+pkgdesc="pypy2"
```
Shizmob: A better package description would be nice. Maybe something like this: `pkgdesc="A fast, compliant alternative implementation of the Python 2 language"`.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+url="https://bitbucket.org/pypy"
+arch="all"
+license="MIT"
+depends="libssl1.0 libcrypto1.0 libffi libbz2 "
```
Shizmob: This *should* be automatically inferred by `abuild` as noted at the end of the build log, although I can't double-check since the Travis build is timing out.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+makedepends="gcc make python linux-headers  libffi-dev pkgconf zlib-dev bzip2-dev
```
Shizmob: At least `gcc`, `make`, `libc-dev` and `tk` can likely be removed from `makedepends`: The former three belong to `build-base`, and the latter *should* be implied by `tk-dev`. Also, does `pypy` not build against LibreSSL? I would suggest `libressl-dev` if it does, because we try to limit the number of packages linking against OpenSSL.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+pkgver=5.7.1
+pkgrel=0
+pkgdesc="pypy2"
+url="https://bitbucket.org/pypy"
+arch="all"
+license="MIT"
+depends="libssl1.0 libcrypto1.0 libffi libbz2 "
+makedepends="gcc make python linux-headers  libffi-dev pkgconf zlib-dev bzip2-dev 
+            sqlite-dev ncurses-dev readline-dev openssl-dev expat-dev openssl-dev gdbm-dev tk tk-dev libc-dev py-cffi paxmark"
+install=""
```
Shizmob: This line can be omitted if it's empty.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+subpackages="$pkgname-dev $pkgname-doc"
+source="https://bitbucket.org/$pkgname/$pkgname/downloads/pypy2-v$pkgver-src.tar.bz2
+        alpine.patch"
+builddir="$srcdir/pypy2-v$pkgver-src/pypy/goal"
+
+
+prepare(){
```
Shizmob: This entire function is part of the default prepare implementation and can be omitted.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+        for i in $source; do
+                case $i in
+                *.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1
+                esac
+        done
+}
+
+build() {
+        cd "$builddir"
+        /usr/bin/python ../../rpython/bin/rpython --opt=jit --nopax targetpypystandalone.py || return 1
```
Shizmob: No need for a full path here, and please explicitly the Python version (`python2` instead of `/usr/bin/python`, in short).

Shizmob: Also applies below: `|| return 1` can be omitted as `set -e` is default in APKBUILDs now.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+}
+
+package() {
+        mkdir -p "$pkgdir"
+        cd "$builddir/../tool/release"
+        ./package.py --archive-name pypy2-v$pkgver --builddir "$srcdir/pypy2-v$pkgver-alpine" || return 1
```
Shizmob: This doesn't seem to be installing PyPy to `$pkgdir`? There is no reference of `$pkgdir` in this command-line at least. Does PyPy do some magic on its own?

Shizmob: Also, the `|| return 1` statement can be omitted.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+        done
+}
+
+build() {
+        cd "$builddir"
+        /usr/bin/python ../../rpython/bin/rpython --opt=jit --nopax targetpypystandalone.py || return 1
+}
+
+package() {
+        mkdir -p "$pkgdir"
```
Shizmob: This can be removed if the install step creates it automatically, which seems to be the case for a lot of build systems. If it's not the case for PyPy it's not a problem, but yeah.

testing/pypy/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+        cd "$builddir/../tool/release"
+        ./package.py --archive-name pypy2-v$pkgver --builddir "$srcdir/pypy2-v$pkgver-alpine" || return 1
+}
+
+
+
+md5sums="73720fd65aa634d60d4633a4c4e38cbb  pypy2-v5.7.1-src.tar.bz2
```
Shizmob: `md5sums=` and `sha256sums=` have been deprecated and can be removed.

Shizmob:
Furthermore, does PyPy have a testsuite? Since recently we implemented a `check()` function in APKBUILDs that runs the test suite (and installs any dependencies for it defined in `checkdepends=`) -- you may need to upgrade your build system to edge to be able to test that, but since this is a PR against edge, that is probably a good idea anyway. If there was and it ran succesfully (or mostly), that would significantly increase the confidence in a working PyPy.

Shizmob:
Last random note: while I do love the beautiful fractals it draws, they are kind of a noise in our build log this way, making it harder to find errors. Could it be disabled somehow?

testing/pypy/APKBUILD
Shizmob: Also, `py-cffi` should probably be changed to `py2-cffi` for clarity.

kaniini:
I prefer not accepting pypy into Alpine because it justifies keeping python 2 around.  pypy3 would, however, be fine.

While, originally, we decided to keep python 2 around because there were many consumers, these consumers are declining and python 2 only has limited upstream maintenance.  Therefore it makes no sense to add pypy.

HorridBro:
Hi, @kaniini  , I don't know why are you saying that python2 customers are declining since many projects are still using it (for example OpenStack). I don't see a problem in adding pypy2 alongside python2.
The customer will choose which one is the best for him.

Regarding pypy3, I'm working on an APKBUILD file for it and once it's done I'll submit a pull request.

Shizmob:
I've got no particular issues myself with adding pypy, especially as pypy3 is somewhat experimental and existing software benefits from pypy.

HorridBro:
I updated the files after the review.Regarding the fractal , it is suppressed , but i kept all other logging for debugging purposes .

testing/pypy/APKBUILD
Shizmob: This still holds, I think?

testing/pypy/APKBUILD
Shizmob: Was `tk` not implied by `tk-dev`?

Shizmob:
Another comment: does PyPy have a testsuite we can run for `check()` implementation? Also CC'ing @ncopa and @kaniini again for opinions on inclusion.

kaniini:
My position is NAK, as I said earlier.  We want less python 2, not more python 2.

kaniini:
If PyPy developers care about their work being included here, they should make PyPy3 stable and concentrate on supporting 3.6.  Hard NAK for any new Python 2 interpreters.

jirutka:
I fully agree with @kaniini. Python 2 must *finally* die.

clandmeter:
Closed due to inactivity and reviewers NAK.

PR #1546, opened 2017-05-19
testing/py-unittest2 + 3 x depends

itoffshore:
* py-contextlib2
* py-traceback2
* py-linecache2
* py-unittest2

### DIFF
--- a/testing/py-traceback2/APKBUILD
+++ b/testing/py-traceback2/APKBUILD
@@ -1,0 +1,53 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-traceback2
+_pkgname=${pkgname#py-*}
+pkgver=1.4.0
+pkgrel=0
+pkgdesc="Backport of the Python3 stdlib traceback module"
+url="https://github.com/testing-cabal/traceback2"
+arch="noarch"
+license="PSF"
+makedepends="python2-dev python3-dev py-setuptools"
+subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
+options="!check" # temp disable until py2-unittest2 is in repos
+#checkdepends="py2-contextlib2 py-fixtures py-testtools py2-unittest2 py2-linecache2 py-extras"
+source="$pkgname-$pkgver.tar.gz::https://github.com/testing-cabal/$_pkgname/archive/$pkgver.tar.gz"
+builddir="$srcdir/$_pkgname-$pkgver"
+
+build() {
+        cd "$builddir"
+        python2 setup.py build
+        python3 setup.py build
+}
+
+check() {
+        cd "$builddir"
+        python2 -m unittest2 discover
+        python3 -m unittest2 discover
+}
+
+package() {
+	mkdir -p "$pkgdir"
+}
+
+_py2() {
+	replaces="$pkgname"
+	_py python2
+}
+
+_py3() {
+	_py python3
+}
+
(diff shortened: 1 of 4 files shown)


itoffshore:
rebased with:

```
build()
...
check()
...
package()
```

testing/py-unittest2/APKBUILD
```diff
@@ -0,0 +1,31 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-unittest2
```
jirutka: This is Python2-only, so it should be named `py2-unittest2`.

testing/py-unittest2/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+_pkgname=${pkgname#py-*}
+pkgver=1.1.0
+pkgrel=0
+pkgdesc="Backport of the Python3 enhanced unittest testing framework"
+url="https://pypi.python.org/pypi/unittest2"
+arch="noarch"
+license="PSF"
+depends="python py-six py-argparse py-traceback2 py-linecache2"
```
jirutka: Please don’t declare versionless python packages as dependencies, it should be python2, py2-six, py2-argparse, …

testing/py-unittest2/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+makedepends="python2-dev py-setuptools"
+source="$pkgname-$pkgver.tar.gz::http://http.debian.net/debian/pool/main/u/$_pkgname/${_pkgname}_${pkgver}.orig.tar.gz"
+builddir="$srcdir/$_pkgname-$pkgver"
+
+build() {
+        cd "$builddir"
+        python setup.py build
```
jirutka: Please don’t rely on `python` symlink, use explicitly `python2`, or `python3`.

testing/py-traceback2/APKBUILD
```diff
@@ -0,0 +1,33 @@
+# Contributor: Stuart Cardall <developer@it-offshore.co.uk>
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=py-traceback2
```
jirutka: This module supports both Python 2 (legacy) and 3, so it must be packaged for Python 3 only (py3-traceback2), or for both Python 2 and 3 (py-traceback2 with subpackages py2-traceback2 and py3-traceback2). If there’s no specific reason for supporting Legacy Python, I’d prefer py3 only package (it’s simpler).

itoffshore: `python2` versions all of these 4 depends are needed so `py-bandit` can be added with a `check()` 

(which will finally allow me to fix another `python` package with a `check()` using `py-bandit`)

itoffshore:
all the names / depends explicitly set to `py2` for:
* py2-contextlib2
* py2-linecache2
* py2-unittest2

py-traceback2 now with `python2` & `python3` versions

testing/py-traceback2/APKBUILD
itoffshore: any reason why this is still not pulled ???

testing/py-traceback2/APKBUILD
```diff
@@ -0,0 +1,53 @@
...
+        python2 -m unittest2 discover
+        python3 -m unittest2 discover
+}
+
+package() {
+	mkdir -p "$pkgdir"
+}
+
+_py2() {
+	replaces="$pkgname"
```
jirutka: It’s a new aport, so `replaces` is not needed.

PR #1551, opened 2017-05-20
main/musl: update upstream s390x patch

tmh1999:
With this update, ocaml built good but ocaml-findlib still
segfault


jirutka:
@ncopa, could/should we merge this _before_ releasing v3.6?

PR #1552, opened 2017-05-20
main/linux-hardened: disable GRKERNSEC_SYSFS_RESTRICT

Shizmob:
The sysfs restrictions lead to [issues with Mesa](https://bugs.alpinelinux.org/issues/7265) and some other applications. Since there is no sysfs toggle for this functionality, we have to disable it for now.


ncopa:
battery status and backlight also needs access to sysfs.

But I am not convinced it is a good idea to disable it for everyone, since you can not easily enable it for server use. For desktop use you still have the options to use vanilla kernel or add some udev/mdev rule (or /etc/local.d script) to set the permissions on the needed entries.

For the long run, it would be nice with a boot option or sysctl knob to disable it.

Shizmob:
I don't think leaving desktop users with only a vanilla kernel is a good idea. I agree that sysctl is a better solution in the end, but this is not something offered right now. You can't do anything in udev or mdev to bypass this at least as far as the Mesa issue goes, as far as I know.

ncopa:
which is the exact entry?

jirutka:
GRKERNSEC_SYSFS_RESTRICT is problem also for running unprivileged LXC containers: https://github.com/lxc/lxc/issues/296#issuecomment-234708658.

ncopa:
A temp workaround might be something like:
````sh
#!/bin/sh

find /sys -type f | grep -v -w efivars | xargs chmod 644
find /sys -type d | grep -v -w efivars | xargs chmod 755
````

Save as `/etc/local.d/fix-sysfs.start`



I use this script to fix backlight on my laptop:
````sh
#!/bin/sh

chmod 755 /sys/class /sys/class/backlight
for link in /sys/class/backlight/*; do
        p=/sys
        target=$(readlink -f $link)
        i=0
        while [ $p != $target ] && [ $i -lt 30 ]; do
                reminder=${target#$p/}
                p=${p}/${reminder%%/*}
                chmod 755 "$p"
                i=$(($i+1))
        done
done
````

The idea is to execute the above script on the proper uevents (via udev rules).

Shizmob:
I suppose, but that just completely undoes any SYSFS restrictions anyway. I guess it's possible, but I'm not sure how we'd integrate it with the packaging.

ncopa:
Apparently there is a boot option to disable it. `grsec_sysfs_restrict=0`

BlueMax:
`grsec_sysfs_restrict=0` doesn't work. There is another report on Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=888702

I'm on edge/i686 with LXD/unprivileged. It works properly with vanilla kernel.

PR #1553, opened 2017-05-20
community/cabal: moved from testing

mitchty:
As it looks like ghc has migrated from testing to community cabal should as well similar to how rust without cargo isn't too useful.

Need to keep the peanut butter and jelly together for a good sandwich.


mitchty:
Fun, looks like the cabal files were updated since this got created. What changed from the versions here and what is there now:
```
# diff hackage-security.cabal*
--- hackage-security.cabal
+++ hackage-security.cabal.orig
@@ -1,6 +1,6 @@
 name:                hackage-security
 version:             0.5.2.2
-x-revision: 4
+x-revision: 2
 -- xrevision:1 integrates
 -- https://github.com/well-typed/hackage-security/commit/e4bff90a82a588ff2d0beedfc50d5fdf75861d48
 synopsis:            Hackage security library
@@ -115,7 +115,7 @@
                        -- 0.4.2 introduces TarIndex, 0.4.4 introduces more
                        -- functionality, 0.5.0 changes type of serialise
                        tar               >= 0.5     && < 0.6,
-                       time              >= 1.2     && < 1.9,
+                       time              >= 1.2     && < 1.7,
                        transformers      >= 0.4     && < 0.6,
                        zlib              >= 0.5     && < 0.7,
                        -- whatever versions are bundled with ghc:
@@ -153,7 +153,9 @@
                        UndecidableInstances
   -- use the new stage1/cross-compile-friendly Quotes subset of TH for new GHCs
   if impl(ghc >= 8.0)
-    other-extensions: TemplateHaskellQuotes
+    -- place holder until Hackage allows to edit in the new extension token
+    -- other-extensions: TemplateHaskellQuotes
+    other-extensions:
   else
     other-extensions: TemplateHaskell

@@ -162,7 +164,7 @@
   if flag(base48)
     build-depends: base >= 4.8
   else
-    build-depends: base < 4.8, old-locale >= 1.0
+    build-depends: old-locale >= 1.0

   -- The URI type got split out off the network package after version 2.5, and
   -- moved to a separate network-uri package. Since we don't need the rest of
@@ -207,7 +209,9 @@

   if impl(ghc >= 7.10)
      other-extensions: AllowAmbiguousTypes
-                       StaticPointers
+--                       StaticPointers
+-- ^^^ Temporarily disabled because Hackage doesn't know yet about this
+-- extension and will therefore reject this package.

 test-suite TestSuite
   type:                exitcode-stdio-1.0
@@ -220,7 +224,7 @@
                        TestSuite.PrivateKeys
                        TestSuite.Util.StrictMVar
   build-depends:       base,
-                       Cabal < 1.25,
+                       Cabal,
                        containers,
                        HUnit,
                        bytestring,
```

community/cabal/APKBUILD
```diff
@@ -2,61 +2,56 @@
...
 makedepends="ghc ghc-dev gmp-dev libffi-dev zlib-dev binutils-gold chrpath"
 subpackages="$pkgname-doc"
 source="https://hackage.haskell.org/package/$pkgname-install-$pkgver/$pkgname-install-$pkgver.tar.gz
-	cabal-0001-force-ld.gold.patch
-	cabal-0002-busybox-od.patch
-	cabal-0003-use-apkbuild-downloads.patch
-
 	https://hackage.haskell.org/package/mtl-2.2.1/mtl-2.2.1.tar.gz
-	https://hackage.haskell.org/package/mtl-2.2.1/mtl.cabal
+	mtl.cabal::https://hackage.haskell.org/package/mtl-2.2.1/revision/1.cabal
```
jirutka: Source file name must be unique, this is not unique across versions. You must include version in file name (e.g. mtl-2.2.1-1.cabal).

mitchty: It has to be mtl.cabal to work, the cabal file is for the package/module being built, not the package/module+version.

This is a basic requirement of how the bootstrap.sh script works for cabal.

mitchty: If the requirement is that this file has to be named mtl-2.2.1.cabal, then this won't ever work. The cabal file here is updated outside of the package-version.tar.gz to allow for newer ghc compilers to be able to bootstrap cabal itself.

As such, this is how it has to be, otherwise we cannot bootstrap cabal without cabal already being installed.

jirutka: This source file name absolutely must be unique, at least for the current version of abuild. You can strip version from file names in `prepare` if it’s needed.

PR #1554, opened 2017-05-20
testing/tiptop: new abuild

mitchty:


testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
+#-*-mode: Shell-script; coding: utf-8;-*-
```
Shizmob: Can you remove this? We'd prefer to not have editor-specific things in APKBUILDs.

testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+pkgname=tiptop
+pkgver=2.3
+pkgrel=0
+pkgdesc="Tiptop: Hardware Performance Counters for the Masses"
+url="http://tiptop.gforge.inria.fr"
+arch="x86_64"
+license="gpl2"
+depends=""
```
Shizmob: This can be omitted if it's empty.

testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
+#-*-mode: Shell-script; coding: utf-8;-*-
+# Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
+pkgname=tiptop
+pkgver=2.3
+pkgrel=0
+pkgdesc="Tiptop: Hardware Performance Counters for the Masses"
+url="http://tiptop.gforge.inria.fr"
+arch="x86_64"
```
Shizmob: Is there a specific reason that it only works on `x86_64`? If you simply haven't tested it, leave it to `all` - our builders will sort it out.

testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+license="gpl2"
```
Shizmob: Please capitalise this to `GPL2`.

testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends=""
+makedepends="byacc flex linux-headers ncurses-dev libxml2-dev"
+subpackages="${pkgname}-doc"
```
Shizmob: Alpine style is to not use braces for variable expansions unless explicitly required.

testing/tiptop/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+	http://${pkgname}.gforge.inria.fr/releases/${pkgname}-${pkgver}.tar.gz
+"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure --prefix=/usr
+	make
+}
+
```
Shizmob: Could you add a smoke-test `check()` that simply runs the resulting binary with e.g. `--version`?

testing/tiptop/APKBUILD
mitchty: It builds on almost any platform, but not every platform has implemented the requisite perf counters. I'm only aware of these existing for x86_64 and i386 on intel cpus. Power cpus might have an implementation as well I've no idea, arm however I know lacks the syscall and requisite hardware support.

I'm not sure what the appropriate call is here, this binary really only works on architectures that support the sys call and implement the requisite hardware to support the sys call in the perf framework. I'll add i386 to the arches as I'm sure that one works too based off the intel references, but I'm not sure having a binary that doesn't work is too useful elsewhere. Thoughts?

Example run on an armhf system:
```
$ tiptop
syscall: Not supported
Could not perform syscall.
Don't know why...
# from strace...
open("/proc/sys/kernel/perf_counter_paranoid", O_RDONLY|O_LARGEFILE) = -1 ENOENT (No such file or directory)
open("/proc/sys/kernel/perf_event_paranoid", O_RDONLY|O_LARGEFILE) = 3
readv(3, [{iov_base="", iov_len=0}, {iov_base="2\n", iov_len=1024}], 2) = 2
_llseek(3, -1, [1], SEEK_CUR)           = 0
close(3)                                = 0
perf_event_open({type=PERF_TYPE_HARDWARE, size=0 /* PERF_ATTR_SIZE_??? */, config=PERF_COUNT_HW_CPU_CYCLES, ...}, 0, -1, -1, 0) = -1 EOPNOTSUPP (Not supported)
```

testing/tiptop/APKBUILD
mitchty: Sadly there is no --version/-v/-V switch to do this with for the binary. Shall I just put !check in the APKBUILD?

testing/tiptop/APKBUILD
Shizmob: Thanks for the insight! `x86` and `x86_64` is fine then, together with a comment saying that the appropriate perf counters are required. We can always expand them later.

testing/tiptop/APKBUILD
Shizmob: What about `--help`?

mitchty: It does have that, but it looks like any arg parsing happens after it tries to use the aforementioned sys call:
```
$ tiptop --help
syscall: Not supported
Could not perform syscall.
Don't know why...
```

Which is a hokey way to go about parsing arguments, but probably OK for x86, I think, it might depend on if you're running in a VM without hardware perf counters to access. We can try it?

PR #1557, opened 2017-05-21
main/alpine-conf: add lvm2 dependency

rgdoliveira:
alpine-conf package provides 'setup-disk' script which uses
'vgchange' command. This command is provided by lvm2 package,
and it was missing as dependency.

### DIFF
--- a/main/alpine-conf/APKBUILD
+++ b/main/alpine-conf/APKBUILD
@@ -1,12 +1,12 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=alpine-conf
 pkgver=3.6.0_rc1
-pkgrel=2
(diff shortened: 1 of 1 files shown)


rgdoliveira:
@ncopa and @jirutka, Can you take a look at this PR and check if it make senses?

ncopa:
NAK,

the vgchange is optional functionality so lvm2 is "optional" package. We don't want pull in lvm2 for everyone.

rgdoliveira:
@ncopa, I see.

Well, when I booted the ppc64le ISO using qemu and run the "setup-alpine" script, it triggers the "setup-disk" script and failed because 'vgchange' command was not found (as lvm2 is not installed). Maybe I am missing something  when running "setup-alpine" script.

PR #1559, opened 2017-05-22
community/yarn: update to 0.24.6

errm:


community/yarn/APKBUILD
```diff
@@ -10,34 +10,14 @@ license="BSD-2"
...
-		-o -iname '*.md~' \
-		-o -iname '*.cmd' \
-		-o -name '*.min.js' \
-		-o -name '*.map' \
-		-o -name 'bower.json' \
-		-o -name 'Dockerfile.*' \
-		-o -name 'Gruntfile.js' \
-		-o -iname 'license*' -a ! -name '*.js' \
-		-o -type f -a -name '.*' \) \
-		-delete
```
jirutka: Why have you removed this?

errm: The yarn tarball no longer includes all these files so we don't need to remove them:

The contents of the tarball now looks like:

```
.
├── LICENSE
├── bin
│   ├── yarn
│   ├── yarn.cmd
│   ├── yarn.js
│   ├── yarnpkg
│   └── yarnpkg.cmd
├── lib
│   ├── v8-compile-cache.js
│   └── yarn-cli.js
└── package.json
```

jirutka: I see `*.cmd` here...

jirutka: I see `*.cmd` here…

errm: yeah we are removing them with the `rm $builddir/bin/*.cmd` on L20

Shizmob: You forgot to remove the `LICENSE` file. If you can remove that one as well, it LGTM. Is it correct that the entire `node_modules` subfolder in the updated package is gone?

errm: Yes the entire `node_modules` folder is gone, there are just 9 files in the new tarball, and we remove the 2 `.cmd` files leaving 7...

I don't think we should remove the `LICENSE` file, we did it before because there were many copies of the same file in `node_modules`, this is just one file, we should respect the licence by including it.

jirutka:
Merged in 3d51e66748ac73ccbcf01d631e9c8162c194dcd2 with few modification. Thanks for your contribution!

errm:
Thanks, your modification makes sense :)

PR #1560, opened 2017-05-22
unmaintained/usbip-utils: make buildable

HRio:
modernize and change kernel version to v4.9


Shizmob:
Hmm, I'm in general not a big fan of how this APKBUILD is written... need to come back to this and see if it can be done in a cleaner way.

HRio:
@Shizmob can a clean-up not be postponed until later, e.g. when moving to testing? better to at-least have something in alpine that can be used? ATM its not usable, but after this change it works quite well, I am using it to get logic analyzers and JTAGs physically connected to a rpi3 into XEN domUs running on my laptop.

Shizmob:
Sure, as long as it stays in unmaintained it's fine by me.

PR #1561, opened 2017-05-23
community/chromium: update to 58.0.3029.110

khanku:


clandmeter:
Can you elaborate about this PR?
where did you get these patches, what do they do?

khanku:
I committed a few too many patches (from the [inox](https://aur.archlinux.org/packages/inox) patchset) I intended to port at a later point (see also [#4452](http://bugs.alpinelinux.org/issues/4452)).d 
The 6 new/changed patches I adapted myself or got from [voidlinux](https://github.com/voidlinux/void-packages/tree/master/srcpkgs/chromium/files/musl-patches).

clandmeter:
I'm not sure we want to apply inox patch set to our main chromium package. I looked at your patches and I see the are blind copies from void which uses a different stack size, so I'm concerned about the quality.

Chromium does not have musl support so it takes time and knowledge to support it on Alpine, adding inox will add another layer of complexity.

If you are really willing to maintain inox on Alpine the only way will be to have an separated inox package in testing and show is you are capable of maintaining it.

cc @ncopa

PR #1573, opened 2017-05-25
testing/colormake: new aport

HRio:


testing/colormake/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+# Maintainer: Henrik Riomar <henrik.riomar@gmail.com>
+pkgname=colormake
+pkgver=0.9.20140503
+pkgrel=0
+pkgdesc="A simple wrapper around make to colorize the output"
+url="http://bre.klaki.net/programs/colormake"
+arch="noarch"
+license="GPL2+"
+depends="perl"
+options="!check"
```
Shizmob: Is the upstream test suite broken or otherwise unsuitable? We'd like `check()` on as many package as possible. :)

HRio: does not contain any checks, will add "PATH=.:${PATH} ./colormake --version" as check(), is that OK?

Shizmob: I'm not sure if the $PATH part is needed, but yes, running the binary is an okay smoke test. :)

HRio:
@Shizmob bash missing from depends and make from checkdepends, fixed in 63ae8ae

PR #1574, opened 2017-05-25
[llvm packages]: add tests and cleanup dependencies

Shizmob:
This PR adds tests to the the LLVM packages and cleans up their dependencies.
It also moves the `llvm-test-utils` package to `llvm-utils` to reflect its more general usage and upstream terminology.


ncopa:
can we close this now that we have llvm5?

jirutka:
IIRC it’s still useful, just need to be rebased and updated.

PR #1577, opened 2017-05-27
testing/scala: new aport

norbjd:
Hello,                                                                                              
                                                                                                    
I have just created a new aport for `scala` (version 2.12.2).                                       
                                                                                                    
This is my first contribution, I followed the guidelines, but I may have done a few mistakes in the `APKBUILD` (at least, some bad practices). I'll be grateful if someone can review it and make constructive criticisms to progress.
                                                                                                    
I have separated `scala-doc` from `scala`, since the documentation is quite heavy (~ 600 MB) because it contains :
- HTML scaladoc for default API + scripts                                                           
- Man pages for scripts                                                                             
- Javadoc JARs                                                                                      
                                                                                                    
For `scala-doc`, I took the `.deb` package as an example.                                           
                                                                                                    
The build can take some time (~ 30 minutes on my laptop). The longest task is documentation generation.

### DIFF
--- a/testing/scala/APKBUILD
+++ b/testing/scala/APKBUILD
@@ -1,111 +1,141 @@
 # Contributor: norbjd
-# Maintainer:
+# Maintainer: norbjd <norbjd@users.noreply.github.com>
 pkgname=scala
 pkgver=2.12.2
 pkgrel=0
-pkgdesc="Scala 2.12.2"
+pkgdesc="The Scala Programming Language"
 url="http://scala-lang.org"
 arch="all"
 license="BSD3"
 
-# bash          : needed for scripts (e.g. scala)
 # openjdk8-jre  : needed for execution 
 depends="bash openjdk8-jre"
 
-# openjdk8      : needed for sbt + compilation
-# sbt           : needed for compilation
-# graphviz      : needed for diagrams generation in scaladoc
+# openjdk8  : needed for sbt + compilation
+# sbt       : needed for compilation
+# graphviz  : needed for diagrams generation in HTML scaladoc
 makedepends="openjdk8 sbt graphviz"
 
 # documentation is separated
-subpackages="${pkgname}-doc"
+subpackages="$pkgname-doc"
 
-source="${pkgname}-${pkgver}.tar.gz::https://github.com/scala/scala/archive/v2.12.2.tar.gz"
-builddir="${srcdir}/${pkgname}-${pkgver}"
+source="$pkgname-$pkgver.tar.gz::https://github.com/$pkgname/$pkgname/archive/v$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
 
 # important directories
-_scala_home="/usr/share/${pkgname}-${pkgver}"
-_scala_docs_home="/usr/share/doc/${pkgname}-${pkgver}"
-_scala_man_home="/usr/share/man/man1"
-_bin_home="/usr/bin"
+_scala_home="/usr/share/$pkgname-$pkgver"
+_scala_docs_home="/usr/share/doc/$pkgname-$pkgver"
 
-# sbt command with JVM options (memory)
+# sbt command with JVM options (memory, to avoid OOM errors)
 _sbt() {
-        sbt -Xms2G -Xmx4G -XX:MaxMetaspaceSize=1G "$@"
+    sbt -Xms2G -Xmx4G -XX:MaxMetaspaceSize=1G "$@"
 }
 
 build() {
-        cd "$builddir"
+    cd "$builddir"
 
-        # use sbt to compile, package, and create doc
-	_sbt ';dist/mkPack;doc;compile:managedResources;packageDoc' || return 1
+    # use sbt to compile, package, and create doc
+    _sbt ';dist/mkPack;doc;compile:managedResources;packageDoc'
 }
 
 check() {
-        cd "$builddir"
+    cd "$builddir"
 
-        # check if all scripts have been created and work
-	./build/pack/bin/scala -version
-	./build/pack/bin/scalac -version
-	./build/pack/bin/scaladoc -version
-	./build/pack/bin/scalap -version
-	./build/pack/bin/fsc -version
+    # check if all scripts have been created and work
+    ./build/pack/bin/scala -version
+    ./build/pack/bin/scalac -version
+    ./build/pack/bin/scaladoc -version
+    ./build/pack/bin/scalap -version
+    ./build/pack/bin/fsc -version
 }
 
 package() {
-        cd "$builddir"
+    cd "$builddir"
 
-        # copy scripts
-	mkdir -p "$pkgdir"/"$_scala_home" || return 1
-	cp -R build/pack/* "$pkgdir"/"$_scala_home"/ || return 1
+    # copy scripts
+    mkdir -p "$pkgdir/$_scala_home"
+    cp -R build/pack/* "$pkgdir/$_scala_home"
 
-        # remove useless specific Windows scripts
-	rm "$pkgdir"/"$_scala_home"/bin/*.bat || return 1
+    # remove useless specific Windows scripts
+    rm "$pkgdir/$_scala_home"/bin/*.bat
 
-        # create symbolic links
-	mkdir -p "$pkgdir"/"$_bin_home" || return 1
-	for _bin in `ls "$pkgdir"/"$_scala_home"/bin/`
-	do
-		ln -sf "$_scala_home"/bin/"$_bin" "$pkgdir"/"$_bin_home"/"$_bin" || return 1
-	done
+    # create symbolic links
+    mkdir -p "$pkgdir"/usr/bin
+
+#    local _bin
+#    for _bin in "$pkgdir/$_scala_home"/bin/*
+#    do
+#        local _bin_basename=$(basename "$_bin")
+#        ln -sf "$_scala_home"/bin/"$_bin_basename" "$pkgdir"/usr/bin/"$_bin_basename"
+#    done
+    
+    ls "$pkgdir/$_scala_home"/bin/ | \
+    xargs -I_bin \
+        ln -sf "$_scala_home"/bin/_bin "$pkgdir"/usr/bin/_bin
+}
+
+man_doc() {
+    cd "$builddir"
+
+    mkdir -p "$subpkgdir"/usr/share/man/man1
+
+    # copy manpages for scripts
+#    local _manfile
+#    for _manfile in $(find "$builddir"/target/scala-dist/resource_managed/main/man/man1/ -type f)
+#    do
+#        gzip "$_manfile"
+#        cp "$_manfile".gz "$subpkgdir"/usr/share/man/man1
+#    done
+
+    find "$builddir"/target/scala-dist/resource_managed/main/man/man1/ -type f \
+        -exec gzip {} \; \
+        -exec cp {}.gz "$subpkgdir"/usr/share/man/man1 \;
+}
+
+html_doc() {
+    cd "$builddir"
+
+    mkdir -p "$subpkgdir/$_scala_docs_home"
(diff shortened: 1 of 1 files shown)


testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+pkgver=2.12.2
+pkgrel=0
+pkgdesc="Scala 2.12.2"
+url="http://scala-lang.org"
+arch="all"
+license="BSD3"
+
+# bash          : needed for scripts (e.g. scala)
+# openjdk8-jre  : needed for execution 
+depends="bash openjdk8-jre"
```
Shizmob: Is the JRE dependency not inferred by abuild automatically? It should do dynamic library analysis on the final executables, but I'm not sure how Scala works. The `bash` dependency reason doesn't need to be documented so thoroughly.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+
+# openjdk8      : needed for sbt + compilation
+# sbt           : needed for compilation
+# graphviz      : needed for diagrams generation in scaladoc
+makedepends="openjdk8 sbt graphviz"
+
+# documentation is separated
+subpackages="${pkgname}-doc"
```
Shizmob: General style is to not use braces for variables unless strictly needed -- this also goes for the variables below.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
+# Contributor: norbjd
+# Maintainer:
```
Shizmob: Are you willing to maintain it too, perhaps? :)

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+pkgname=scala
+pkgver=2.12.2
+pkgrel=0
+pkgdesc="Scala 2.12.2"
```
Shizmob: I'd go with a somewhat more descriptive description here, even just `The Scala programming language` should do.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+
+# openjdk8      : needed for sbt + compilation
+# sbt           : needed for compilation
+# graphviz      : needed for diagrams generation in scaladoc
+makedepends="openjdk8 sbt graphviz"
+
+# documentation is separated
+subpackages="${pkgname}-doc"
+
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/scala/scala/archive/v2.12.2.tar.gz"
```
Shizmob: It might be more maintainable to make the URL use `$pkgname/archive/v$pkgver.tar.gz`.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+# sbt command with JVM options (memory)
+_sbt() {
+        sbt -Xms2G -Xmx4G -XX:MaxMetaspaceSize=1G "$@"
+}
+
+build() {
+        cd "$builddir"
+
+        # use sbt to compile, package, and create doc
+	_sbt ';dist/mkPack;doc;compile:managedResources;packageDoc' || return 1
```
Shizmob: The `|| return 1` clauses here can be omitted, as `abuild` does `set -e` automatically these days. This also goes for the instances below.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+
+check() {
+        cd "$builddir"
+
+        # check if all scripts have been created and work
+	./build/pack/bin/scala -version
+	./build/pack/bin/scalac -version
+	./build/pack/bin/scaladoc -version
+	./build/pack/bin/scalap -version
+	./build/pack/bin/fsc -version
```
Shizmob: Nice smoke test. Does Scala offer a test suite we can too, perhaps? Do you know?

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+}
+
+package() {
+        cd "$builddir"
+
+        # copy scripts
```
Shizmob: Seems to be an indentation issue here.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+makedepends="openjdk8 sbt graphviz"
+
+# documentation is separated
+subpackages="${pkgname}-doc"
+
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/scala/scala/archive/v2.12.2.tar.gz"
+builddir="${srcdir}/${pkgname}-${pkgver}"
+
+# important directories
+_scala_home="/usr/share/${pkgname}-${pkgver}"
```
Shizmob: Does the home need to be versioned? I don't particularly see why. Same for the doc location.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+_scala_docs_home="/usr/share/doc/${pkgname}-${pkgver}"
+_scala_man_home="/usr/share/man/man1"
+_bin_home="/usr/bin"
```
Shizmob: These two are somewhat obvious and shouldn't really be expected to be changed anytime soon - I don't think variables are needed for this.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+        # copy scripts
+	mkdir -p "$pkgdir"/"$_scala_home" || return 1
+	cp -R build/pack/* "$pkgdir"/"$_scala_home"/ || return 1
+
+        # remove useless specific Windows scripts
+	rm "$pkgdir"/"$_scala_home"/bin/*.bat || return 1
+
+        # create symbolic links
+	mkdir -p "$pkgdir"/"$_bin_home" || return 1
+	for _bin in `ls "$pkgdir"/"$_scala_home"/bin/`
```
Shizmob: I'd use local variable here: `local bin; for bin in ...`. Also, you can use `for bin in "$pkgdir"/"$scala_home"/bin/*` instead of a subprocess `ls` invocation.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+    	mkdir -p "$subpkgdir"/"$_scala_docs_home"/api
+    	mkdir -p "$subpkgdir"/"$_scala_docs_home"/api/jars
+    	mkdir -p "$subpkgdir"/"$_scala_docs_home"/tools
+    	mkdir -p "$subpkgdir"/"$_scala_man_home"
+
+        # copy default doc resources (licenses, README)
+  	cp -R "$builddir"/doc/* "$subpkgdir"/"$_scala_docs_home"/ || return 1
+
+        # copy HTML scaladoc for default API
+        cp -R "$builddir"/build/scaladoc/compiler       "$subpkgdir"/"$_scala_docs_home"/api/compiler || return 1
```
Shizmob: Some more non-tab indentation here.

testing/scala/APKBUILD
```diff
@@ -0,0 +1,111 @@
...
+        cp -R "$builddir"/build/scaladoc/library        "$subpkgdir"/"$_scala_docs_home"/api/library || return 1
+        cp -R "$builddir"/build/scaladoc/reflect        "$subpkgdir"/"$_scala_docs_home"/api/reflect || return 1
+        cp -R "$builddir"/build/scaladoc/scalap         "$subpkgdir"/"$_scala_docs_home"/api/scalap || return 1
+
+        # copy HTML doc for tools (scripts)
+        cp -R "$builddir"/src/manual/scala/tools/docutil/resources/* "$subpkgdir"/"$_scala_docs_home"/tools/ || return 1
+    	cp "$builddir"/target/scala-dist/resource_managed/main/doc/tools/* "$subpkgdir"/"$_scala_docs_home"/tools/ || return 1
+
+        # copy manpages for scripts
+	for _manfile in `find "$builddir"/target/scala-dist/resource_managed/main/man/man1/ -type f`
```
Shizmob: Same comments for-loop wise as above apply here.

testing/scala/APKBUILD
jirutka: No, Scala is written in Java, not C/C++. ;)

norbjd:
Hello @Shizmob,

Thanks for the review ! I changed some things according to your remarks :

- Syntax (remove braces, indentation)
- Replace `for` loops with `xargs` or `find ... -exec`
- Remove useless variables (`_scala_man_home` or `_bin_home`). I chose to keep `_scala_home` and `_scala_docs_home` to avoid repetition of `/usr/share/$pkgname-$pkgver` or `/usr/share/doc/$pkgname-$pkgver` (less readable)

For the `check` operation, I think those tests are sufficient. There are tests of the source code (unit tests, integration tests, ...) but to my opinion, it's not useful to run them, since the source code is from an official release (therefore it should pass all tests).

Finally, about the documentation : It could be useful to have also a local HTML documentation corresponding **exactly** to the version of Scala packaged. Same thing about the Javadoc JARs. For users who don't need doc, the `scala-doc` package can simply be not installed.

jirutka:
@Shizmob It downloads a lot of dependencies from Maven repository during build.

norbjd:
> It downloads a lot of dependencies from Maven repository during build.

Indeed, SBT (Scala Build Tool) needs some artifacts to properly build the Scala project (plugins for building + of course dependencies). Can we do something about that (maybe use a cache Maven repository for dependencies) ?

Shizmob:
We ideally want all dependencies specified in the APKBUILD so they can be properly verified and checksummed. Is there a way to efficiently solve this? The package should buildable without internet access, basically.

norbjd:
The dependencies for the build here are a hundred of JARs. I've got the exact list by looking into my cleaned local repository after doing the build, but how can I specify those JARs in the `APKBUILD`, since they are files and not "packages" like `openjdk8-jre` or `sbt` ?

I can surely download them, verify their checksum before the build, and then put them to the local cache repository, but this may provoke difficulties to maintain the package, since some dependencies (and versions) could change from a version of Scala to another.

I looked for a similar example using `sbt` and already packaged in Alpine, I found this : https://github.com/alpinelinux/aports/blob/master/community/java-snappy/APKBUILD. I've built it, the JARs **are downloaded during the build**.

I am a little confused, please let me know the preferred way. To my opinion, downloading the JARs outside of `sbt` is not recommended, since `sbt` is used to manage dependencies and their versions (and of course building).

jirutka:
> I found this : https://github.com/alpinelinux/aports/blob/master/community/java-snappy/APKBUILD. I've built it, the JARs are downloaded during the build.

Yes, java-snappy violates this rule, I should fix it too.

This is evergreen problem. It’s indeed nonsense to declare all these JARs in APKBUILD and duplicate job of sbt. We just need to somehow ensure that everytime you build the specific version exactly the same source files will be used.

I’ll look at it this later.

norbjd:
I tried to add all the dependencies JARs one by one in the `APKBUILD`, but this is too difficult because there are more than a hundred JARs, and it's sometimes hard to find the URL to download a specific JAR. Moreover, I think this solution is too complicated to maintain.

So I decided to test an other solution. After running `sbt update` (download all necessary stuff for building) from a clean environment, I just packaged the downloaded dependencies to a `tar.gz` archive, and add this to the `APKBUILD` sources. I found that `sbt` need 2 directories :
- the `sbt.boot.directory` (`.boot`), containing among others, JARs relative to a previous version of `scala`, here 2.10.6 (required for `sbt` to work)
- the ivy cache (`.ivy`), containing build dependencies for `scala` (defined in the `build.sbt`)

When running `sbt` with `-no-share` and `-offline` options, those two directories must be in `project` folder.  So unarchiving the `tar.gz` in the `project` folder BEFORE building `scala` binaries with `sbt` allows to build the apk without Internet access, as required, and moreover, multiple builds of the package will be always based on the same source files (packaged in the `.tar.gz`).

The only problem I see here is the need to upload the `tar.gz` containing all dependencies. Could it be put in the folder where `APKBUILD` is ? Or is it better to put it somewhere on the web ? The archive size is 103 MB.

PR #1578, opened 2017-05-27
testing/firefox: update to 53.0.3

khanku:
Update to firefox 53.0.3 which requires rust (could be disabled still but will be required in 54 as per https://bugzilla.mozilla.org/show_bug.cgi?id=1284816) and dropped the mozilla SDK as per https://bugzilla.mozilla.org/show_bug.cgi?id=1333826.

Patches:
- fix-bug-1261392.patch is from https://hg.mozilla.org/integration/mozilla-inbound/rev/bc44cbe93310
- fix-bug-1338655.patch is from https://reviewboard.mozilla.org/r/111682/diff/
- fix-rust-target.patch is from myself, someone with more musl/rust experience probably should review it.


testing/firefox/APKBUILD
```diff
@@ -10,9 +10,13 @@ url="http://www.firefox.com"
 arch="all !aarch64"
 license="GPL LGPL MPL"
 depends=""
-depends_dev="
+makedepends="$depends_dev
 	alsa-lib-dev
+	autoconf2.13
```
Shizmob: Are all these added dependency actually needed? I don't recall them being so before.

khanku:
@Shizmob as far as I can tell the SDK is gone, see https://bugzilla.mozilla.org/show_bug.cgi?id=1333826

testing/firefox/APKBUILD
khanku: Those dependencies use to be in "depends_dev" and were included in "makedepends" with "$depends_dev".

testing/firefox/APKBUILD
```diff
@@ -25,32 +29,27 @@ depends_dev="
...
 	sed
+	startup-notification-dev
+	wireless-tools-dev
 	yasm
 	zip
 	"
 
 install=""
-subpackages="$pkgname-dev"
+subpackages=""
```
Shizmob: This line can be omitted if it's empty.

testing/firefox/APKBUILD
```diff
@@ -10,9 +10,13 @@ url="http://www.firefox.com"
 arch="all !aarch64"
 license="GPL LGPL MPL"
 depends=""
-depends_dev="
+makedepends="$depends_dev
```
Shizmob: Now that `depends_dev` is gone, it doesn't need to be in `makedepends` either.

testing/firefox/fix-rust-target.patch
```diff
@@ -0,0 +1,11 @@
+--- a/build/moz.configure/rust.configure	2017-05-27 10:48:23.927528005 +0000
++++ b/build/moz.configure/rust.configure	2017-05-27 10:52:42.973790729 +0000
+@@ -130,7 +130,7 @@
+             ('arm', 'Linux'): 'armv7-unknown-linux-gnueabihf',
+             ('sparc64', 'Linux'): 'sparc64-unknown-linux-gnu',
+             ('x86', 'Linux'): 'i686-unknown-linux-gnu',
+-            ('x86_64', 'Linux'): 'x86_64-unknown-linux-gnu',
```
Shizmob: We build Firefox for more arches than just x86_64, so if you could fix those up too that would be great.

khanku: This might prove problematic as there is no rust for any other architecture than x86_x64 at the moment, likely because the void linux supports only x86_64, see https://github.com/alpinelinux/aports/blob/master/community/rust/APKBUILD#L26.

We could use "--disable-rust" but that will stop working with version 54 (see https://bugzilla.mozilla.org/show_bug.cgi?id=1284816)

Shizmob: iight. I'm going to work on enabling Rust on other architectures - Void's thing is not an issue as we can cross-bootstrap arches now that we have our own binaries, it's just used out of convenience as it works with our build servers until we have our own bootstrap.

khanku: I extended the patch to cover x86 and armhf; aarch64 is explicitly disabled in the APKBUILD and sparc is unsupported.
Obviously, with no rust, I could not test those.

PR #1587, opened 2017-05-29
main/py-requests: upgrade to 2.16.5

fabaff:


main/py-requests/APKBUILD
```diff
@@ -2,12 +2,13 @@
...
 _pkgname=requests
-pkgver=2.13.0
+pkgver=2.16.5
 pkgrel=0
 pkgdesc="A HTTP request library for Python"
 url="http://www.python-requests.org/"
 arch="noarch"
-license="ASL-2.0"
+license="ASL 2.0"
+options="!check"
```
Shizmob: No upstream testsuite? If so, please add a comment indicating it.

Shizmob: The same comment as in #1588 for the license applies.

fabaff: Here the license matches the current [convention](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license).

Comment added

Shizmob:
👍 LGTM now.

PR #1588, opened 2017-05-29
main/py-chardet: upgrade to 3.0.3

fabaff:


main/py-chardet/APKBUILD
```diff
@@ -2,12 +2,12 @@
...
 pkgname=py-$_pkgname
-pkgver=2.3.0
-pkgrel=3
+pkgver=3.0.3
+pkgrel=0
 pkgdesc="The Universal Character Encoding Detector"
 url="https://github.com/chardet/chardet/"
 arch="noarch"
-license="LGPL-2.1"
+license="LGPL 2.1"
```
Shizmob: What's the reason for this? I'm unsure what we currently use as license separator, but I think it's space in a few packages.

fabaff: No, it should be [nothing](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license). Seems I can't remember what I wrote 3.5 years back.

PR #1593, opened 2017-05-29
testing/idris: Fix build issues when newer packages present

mitchty:
Misc fixes relating to use of freeze files and cabal.


testing/idris/APKBUILD
```diff
@@ -18,19 +15,20 @@ source="$pkgname-$pkgver.tar.gz::https://github.com/idris-lang/Idris-dev/archive
 	cabal.config"
 builddir="$srcdir/Idris-dev-$pkgver"
 
-export CABAL_CONFIG="$srcdir/cabal/cabal.config"
+CABAL_HOME="$srcdir/cabal"
+export CABAL_HOME
```
Shizmob: Does this need to be exported? Do any sub-processes rely on it?

testing/idris/APKBUILD
```diff
@@ -51,30 +49,34 @@ prepare() {
...
 	# Create Cabal sandbox that will be used for all subsequent cabal
 	# invocations. This is something like Python virtual env.
-	cabal sandbox init
+	HOME="$CABAL_HOME" cabal sandbox init
 
 	# Fetch all dependencies.
-	# XXX: This should always fetch the same versions as defined in
-	# our cabal.config, but Cabal does not verify packages integrity!
-	cabal fetch libffi .
+	for pkg in $(cat cabal.config | sed -e 's/constraints[:]//g' -e 's/--.*//g' | tr -d '[:space:]' | sed -e 's/[,]/ /g' -e 's/==/-/g'); do
```
Shizmob: Maybe we should move this pipeline into a special function.

testing/idris/APKBUILD
mitchty: Nope, i'll nuke the export.

testing/idris/APKBUILD
mitchty: Sure, its a hack for now, i'm looking at fixing cabal fetch to take constraints into play as noted in that bug. But i'm a crappy haskell developer so it might be some time.

testing/idris/APKBUILD
Shizmob: Since it's a non-standard variable, could you prefix it with `_`, like `_cabal_home`?

testing/idris/APKBUILD
Shizmob: Roger.

mitchty: Ok split it apart into 3 functions as I noticed my second copy/paste was wrong, now it actually passes --constraint=package==version properly. It worked more due to dumb luck before but this is what I was trying to accomplish to tell the cabal install constraint solver to use the freeze packages and not what the newest that might be in the index.

testing/idris/APKBUILD
mitchty: Done

nmeum:
I would personally prefer it if helper function like `cabal_update` were prefixed with an underscore other than that the update looks fine to me. Would merged it if you make that change.

jirutka:
`cabal_update` is not a split function, it’s a helper function that is intended to be called from CLI (`abuild cabal_update`), so it should not be prefixed with underscore.

This is not the first package when we need some special procedure for bumping, so it’d be better to define special function like `bump` that will be called by `abump` when bumping the package.

nmeum:
> cabal_update is not a split function, it’s a helper function that is intended to be called from CLI (abuild cabal_update), so it should not be prefixed with underscore.

Ah, well didn't know that (sorry!). You are right, it's fine in that case.

mitchty:
Just to be sure, is there anything needed on my end change wise or is everything ok as it currently stands?

nmeum:
Well…looking at the aport again the comment written by @jirutka only applies to `cabal_update`.`cabal_constraints_prime`, `cabal_pkgs` and `cabal_constraints` are in fact local functions, right? In that case I would indeed suggest prefixing them with an underscore. @jirutka what's your opinion on this?

nmeum:
Merged! Thanks for your work on this and sorry it took so long to get it merged :/

nmeum:
@mitchty Totally forgot to inform you about this: The build failed on armhf, thus idris was disabled on that arch for now (see c2fc85c203693d141a255af17efc02ed02237811). Any idea what could have caused this?

mitchty:
@nmeum hard to say, any logs I can look at?

nmeum:
> any logs I can look at?

http://tpaste.us/Vbxz

mitchty:
Curious on this bit:

```
cabal: Missing dependencies on foreign libraries:
* Missing C libraries: ffi, ffi
This problem can usually be solved by installing the system packages that
provide these libraries (you may need the "-dev" versions). If the libraries
are already installed but in a non-standard location then you can use the
flags --extra-include-dirs= and --extra-lib-dirs= to specify where they are.
cabal: Leaving directory '/tmp/cabal-tmp-29534/libffi-0.1'
Updating world file...
cabal: Error: some packages failed to install:
```

Shouldn't be hard to fix, but need to reproduce it, my arm builder seems to be building it fine (takes a while though). At least it worked in June...ish based on the file dates.

What version of alpine linux is run on the builders?

nmeum:
> What version of alpine linux is run on the builders?

The builder which emitted the error log linked above runs Alpine Linux edge.

mitchty:
Cool i'll see if i can't replicate it that way.

PR #1600, opened 2017-05-31
testing/ncmpcpp: upgrade to 0.8

mmonnier:


testing/ncmpcpp/APKBUILD
```diff
@@ -27,19 +27,15 @@ build () {
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
 		--localstatedir=/var \
-		--disable-unicode \
 		--enable-clock \
 		--with-taglib \
-		--with-curl \
```
Shizmob: Why was `--with-curl` removed?

mmonnier: The libcurl dependency is no longer optional.

PR #1603, opened 2017-05-31
main/py-django: add py3 subpackage

almikhayl:
main/py-django: add py3 subpackage

### DIFF
--- a/main/py-django/APKBUILD
+++ b/main/py-django/APKBUILD
@@ -3,16 +3,16 @@
 pkgname=py-django
 _pkgname=Django
 pkgver=1.11
-pkgrel=0
+pkgrel=1
 pkgdesc="A high-level Python Web framework"
 url="http://djangoproject.com/"
 arch="noarch"
 license="BSD"
-depends="python2"
+depends=""
 depends_dev=""
-makedepends="python2-dev py-setuptools"
+makedepends="python2-dev python3-dev py-setuptools"
 install=""
-subpackages=""
+subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
 source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
 
 builddir="$srcdir"/$_pkgname-$pkgver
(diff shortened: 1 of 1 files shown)


main/py-django/APKBUILD
```diff
@@ -3,16 +3,16 @@
...
 _pkgname=Django
 pkgver=1.11
-pkgrel=0
+pkgrel=1
 pkgdesc="A high-level Python Web framework"
 url="http://djangoproject.com/"
 arch="noarch"
 license="BSD"
-depends="python2"
+depends=""
```
Shizmob: You can omit this line if it's now empty.

main/py-django/APKBUILD
```diff
@@ -27,12 +27,33 @@ builddir="$srcdir"/$_pkgname-$pkgver
 
 build() {
 	cd "$builddir"
-	python2 setup.py build || return 1
+	python2 setup.py --quiet build || return 1
```
Shizmob: `|| return 1` statements can be omitted now.

main/py-django/APKBUILD
```diff
@@ -27,12 +27,33 @@ builddir="$srcdir"/$_pkgname-$pkgver
...
+	python3 setup.py --quiet build || return 1
 }
 
 package() {
 	cd "$builddir"
-	python2 setup.py install --root "$pkgdir" --skip-build || return 1
+	python2 setup.py --quiet install --root "$pkgdir" --skip-build || return 1
```
Shizmob: Likewise, `|| return 1` statements are unneeded.

main/py-django/APKBUILD
almikhayl: Thanks for suggestion. It seems that at least not so few packages providing py3 and py2 subpackages use empty but present `depends` variable. For example main/py-yaml, main/py-tz, py-asn1. For now I will leave it as is.

jirutka:
Django installs `/usr/bin/django-admin`, so you’d end up with conflicting file between py2 and py3 subpackage. Please add version suffix (`-2`, `-3`) to it and symlink unsuffixed to the py3 variant.

PR #1610, opened 2017-06-01
testing/perl-config-properties: new aport

schipplock:
http://search.cpan.org/~salva/Config-Properties-1.80/lib/Config/Properties.pm
Config::Properties - Read and write property files

### DIFF
--- a/testing/perl-config-properties/APKBUILD
+++ b/testing/perl-config-properties/APKBUILD
@@ -1,0 +1,47 @@
+# Contributor: Andreas Schipplock <andreas@schipplock.de>
+# Maintainer: Andreas Schipplock <andreas@schipplock.de>
+pkgname=perl-config-properties
+_pkgreal=Config-Properties
+pkgver=1.80
+pkgrel=0
+pkgdesc="Read and write property files"
+url="http://search.cpan.org/dist/Config-Properties/"
+arch="noarch"
+license="GPL PerlArtistic"
+cpandepends=""
+cpanmakedepends=""
+cpancheckdepends=""
+depends="$cpandepends"
+makedepends="perl-dev $cpanmakedepends"
+checkdepends="$cpancheckdepends"
+subpackages="$pkgname-doc"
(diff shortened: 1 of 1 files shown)


testing/perl-config-properties/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+# Maintainer: Andreas Schipplock <andreas@schipplock.de>
+pkgname=perl-config-properties
+_pkgreal=Config-Properties
+pkgver=1.80
+pkgrel=0
+pkgdesc="Read and write property files"
+url="http://search.cpan.org/dist/Config-Properties/"
+arch="noarch"
+license="GPL PerlArtistic"
+cpandepends=""
```
Shizmob: If lines are empty (or empty as a result of containing empty variables, you can safely omit then.

testing/perl-config-properties/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+cpancheckdepends=""
+depends="$cpandepends"
+makedepends="perl-dev $cpanmakedepends"
+checkdepends="$cpancheckdepends"
+subpackages="$pkgname-doc"
+source="http://search.cpan.org/CPAN/authors/id/S/SA/SALVA/$_pkgreal-$pkgver.tar.gz"
+builddir="$srcdir/$_pkgreal-$pkgver"
+
+prepare() {
+	default_prepare || return 1
```
Shizmob: Could you remove the `|| return 1` here? It should not be needed.

testing/perl-config-properties/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+
+build() {
+	cd "$builddir"
+	export CFLAGS=$(perl -MConfig -E 'say $Config{ccflags}')
+	make
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install || return 1
```
Shizmob: Likewise, the `|| return 1` should not be needed.

Shizmob:
Squashed and merged, thanks for contributing!

PR #1622, opened 2017-06-06
main/musl: Include patch to musl to getgr, getpw, and getspnam to set errno on error

r0p0s3c:
Without this, shadow utilities (user/group add/mod) do not cope with
groups that have large numbers of members, as the memory reallocation
code assumes errno will be set to ERANGE in such cases.

My plan is to also push this upstream to musl, so this is should only be required in the meantime


ncopa:
LGTM, but I want hear what musl devs says about it first.

jirutka:
Also pkgrel must be incremented.

r0p0s3c:
Apologies, done. I haven't heard back from the musl devs yet, btw.

r0p0s3c:
This has been resolved upstream in commit [2d7d05f](https://git.musl-libc.org/cgit/musl/commit/?id=2d7d05f031e014068a61d3076c6178513395d2ae) so no longer required

ncopa:
do we need to back port that to alpine `3.6-stable`?

r0p0s3c:
Good point. I presume even edge only brings in official musl releases also? Since I am not sure of the timing of that, is the best strategy to submit PRs/patches for both stable and edge?

PR #1637, opened 2017-06-09
testing/vips: upgrade to 8.5.6

AVVS:
libvips moved to https://github.com/jcupitt/libvips, APKBUILD now uses releases from there


ncopa:
i merged the patches. thanks!

PR #1644, opened 2017-06-09
main/gnutls: upgrade to 3.5.13

misery:
Maybe backport to v3.6? http://gnutls.org/security.html#GNUTLS-SA-2017-4


ncopa:
bet way to make sure we dont forget 3.6-stable is to file a bug on https:/bugs.alpinelinux.org with target `3.6.2`

PR #1675, opened 2017-06-13
testing/rspamd upgraded to 1.6.3

TBK0000:


TBK0000:
@ncopa Is there anything I can do to speed up the process?

The tweaks I have made to the build script are a result of this commit:
https://github.com/vstakhov/rspamd/commit/17a6ce2ae11994e60f88e934edf2f62ce799089c

tmpfile:
abuild now runs with `set -e` so we don't need any ` || return 1` now as it's implicit by default.
So you should remove all instances of ` || return 1`.
Apart from that it LGTM.
Thanks

TBK0000:
@tmpfile Fixed 👍

fabled:
Another PR with same goal was merged, can you check what it contains and rebase and create new PR for any other pending changes. Thanks.

TBK0000:
@fabled I have create a PR https://github.com/alpinelinux/aports/pull/2213 to fix an issue with the current build.

PR #1689, opened 2017-06-14
testing/weston: change arch to "all"

ollieparanoid:
Changed arch to "all", so it can be built on aarch64 (and I see no reason, why it should not work on other architectures).
I did not increase the pkgrel, because a rebuild should not be necessary for this.

### DIFF
--- a/testing/weston/APKBUILD
+++ b/testing/weston/APKBUILD
@@ -7,7 +7,7 @@ _libname=lib$pkgname
 _libdir=$_libname-${pkgver%%.*}
 pkgdesc="The reference Wayland server"
 url="http://wayland.freedesktop.org"
-arch="x86_64 armhf"
+arch="all"
 license="MIT"
 depends=""
 makedepends="wayland-protocols libxkbcommon-dev xkeyboard-config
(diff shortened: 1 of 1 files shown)


ollieparanoid:
This commit did not change anything regarding weston's `check()` function. Strange that it fails now. What is the policy here? Will this PR get accepted, regardless of this, or do I have to fix the testsuite (although I didn't change anything that could cause it to fail, meaning it must have failed before)?

Thanks!

jirutka:
I think that the tests in check phase failed even before, but it was ignored – we run abuild with `set -e` now.

ollieparanoid:
That makes sense. I was able to get down to 7 failures of 23 tests with the following check function:
```
check() {
    export XDG_RUNTIME_DIR="$builddir/.xdg-runtime-dir"
    mkdir -p "$XDG_RUNTIME_DIR"
    mkdir -p "/tmp/.X11-unix/" # hardcoded in xwayland
    make -C "$builddir" check
    return 0
}
```

Creating the `/tmp/.X11-unix/` folder seems ugly to me though, do we really want to do this? I've noticed, that Arch does not have a `check()` function for `weston`, so this would be the easy way out.

Would it be acceptable to use `!check` in the `APKBUILD`?

kaniini:
!check is probably ok for now, but we should fix the tests too, as the idea is to use tests as an early canary for finding breakages.

ollieparanoid:
Okay, I did the lazy version with `!check` for now. I think the right way to fix this is talk to upstream (which I might do eventually).

PR #1693, opened 2017-06-15
main/userspace-rcu: upgrade to 0.10.0

fabaff:


ncopa:
breaks abi. a number of packages needs rebuild

PR #1697, opened 2017-06-15
Bug #7404 TLS negotiation error in OpenJDK 8 u131

shatil:
Fixes an OpenJDK 8 regression discovered in docker-library/openjdk#115
on Alpine Linux 3.5 (u121) and 3.6 (u131) that causes TLS negotiation
errors for some clients.

Root cause appears to be OpenJDK announcing support for NIST curves the
underlying NSS library does doesn't. This patch limits OpenJDK's
announcement to elliptic curves 23 (secp256r1), 24 (secp384r1), and 25
(secp521r1).

Related issues:

* https://github.com/docker-library/openjdk/issues/115
* https://bugs.alpinelinux.org/issues/7404
* https://access.redhat.com/discussions/2339811
* https://bugzilla.redhat.com/show_bug.cgi?id=1022017
* https://bugzilla.redhat.com/show_bug.cgi?id=1348525


shatil:
Peachy.

```

The job has been terminated
```

Raw log looks okay, though, so far.

PR #1713, opened 2017-06-17
main/glew: upgrade to 2.0.0

tmpfile:


fabled:
SONAME changes, so all packages depending on glew needs rebuild after this.

PR #1714, opened 2017-06-17
main/guile: upgrade to 2.2.2

tmpfile:


tmpfile:
`strip: Unable to recognise the format of the input file `./usr/lib/guile/2.2/ccache/language/tree-il.go'`

Should I add options="!strip"?

kaniini:
it's a weird situation, i am not sure.  we may want to fix this for real instead of just disabling strip.

fabled:
Seems the `.go` files are ELF objects of type `no machine`. This makes the default `strip` fail. This is used to work before abuild was in `set -e` mode, so I wonder if we just fix abuild to ignore strip return value, or to detect the ELF binary machine type and ignore these ones. Alternative is to use `option= "!strip"`, but then the real ELF binaries would need to be manually stripped.
@ncopa Any ideas here?

ncopa:
I wonder if we can make `scanelf` skip the `no machine`  ELFs or maybe use `grep -v` or similar to filter those out in abuild?

fabled:
Apparently `scanelf` does not support filtering or displaying the machine type easily. Another option is to print the OSABI (`-I`) and filter based on that: the `.go` files are distinctively of type `standalone`. Unfortunately scanelf has a bug making it not display that correctly, but it's fixable.

tmpfile:
Thanks fabled. Good work.

fabled:
This built fine on my LXC, but on real builders it fails with:
```
/bin/bash: line 6: 16478 Segmentation fault      GUILE_AUTO_COMPILE=0 ../meta/build-env guild compile --target="powerpc64le-alpine-linux-musl" -O1 -L "/home/buildozer/aports/main/guile/src/guile-2.2.2/module" -L "/home/buildozer/aports/main/guile/src/guile-2.2.2/guile-readline" -o "language/tree-il/peval.go" "../module/language/tree-il/peval.scm"
make[2]: *** [Makefile:1928: language/tree-il/peval.go] Error 139
```

Any ideas?

fabled:
Ok, now it built on 3rd or 4th attempt. Seems it's random build failure. Possible related to PIE.

kaniini:
This is related to recursion overflow.  s390x is not able to build guile 2.2, at all.

kaniini:
I suggest we downgrade guile back to 2.0.14, 2.2 isn't ready for primetime from what I see.

PR #1718, opened 2017-06-17
main/nasm: upgrade to 2.13.01

tmpfile:


fabled:
Build fails with:
> make: *** No rule to make target 'nasmlib.o'.  Stop.

Care to fix the issue?

tmpfile:
That's because they moved nasmlib.h inside a directory and split nasmlib.c into many other .c files.

What do you want me to do?
I will just remove the nasmlib.o target for now. If you want me to add any other target, let me know.

fabled:
I suppose there has been some parallel build issue that got fixed by building that specific target first. If the changelog indicates parallel build problem has been fixed, what you did is right solution. We'll also test build this on a 24-cpu box.

PR #1720, opened 2017-06-17
main/seabios: upgrade to 1.10.2

tmpfile:


tmpfile:
```
Error!  ROM doesn't fit (132160 > 131072)
   You have to either increase the size (CONFIG_ROM_SIZE)
   or turn off some features (such as hardware support not
   needed) to make it fit.  Trying a more recent gcc version
   might work too.
```

I'm not sure how to proceed.

PR #1722, opened 2017-06-18
testing/tsung: new aport

okeuday:
http://www.process-one.net/en/tsung/
Tsung is a high-performance benchmark framework for various protocols including HTTP, XMPP, LDAP, etc.

This package is using the latest git commit which is considered "stable" pre-release (without the version officially incremented).  A ".1" suffix was added to the (last) version to signify this, since "rc1" was not accepted by abuild.

### DIFF
--- a/testing/tsung/0005-Do-not-install-manpages-in-main-install.patch
+++ b/testing/tsung/0005-Do-not-install-manpages-in-main-install.patch
@@ -1,0 +1,36 @@
+--- tsung-445d78213bb4360d78df3753b3e23e6a28d701f0/Makefile.in
++++ tsung-445d78213bb4360d78df3753b3e23e6a28d701f0/Makefile.in
+@@ -211,6 +211,12 @@
+ 	-rm -f $(EBIN)/*.beam tsung.sh tsung.spec tsung.xml tsung.sh tsung-recorder.sh
+ 	-rm -f *.xml config.log src/test/*.xml src/test/usersdb.csv
+ 
++install_doc: tsung doc install_recorder install_controller $(CONFFILE)
++	install -d $(DESTDIR)$(MAN_DIR)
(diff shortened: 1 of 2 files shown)


okeuday:
The error in the Travis build is something I haven't seen locally, which says `py-matplotlib` depends on `py2-tkinter` .  This is strange, since it is called `python2-tkinter` everywhere, though that doesn't follow the common guidelines for python dependencies.

tmpfile:
This was fixed on https://github.com/alpinelinux/aports/commit/d5409f57dd274647052e4a4d2057e1232a2a36a6
But the pkgrel wasn't bumped, so the package didn't get rebuilt.

okeuday:
@tmpfile I attempted to bump the py-matplotlib pkgrel in this pull request, and the build did get further, though it required an additional package which has been added to the tsung makedepends.  However, the master branch also had a separate commit to bump the py-matplotlib pkgrel and the same error occurs now.  I have made sure there is no conflict in this pull request, but somehow it seems to have impacted the py-matplotlib pkgrel where it needs another bump?

okeuday:
This package should be fine now, but tell me if there is some problem I am unaware of.

tmpfile:
Since https://github.com/alpinelinux/abuild/commit/36d5193776180385a39626a83241822736a5f6b8 has been merged there's no need for || return 1.
It's implicit to every command.

tmpfile:
It also looks like this package should be set to `noarch` instead of `all`.

tmpfile:
You can also remove the prepare() function completely since it's not doing anything besides the default.

okeuday:
@tmpfile The single commit has been updated based on your feedback.  Setting the `arch` to `noarch` could break in the future if the `erlang` package ever enables hipe by default, but it is best to not have hipe enabled (if someone wanted it, it would require a separate erlang package name).

PR #1725, opened 2017-06-18
community/suitesparse: build at -O3

tmpfile:
The bug that prevented this has been fixed upstream.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71505


fabled:
That bug says it's duplicate of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71702 which is still open.

tmpfile:
Yes, but the specific case that prevented compilation at -O3 was fixed.
The bug still opened is a more general case that doesn't apply here.

Travis confirms, build passed.

PR #1731, opened 2017-06-19
testing/[various]: remove trailing whitespace in pkgdesc

tmpfile:


tmpfile:
Fails because it touches icingaweb2-module-businessprocess which needs the fix from https://github.com/alpinelinux/aports/pull/1728

PR #1736, opened 2017-06-20
main/font-alias: fix build

tmpfile:
Fixes https://bugs.alpinelinux.org/issues/7446 as per request.


tmpfile:
I received a request to fix this package and here it is.
We should also look into making `configure: WARNING: unrecognized options: [...]` an error, returning 1 and failing the build.
This would avoid similar issues in the future.

rnalrd:
Sorry, @fcolista pushed a fix before checking for pending PR

tmpfile:
No problem. It's good that it was fixed.

PR #1740, opened 2017-06-20
Audit updates

tych0:
Hi guys,

Here's some updates to the audit package (thanks @doughdemon for rebasing and updating the patches, hopefully we won't have to carry these for very long :).


tych0:
Thanks!

PR #1741, opened 2017-06-20
testing/cloudi: new aport

okeuday:
http://cloudi.org/
CloudI is an open-source private cloud computing framework for efficient, scalable, and stable soft-realtime event processing.

The configuration is all done within files that live in the directory `/etc/cloudi/` (so a `cloudi.confd` file wouldn't have anything to configure, since command line options aren't used with the `cloudi` executable).  The logging (including a pid file) is put into `/var/log/cloudi/`.  The `cloudi` executable is a start/stop script that needs to function synchronously to wait for other OS processes to terminate (external services) while enforcing its own timeout (a max of 60 seconds) and it needs to start the Erlang release in the background in a way where the pidfile is valid, which is why the `cloudi.initd` file is not using `start-stop-daemon`.  Some of the `makedepends` get dynamically linked and become runtime dependencies, but my understanding is that `apk` understands those packages need to exist on the installation as a dependency, even if they don't exist in the `depends` variable.

### DIFF
--- a/testing/cloudi/APKBUILD
+++ b/testing/cloudi/APKBUILD
@@ -1,0 +1,84 @@
+# Contributor: Michael Truog <mjtruog@gmail.com>
+# Maintainer: Michael Truog <mjtruog@gmail.com>
+
+pkgname=cloudi
+pkgver=1.7.1
+pkgrel=0
+pkgdesc="CloudI is an open-source private cloud computing framework for efficient, scalable, and stable soft-realtime event processing."
+url="http://cloudi.org/"
+license="MIT"
+arch="all"
+depends="g++"
+makedepends="autoconf
+             automake
+             binutils-dev
+             boost-dev
+             boost-system
+             boost-thread
+             erlang
+             erlang-asn1
+             erlang-common-test
+             erlang-crypto
+             erlang-dev
+             erlang-erl-interface
+             erlang-eunit
+             erlang-inets
+             erlang-jinterface
+             erlang-public-key
+             erlang-reltool
+             erlang-sasl
+             erlang-snmp
+             erlang-ssl
+             erlang-syntax-tools
+             erlang-tools
+             erlang-xmerl
+             gmp-dev
+             libexecinfo-dev
+             libtool
+             nodejs
+             openjdk8
+             perl
+             php7
+             python2
+             python2-dev
+             ruby
+             "
+install=""
+subpackages=""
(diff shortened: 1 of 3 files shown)


okeuday:
The Travis build happens to have mysql running and probably postgresql too, so the tests get angry due to wanting to use it for testing.  Will update the package to not run these tests.

okeuday:
The make usage (in the `APKBUILD` file) currently requires `-j1` but that should improve in the future.

testing/cloudi/APKBUILD
```diff
@@ -0,0 +1,84 @@
...
+subpackages=""
+source="http://downloads.sourceforge.net/$pkgname/$pkgname-$pkgver.tar.gz
+        0005-Disable-pgsql-mysql-tests-for-aports-travis-build.patch
+        $pkgname.initd"
+
+builddir="$srcdir/$pkgname-$pkgver/src"
+
+build() {
+	cd "$builddir"
+	./autogen.sh
```
rnalrd: Should go into prepare()

testing/cloudi/APKBUILD
```diff
@@ -0,0 +1,84 @@
+# Contributor: Michael Truog <mjtruog@gmail.com>
+# Maintainer: Michael Truog <mjtruog@gmail.com>
+
+pkgname=cloudi
+pkgver=1.7.1
+pkgrel=0
+pkgdesc="CloudI is an open-source private cloud computing framework for efficient, scalable, and stable soft-realtime event processing."
```
rnalrd: Line shortened

rnalrd:
Actually doesn't build on buildservers:

http://build.alpinelinux.org/buildlogs/build-edge-x86/testing/cloudi/cloudi-1.7.1-r0.log
http://build.alpinelinux.org/buildlogs/build-edge-x86_64/testing/cloudi/cloudi-1.7.1-r0.log

Would you please fix it?

Thanks!

jirutka:
@rnalrd Have you examined structure of files in the package? It’s quite clear from the files list in the [log](https://travis-ci.org/alpinelinux/aports/builds/245106203#L3853) that this abuild is not very good.

okeuday:
@jirutka The directory structure is very similar to `erlang`, so you would also be asserting the `erlang` abuild isn't very good.  Doing things this way keeps all versions referenced explicitly when they become dependencies elsewhere.

@rnalrd Will make a pull request for fixing the problem on the buildservers.  A test is having problems in that environment for some reason.

PR #1748, opened 2017-06-21
testing/perl-lockfile-simple: new aport

TBK0000:


testing/perl-lockfile-simple/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgname=perl-lockfile-simple
+_pkgreal=LockFile-Simple
+__pkgreal=lockfile-simple
+pkgver=0.208
+pkgrel=0
+pkgdesc="Simple file locking scheme"
+url="http://search.cpan.org/dist/LockFile-Simple/"
+arch="noarch"
+license="GPL PerlArtistic"
+options="!check"
```
ncopa: why don't you move the `make test` to a  `check` function?

TBK0000:
@ncopa Thanks for the advice. I have made the change.

PR #1754, opened 2017-06-21
community/asciinema: upgrade to 1.4.0 and enable on ppc64le

rgdoliveira:
The upgrade from version 1.2.0 to 1.3.0 brings back the original Python implementation
of asciinema[1] and it supports only Python 3.

This patch changes the build to reflect the changes and also enable build on ppc64le.

[1] http://blog.asciinema.org/post/and-now-for-something-completely-different/

### DIFF
--- a/community/asciinema/APKBUILD
+++ b/community/asciinema/APKBUILD
@@ -1,32 +1,25 @@
 # Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=asciinema
-pkgver=1.2.0
+pkgver=1.4.0
 pkgrel=0
 pkgdesc="A Command line recorder for asciinema.org service"
 url="https://github.com/asciinema/asciinema/"
-arch="all !aarch64 !ppc64le !s390x"
+arch="all !aarch64 !s390x"
 license="GPLv3"
-depends="curl"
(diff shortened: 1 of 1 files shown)


jirutka:
I really like this blog post about disillusionment from crappy Go. 😸

However that’s not a reason for adding `py-` prefix to the package. It’s a CLI program that happens to be written in Python, not a Python library.

rgdoliveira:
@jirutka yeah, make senses! I applied the changes you suggested and if I missed something, please tell me

jirutka:
Please remove py3 subpackage, i.e. make it straightforward like any non-python pkg.

rgdoliveira:
I removed the py3 subpackage and sent a new version of the patch

community/asciinema/APKBUILD
```diff
@@ -1,32 +1,25 @@
...
-arch="all !aarch64 !ppc64le !s390x"
+arch="all !aarch64 !s390x"
 license="GPLv3"
-depends="curl"
-depends_dev=""
-makedepends="go python2"
-subpackages="$pkgname-doc"
-options="!strip"
+depends="ncurses"
+makedepends="python3-dev py-setuptools"
```
jirutka: py-setuptools is only for python2. There's no py3 variant, because setuptools is included in stdlib of Python 3. So you should omit it.

rgdoliveira:
done!

PR #1756, opened 2017-06-22
community/arpack: disable check() in ppc64le for now

rgdoliveira:
Tests from arpack are hanging on ppc64le with abuild but tests work when
running manually (make check) and all tests pass. It can be an issue in
fakeroot. Disabling ppc64le tests for now.


jirutka:
I’ve tried to disable fakeroot for check on ppc64le, let’s wait what will happen.

rgdoliveira:
@jirutka seems that  it did not work as we expect, can we disable tests for now on ppc64le with this patch?

PR #1763, opened 2017-06-23
testing/cloudi: new aport

okeuday:
http://cloudi.org/
CloudI is an open-source private cloud computing framework for efficient, scalable, and stable soft-realtime event processing.

This pull-request is a continuation of https://github.com/alpinelinux/aports/pull/1741 .  The tests that use postgresql and mysql are disabled to avoid them failing when they attempt to test these databases (Travis has the databases running, but they require a database be created for the tests to pass).  The cloudi_service_filesystem test is disabled due to the x86_64 and x86 buildservers not wanting /tmp to be used for temporary files during testing.  These changes are applied within the `0005-Disable-tests-for-aports-buildservers.patch` patch.

### DIFF
--- a/testing/cloudi/APKBUILD
+++ b/testing/cloudi/APKBUILD
@@ -1,0 +1,118 @@
+# Contributor: Michael Truog <mjtruog@gmail.com>
+# Maintainer: Michael Truog <mjtruog@gmail.com>
+#
+# INFO
+#
+# Why is the cloudi.initd file not using the openrc default start-stop-daemon?
+#
+#   With the Erlang VM it is possible to use the -heart erl command line
+#   argument (which is possible with cloudi configuration of the
+#   /etc/cloudi/vm.args file).  To allow the use of the heartbeat monitoring
+#   of the Erlang runtime system, it is necessary to rely upon the cloudi
+#   script for handling the cloudi pid and not utilize openrc for pidfile
+#   creation or signal-based termination.
+#
+#   With a typical Erlang VM release, it is normally expected to provide
+#   the ability to attach to the Erlang VM shell of the running node using
+#   pipes.  The setup of the Erlang VM shell pipes needs to be handled with
+#   Erlang VM execution that also handles detaching the daemon process.
+#   If the openrc default start-stop-daemon execution was used instead,
+#   it would not be possible to easily setup the Erlang VM shell pipes.
+#
+#   The termination of cloudi utilizes a 65000 millisecond timeout, allowing
+#   all services a maximum of 60000 milliseconds for termination with
+#   5000 millisecond for internal Erlang VM delays.  The cloudi script does
+#   block for this termination to complete.
+#
+#   The pid file path is set during the configuration step
+#   (shown below with the assignment of CLOUDI_PID_FILE).
+
+pkgname=cloudi
+pkgver=1.7.2_rc1
+_srcver=70addfda7133ac209157d48e38be70be7597ce8a
+pkgrel=0
+pkgdesc="Cloud computing framework for efficient, scalable, and stable soft-realtime event processing."
+url="http://cloudi.org/"
+license="MIT"
+arch="all"
+depends="g++"
+makedepends="autoconf
+             automake
+             binutils-dev
+             boost-dev
+             boost-system
+             boost-thread
+             erlang
+             erlang-asn1
+             erlang-common-test
+             erlang-crypto
+             erlang-dev
+             erlang-erl-interface
+             erlang-eunit
+             erlang-inets
+             erlang-jinterface
+             erlang-public-key
+             erlang-reltool
+             erlang-sasl
+             erlang-snmp
+             erlang-ssl
+             erlang-syntax-tools
+             erlang-tools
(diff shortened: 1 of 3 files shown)


tmpfile:
>  The cloudi_service_filesystem test is disabled due to the x86_64 and x86 buildservers not wanting /tmp to be used for temporary files during testing.

Instead of disabling the tests you could try changing TMPDIR.
Example: https://github.com/alpinelinux/aports/blob/master/community/julia/APKBUILD#L138-L141

okeuday:
@tmpfile I disabled `cloudi_service_filesystem` because I know the test has `/tmp/` hardcoded, though the next version will utilize `TMPDIR` to be able to use configuration like this (due to the change at https://github.com/CloudI/CloudI/commit/0124de9c90f886d86d2dbe851501ea004828036c)

okeuday:
@tmpfile I would like to iterate (gradually improve) this package, but it is unclear how to do so, if there are no further comments and it doesn't get merged.  Should I change this in some way now, or is it possible to test it out after the merge?

tmpfile:
LGTM, but we have to wait for someone with commit access to approve and merge.
I notice Alpine is more active during weekdays. Try to catch someone online and link to this issue.
Sometimes I'm also disappointed like you, but we have to understand it's a lot of work for just a few people.
In some periods everything gets reviewed really soon in a matter of hours, but in other periods it takes a few weeks...
But if you talk to someone and link your PR, you can be sure they will try their best to address it.
HTH

okeuday:
@tmpfile Ok, will follow-up an a weekday.  Thanks!

testing/cloudi/APKBUILD
```diff
@@ -0,0 +1,91 @@
...
+             openjdk8
+             perl
+             php7
+             python2
+             python2-dev
+             ruby
+             "
+install=""
+subpackages=""
+source="https://github.com/$pkgname/$pkgname/archive/$_srcver.tar.gz
```
jirutka: Why do you use git rev instead of pkgver?

testing/cloudi/cloudi.initd
```diff
@@ -0,0 +1,51 @@
...
+# not necessary to utilize a user other than root for executing the cloudi
+# script.
+
+# pidfile path is set during configuration step before installation
+#pidfile=/run/cloudi.pid
+
+# These comments are provided here to encourage comment from the alpine
+# developers that are currently blocking the merge of this port, since there
+# has been no feedback or communication about necessary changes or
+# potential problems.
```
jirutka: Uh, init file is not a diary… Please put this useful yet verbose description to the APKBUILD and leave just brief note here referencing detail description in APKBUILD.

testing/cloudi/cloudi.initd
```diff
@@ -0,0 +1,51 @@
...
+}
+start() {
+	$command start
+}
+stop() {
+	$command stop
+}
+restart() {
+	$command restart
+}
```
jirutka: Does `/etc/init.d/cloudi status` work?

testing/cloudi/APKBUILD
okeuday: I switched this from the pkgver to a git rev for testing purposes only.  That allows the use of recent changes for the pidfile path provided to configure and newer cloudi script use.  I went over the older (`1.7.1` release and previous) `cloudi` script and the `nodetool` script to simplify and improve the tasks they do.  The one thing that remains is that distributed Erlang is used by CloudI and must be started.  CloudI starts epmd if it hasn't been started already, but that isn't ideal as a long-term solution, that is only a way of making sure stuff always works in any deployment scenario.  The `community/erlang` package really needs a script to handle epmd startup, so it is given more visibility as a dependency, though that is outside the scope of this pull request (this package works either way).

testing/cloudi/cloudi.initd
okeuday: Sure, can move this information to the APKBUILD file.

testing/cloudi/cloudi.initd
okeuday: I will add a status function to make sure that works properly.

okeuday:
@jirutka I have updated the commit based on your feedback.  Please tell me if other changes are necessary.

okeuday:
@jirutka I have updated the package version to use the proper release candidate format for Alpine/Gentoo which can help make it clearer why a repo commit is used.  The update also improves the file permissions used in the installation so the root runtime can not be easily abused.

clandmeter:
I just disabled this build on infra, it was failing and after a tried to restart the build it took very long and was blocking the builders (CT was processing for 4180 seconds...).  Can you check local what is going on? are you using an alpine edge dev environment to test your builds?

okeuday:
@clandmeter Yes, I have been testing the build of cloudi with x86 and haven't seen this problem there (or in Travis builds, or with local builds).  However, this error relates to the cloudi test of cloudi_service_quorum, due to a random number that likely has a rounding problem.  Discussed with @kaniini and will disable the cloudi_service_quorum for now in a separate pull request.

PR #1765, opened 2017-06-24
main/cjdns: update to 20

scadu:
Also get rid of bash in makedepends


scadu:
Can we marge this one? @Shizmob @jirutka

PR #1775, opened 2017-06-25
main/python3: upgrade to 3.6.2, optimize build, cleanup

scadu:
Spitted into two commits since https://github.com/scadu/aports/commit/dcf6247dc0426b59bcbce8deefcd852dc7e02820 which fixes issue reported by @awilfox on IRC:

> I am building packages in a chroot.  the host system is 64-bit Debian, the chroot is 32-bit Alpine, but it says: checking host system type... powerpc64-unknown-linux-gnu
this is definitely not desireable.  it should be using --build --host like any other

The second commit https://github.com/scadu/aports/commit/6c482bda93672107fa892d2a1ed213da31ce2b39 actually enables optimization options somehow missed during previous builds. 

Travis test will fail here. Unfortunately it requires more time, however the boost is noticeably: 30% speed boost in some parts of the stdlib -- I think it's worth it.

### DIFF
--- a/main/python3/APKBUILD
+++ b/main/python3/APKBUILD
@@ -12,33 +12,30 @@ license="custom"
 provides="py3-pip"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-tests::noarch
 	$pkgname-tkinter:tkinter"
-depends=""
 makedepends="expat-dev libressl-dev zlib-dev ncurses-dev bzip2-dev xz-dev
 	sqlite-dev libffi-dev tcl-dev linux-headers gdbm-dev readline-dev
 	tk tk-dev"
 source="http://www.python.org/ftp/python/$pkgver/Python-$pkgver.tar.xz
 	musl-find_library.patch
-	fix-xattrs-glibc.patch
-	"
+	fix-xattrs-glibc.patch"
 builddir="$srcdir/Python-$pkgver"
 
 prepare() {
-	default_prepare || return 1
+	default_prepare
 
 	cd "$builddir"
 	# force system libs
 	rm -r Modules/expat \
 		Modules/zlib \
 		Modules/_ctypes/darwin* \
-		Modules/_ctypes/libffi* \
-		|| return 1
+		Modules/_ctypes/libffi*
 }
 
 build() {
 	cd "$builddir"
 
 	./configure \
-		--prefix=/usr
+		--prefix=/usr \
 		--disable-rpath \
 		--enable-ipv6 \
 		--enable-loadable-sqlite-extensions \
@@ -50,26 +47,23 @@ build() {
 		--with-system-expat \
 		--with-system-ffi \
 		--with-threads \
-		|| return 1
 
-	make EXTRA_CFLAGS="$CFLAGS" || return 1
+	make EXTRA_CFLAGS="$CFLAGS"
 }
 
 package() {
 	cd "$builddir"
-	make -j1 DESTDIR="$pkgdir" EXTRA_CFLAGS="$CFLAGS" install maninstall \
-		|| return 1
+	make -j1 DESTDIR="$pkgdir" EXTRA_CFLAGS="$CFLAGS" install maninstall
 	install -Dm644 LICENSE "$pkgdir"/usr/share/licenses/$pkgname/LICENSE
-
 }
 
 dev() {
-	default_dev || return 1
+	default_dev
 
 	# pyconfig.h is needed runtime so we move it back
 	mkdir -p "$pkgdir"/usr/include/python${_basever}m
 	mv "$subpkgdir"/usr/include/python${_basever}m/pyconfig.h \
-		"$pkgdir"/usr/include/python${_basever}m/
+	    "$pkgdir"/usr/include/python${_basever}m/
 }
 
 tests() {
@@ -93,6 +87,6 @@ tkinter() {
 	mv "$pkgdir"/$libdir/_tkinter.*.so "$subpkgdir"/$libdir/
 }
 
-sha512sums="8605fb7019386fec227d4b06d06f00ae500a8a89df289bfe6141bb56196c75483a60cc0ee553930742b31cefce68add5ccf226e0f27b7b915f5026d597e1ac29  Python-3.6.1.tar.xz
+sha512sums="a8270a09a9e9b39f69ece6cdade2fa964665d2107b5acbad4453f1b921107b329c697c137185928fb4a576fc0f2ae2a98dbf26a8b7ea17219e990ddbc216db8b  Python-3.6.2.tar.xz
 ab8eaa2858d5109049b1f9f553198d40e0ef8d78211ad6455f7b491af525bffb16738fed60fc84e960c4889568d25753b9e4a1494834fea48291b33f07000ec2  musl-find_library.patch
 37b6ee5d0d5de43799316aa111423ba5a666c17dc7f81b04c330f59c1d1565540eac4c585abe2199bbed52ebe7426001edb1c53bd0a17486a2a8e052d0f494ad  fix-xattrs-glibc.patch"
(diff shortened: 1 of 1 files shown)


awilfox:
I can confirm this solves our chroot build/host type issue.

The optimisations are indeed a very good idea, but see docker-library/python#160.  When I applied this to our aforementioned PowerPC builder, I had some very noteworthy results:

1. The build process took over two hours.  Of course, it is a dual-processor 1.8 GHz Power4, so it is not exactly a shiny new build cluster.
2. When `abuild` hit the package phase, `make clean` was run by Python's build system, and the entire thing started over again.  I started the build at 20:58 and by 01:15, it was just wrapping up.
3. Unfortunately while making the apk files, the chroot ran out of disk space.  No worries, I thought; just increase the size of the volume and run `abuild rootpkg` again.  Of course, that started the build over again, since Python did `make clean` again...
4. And then this happened:

```
/tmp/ccObloFK.ltrans4.ltrans.o: In function `_GLOBAL__sub_I_65535_0_pgen.c':
<artificial>:(.text.startup+0x24): undefined reference to `__gcov_init'
/usr/lib/gcc/powerpc-foxkit-linux-musl/6.3.0/../../../../powerpc-foxkit-linux-musl/bin/ld: BFD (GNU Binutils) 2.28 assertion fail elf32-ppc.c:8923
/usr/lib/gcc/powerpc-foxkit-linux-musl/6.3.0/../../../../powerpc-foxkit-linux-musl/bin/ld: BFD (GNU Binutils) 2.28 assertion fail elf32-ppc.c:8923
/tmp/ccObloFK.ltrans5.ltrans.o:/home/awilcox/aports/main/python3/src/Python-3.6.1/Parser/pgen.c:254: more undefined references follow
make: *** [Makefile:813: Parser/pgen] Error 1
>>> ERROR: python3*: package failed
>>> ERROR: python3: rootpkg failed
```

At this point I was having major Gentoo flashbacks and gave up.  :wink:  Maybe --enable-optimizations is not stable on PowerPC or maybe it is incompatible with GCC 6.3.0.  Not sure how to tell, and not sure I want to spend the machine cycles and wall time on debugging it at this time.  But do feel free to try and figure out, if you want, because it is a good thing to do.

scadu:
The first three "problems" are the result of enabled flags. That's how it works and it's explained in linked docker-library issue. 

If it comes to the last one: I'm afraid I have no access to any ppc machine, and even so debugging this would be a magic. We could omit those flags on ppc till we figure it out. Thanks.

scadu:
Maybe we should apply ` --with-lto` and `--enable-optimizations` only for x86` and x86_64.
@jirutka, what do you think?

awilfox:
LTO works fine on PowerPC.  Can't speak for any other architecture, bootstrapping takes a while.

main/python3/APKBUILD
```diff
@@ -40,17 +40,17 @@ build() {
...
-		--enable-optimize \
-		--enable-lto \
+		--enable-optimizations \
+		--with-lto \
 		--with-computed-gotos \
 		--with-dbmliborder=gdbm:ndbm \
 		--with-system-expat \
 		--with-system-ffi \
 		--with-threads \
+		--with-ensurepip=install \
```
ncopa: Why so you enable ensurepip? see http://bugs.alpinelinux.org/issues/5381

scadu: Looks like I forgot to remove this line once I finished playing with APKBUILD.

scadu:
3.6.2 build with `--with-lto` and `--enable-optimizations` took 26 minutes on Xeon E31220 with 16G RAM and HDD. @ncopa would it be possible to merge all of this?

main/python3/APKBUILD
```diff
@@ -40,17 +40,17 @@ build() {
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
-		--disable-rpath \
```
ncopa: why was `--disable-rpath` removed?

main/python3/APKBUILD
ncopa: i think the entire commit can be squashed with the previous

main/python3/APKBUILD
```diff
@@ -50,7 +50,6 @@ build() {
 		--with-system-expat \
 		--with-system-ffi \
 		--with-threads \
-		--with-ensurepip=install \
 		
 
 	make EXTRA_CFLAGS="$CFLAGS"
```
ncopa: I think the commit should be squashed with the previous (`git rebase -i` and set this to `fixup`)

main/python3/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Contributor: Kiyoshi Aman <kiyoshi.aman@gmail.com>
 
 pkgname=python3
-pkgver=3.6.1
+pkgver=3.6.2
 _basever="${pkgver%.*}"
 pkgrel=5
```
ncopa: we should set `pkgrel=0` when we bump `pkgver`

scadu:
Following tests failed for me and @awilfox 

```
 test__locale failed
 test_asyncore skipped
 test_capi failed
 test_curses skipped (resource denied)
 test_devpoll skipped
 test_gdb skipped
 test_kqueue skipped
 test_locale failed
 test_msilib skipped
 test_multiprocessing_fork skipped
 test_multiprocessing_forkserver skipped
 test_multiprocessing_main_handling skipped
 test_multiprocessing_spawn skipped
 test_nis skipped
 test_os failed
 test_ossaudiodev skipped (resource denied)
 test_posix failed
 test_re failed
 test_resource failed
 test_shutil failed
 test_smtpnet skipped (resource denied)
 test_socketserver skipped (resource denied)
 test_startfile skipped
 test_strptime failed
 test_subprocess skipped
 test_threadsignals failed
 test_time failed
 test_timeout skipped (resource denied)
 test_tix skipped (resource denied)
 test_tk skipped (resource denied)
 test_ttk_guionly skipped (resource denied)
 test_urllib2net skipped (resource denied)
 test_urllibnet skipped (resource denied)
 test_winconsoleio skipped
 test_winreg skipped
 test_winsound skipped (resource denied)
 test_xmlrpc_net skipped (resource denied)
 test_zipfile64 skipped (resource denied)
```

Quoting @awilfox 

> test_locale is meant to fail; musl doesn't support strcoll (it is stub) nor strxfrm (which uses strcoll) nor nl_langinfo without you writing your own (which is why '.' != ',')

Probably the rest of tests failed due to musl too since it passes on glibc-based distro (e.g. Debian Jessie). There is no distro to compare since Void do not use optimization and they are still on 3.5.

main/python3/APKBUILD
```diff
@@ -38,13 +38,13 @@ build() {
 	cd "$builddir"
 
 	./configure \
-		--prefix=/usr \
+		--prefix=/usr
```
mingwandroid: You probably don't want to remove the line continuation here.

scadu: Already fixed. Please see [204198c](https://github.com/alpinelinux/aports/pull/1775/commits/204198c974ccdc09154075cfe7d8eb617e73f85d#diff-6e86de19c7a7623f0f4c6030b8eb5940R38)

main/python3/APKBUILD
scadu: Done.

main/python3/APKBUILD
scadu: Done.

main/python3/APKBUILD
scadu: Shouldn't be included in the commit. Restored.

fabled:
I had to remove optimizations due to failure on x86 (probably also aarch64). See commit 3778f5bafc528d05f991945fb6dc5fa9781d1873.

PR #1776, opened 2017-06-26
testing/audit: promote to main

tych0:
Hi all,

I'd like to promote audit to main. I'm happy to keep maintaining it. I've tested it (and we're going to include it in LinuxKit). Let me know what needs to be done!


kaniini:
works for me if you want to keep maintainership.  merged.

PR #1797, opened 2017-06-30
main/xen: fix tmux option

radhus:
Tmux has renamed the set-remain-on-exit option to simply remain-on-exit
since tmux 2.4. The functionality is needed to make sure console logs
are kept if a domain dies.


radhus:
@kaniini Note that this is currently broken in v3.6 stable. Do you want me to cherry-pick a fix to 3.6-stable after this is merged and send a new PR?

kaniini:
@radhus works for me.

PR #1812, opened 2017-06-30
main/xen: fix tmux option

radhus:
Tmux has renamed the set-remain-on-exit option to simply remain-on-exit
since tmux 2.4. The functionality is needed to make sure console logs
are kept if a domain dies.

(cherry picked from commit bcd5c26f74d34fa70d0c069be434a465efa7ab53)


radhus:
@kaniini this follows #1797 :)

PR #1817, opened 2017-07-02
new aport: homebank

7heo:
Free personal finance software, money management

### DIFF
--- a/testing/homebank/APKBUILD
+++ b/testing/homebank/APKBUILD
@@ -1,0 +1,32 @@
+# Contributor: 7heo <7heo@mail.com>
+# Maintainer: 7heo <7heo@mail.com>
+pkgname=homebank
+pkgver=5.1.5
+pkgrel=0
+pkgdesc="Free personal finance software, money management"
+url="http://homebank.free.fr/"
+arch="all"
+license="GPL"
+makedepends="bash gtk+3.0-dev libsoup-dev libofx-dev"
+install=""
+subpackages="$pkgname-lang"
+source="http://homebank.free.fr/public/$pkgname-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure --prefix=/usr
+	make
+}
+
+check() {
+	"$builddir"/src/homebank --version | grep -q "^$pkgname $pkgver$"
+}
+
+package() {
+	cd "$builddir"
+	chmod a+x install-sh
+	make DESTDIR="$pkgdir" install
+}
+
+sha512sums="674a6385f427ba23039745b476d7474c9f41f0183d5a41d6fa340fe36cc9b0b3a4a3da30dfecc4c4c3b8bfc72aba751799a5750b71eaf28b13333475d31e92e0  homebank-5.1.5.tar.gz"


tmpfile:
> new aport: homebank

`testing/homebank: new aport` is more appropriate


> _pkgver=$(echo $pkgver | sed 's/\.[0-9]\+$/.x/')

echo, pipe, sed... why not keep this simple?
`_pkgver=${pkgver%.*}.x`


> md5sums="..........."
> sha256sums="..........."

That's cruft from old times, you should remove it.
Abuild only checks sha512sums now.

7heo:
Thanks for the review @tmpfile.

testing/homebank/APKBUILD
```diff
@@ -0,0 +1,37 @@
+# Contributor:
+# Maintainer: 7heo <7heo@mail.com>
+pkgname=homebank
+pkgver=5.1.5
+pkgrel=0
+sourcerev=56 # bump pkgrel when you change this. Get it form http://bazaar.launchpad.net/~mdoyen/homebank/5.1.x/changes
```
jirutka: Non-standard global variables (i.e. not defined by abuild) should be prefixed with underscore (i.e. `_sourcerev`).

testing/homebank/APKBUILD
```diff
@@ -0,0 +1,37 @@
...
+makedepends="bash gtk+3.0-dev libsoup-dev libofx-dev"
+install=""
+subpackages="$pkgname-lang"
+options="!check"
+source="$pkgname-$pkgver-$sourcerev.tar.gz::http://bazaar.launchpad.net/~mdoyen/$pkgname/${pkgver%.*}.x/tarball/$sourcerev"
+builddir="$srcdir/$pkgname-$pkgver"
+
+unpack() {
+	default_unpack
+	mv "$srcdir/~mdoyen/$pkgname/${pkgver%.*}.x" "$builddir"
```
jirutka: We usually do this in `prepare` function, not `unpack`.

testing/homebank/APKBUILD
```diff
@@ -0,0 +1,37 @@
...
+sourcerev=56 # bump pkgrel when you change this. Get it form http://bazaar.launchpad.net/~mdoyen/homebank/5.1.x/changes
+pkgdesc="Free personal finance software, money management"
+url="http://homebank.free.fr/"
+arch="all"
+license="GPL"
+depends="libsoup"
+makedepends="bash gtk+3.0-dev libsoup-dev libofx-dev"
+install=""
+subpackages="$pkgname-lang"
+options="!check"
```
jirutka: This is an application, right? If upstream does not provide any tests, then you should add at least trivial check that just runs the executable to ensure that it’s not totally broken (e.g. `homebank --version`).

testing/homebank/APKBUILD
```diff
@@ -0,0 +1,37 @@
+# Contributor:
```
jirutka: Since you’ve created this abuild, you’re the first contributor. ;)

7heo:
@jirutka Thanks for the review; however I have to solve the problem of the elusive checksum in the tarball first. So I'll do the modifications you requested; but it will still fail the tests...

7heo:
Changes requested have been performed. For the record, the current fingerprint is `bb905fc8c022de6dee5d68566ac4f67b5411728ced19075c169eb7905cb66a68f538ea496a59d435422cfaa65449cffc1d4acf039ffd2c3d87df4683b358d5c5`.

jirutka:
Why do you download it from this obscure bazaar.launchpad.net thing? There’s a link to release tarball on their website http://homebank.free.fr/, for 5.1.5 it’s http://homebank.free.fr/public/homebank-5.1.5.tar.gz.

7heo:
@jirutka Very good point. The habit of using github as a source I suppose.

jirutka:
Well, that’s a good habit in case of GitHub, but definitely not with launchpad.net. :P

7heo:
I realized that now. I'm taking care of my server and then I will push a new version of homebank with your suggestions factored in ;) Thanks a lot for the hint!

jirutka:
Shouldn’t be /usr/share/homebank/help moved into -doc subpkg? I have no experience with desktop packages, so not sure.

PR #1838, opened 2017-07-04
Removed Down mirror from mirrors.yaml

inpothet:
Removed LASCA - Institute of Computing - UNICAMP as the mirror has been down for some time


tmpfile:
Why didn't check() catch this?
https://github.com/alpinelinux/aports/blob/master/main/alpine-mirrors/APKBUILD#L21-L36

inpothet:
Mabye because the webserver returns an 404 ?

ncopa:
fixed with commit eddf9ac1b4deaaca926fe8bd1d7f658b0929f345

PR #1840, opened 2017-07-04
alpine-mirrors: remove broken unicamp mirror

Ikke:
The mirror is down, remove it from the list of mirrors. 

A user reported on IRC this mirror has been down for a while.


ncopa:
duplicates #1838

ncopa:
fixed with commit eddf9ac1b4deaaca926fe8bd1d7f658b0929f345

PR #1843, opened 2017-07-05
testing/py3-pygpgme: new aport

TBK0000:


fcolista:
Hi, thanks for your contribution!
We have a good reference for python APKBUILD that you can find here:
https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python
Few things:

- py-setuptools  is not needed since is for python2 while you are using python3
- gcc can be removed also from makedepends, since it's used by default
- There is a reason why you skip the check by putting "options="!check" ?
- Please adjust pkgname and _pkgname according with the APKBUILD's example beforementioned.
- there's no maintainer set
Thanks.

.: Francesco

TBK0000:
@fcolista Thanks for the advice and the link, I have been looking for exampes.

I have pushed the correction.

PR #1845, opened 2017-07-05
testing/gcompat: Create symlink for lib64

Sodki:
Some precompiled binaries use /lib64 instead of /lib, so this change is needed to make them work. This workaround is also used in musl/libc6-compat. This is also a recommended behaviour in gcompat's documentation.

### DIFF
--- a/testing/gcompat/APKBUILD
+++ b/testing/gcompat/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=gcompat
 pkgver=0.0.1
-pkgrel=0
+pkgrel=1
 pkgdesc="enhanced glibc compatibility layer"
 url="http://github.com/kaniini/gcompat"
 arch="all"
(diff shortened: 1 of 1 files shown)


errm:
I have tested building package on my system and the changes seem to be working as intended...

fabled:
How about then just depending on `libc6-compat` because now you can't install them both at the same time due to both trying to install the same file.

fabled:
Alternative is to ship that symlink in `alpine-baselayout`. @ncopa Any comments?

errm:
@fabled am I misunderstanding the meaning of `replaces` ... I thought that meant that this package is meant as a replacement for `libc6-compat` rather than to be installed at the same time?

fabled:
Ah, missed that. I would still rather see this as `provides="libc6-compat=$pkgver"` or `depends="!libc6-compat". The are subtle reasons for that. Does this package provide all other symlinks as libc6-compat?

errm:
So there are a bunch of links in `libc6-compat` that are not in the `gcompat` package

```
/lib/libpthread.so.0 -> /lib/libc.musl-x86_64.so.1
/lib/libutil.so.1 -> /lib/libc.musl-x86_64.so.1
/lib/librt.so.1 -> /lib/libc.musl-x86_64.so.1
/lib/libcrypt.so.1 -> /lib/libc.musl-x86_64.so.1
/lib/libc.so.6 -> /lib/libc.musl-x86_64.so.1
/lib/libm.so.6 -> /lib/libc.musl-x86_64.so.1
```
I think they might not be needed due to the way the gcompat loader works...

kaniini:
At start time, the loader takes care of those details, thanks to musl suppressing load_one_library() calls for those library names internally.

I believe the symlinks are needed for dlopen(), but not certain where they should point yet.  It may make sense to generate stubs for those specific binaries.

kaniini:
I took a slightly different approach, but should fix this problem.

PR #1846, opened 2017-07-06
community/hugo: update to 0.24.1

lexsca:
This is to update Hugo to the current version and official repo.

It's also a first pass at adding tests. Alas, the Hugo tests aren't
designed to be run outside of a git repo. Because of this, some of
the tests are removed from the build directory, and the hard coded
version of a failing test is set to the correct version.

Ideally, the tests should run wihtout being modified. However, this
is adding some testing where none existed, so it should help a bit.


lexsca:
Please let me know if any changes are needed to get this merged.

PR #1848, opened 2017-07-06
testing/postgresql-pglogical: upgrade to 2.0.1

gmile:


gmile:
Also, attaching a patch allows building pglogical on alpine. For more info on the patch see https://github.com/2ndQuadrant/pglogical/issues/99#issuecomment-312780669

testing/postgresql-pglogical/APKBUILD
```diff
@@ -11,9 +11,14 @@ arch="all"
 license="custom"
 depends="postgresql>=$_pgver.0"
 makedepends="libedit-dev libxml2-dev postgresql-dev>=$_pgver.0 zlib-dev"
-source="http://packages.2ndquadrant.com/pglogical/tarballs/$_pkgname-$pkgver.tar.bz2"
+source="http://packages.2ndquadrant.com/pglogical/tarballs/$_pkgname-$pkgver.tar.bz2
+        001-fix-stdin-handling.patch"
 builddir="$srcdir/$_pkgname-$pkgver"
 
+prepare() {
+  default_prepare || return 1
```
jirutka: This is redundant, `default_prepare` is implicitly called when you don't define custom `prepare` function.

gmile: @jirutka thanks, I didn't know that. I've force-pushed the change with `prepare()` removal

PR #1849, opened 2017-07-06
testing/spotifyd: new aport

dawidd6:
An open source Spotify client running as a UNIX daemon.


ncopa:
We need to figure out how to deal with cargo properly.
https://github.com/alpinelinux/aports/blob/master/community/cargo/APKBUILD#L19-L28

PR #1856, opened 2017-07-07
community/php5: upgrade 5.6.31

andypost:
Security release http://php.net/archive/2017.php#id2017-07-06-4


andypost:
Next security release https://github.com/alpinelinux/aports/pull/2634

PR #1874, opened 2017-07-07
main/libnfs: upgrade to 2.0.0

tmpfile:


fcolista:
This upgrade breaks ABI, other packages must be rebuilt against this new version.
The old version (1.11.0) has libnfs.so.8, the new one has libnfs.so.11.
Packages: kodi, mpd, qemu.

ncopa:
FYI. this upgrade breaks ABI. we need to rebuild all applications linked to libnfs. (AFAICS qemu, mpd and kodi)

PR #1882, opened 2017-07-08
community/cfengine: fix project url and modernize APKBUILD

rgdoliveira:


ncopa:
pkgrel needs to be bumped when project url changes. It is displayed with `apk info`.

PR #1888, opened 2017-07-09
main/freetype: upgrade url and modernize APKBUILD

tmpfile:


main/freetype/APKBUILD
```diff
@@ -22,11 +22,6 @@ source="http://download.savannah.gnu.org/releases/freetype/freetype-$pkgver.tar.
 	freetype-profile.sh
 	"
 
-# secfixes:
-#   2.7.1-r1:
-#     - CVE-2017-8105
-#     - CVE-2017-8287
-
```
ncopa: why do you remove those?

main/freetype/APKBUILD
```diff
@@ -2,15 +2,15 @@
...
 pkgdesc="TrueType font rendering library"
-url="http://freetype.sourceforge.net"
+url="https://www.freetype.org/"
 arch="all"
 license="GPL"
 options="!check"
 depends=""
 depends_dev=""
-makedepends="$depends_dev zlib-dev libpng-dev bzip2-dev"
+makedepends="$depends_dev zlib-dev libpng-dev bzip2-dev harfbuzz-dev"
```
ncopa: harfbuzz depends on freetype so this is not acceptable.

Possible solution if we really *need* harfbuzz:
https://github.com/Homebrew/legacy-homebrew/issues/27847#issuecomment-38633313

main/freetype/APKBUILD
```diff
@@ -38,21 +33,16 @@ build() {
...
 	cd "$builddir"
-	make DESTDIR="$pkgdir" install || return 1
+	make DESTDIR="$pkgdir" install
 
 	install -Dm644 "$srcdir"/freetype-profile.sh \
 		"$pkgdir"/etc/profile.d/freetype.sh
-
-	# for compat. This should be removed once all apps are properly using
-	# pkg-config
-	ln -s freetype2 "$pkgdir"/usr/include/freetype
```
ncopa: Have you verified that all apps are properly using pkg-config?

main/freetype/APKBUILD
tmpfile: I think they only make the APKBUILD messier.
Isn't the place of that comments on alpine-secdb?
I think it fits better there.
But if you want I can re-add those comments here, I just don't like it personally.

main/freetype/APKBUILD
tmpfile: I don't think we really need harfbuzz.
Because of the circular dep I will just remove it.

main/freetype/APKBUILD
tmpfile: I didn't verify.
I just assumed that was old cruft because it was from 2013.

How can we check to be sure?

main/freetype/APKBUILD
ncopa: alpine-secdb is generated from those comments, so if you remove it from APKBUILD you remove it from alpine-secdb.

ncopa: we should keep the secfix comment.

tmpfile: I don't think it's nice to have that information there because it makes the APKBUILD a bit dirtier.
But okay... if that's how alpine-secdb is generated I have no option.
I'm readding the comments.

tmpfile:
All issues addressed except for the symlink.

If all apps are properly using pkg-config then this PR can be merged.
If not all apps are properly using pkg-config then I'll readd the symlink.

How can we verify if all apps are properly using pkg-config?

ncopa:
> @tmpfile  How can we verify if all apps are properly using pkg-config?

I suppose the only way is to test build all the packages that has freetype-dev in its makedepends (or depends on `so:libfreetype.so.6`)

The symlink was introduced in commit a84e5824cdb from 2013. I suppose we could testbuild a few packages so make sure that not everything is broken, then we remove the symlink and fix breakages when we rebuild world for the 3.7 release.

PR #1891, opened 2017-07-10
testing/eigen: new aport

chambbj:
http://eigen.tuxfamily.org/
C++ template library for linear algebra

### DIFF
--- a/testing/eigen/APKBUILD
+++ b/testing/eigen/APKBUILD
@@ -1,0 +1,32 @@
+# Contributor: Bradley J Chambers <brad.chambers@gmail.com>
+# Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+pkgname=eigen
+pkgver=3.3.4
+pkgrel=0
+pkgdesc="Eigen is a C++ template library for linear algebra"
+url="http://eigen.tuxfamily.org/index.php?title=Main_Page"
+arch="noarch"
+license="MPL2"
+depends=""
+makedepends=""
+install=""
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgname"
+
+build() {
+        return 0
+}
+
+prepare() {
+        mv "$builddir"-* "$builddir"  # directory name contains hash
+        default_prepare
+}
+
+package() {
+        mkdir -p "$pkgdir"/usr/include/eigen3
+        cp -r "$builddir"/Eigen "$pkgdir"/usr/include/eigen3
+        cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
+}
+
+sha512sums="4077a5c3b95e3573774ccd3fe6c7233cb4b83db2358c19b43ea796925bd0201451d8632bddc5d68b1b57bbf67c5473a8908926eed065a745689a2acec9711d5c  eigen-3.3.4.tar.gz"


ncopa:
````diff
diff --git a/testing/eigen/APKBUILD b/testing/eigen/APKBUILD
index c8a9075ed5..d0ee6c172b 100644
--- a/testing/eigen/APKBUILD
+++ b/testing/eigen/APKBUILD
@@ -10,7 +10,7 @@ license="MPL2"
 depends=""
 makedepends=""
 install=""
-subpackages="$pkgname-dev"
+subpackages=""
 source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-eigen-5a0156e40feb"
 
````

File: testing/eigen/APKBUILD [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc="Eigen is a C++ template library for linear algebra"
+url="http://eigen.tuxfamily.org/index.php?title=Main_Page"
+arch="all"
+license="MPL2"
+depends=""
+makedepends=""
+install=""
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-eigen-5a0156e40feb"
```
jirutka: This hash in directory name is not good, it’d complicate bumping. You can workaround it using e.g.:

```sh
builddir="$srcdir/$pkgname-$pkgname"

prepare() {
	mv "$builddir"-* "$builddir"  # directory name contains hash
	default_prepare
}
```

testing/eigen/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+
+build() {
+        return 0
+}
+
+package() {
+        mkdir -p "$pkgdir"/usr/include
+        cp -r "$builddir"/Eigen "$pkgdir"/usr/include
+        cp -r "$builddir"/unsupported "$pkgdir"/usr/include
```
jirutka: What is this?

testing/eigen/APKBUILD
chambbj: Great suggestion. I'll look into this.

testing/eigen/APKBUILD
chambbj: Eigen is a header only library. Nothing actually needs to be built. This is just moving the extracted headers to their intended home in fakeroot.

chambbj:
@ncopa My thought was that `eigen-dev` is the only package that people would ever use. Eigen is a header only library that a developer may wish to use in their own application development.

testing/eigen/APKBUILD
jirutka: I understand that, but why it’s named `unsupported`? This is quite general name, it should be inside `/usr/include/Eigen` directory… if it _really_ belongs to `/usr/include`, there are even CMakeList.txt files, tests etc.

testing/eigen/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Bradley J Chambers <brad.chambers@gmail.com>
+# Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+pkgname=eigen
+pkgver=3.3.4
+pkgrel=0
+pkgdesc="Eigen is a C++ template library for linear algebra"
+url="http://eigen.tuxfamily.org/index.php?title=Main_Page"
+arch="all"
```
jirutka: Have you read the `abuild` output?

```
>>> WARNING: eigen*: No arch specific binaries found so arch should probably be set to "noarch"
```

chambbj: Fixed

testing/eigen/APKBUILD
chambbj: Nope, it doesn't belong inside `/usr/include`. That was my mistake. Now, both `Eigen` and `unsupported` are installed to `/usr/include/eigen3`. For more on `unsupported` see [here](https://eigen.tuxfamily.org/dox/unsupported/index.html).

chambbj:
@jirutka @ncopa I believe all items have been addressed. If you are happy with them, I can squash.

testing/eigen/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+
+prepare() {
+        mv "$builddir"-* "$builddir"  # directory name contains hash
+        default_prepare
+}
+
+package() {
+        mkdir -p "$pkgdir"/usr/include/eigen3
+        cp -r "$builddir"/Eigen "$pkgdir"/usr/include/eigen3
+        cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
```
ncopa: Do we want support something that is named `unsupported`?

PR #1893, opened 2017-07-10
testing/nitro: new aport

chambbj:
https://github.com/hobu/nitro
Nitro is a library for reading/writing NITF files.


testing/nitro/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+# Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+pkgname=nitro
+pkgver=2.7
+_pkgver="$pkgver"dev-1
+pkgrel=0
+pkgdesc="The Nitro NITF project"
+url="https://github.com/hobu/nitro"
+arch="all"
+license="GPL3"
+depends="libstdc++ libgcc musl"
```
ncopa: Are those dependencies really needed? They should be automatically detected.

ncopa: They are automatically detected:
https://travis-ci.org/alpinelinux/aports/builds/252150974#L7304-L7306

chambbj: Thanks for catching this. I'll make the change.

chambbj:
@ncopa We should be good now unless you prefer to have the entire `depends=""` line removed when not needed.

PR #1897, opened 2017-07-10
community/fvwm: fix project url

rgdoliveira:


community/fvwm/APKBUILD
```diff
@@ -2,9 +2,9 @@
 # Maintainer: ScrumpyJack <scrumpyjack@st.ilet.to>
 pkgname=fvwm
 pkgver=2.6.7
-pkgrel=0
+pkgrel=1
 pkgdesc="FVWM is a powerful desktop window manager for the X  Window system"
-url="http://fvwm.org"
```
fabled: This URL is still valid, and seems to be the home page of the project. Why would we want to show the source code browser instead of the real homepage?

rgdoliveira: hm, weird. When I tried the link it did not work. Maybe the website got down.

I will close this PR, thanks for reviewing it.

PR #1900, opened 2017-07-11
main/sshfs: upgrade to 3.0.0

tmpfile:

PR #1902, opened 2017-07-11
main/fuse: upgrade to 3.1.0

tmpfile:


fcolista:
This upgrade breaks ABI, there are those packages that must be rebuilt with the new version:
(sshfs rebuild is already pushed in commit #1900 )

- **main:**
avfs
encfs
gvfs
kamailio
py-fuse
truecrypt
unionfs-fuse

- **community:**
archivemount
bmon
fuse-exfat
luufs
lxcfs
obnam
tilda

- **testing:**
geth
i3status
lizardfs
mergerfs
mhddfs

I'll bump the pkgver for those other packages.

fcolista:
NACK.
fuse3 ABI breakage causes encfs to not work, and it was the second package I've tested while bumping the pkgrel...so I don't know about the others...so this patch cannot be applied as-is.
A possibile workaround might be rename fuse to fuse2, and use fuse with new version (this is what Arch does. Void is sticking to 2.9.7, Gentoo, Suse and Fedora has two different packages as well).

nmeum:
fuse 3 should be added as a separate package so we can still compile packages still using the old fuse 2 API.

nmeum:
I added a new aport `testing/fuse3` with your changes. If you take over maintainer ship for this aport and upgrade it to 3.2.0 we can move it to main and merge #1900.

PR #1911, opened 2017-07-12
main/[various]: remove contributor lines

tmpfile:


ncopa:
This will not be merged. I don't want remove the contributor lines unless there are a good reason to do so.

PR #1915, opened 2017-07-13
community/zstd: fix tests in ppc64le

rgdoliveira:
some zstd tests are breaking in ppc64le because it defines the max
number of supported cores as 128 (ZSTDMT_NBTHREADS_MAX 128) but the
Alpine ppc64le machine has 160 cores.

I am increasing the max core value to 256 for now and opened a bug[1]
in zstd project.

[1] https://github.com/facebook/zstd/issues/760


rgdoliveira:
@ncopa can you review this one please? It is the case we discussed today.

PR #1937, opened 2017-07-15
testing/py-sphinxcontrib-websupport: new aport

misery:


misery:
I'm afraid that this package is mandatory for py-sphinx - which ist in main already. :-(

PR #1941, opened 2017-07-15
main/putty: upgrade to 0.70

misery:


tmpfile:
> source="http://the.earth.li/~sgtatham/putty/latest/putty-$pkgver.tar.gz"

latest should be changed to $pkgver
If it isn't the APKBUILD will fetch an invalid url after latest stops being pkgver (as it happens now).

misery:
Fixed

PR #1945, opened 2017-07-18
testing/pcsc-cyberjack: new aport

misery:


fabled:
Add `--disable-static` unless there's good reason for the static library.

misery:
fixed

PR #1949, opened 2017-07-19
testing/lttng-tools: enable build on ppc64le

rgdoliveira:
Some tests from lttng-tools were failing because the ppc64le alpinebox
has 160 cores and some lttng-tools stress tests were running out of
file descriptors. I increased the ulimit hard limit value in the host
to fix this issue.


rgdoliveira:
@ncopa Increasing the alpinebox ppc64le hard limit fixed the issue in my container, think it will work with the builder too

PR #1963, opened 2017-07-21
testing/py-backports-csv: new aport

tboerger:
This new package is required for py-cli_helpers which is a new
dependency required for the update of mycli/pgcli

### DIFF
--- a/testing/py-backports-csv/APKBUILD
+++ b/testing/py-backports-csv/APKBUILD
@@ -1,0 +1,54 @@
+# Contributor: Thomas Boerger <thomas@webhippie.de>
+# Maintainer: Thomas Boerger <thomas@webhippie.de>
+pkgname=py-backports-csv
+_pkgname=backports.csv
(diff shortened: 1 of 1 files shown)


jirutka:
terminaltables supports both Python 2 (legacy) and 3, so it must be packaged for Python 3 only (py3-terminaltables), or for both Python 2 and 3 (py-terminaltables with subpackages py2-terminaltables and py3-terminaltables). If there’s no specific reason for supporting Legacy Python, I’d prefer py3 only package, it’s simpler.

jirutka:
backports.csv supports both Python 2 (legacy) and 3, so it must be packaged for Python 3 only (py3-backports-csv), or for both Python 2 and 3 (py-backports-csv with subpackages py2-backports-csv and py3-backports-csv). If there’s no specific reason for supporting Legacy Python, I’d prefer py3 only package (it’s simpler).

tboerger:

PR #1964, opened 2017-07-21
testing/py-cli_helpers: new aport

tboerger:
This new package is required for the update of mycli/pgcli


tboerger:
This depends on https://github.com/alpinelinux/aports/pull/1962 and https://github.com/alpinelinux/aports/pull/1963

jirutka:
cli_helpers supports both Python 2 (legacy) and 3, so it must be packaged for Python 3 only (py3-cli_helpers), or for both Python 2 and 3 (py-cli_helpers with subpackages py2-cli_helpers and py3-cli_helpers). If there’s no specific reason for supporting Legacy Python, I’d prefer py3 only package, it’s simpler.

tboerger:

PR #1965, opened 2017-07-21
testing/py-pgspecial: upgrade to 1.8.0

tboerger:
The old version had been flagged and the upgrade is required for an
update to pgcli.


jirutka:
Please change python2 to python3 and try if it works under Python 3.6 (upstream declares and tests compatibility for 2.6, 2.7, 3.3, 3.4).

tboerger:

PR #1966, opened 2017-07-21
testing/py-prompt_toolkit: upgrade to 1.0.14

tboerger:
The old version had been flagged and the upgrade is required for an
update to pgcli/mycli.


jirutka:
Please change python2 to python3 and rename package to py3-prompt_toolkit, or add py2-prompt_toolkit and py3-prompt_toolkit subpackages (see [wiki](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python#Multiversion_package_without_executables)).

tboerger:

PR #1981, opened 2017-07-24
Rename gst-plugins-base1-dev dependency to gst-plugins-base-dev

PureTryOut:
The gst-plugins-base1-dev dependency was renamed to gst-plugins-base-dev, so I've updated farstream to use this.


ncopa:
i fixed the commit message a bit

PureTryOut:
"main/farstram"

Not so sure about that ;)

ncopa:
:man_facepalming: mondays....

PR #1983, opened 2017-07-24
July 2017 merge from Adélie

awilfox:
These are the changes from Adélie that we feel are appropriate for inclusion in Alpine.

Mostly they are simple fixes/modernisations of packages used by base X.Org.  There is a bump for ca-certificates since Debian has removed the version Alpine uses from their mirror.  There are also three packages that gain testing support.

Let me know if there are any issues with any of the patches.

All the best,
--arw


kaniini:
@awilfox some build failures: main/attr (fails testsuite), main/ca-certificates (missing python build dependency), main/unzip (404), main/paxmark (needs checksums fixed)

awilfox:
@jirutka: I have fixed the commit message for unzip.

@kaniini:

* main/attr builds and tests fine here on x86_64 and PowerPC.  On what architectures does it fail?
* main/ca-certificates has python3 in makedepends.  What is the error?
* main/paxmark is fixed in latest push.
* I don't know why main/unzip has a 404 error.  The file should be hosted on Alpine's own infrastructure (this appears to be to avoid FTP).

awilfox:
Okay, I have finally figured out how to view Travis CI logs and I think I see the problems.  But I am not sure how to fix them.

* main/attr fails because `ls` in busybox does not have the same output as `ls` in coreutils.  Adélie is using coreutils so I did not notice the bug.  Perhaps that commit cannot be merged then.  Alternatively coreutils could be added to checkdepends.
* main/ca-certificates fails because the Alpine Python 3 package does not provide /usr/bin/python, only the Python 2 package does.  Again, since Adélie only ships Python 3, we provide the symlink and that was masking the error on our build infrastructure.  I am not sure how to resolve this.

kaniini:
Hmm.

- main/attr: lets add coreutils to `$checkdepends`.

- main/ca-certificates: lets patch mozilla/Makefile to use python3 explicitly.

Other than that, it seems fine.

kaniini:
Actually @jirutka pointed out that adding coreutils to $checkdepends would be problematic.  Lets just disable the broken tests.

jirutka:
> the Alpine Python 3 package does not provide /usr/bin/python, only the Python 2 package does.

Yes, that’s correct, this is highly recommended by upstream, see [PEP 394](https://www.python.org/dev/peps/pep-0394/).
If Adélie doesn’t ship Python 2, then it makes sense to symlink `python3` to `python`, but that’s not Alpine’s case.

awilfox:
I have dropped the ca-certificates patch since it seems to be better handled in tree now.  Additionally, I have dropped the attr test suite addition until I find a way to disable the failing tests.

awilfox:
As discussed on IRC I have additionally added two commits that fix cross-compilation.

jirutka:
Merged, thanks for your contribution!

(GH PR Closer didn’t recognize this PR, strange 😕)

PR #1986, opened 2017-07-24
main/nfs-utils: Patch idmapd to remove dependency on DNOTIFY

jantatje:
idmapd requires CONFIG_DNOTIFY, which is not enabled in
linux-vanilla or linux-hardened. This is fixed by applying this
patch (https://patchwork.kernel.org/patch/9802949/) by Alan Swanson
which uses INOTIFY instead.


jirutka:
Please add original patch header and information about origin of the patch (that patchwork url) into the patch file.

PR #1990, opened 2017-07-24
Try to fix gvfs-fuse and bring back xfce-polkit

tmpfile:
I noticed a problem in Xfce.
If you mount an external NTFS hard-drive by opening Thunar (with thunar-volman, gvfs and gvfs-fuse installed) the drive will be mounted read-only.
Hard-drives that are not NTFS work fine.
And mounting from the terminal with ntfs-3g also works fine.
So the problem is somewhere in thunar, it's not calling ntfs-3g but most likely the kernel ntfs support (which lacks write functionality).
So I think the problem is with gvfs-fuse.

I found some discussions online that point to gvfs-fuse needing fuse-utils as dependency.
This is what I try to accomplish in this PR.

And I also moved xfce-polkit back from unmaintained because it's necessary to mount from thunar as normal user, without root.

### DIFF
--- a/main/gvfs/APKBUILD
+++ b/main/gvfs/APKBUILD
@@ -1,19 +1,18 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=gvfs
-pkgver=1.32.1
+pkgver=1.33.3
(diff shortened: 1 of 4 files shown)


tmpfile:
Turns out this had nothing to do with the issue.
What was needed was to make a symlink (ntfs-3g -> mount.ntfs)

PR #1991, opened 2017-07-24
testing/container-linux-config-transpiler: new aport

errm:


errm:
Closing this because upstream now release a static bin that works fine on alpine... so no need for this package I guess ....

PR #1997, opened 2017-07-25
testing/skopeo: new aport

errm:


testing/skopeo/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+# Maintainer: Ed Robinson <ed@reevoo.com>
+pkgname=skopeo
+pkgver=0.1.23
+pkgrel=0
+pkgdesc="Work with remote images registries - retrieving information, images, signing content"
+url="https://github.com/projectatomic/skopeo"
+arch="x86_64"
+license="ASL 2.0"
+makedepends="go btrfs-progs-dev gnupg gpgme-dev ostree-dev lvm2-dev bash"
+options="!strip"
```
Shizmob: Why is this needed? Can it be commented?

testing/skopeo/APKBUILD
```diff
@@ -0,0 +1,39 @@
+# Contributor: Ed Robinson <ed@reevoo.com>
+# Maintainer: Ed Robinson <ed@reevoo.com>
+pkgname=skopeo
+pkgver=0.1.23
+pkgrel=0
+pkgdesc="Work with remote images registries - retrieving information, images, signing content"
+url="https://github.com/projectatomic/skopeo"
+arch="x86_64"
```
Shizmob: Does this work for other arches too?

testing/skopeo/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+
+build() {
+	cd ${_projectatomic}/${pkgname}
+	make binary-local
+}
+
+package() {
+	cd "$builddir"
+	install -Dm755 skopeo "$pkgdir"/usr/bin/skopeo
+	install -Dm644 LICENSE "$pkgdir"/usr/share/licenses/$pkgname/LICENSE
```
Shizmob: We generally aim to avoid installing licenses if they are standard ones, afaik.

testing/skopeo/APKBUILD
errm: because this is go, you cannot strip the binary in the normal way...

errm: IIRC this is just set without comment on every other go based package

testing/skopeo/APKBUILD
errm: It should do, but I have no way of testing, I only have an Intel machine to test on.

testing/skopeo/APKBUILD
errm: I am not a lawyer, but in my view not including the license in the package would break 4 (a) of ASL 2.0

We are installing the licence in some other packages IIRC

testing/skopeo/APKBUILD
errm: e.g. https://github.com/alpinelinux/aports/search?p=1&q=%21strip+go&type=&utf8=%E2%9C%93

jirutka: We used to not strip go binaries because of some FUD, but then we realised that they can be normally striped as any other binary. However, it remained unchanged in existing abuilds. If you're not sure that it doesn't work stripped, then please remove `!strip`.

testing/skopeo/APKBUILD
jirutka: In any case we split licenses into a -doc subpackage (add `subpackages=$pkgname-doc`).

errm:
@jirutka I have made some changes PTAL

errm:
🤷‍♂️

clandmeter:
Sorry but this seemed to have merged with new version of LXC.

PR #2015, opened 2017-07-28
Tomcat native openssl

amadigan:
I realize that Alpine has switched to LibreSSL (save for easy-rsa, also in community), but the upstream for this package does not yet support LibreSSL (https://bz.apache.org/bugzilla/show_bug.cgi?id=58434). The default Tomcat install logs an error (but does start) when the current tomcat-native is used. For various reasons it is useful to be able to run SSL in Tomcat, and I think it would be better to avoid having to choose between using tomcat-native and using SSL in Tomcat.

Sorry for submitting the PR twice, I realized after that tomcat-native did have tests, but there were some issues with the ant build.xml that executes them. I've added a patch to enable the tests, which do pass. I also tested manually with Tomcat 8.5.16 with SSL enabled.

### DIFF
--- a/community/tomcat-native/APKBUILD
+++ b/community/tomcat-native/APKBUILD
@@ -9,14 +9,17 @@ url="http://tomcat.apache.org/native-doc/"
 arch="all"
 license="ASL-2.0"
 depends="openjdk8-jre-base openssl"
-makedepends="apr-dev chrpath openjdk8 openssl-dev"
+makedepends="apr-dev chrpath openjdk8 openssl-dev apache-ant"
 subpackages="$pkgname-dev"
-source="http://www-eu.apache.org/dist/tomcat/tomcat-connectors/native/$pkgver/source/$pkgname-$pkgver-src.tar.gz"
-builddir="$srcdir/$pkgname-$pkgver-src/native"
-options=!check
+source="http://www-eu.apache.org/dist/tomcat/tomcat-connectors/native/$pkgver/source/$pkgname-$pkgver-src.tar.gz enable-tests.patch"
+builddir="$srcdir/$pkgname-$pkgver-src"
+
+prepare()	{
+	default_prepare || return 1
+}
 
 build() {
-	cd "$builddir"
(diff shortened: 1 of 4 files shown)


community/tomcat-native/APKBUILD
```diff
@@ -1,45 +1,55 @@
...
-pkgrel=0
+pkgrel=1
 pkgdesc="Native resources optional component for Apache Tomcat"
 url="http://tomcat.apache.org/native-doc/"
 arch="all"
 license="ASL-2.0"
-depends="openjdk8-jre-base"
-makedepends="apr-dev chrpath openjdk8"
+depends="openjdk8-jre-base openssl"
+makedepends="apr-dev chrpath openjdk8 openssl-dev apache-ant"
```
jirutka: `apache-ant` should be in `checkdepends`.

community/tomcat-native/APKBUILD
```diff
@@ -1,45 +1,55 @@
...
 subpackages="$pkgname-dev"
-source="http://www-eu.apache.org/dist/tomcat/tomcat-connectors/native/$pkgver/source/$pkgname-$pkgver-src.tar.gz"
-builddir="$srcdir/$pkgname-$pkgver-src/native"
+source="http://www-eu.apache.org/dist/tomcat/tomcat-connectors/native/$pkgver/source/$pkgname-$pkgver-src.tar.gz enable-tests.patch"
```
jirutka: Please put each source on a separate line.

jirutka:
I’d prefer to split this package into two variants – tomcat-native w/o TLS and tomcat-native-tls with TLS, each with `depends="!<other>"` and `replaces="<other>"` (`<other>` means `$pkgname` or `$pkgname-tls`).

amadigan:
I'll update with the fixes and split this to a "tomcat-native-openssl" package (is that name OK?). I'll also see if I can get the tests to pass without SSL.

amadigan:
Should I close the request?

PR #2041, opened 2017-07-29
testing/bird: upgrade to 1.6.3

Fusl:


testing/bird/APKBUILD
```diff
@@ -1,8 +1,9 @@
 # Contributor: Francesco Zanini <francesco@zanini.me>
+# Contributor: Katie Holly <holly@fuslvz.ws>
 # Maintainer: Francesco Zanini <francesco@zanini.me>
-pkgname=bird
-pkgver=1.6.0
-pkgrel=2
+pkgname="bird"
```
Shizmob: This extra quoting isn't really necessary and not in line with the rest of the APKBUILDs. I'd prefer it without, personally.

testing/bird/APKBUILD
```diff
@@ -11,17 +12,16 @@ depends=""
...
+	ftp://bird.network.cz/pub/bird/bird-$pkgver.tar.gz
+	bird.initd
+	"
+builddir="$srcdir/bird-$pkgver"
 
 prepare() {
 	cd "$builddir"
-	default_prepare || return 1
-	update_config_guess || return 1
+	default_prepare
```
Shizmob: 👌

testing/bird/APKBUILD
```diff
@@ -32,49 +32,39 @@ _build() {
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
 		--localstatedir=/var \
-		$@ \
-		|| return 1
-	make || return 1
+		$@
```
Shizmob: Could you change it to `"$@"` here? According to POSIX, `"$@"` is the only way to reproduce the received argument line with proper quoting itself. It's a bit of a special case.

testing/bird/APKBUILD
```diff
@@ -32,49 +32,39 @@ _build() {
...
 
-
 build() {
 	cd "$builddir"
 	mkdir -p ipv6 ipv4
-	cd "$builddir"/ipv6
-	_build --enable-ipv6 || return 1
-	cd "$builddir"/ipv4
-	_build || return 1
+	cd "$builddir/ipv6"
```
Shizmob: Same as above, I'm not a big fan of the unnecessary quoting, although it's not a huge deal here. Existing style is to only quote variables, not the entire thing that includes the variable.

PR #2042, opened 2017-07-30
community/ghc: upgrade to 8.2.1

mitchty:
There is only one package that depends on ghc right now, and that is
community/cabal. The move to community and the subsequent cabal file
bounds that added made the older cabal-install build with 8.2.1 so
no real reason to hold off upgrading ghc.

Removed the haskell-* provides as that was originally a plan to follow
what arch was doing for their haskell setup. In lieu of that not
providing value removing them entirely.

The ghc test suite was updated to use python 3 from python 2.

Finally really late in the release candidate cycle a change to ld
detection necessitated the need to pass --disable-ld-override to the
configure of ghc.

Without that in place there appears to be some shenanigans going on
in exported symbols.

Ghc upstream ticket:
https://ghc.haskell.org/trac/ghc/ticket/13958

Offending commmit:
https://git.haskell.org/ghc.git/commitdiff/2785ef0e31a123400da950ffafebe6cb1ce3f4eb


mitchty:
Before this is merged I want to have some talks to the stack folks as they might be affected if ghc gets upgraded from 8.0.2 to 8.2.1:
https://github.com/commercialhaskell/stack/issues/3045

Just want to make sure we don't negatively impact downstream users. cc @borsboom

borsboom:
Ideal, from my point of view, would be a way to install multiple version of GHC, since GHC is a project that tends to have breaking changes with every major release.  So e.g. packages `ghc-8.2.1` and `ghc-8.0.2`, and maybe a way to select the default (similar to Debian's update-alternatives).  I don't know Alpine's package management system well enough to know whether that's practical.

Stack will probably stick with building using ghc 8.0.2 until the next Stackage LTS release that uses GHC 8.2.1 comes out.  Either way, though, I don't think that should block Alpine from upgrading to the latest GHC.

mitchty:
Do you mean installing multiple versions simultaneously or just the ability to be able to select a specific version?

I know there is/was some recent discussion on the mailing list to allow for something similar to update-alternatives. But given I'm more just a maintainer for ghc I'm not the best to speak for that getting added as a feature into Alpine Linux. I'll ask on irc what might or is planned as something similar for ghc would be ideal due to the breaking changes llvm tends to throw into its IR representation which ghc tends to have issues with as well.

Just wanted to be sure this wouldn't negatively impact stack builds, thanks for the information @borsboom !

borsboom:
> Do you mean installing multiple versions simultaneously or just the ability to be able to select a specific version?

I think the former would be ideal, although the latter would still be useful.  It's how I wish all distros would package GHC (similar to how they support multiple python or perl major versions, since those also have changes that break the whole ecosystem, just like major GHC versions usually do, although we're a lot quicker about updating the ecosystem).

mitchty:
Roger that, I'll ask around. I was setting up some personal CI to do something similar to hvr's ppa and install to /opt/ghc/$version to keep track of upstream git changes. I can probably test out doing stuff like that to provide an alternative apk repo to install ghc-$version alongside of the official releases.

Wouldn't be ideal but it should allow for that use case in the interim at least. Well assuming you trust the compiled binaries I build, but thats another matter entirely.

borsboom:
That sounds great, thanks!  Alternatively, if you have a process for building bindists that can install with the standard `./configure --prefix=PATH && make install` process, that would work really well and could _also_ be used by `stack setup` for Stack to be able to manage multiple GHC installations itself.  That's basically all we'd need to be able to have Stack officially support Alpine Linux (https://github.com/commercialhaskell/stack/issues/2387).

mitchty:
I forgot to add the updated 0000-alpine.patch. Now fixed, and the travis builder appears to not have community packages but it always times out anyway.

I'll work on getting the multiple ghc stuff setup too, it wouldn't be too hard to set it up to build bindists as well. The only annoying part would be armhf takes almost a full day to compile on my one box. I'll get some sort of ci setup going over the weekend and give it a go.

borsboom:
Since I'm making requests already, I'll toss another in here: can we get x86 (i386) packages for GHC too?  Then I can start building static 32-bit `stack` executables!.

mitchty:
Yep, I'm working on it, but I have to do it a bit different than my initial port so... its taking longer. Someone asked on the mailing list for it too, it was just never a priority as all I needed was armhf and x64 for my own nefarious needs. I'll see if I can't poke at it again this weekend, its mostly me needing to understand the cross compile setup on alpine linux.

mitchty:
Actually hold off on this for a bit, hitting some fun linking errors on armhf. Related to https://sourceware.org/bugzilla/show_bug.cgi?id=16177

mitchty:
So the problem appears that community/llvm39 is built against a prior libc++, the abi changed for llvm4.0 so I get lots of linker erorrs.

So this PR shouldn't get merged until community/llvm39 and probably community/llvm37 are rebuilt. There is a couple of changes I did find in debugging but I can add that in tonight.

fabled:
Thanks for working on this and the updates.

mitchty:
That got further at least on arm, new fun!

```
"inplace/bin/ghc-stage2" -hisuf dyn_hi -osuf  dyn_o -hcsuf dyn_hc -fPIC -dynamic  -H32m -O -Wall      -hide-all-packages -i -iutils/
haddock/driver -iutils/haddock/haddock-api/src -iutils/haddock/haddock-library/vendor/attoparsec-0.13.1.0 -iutils/haddock/haddock-li
brary/src -iutils/haddock/dist/build -Iutils/haddock/dist/build -iutils/haddock/dist/build/haddock/autogen -Iutils/haddock/dist/buil
d/haddock/autogen    -optP-DIN_GHC_TREE -optP-include -optPutils/haddock/dist/build/haddock/autogen/cabal_macros.h -package-id base-
4.10.0.0 -package-id filepath-1.4.1.2 -package-id directory-1.3.0.2 -package-id containers-0.5.10.2 -package-iddeepseq-1.4.3.0 -pack
age-id array-0.5.2.0 -package-id xhtml-3000.2.2 -package-id Cabal-2.0.0.2 -package-id ghc-boot-8.2.1 -package-id ghc-8.2.1 -package-
id bytestring-0.10.8.2 -package-id transformers-0.5.2.0 -funbox-strict-fields -Wall -fwarn-tabs -O2 -threaded -XHaskell2010  -no-use
r-package-db -rtsopts -Wno-unused-imports -Wno-deprecations     -Wnoncanonical-monad-instances  -odir utils/haddock/dist/build -hidi
r utils/haddock/dist/build -stubdir utils/haddock/dist/build -c utils/haddock/haddock-library/src/Documentation/Haddock/Types.hs -o
utils/haddock/dist/build/Documentation/Haddock/Types.dyn_o
make[1]: *** [utils/haddock/ghc.mk:21: utils/haddock/dist/build/Haddock/Backends/Hyperlinker/Types.dyn_o] Segmentation fault
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [utils/haddock/ghc.mk:21: utils/haddock/dist/build/ResponseFile.dyn_o] Segmentation fault
make[1]: *** [utils/haddock/ghc.mk:21: utils/haddock/dist/build/Haddock/GhcUtils.dyn_o] Segmentation fault
make[1]: *** [utils/haddock/ghc.mk:21: utils/haddock/dist/build/Documentation/Haddock/Types.dyn_o] Segmentation fault
make: *** [Makefile:127: all] Error 2
```

Wondering if its just stack/ulimit -a related, I'll find out shortly.

nmeum:
> Wondering if its just stack/ulimit -a related, I'll find out shortly.

Did you figure out if that's the case?

mitchty:
Sadly no, I've been trying to bisect ghc from when 8.0.2 worked to 8.2.1 failed and then figure out what to fix. But its slow going with arm builds of things and I often end up having to debug a lot of other things.

mitchty:
Going to close this out in lieu of 8.4.

neongreen:
Everybody who was subscribed to this PR can now subscribe to https://github.com/alpinelinux/aports/pull/4255.

mitchty:
https://github.com/alpinelinux/aports/pull/4255/commits is the 8.4.2 PR

PR #2043, opened 2017-07-31
community/whois: update to 5.2.17

itoffshore:
https://github.com/rfc1036/whois/blob/next/debian/changelog


fabled:
Seems this was updated earlier in July 29 already.

PR #2050, opened 2017-07-31
Removing myself as a maintainer

olivier-mauras:


nmeum:
Sad to see you go. @ncopa should we also move these packages back to `testing/` since they don't have a maintainer anymore?

jirutka:
I’d prefer _not_ to move them back to testing. I’m willing to take over borgbackup and py-pygit2, if no one else wants them. Who’s next volunteer?

nmeum:
I believe it is stupid to have a maintainer comment in those aports pointing to @olivier-mauras even though he is not activity maintaining those packages. We sadly already have several packages in `main/` and `community/` that do not have a maintainer, so even though this sucks it isn't new at all.

Therefore I removed @olivier-mauras from all those packages without removing the maintainer comment entirely and bumped the pkgrels to trigger a rebuild since the maintainer information is also stored in the resulting apk package.

This should make it easier for other people to take over maintainership for those packages since it is very clear now that they are unmaintained, so @jirutka if you would like to maintain `borgbackup` and `py-pygit2` go ahead.

PR #2057, opened 2017-07-31
Add community/consolekit2

PureTryOut:
As part of a larger effort to [package Plasma Mobile](https://github.com/postmarketOS/pmbootstrap/tree/37981fccd0635b13a9d23c27ea0ce3989a559c54/aports/kde) for the [postmarketOS](https://postmarketos.org) project, and thus also for Alpine Linux, I'd like to add the consolekit2 package. It's a fork of the consolekit already packaged, with the difference of actually being maintained. It should be a drop in replacement.

The patch included makes it build on systems other than x86_64 (specifically armhf), but is already upstreamed in consolekit2 and can be removed in a next release.

Since this is the first package I'm submitting, ever, there is probably some stuff which could be improved in here.

### DIFF
--- a/testing/consolekit2/APKBUILD
+++ b/testing/consolekit2/APKBUILD
@@ -1,0 +1,38 @@
+# Contributor: Bart Ribbers <bart.ribbers@openmailbox.org>
+# Maintainer: Bart Ribbers <bart.ribbers@openmailbox.org>
+pkgbase=ConsoleKit2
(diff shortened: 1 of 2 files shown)


kaniini:
hi, can you squash this down into a single commit?

also, normally WIP packages go into testing, but if you say it is working, I am fine with accepting it directly into community.

PR #2064, opened 2017-08-02
testing/hexer: new aport

chambbj:
https://github.com/hobu/hexer
LAS and OGR hexagonal density and boundary surface generation.


testing/hexer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+arch="all"
+license="LGPL2+"
+options="!check"
+depends="libgcc libstdc++ musl gdal"
+makedepends="cmake gdal-dev"
+install=""
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::https://github.com/hobu/$pkgname/archive/$pkgver.tar.gz
+        10-missing-includes.patch
+        20-backtrace-guards.patch"
```
ncopa: are those patches submitted upstream?

testing/hexer/20-backtrace-guards.patch
```diff
@@ -0,0 +1,32 @@
...
+index 92d810b..9ea3dc7 100644
+--- a/apps/Utils.cpp
++++ b/apps/Utils.cpp
+@@ -44,7 +44,9 @@
+ #include <cxxabi.h>
+ #include <sys/ioctl.h>
+ #include <sys/wait.h>  // WIFEXITED, WEXITSTATUS
++#ifdef __GLIBC__
+ #include <execinfo.h> // backtrace
++#endif
```
ncopa: It would probably be better to have a compile test that checks for `execinfo.h` and set `HAVE_EXECINFO_H` or similar and then do:
````C
#ifdef HAVE_EXECINFO_H
#include <execinfo.h>
#endif
````

File: testing/hexer/10-missing-includes.patch [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+diff --git a/include/hexer/Utils.hpp b/include/hexer/Utils.hpp
+index 8c26fd3..fd7ed30 100644
+--- a/include/hexer/Utils.hpp
++++ b/include/hexer/Utils.hpp
+@@ -15,6 +15,9 @@
+ 
+ #pragma once
+ 
++#include <cmath>
++#include <limits>
```
ncopa: This change can likely go upstream as is.

testing/hexer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+        mkdir -p "$builddir"
```
ncopa: This should not be needed

testing/hexer/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+}
+
+package() {
+        cd "$builddir"/build
+        make DESTDIR="$pkgdir" install
+}
+
+prepare() {
+        default_prepare
+}
```
ncopa: You can remove this `prepare()` function since this is exactly what the abuild provided `prepare()` does.

testing/hexer/APKBUILD
chambbj: Not yet, but I can make it happen.

chambbj: Refreshing my memory now, the first patch is already in master, so I just need to coordinate a new release. The second is somewhat specific to this effort to build on Alpine, so I'll need to address that separately, taking your suggested compile test into consideration.

chambbj:
@ncopa Requested changes made. I will make sure these get submitted to the upstream repo as well.

PR #2065, opened 2017-08-02
testing/cloudi: enable with patch fix

okeuday:
Disable the cloudi_service_quorum test due to failures on alpine build servers
related to a random number rounding problem.

cc @kaniini


okeuday:
@kaniini Thank you for helping guide this!  Based on the build logs the testing/cloudi package is ok now and will be used for further testing.  Thanks!

PR #2074, opened 2017-08-03
main/libatomic_ops: upgrade to 7.4.6

misery:
* Modernize abuild
* Enable shared library
* Use source tarball


nmeum:
Thanks for your patch! Just one small suggestion: It is usually helpful if you split commits upgrading the packages and commits modernizing the APKBUILD.

PR #2086, opened 2017-08-05
main/git: Upgrade to 2.14.0

andypost:
Ref https://lkml.org/lkml/2017/8/4/481

### DIFF
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=git
-pkgver=2.13.3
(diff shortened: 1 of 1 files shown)


andypost:
Form release notes
> Git can now be built with PCRE v2 instead of v1 of the PCRE
   library. Replace USE_LIBPCRE=YesPlease with USE_LIBPCRE2=YesPlease
   in existing build scripts to build against the new version.  As the
   upstream PCRE maintainer has abandoned v1 maintenance for all but
   the most critical bug fixes, use of v2 is recommended.
So gonna push this update as well

andypost:
> pcre2-dev (missing)
so reverting

ncopa:
I think we should move pcre2 from community to main and use pcre2 for git.

andypost:
now build went fine!

PR #2087, opened 2017-08-06
testing/jo: upgrade to 1.1 and modernize

rgdoliveira:


fcolista:
Hi @rgdoliveira !
Would you mind to move the "make check" inside a check() function, rather than keep it in build()? 
Thanks!

rgdoliveira:
sure @fcolista! just did that, thanks for the review.

PR #2094, opened 2017-08-06
testing/colormake: snapshot from 20170221

HRio:
While at it add symlinks:
 clmake like colormake but piped through less
 colormake-short and clmake-short versions of the commands will massage the output so lines do not wrap

### DIFF
--- a/testing/colormake/APKBUILD
+++ b/testing/colormake/APKBUILD
@@ -1,17 +1,18 @@
 # Contributor: Henrik Riomar <henrik.riomar@gmail.com>
 # Maintainer: Henrik Riomar <henrik.riomar@gmail.com>
 pkgname=colormake
-pkgver=0.9.20140503
+pkgver=0.9.20170221
 pkgrel=0
+_sha=93dd19b5143b124208caab5e9b6d57b5e6ec02cb
 pkgdesc="A simple wrapper around make to colorize the output"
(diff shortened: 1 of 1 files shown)


testing/colormake/APKBUILD
```diff
@@ -1,17 +1,18 @@
 # Contributor: Henrik Riomar <henrik.riomar@gmail.com>
 # Maintainer: Henrik Riomar <henrik.riomar@gmail.com>
 pkgname=colormake
-pkgver=0.9.20140503
+pkgver=0.9.20170221
 pkgrel=0
+_sha=93dd19b5143b124208caab5e9b6d57b5e6ec02cb
```
ncopa: I want avoid adding random git snapshots if possible. Could you ask if upstream could make a new tag?

HRio: @ncopa sure done https://github.com/pagekite/Colormake/issues/22

HRio: @ncopa upstream does not seem to be active on github at the moment: https://github.com/pagekite

93dd19b5143b124208caab5e9b6d57b5e6ec02cb is the latest in HEAD (not moved since 20170221), do you think we can merge this PR while waiting for upstream? I am watching the upstream repo for changes, so if it changes I can prepare a new PR.

PR #2099, opened 2017-08-08
Package libpipeline and man-db, courtesy Mx Wilcox.

Aerdan:
I've found mandoc (mdocml) to be deficient, primarily as mandoc's `man` produces cryptic error messages regarding its database. Consequently, here's an alternate `man` implementation.


fabled:
@awilfox You ok to maintain these on Alpine too?

awilfox:
Yes.  I will be glad to maintain these packages on Alpine as well.

jirutka:
Please split this into two separate commits, one per abuild, and follow our rules for git commit messages (`<repo>/<pkgname>: new aport`).

PR #2108, opened 2017-08-09
testing/aws-cli: upgrade to 1.11.130, also upgraded related library

vakartel:
testing/py-botocore: upgrade to 1.5.93

### DIFF
--- a/testing/aws-cli/APKBUILD
+++ b/testing/aws-cli/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Valery Kartel <valery.kartel@gmail.com>
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 pkgname=aws-cli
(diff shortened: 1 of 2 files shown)


fcolista:
Hi @vakartel thanks for your contribution!
Do you mind to split the patch in two different commits?
It's cleaner have one commit for each upgrade.
Thanks a lot!
.: Francesco

vakartel:
@fcolista I just pushed py-botocore upgrade there to not create new branch in git. This library is used by aws-cli only.

fcolista:
No problem @vakartel  I've merged it.
I don't want to be finicky :) Just keep into consideration that now py-botocore is used by aws-cli only, but it's not going to be like this always.
Having a separate commit for each package should be the standard, please follow this procedure next time.
That said, thanks again for your contribution!

PR #2109, opened 2017-08-09
main/wayland: upgrade to 1.14.0, split libraries. main/wayland-protocols: upgrade to 1.10 

vakartel:


ncopa:
I pushed this but it fails !?

Documentation fails to build and tests fails.

PR #2110, opened 2017-08-09
testing/weston: upgrade to 3.0.0

vakartel:
configure: error: Package requirements (wayland-protocols >= 1.8) were not met

So #2109  must be accepted first


testing/weston/timespec.patch
```diff
@@ -0,0 +1,10 @@
+--- a/tests/timespec-test.c
++++ b/tests/timespec-test.c
+@@ -25,6 +25,7 @@
+ 
+ #include "config.h"
+ 
++#include <time.h>
```
ncopa: is this reported upstream?

PR #2113, opened 2017-08-09
community/rethinkdb: upgrade to 2.3.6

rgdoliveira:


ncopa:
This fails on kernel with pax

PR #2126, opened 2017-08-11
main/syslog-ng: Added new subpackage which contains JSON plugin for s…

talien:
…yslog-ng


ncopa:
fixed it for you. thanks!

PR #2133, opened 2017-08-11
testing/ndctl: upgrade to 57.1 and modernize

rgdoliveira:
Upgrade and remove a patch that is already applied upstream


testing/ndctl/APKBUILD
```diff
@@ -1,42 +1,32 @@
...
+source="$pkgname-$pkgver.tar.gz::https://github.com/pmem/$pkgname/archive/v$pkgver.tar.gz"
 
 builddir="$srcdir"/ndctl-$pkgver
 
 build() {
 	cd "$builddir"
-
-	./autogen.sh || return 1
-
+	./autogen.sh
```
nmeum: I manually did this upstream but just so you know for feature patch submission: Please invoke `./autogen.sh` in `prepare()` not in `build()`. Thanks!

rgdoliveira: Thanks for the review @nmeum . I will remember that too for future patches!

PR #2135, opened 2017-08-12
August 2017 mostly-test-focused merge from Adélie

awilfox:
Adds tests to over 20 packages, marks some as not having test suites, and modernises a few APKBUILDs.

I am not entirely sure if all of these changes fully meet the Alpine standards because inside the tree there isn't a lot of consistency.  I went with what I felt was right, so if they need to be changed, just let me know.


main/ncurses/APKBUILD
```diff
@@ -8,10 +8,11 @@ url="http://www.gnu.org/software/ncurses/"
 arch="all"
 license=MIT
 depends=
+options="!check"  # "tests" are actual demo programs, not test suite
 makedepends_build="ncurses"
 source="ftp://invisible-island.net/ncurses/current/ncurses-$_ver.tgz"
 subpackages="$pkgname-static $pkgname-dev $pkgname-doc
-	$pkgname-terminfo-base:base $pkgname-terminfo $pkgname-libs"
+	$pkgname-terminfo-base:base:noarch $pkgname-terminfo:terminfo:noarch $pkgname-libs"
```
ncopa: We should bump pkgrel, also due to the change of pkgdesc below.

main/autoconf/APKBUILD
```diff
@@ -8,7 +8,9 @@ license="GPL2+"
 url="http://www.gnu.org/software/autoconf"
 depends="m4 perl"
 subpackages="$pkgname-doc"
-source="ftp://ftp.gnu.org/pub/gnu/$pkgname/$pkgname-$pkgver.tar.gz"
+source="ftp://ftp.gnu.org/pub/gnu/$pkgname/$pkgname-$pkgver.tar.gz
+	autoconf-2.69-fix-perl-regex.patch
+	"
```
ncopa: this needs pkgrel bump

nmeum:
I would really like to see this getting merged. Any idea why it currently fails to build using travis? Last time I checked the checkdepends weren't installed by travis.

awilfox:
* Commit messages fixed.
* Another 15 or so packages have test suites.
* Removed all the ones causing failures, until Travis can be fixed.

awilfox:
Would it be better to split this into one commit per PR?  :disappointed:

main/musl/APKBUILD
```diff
@@ -12,7 +12,7 @@ depends_dev="!uclibc-dev"
 makedepends="$depends_dev"
 subpackages="$pkgname-dev $pkgname-dbg libc6-compat:compat:noarch"
 case "$BOOTSTRAP" in
-nocc)	pkgname="musl-dev"; subpackages="";;
+nocc)	pkgname="musl-dev"; subpackages=""; arch="noarch";;
```
fabled: The musl package has headers which are architecture specific, this should not be done.

fabled:
Pushed all others except the musl one.

fabled:
Required moving dejagnu to main; reverted `main/check` as it fails to pass checks.

PR #2143, opened 2017-08-15
Add freedreno support to mesa and libdrm

opendata26:
I have tested these changes with my z2 tablet so they should work fine for other adreno devices.

### DIFF
--- a/main/libdrm/APKBUILD
+++ b/main/libdrm/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=libdrm
 pkgver=2.4.81
-pkgrel=0
+pkgrel=1
 pkgdesc="Userspace interface to kernel DRM services"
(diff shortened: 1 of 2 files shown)


opendata26:
@ncopa  done

ncopa:
can you squash it into the previous commit and `git push --force`?

jirutka:
This PR contains changes in two separate abuilds, so there should be two separate commits (with log title in format `<repo>/<pkgname>: …`), one for each abuild.

opendata26:
Done

opendata26:
@jirutka look ok?

main/libdrm/APKBUILD
```diff
@@ -23,6 +23,7 @@ build() {
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
 		--localstatedir=/var \
+                --enable-freedreno \
```
jirutka: There are spaces instead of tabs. Please use tabs for indentation (and alignment) in APKBUILDs.

PR #11, opened 2015-04-26
Mono v4.0

haf:
Hi,

This PR adds the alpha release of mono to testing.

Do you have a good way of systematically testing your packages? In particular with regards to the grsecurity kernel module / JIT.

PR #2144, opened 2017-08-15
testing/softhsm: new aport

silveraid:
http://www.softhsm.org/
cryptographic store accessible through a PKCS #11

### DIFF
--- a/testing/softhsm/APKBUILD
+++ b/testing/softhsm/APKBUILD
@@ -28,7 +28,8 @@ build() {
                 --prefix=/usr \
                 --sysconfdir=/etc \
                 --localstatedir=/var \
-                --disable-gost
+                --disable-gost \
+                --disable-static
         make
 }
 
--- a/testing/softhsm/APKBUILD
+++ b/testing/softhsm/APKBUILD
@@ -1,0 +1,38 @@
+# Contributor: Frank Felhoffer <silveraid@hackme.ca>
+# Maintainer: Frank Felhoffer <silveraid@hackme.ca>
+pkgname=softhsm
+pkgver=2.3.0
+pkgrel=0
+pkgdesc="cryptographic store accessible through a PKCS #11"
+url="http://www.softhsm.org/"
+arch="all"
+license="BSD2"
+depends="openssl"
+makedepends="openssl-dev automake autoconf libtool file"
+install=""
+subpackages="$pkgname-doc"
+source="softhsm-$pkgver.tar.gz::https://github.com/opendnssec/SoftHSMv2/archive/$pkgver.tar.gz"
+builddir="$srcdir/SoftHSMv2-2.3.0"
+
+prepare() {
+        cd "$builddir"
+        sh autogen.sh || return 1
+}
+
+build() {
+        cd "$builddir"
+        ./configure \
+                --prefix=/usr \
+                --sysconfdir=/etc \
+                --localstatedir=/var \
+                || return 1
+        make -j8 || return 1
+}
+
+package() {
+        cd "$builddir"
+        make -j1 DESTDIR="$pkgdir/" install
+}
+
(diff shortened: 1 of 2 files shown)


testing/softhsm/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+}
+
+build() {
+        cd "$builddir"
+        ./configure \
+                --prefix=/usr \
+                --sysconfdir=/etc \
+                --localstatedir=/var \
+                || return 1
+        make -j8 || return 1
```
ncopa: we should not specify `-j8` to make. The build server should have a good number set via `MAKEFLAGS` env var. Some builder will have higher number and some will have lower.

testing/softhsm/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+depends="openssl"
+makedepends="openssl-dev automake autoconf libtool file"
+install=""
+subpackages="$pkgname-doc"
+source="softhsm-$pkgver.tar.gz::https://github.com/opendnssec/SoftHSMv2/archive/$pkgver.tar.gz"
+builddir="$srcdir/SoftHSMv2-2.3.0"
+
+prepare() {
+        cd "$builddir"
+        sh autogen.sh || return 1
```
ncopa: we dont need the `|| return 1`

testing/softhsm/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+}
+
+build() {
+        cd "$builddir"
+        ./configure \
+                --prefix=/usr \
+                --sysconfdir=/etc \
+                --localstatedir=/var \
+                || return 1
```
ncopa: the `|| return 1` should be removed

testing/softhsm/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+# Maintainer: Frank Felhoffer <silveraid@hackme.ca>
+pkgname=softhsm
+pkgver=2.3.0
+pkgrel=0
+pkgdesc="cryptographic store accessible through a PKCS #11"
+url="http://www.softhsm.org/"
+arch="all"
+license="BSD2"
+depends="openssl"
+makedepends="openssl-dev automake autoconf libtool file"
```
ncopa: Does this build with `libressl-dev`? If not, can you please file a bug upstream? We want avoid openssl if possible.

silveraid: I was able to compile it with libressl without any problem.

testing/softhsm/APKBUILD
silveraid: Removed.

testing/softhsm/APKBUILD
silveraid: Removed.

testing/softhsm/APKBUILD
silveraid: Removed.

silveraid:
All requested changes have been completed.

gbolo:
+1 on this new aport! Considering alpine is _"security-oriented"_, it would be nice to have some **pkcs11** related packages available!

testing/softhsm/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+        sh autogen.sh
+}
+
+build() {
+        cd "$builddir"
+        ./configure \
+                --prefix=/usr \
+                --sysconfdir=/etc \
+                --localstatedir=/var \
+                --disable-gost
```
fabled: Could you add `--disable-static` too? Otherwise the large and useless `.a` gets bundled in.

silveraid: Added, thanks for the suggestion.

silveraid:
All requested changes have been completed.

PR #2145, opened 2017-08-16
unmaintained/js: Fixed build. Backported patches from fedora dist

sergey-safarov:


sergey-safarov:
@Ikke could you review PR update.
If all ok, then I squash commit into one and add package back after commit 3ce3c4fd596debefbad77328a9b62a39eccf753c

PR #2148, opened 2017-08-16
community/namecoin: fix namecoin checksum

rgdoliveira:


jirutka:
@fcolista Have you investigated why is the checksum different, i.e. what has been changed between the original tarball and the last one?

PR #2156, opened 2017-08-20
community/perl-io-tee: update to 0.65

itoffshore:
community/perl-io-tee: update to `0.65`


nmeum:
Thanks for your patch! One minor suggestion: Please make sure you use `upgrade` instead of `update` in the commit message 😊

itoffshore:
ok will do ;o)

PR #2163, opened 2017-08-22
main/gnutls: add check

nmeum:


main/gnutls/APKBUILD
```diff
@@ -41,6 +42,11 @@ build() {
 	make || return 1
 }
 
+check() {
+	make -C "$builddir" check || \
+		cat "$builddir"/tests/test-suite.log
```
misery: I think the whole build should fail if the check fails

nmeum: yes, indeed. The cat invocation was only added to debug a build failure because the tests were [failing on the builders](http://build.alpinelinux.org/buildlogs/build-edge-x86_64/main/gnutls/gnutls-3.5.13-r1.log) and I wanted to find out why. However, sadly the build failure can't be reproduced on the travis ci runner.

PR #2166, opened 2017-08-22
main/consolekit: replaced by consolekit2

PureTryOut:
As discussed in the IRC on #alpine-devel, replacing the unmaintained consolekit by the maintained, drop-in, replacement: consolekit2 (and thus also moving consolekit2 from testing to main).

Tested it myself with the Mate desktop (which has a dependency on consolekit) and it's showing no issues, but this could probably use some testing by others.


main/consolekit2/APKBUILD
```diff
@@ -1,24 +1,28 @@
...
-pkgdesc="A framework for defining and tracking users, login sessions, and seats. "
+pkgrel=2
+pkgdesc="A framework for defining and tracking users, login sessions, and seats"
+provides=consolekit
+replaces=consolekit
 arch=all
 url="https://consolekit2.github.io/ConsoleKit2"
 license=GPL2
-depends="gettext glib zlib polkit eudev libdrm cgmanager libnih"
+depends="gettext glib zlib polkit eudev libdrm libnih"
```
ncopa: I don't think gettext, glib, zlib, libdrm and libnih needs to be in depends. They should be automatically pulled in.

main/consolekit2/APKBUILD
```diff
@@ -1,24 +1,28 @@
...
-		ac_disable_static.patch"
-subpackages="$pkgname-doc $pkgname-lang"
+		ac_disable_static.patch
+		0001-busybox-reboot-and-poweroff-support.patch"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
 
 prepare() {
 	cd "$srcdir"/$pkgbase-$pkgver
 	patch -p1 < "$srcdir"/ac_disable_static.patch
+	patch -p1 < "$srcdir"/0001-busybox-reboot-and-poweroff-support.patch
```
ncopa: we should run `default_prepare` instead of manually running patch.

main/consolekit/pam-foreground-compat.ck
```diff
@@ -1,17 +0,0 @@
-#!/bin/sh
```
ncopa: Are you 100% sure that this is no longer needed? see commit 85314dd187611074fc409c6cee98720751e499f6

ncopa:
Thank you for working on this! :heart:

main/consolekit2/APKBUILD
```diff
@@ -1,24 +1,31 @@
...
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
 pkgbase=ConsoleKit2
 pkgname=consolekit2
 pkgver=1.2.0
-pkgrel=1
-pkgdesc="A framework for defining and tracking users, login sessions, and seats. "
+pkgrel=2
+pkgdesc="A framework for defining and tracking users, login sessions, and seats"
+provides=consolekit
```
fabled: We might want to do `provides="consolekit=$pkgver"` so this gets automatically installed to satisfy `consolekit` dependency.

main/consolekit2/APKBUILD
```diff
@@ -1,24 +1,31 @@
...
+		ac_disable_static.patch
+		0001-busybox-reboot-and-poweroff-support.patch
+		pam-foreground-compat.ck"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
+builddir="$srcdir"/$pkgbase-$pkgver
 
 prepare() {
-	cd "$srcdir"/$pkgbase-$pkgver
-	patch -p1 < "$srcdir"/ac_disable_static.patch
+	default_prepare || return
```
fabled: The `|| return` is not needed as abuild runs functions with `set -e` now.

PR #2179, opened 2017-08-25
testing/vips: upgrade to 8.5.8 and modernize

rgdoliveira:


testing/vips/APKBUILD
```diff
@@ -1,21 +1,19 @@
...
 depends_dev="libjpeg-turbo-dev libexif-dev lcms2-dev"
 makedepends="$depends_dev fftw-dev giflib-dev glib-dev libpng-dev libwebp-dev
 	expat-dev orc-dev tiff-dev"
-install=""
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang $pkgname-tools"
-source="https://github.com/jcupitt/libvips/releases/download/v$pkgver/vips-$pkgver.tar.gz"
+source="https://github.com/jcupitt/libvips/releases/download/v$pkgver/$pkgname-$pkgver.tar.gz"
 
-builddir="$srcdir"/vips-$pkgver
+builddir="$srcdir/$pkgname-$pkgver"
```
ncopa: its a matter of taste, but I prefer to avoid using `$pkgname` in `builddir` and `source` because I think it provides no value and it makes it harder to rename the package if we ever do an `vips2` package or similar.

PR #2187, opened 2017-08-26
main/beecrypt: modernize and add check

rgdoliveira:


rgdoliveira:
Hmm, a test is failing on x86_64. I tried the build on ppc64le and all tests passed.

main/beecrypt/APKBUILD
```diff
@@ -34,14 +28,18 @@ build() {
...
 }
 
 package() {
-	cd "$_builddir"
-	make libaltdir=/usr/lib DESTDIR="$pkgdir" install || return 1
+	cd "$builddir"
+	make libaltdir=/usr/lib DESTDIR="$pkgdir" install
 }
 
 md5sums="8441c014170823f2dff97e33df55af1e  beecrypt-4.2.1.tar.gz"
```
tmpfile: You can also remove md5sums and sha256sums now that abuild ignores them.

ncopa:
It looks like it is fortify that catches something:
````
warning: Can't read pathname for load map: No error information.
Core was generated by `/home/ncopa/aports/main/beecrypt/src/beecrypt-4.2.1/tests/.libs/lt-testrsa'.
Program terminated with signal SIGILL, Illegal instruction.
#0  memcpy (__n=72, __os=0x72a36bd16328, __od=0x72a36bd162e8) at /usr/include/fortify/string.h:48
48                      __builtin_trap();
[Current thread is 1 (LWP 8493)]
(gdb) bt
#0  memcpy (__n=72, __os=0x72a36bd16328, __od=0x72a36bd162e8) at /usr/include/fortify/string.h:48
#1  mpndivmod (result=result@entry=0x72a36bd162e0, xsize=xsize@entry=9, xdata=xdata@entry=0x72a36bd16328, ysize=ysize@entry=4, ydata=0x3a05de82e40, 
    workspace=workspace@entry=0x72a36bd16370) at mp.c:1372
#2  0x000072a36c2a5ae6 in rsakpMake (kp=0x7a42bd0faff0, rgc=0x7a42bd0fafe0, bits=512) at rsakp.c:74
#3  0x000003a05bacbb8b in main () at testrsa.c:89
````

ncopa:
reported upstream: https://sourceforge.net/p/beecrypt/support-requests/19/

IIRC the only thing that has ever used this library was vserver which we haven't supported for a long time. It does not look like anything uses beecrypt anymore. I think we can move it to unmaintained.

PR #2188, opened 2017-08-26
main/squid: support for ext_session_acl

ssut:


fabled:
Applied with `pkgrel` bump.

PR #2189, opened 2017-08-26
testing/upmpdcli: add sysconfdir to move from /usr/etc to /etc

misery:


fabled:
Applied with `pkgrel` bump.

PR #2190, opened 2017-08-26
community/bareos: upgrade to 16.2.6, fixes

vakartel:
- many fixes in APKBUILD
- add perl as dependency, because some scripts use it
- add bat subpackage (qt administration tool)
- remove conf.d files because they are obsolete and incorrect
- rewrite init.d scripts
- fixes in (pre|post)-install scripts


vakartel:
to-do: 
- split database backends to subpackages
- write post-install info about initial config and database creation.

jirutka:
I've already asked you at least three times to split large changes/refactoring into multiple commits, each focused on single (type of) change. Such mega-commits like this one are very hard to review, not just for reviewer, but for everyone who wants to know what has been changed.

fcolista:
@vakartel I'm going to merge those patch. Please keep in mind what @jirutka said, this will help us in merging the patch quickier and to have right and clean comments on git tree's commits.
Thanks.

PR #2199, opened 2017-08-27
community/go: upgrade to 1.9

nmeum:
Also remove cover:

> For Go releases 1.5 and later, this tool lives in the standard
repository. The code here is not maintained.

See: https://godoc.org/golang.org/x/tools/cmd/cover

### DIFF
--- a/community/go/APKBUILD
+++ b/community/go/APKBUILD
@@ -1,9 +1,9 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Eivind Uggedal <eivind@uggedal.com>
 pkgname=go
-pkgver=1.8.3
+pkgver=1.9
 # This should be the latest commit on the corresponding release branch
-_toolsver="5682db0e919ed9cfc6f52ac32e170511a106eb3b"
+_toolsver="5d2fd3ccab986d52112bf301d47a819783339d0e"
 pkgrel=0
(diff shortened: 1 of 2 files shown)


ncopa:
cover is not removed. It is still built and shipped.

````
ncopa-edge-x86_64:~/aports/community/go$ find pkg -name cover
pkg/go-cross-openbsd/usr/lib/go/pkg/tool/openbsd_amd64/cover
pkg/go-cross-openbsd/usr/lib/go/pkg/tool/openbsd_386/cover
pkg/go-cross-darwin/usr/lib/go/pkg/tool/darwin_amd64/cover
pkg/go-cross-darwin/usr/lib/go/pkg/tool/darwin_386/cover
pkg/go/usr/lib/go/src/cmd/cover
pkg/go/usr/lib/go/pkg/tool/linux_amd64/cover
pkg/go-cross-freebsd/usr/lib/go/pkg/tool/freebsd_amd64/cover
pkg/go-cross-freebsd/usr/lib/go/pkg/tool/freebsd_386/cover
````

nmeum:
Well cover is not entirely removed since it lives in the standard library now. Only the version from go tools was removed since it is no longer maintained, the change made in this commit should therefore result in the removal of `/usr/bin/cover`. According to the checkpkg output on the travis CI runner that binary is indeed removed.

ncopa:
ok. thanks or the clarification

PR #2200, opened 2017-08-27
main/dovecot: upgrade to 2.2.32, pidgeonhole-plugin upgrade to 0.4.20, fixes

vakartel:
- cosmetic fixes in init-script

- move /usr/libexec/dovecot to /usr/lib/dovecot

- add pidgeonhole (sieve) ldap storage plugin as subpackage

- move all sieve-related stuff to pidgeonhole-plugin subpackage,
  so now dovecot package have no sieve-dependent libs and binaries.

- move all ldap-related libs to dovecot-ldap subpackage,
  so now dovecot package doesn't depends on libldap.


vakartel:
@fabled 
I saw in many other packages binaries from /usr/libexec moved to /usr/lib/<pkgname>. So I made the same.

PR #2205, opened 2017-08-28
main/confuse: upgrade to 3.2.1 and modernize

rgdoliveira:


ncopa:
it looks like this breaks ABI?

ncopa:
It was the 3.2 release that broke ABI but they didnt bump SO version. https://github.com/martinh/libconfuse/issues/101

We need to rebuild everything using confuse.

PR #2222, opened 2017-08-31
main/jq: add check function

rgdoliveira:


fabled:
Seems all test cases fail. Any idea why?

rgdoliveira:
Not yet, I need to take a look. 
In ppc64le all tests passed, maybe missing some dependency?

fabled:
```

FAIL: tests/mantest
===================

mktemp: Invalid argument
Your OS does not support mktemp(1) -d

FAIL: tests/jqtest
==================

mktemp: Invalid argument
Your OS does not support mktemp(1) -d

FAIL: tests/onigtest
====================

mktemp: Invalid argument
Your OS does not support mktemp(1) -d

FAIL: tests/shtest
==================

mktemp: Invalid argument
Your OS does not support mktemp(1) -d
```
Maybe needs `checkdepends=coreutils`. But it's weird since busybox mktemp does support `-d`. Looking at `tests/setup` it is calling `mktemp -d -t jqXXXXX`. This is missing one `X` (there should be six of them). Perhaps file upstream bug and ask fixing this?

rgdoliveira:
@fabled you are right. 
We need coreutils in checkdepends to test work. I double-checked it on ppc64le and I had coreutils installed (that's why it worked)

fabled:
We would prefer fixing `make check`. The following patch fixes it, and should be submitted upstream.

```
diff -ru jq-1.5.orig/tests/setup jq-1.5/tests/setup
--- jq-1.5.orig/tests/setup     2015-08-16 04:37:35.000000000 +0000
+++ jq-1.5/tests/setup  2017-09-07 07:07:55.035838903 +0000
@@ -29,7 +29,7 @@
     fi
 }
 trap clean EXIT
-d=`mktemp -d -t jqXXXXX || true`
+d=`mktemp -d -t jqXXXXXX || true`
 if [ -z "$d" ]; then
     echo "Your OS does not support mktemp(1) -d" 1>&2
     exit 1
```

rgdoliveira:
@fabled found a patch already merged upstream that fix our issue. Thanks for reviewing it.

ncopa:
we can just backport https://github.com/stedolan/jq/commit/b7a219a91dc1094022478dd9bc0f60275367ce25

rgdoliveira:
@ncopa that's what I did, thanks!

PR #2230, opened 2017-09-02
main/libsigc++: modernize APKBUILD and add check

rgdoliveira:


fabled:
Seems 2.99.x is the unstable tree leading ultimately to 3.0 release. Current policy is to have stable release only unless there's compelling reasons otherwise.

rgdoliveira:
Thanks for reviewing it @fabled. Yeah, it is an unstable version, my mistake.
I removed the upgrade from the patch.

PR #2244, opened 2017-09-04
main/libdrm: upgrade to 2.4.83 and add check

tmpfile:


tmpfile:
Without cunit-dev in the makedepends one test is ignored.
> configure: WARNING: Could not find cunit library. Disabling amdgpu tests
But cunit is in community and libdrm in main so libdrm can't depend on cunit.
Solution is to move cunit to main or libdrm to community.

ncopa:
I moved cunit to main but it still fails:

````
  CC       amdgpu_test-vcn_tests.o
uvd_enc_tests.c:39:20: fatal error: uve_ib.h: No such file or directory
 #include "uve_ib.h"
                    ^
compilation terminated.
````

tmpfile:
I found a fix from upstream.
https://cgit.freedesktop.org/mesa/drm/commit/?id=05a830d382382905d87b55af90aa109fb76e4c16

But it touches Makefile.am so we'd need to add automake and autoconf to makedepends, plus call them later to regenerate config files.

So I think the best for now is to comment the check() function and just uncomment when 2.4.84 is released.

tmpfile:
Upstream had a fix but it touches Makefile.am so we need now automake and autoconf in makedepends, and also need to run aclocal, autoconf and automake and prepare to regenerate config files.
I added comments to remind us that when libdrm 2.4.84 and the fix is merged we can go without those extra steps.

tmpfile:
Upstream patch didn't work either.
I give up for now, when 2.4.84 is released I come back and upgrade libdrm + add check().
Let it rest until then.

ncopa:
Upstream patch will make sure that the missing header gets included in the generated tarball. Since we build from the generated release tarball, the file is missing and we need a patch that adds the file.

https://cgit.freedesktop.org/mesa/drm/commit/tests/amdgpu/uve_ib.h?id=9d133dd08720d80dfc8ce098bf0972ebd05c454d

tmpfile:
Good job, thanks.

PR #2245, opened 2017-09-04
main/util-linux: upgrade to 2.30.1

tmpfile:


main/util-linux/APKBUILD
```diff
@@ -2,26 +2,23 @@
...
 esac
 
-pkgrel=2
+pkgrel=0
 pkgdesc="Random collection of Linux utilities"
 url="http://git.kernel.org/cgit/utils/util-linux/util-linux.git"
 arch="all"
 license="GPL2 GPL2+ LGPL2+ BSD Public Domain"
 depends="findmnt"
-# use GNU sed til bb sed is fixed. Also GNU tar is needed
```
fabled: I wonder what the sed/tar issues have been. Any idea when this has been fixed?

main/util-linux/APKBUILD
```diff
@@ -82,68 +77,59 @@ dev() {
 }
 
 bashcomp() {
-	depends=""
```
fabled: This should not be removed. It is necessary to reset `depends` manually in order to not inherit the parent package's manual dependencies. Same applies for other split functions and `replaces`.

main/util-linux/APKBUILD
tmpfile: No idea what the problem was or when it has been fixed.
The comment about sed has been there since 2009.
https://github.com/alpinelinux/aports/commit/c53dca816490c4c81e0e2c5d4adc5c3a0701ca8d
And the comment about tar was added in 2012.
https://github.com/alpinelinux/aports/commit/ad89f93ea4166fc1716014b7db728867d2140442

As those were very old issues, I thought they should be resolved by now, so I tried to compile with BusyBox utils and it worked.

main/util-linux/APKBUILD
tmpfile: I fixed it.

PR #2248, opened 2017-09-05
main/bc: upgrade to 1.07.1

tmpfile:


tmpfile:
bc now depends on ed
The solution is to move ed from testing to main

PR #2265, opened 2017-09-06
main/hunspell: upgrade to 1.6.2

tmpfile:


main/hunspell/APKBUILD
```diff
@@ -1,21 +1,18 @@
...
 makedepends="gettext-dev ncurses-dev readline-dev autoconf automake libtool"
-source="$pkgname-$pkgver.tar.gz::https://github.com/hunspell/hunspell/archive/v$pkgver.tar.gz
-	"
+source="$pkgname-$pkgver.tar.gz::https://github.com/hunspell/hunspell/archive/v$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
 
-builddir="$srcdir"/$pkgname-$pkgver
 prepare() {
 	cd "$builddir"
-	default_prepare
```
fabled: Any particular reason why you are removing this?

tmpfile: I tested without it and it builds fine. I think the smaller an APKBUILD the better.
But if it's needed I can add it back.

fabled: Yes, it builds. But if a patch is ever added to `$sources` it will not be applied without this. Generally we've wanted to keep `default_prepare` so we don't get patches left out accidentally if they are added.

tmpfile: Good to know.
I fixed it.

PR #2267, opened 2017-09-06
main/rest: modernize APKBUILD and add check

rgdoliveira:


fabled:
```
make[3]: Entering directory '/home/tteras/aports/main/rest/src/rest-0.8.0/tests'
PASS: oauth2
../build/test-driver: line 107: 11415 Trace/breakpoint trap   "$@" > $log_file 2>&1
XFAIL: xml
PASS: proxy
PASS: proxy-continuous
PASS: custom-serialize
PASS: lastfm
PASS: oauth-async
PASS: flickr
PASS: threaded
PASS: oauth
============================================================================
Testsuite summary for rest 0.8.0
============================================================================
# TOTAL: 10
# PASS:  9
# SKIP:  0
# XFAIL: 1
# FAIL:  0
# XPASS: 0
# ERROR: 0
```
Packaging still is success, so sounds like `make check` does not return an error. I'll probably apply this, but would be good to check what's the failing use case, and why it did not abort build.

PR #2272, opened 2017-09-06
testing/nitro: upgrade to 2.7dev-2

chambbj:


chambbj:
Thanks for catching this @fabled. It is now fixed.

PR #2277, opened 2017-09-07
main/tdb: upgrade to 1.3.15, modernize and add check

rgdoliveira:
Also add doc tdb package

### DIFF
--- a/main/tdb/APKBUILD
+++ b/main/tdb/APKBUILD
@@ -1,32 +1,21 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=tdb
-pkgver=1.3.14
+pkgver=1.3.15
 pkgrel=0
 pkgdesc="The tdb library"
 url="http://tdb.samba.org/"
 arch="all"
 license="LGPLv3+"
-depends=""
 depends_dev="python2"
-makedepends="$depends_dev python2-dev"
-install=""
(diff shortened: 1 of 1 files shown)


main/tdb/APKBUILD
```diff
@@ -1,32 +1,21 @@
...
 pkgdesc="The tdb library"
 url="http://tdb.samba.org/"
 arch="all"
 license="LGPLv3+"
-depends=""
 depends_dev="python2"
 makedepends="$depends_dev python2-dev"
-install=""
-subpackages="$pkgname-dev py-$pkgname:_py $pkgname-libs"
+subpackages="$pkgname-dev py-$pkgname:_py $pkgname-libs $pkgname-doc"
```
fabled: Adding `-doc` fails the build. There's no files the `default_doc` split function would move to doc package. Why was this added?

fabled: configure says:
```
Checking for program xsltproc                                                     : not found 
Checking for stylesheet http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl : not found 
A local copy of the docbook.xsl wasn't found on your system consider installing package like docbook-xsl
```
Probably needs that as `makedepends=docbook-xls` added if documentation is wanted.

rgdoliveira:
Fixed. I added the docbook-xsl in makedepends and now the -doc package is created fine.
Thanks for the review

PR #2284, opened 2017-09-08
testing/virt-what: fix bashisms

HRio:
Add Remove-bashisms.patch from Debian

Drop: skip-ldoms-test-if-not-sparc64.patch not needed after bashisms fix


HRio:
Note: the last change depends on the main/checkbashisms fix to work.

HRio:
@ncopa can you as maintainer of main/checkbashisms check that part of this PR? Thanks.

PR #2285, opened 2017-09-08
testing/nitro: bump pkgrel

chambbj:
The previous commit updated _pkgver, but not pkgver, hence no updates to the
package in the testing repository. This commit bumps the pkgrel to pick up the
change and trigger builds.

### DIFF
--- a/testing/nitro/APKBUILD
+++ b/testing/nitro/APKBUILD
@@ -1,9 +1,9 @@
 # Contributor: Bradley J Chambers <brad.chambers@gmail.com>
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
 pkgname=nitro
-pkgver=2.7
-_pkgver="$pkgver"dev-2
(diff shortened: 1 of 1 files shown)


ncopa:
I wonder if we could translate the `dev-N` to something apk understands. For example:

````sh
pkgver=2.7_beta2
_ver=${pkgver/_beta/dev-}
````

That way we dont need to touch the pkgrel which is supposed to only have changes of the APKBUILD itself and not include changes for the upstream package.

Currently, if you do `apk version nitro` you'll get `2.7-r1` and there is no indication that it matches the `2` in `dev-2`.

What do you think?

chambbj:
I'd be fine with that. I think I just remembered that apk didn't like `dev` and didn't consider that there were alternatives like `beta`. I'll give it a go.

chambbj:
@ncopa I'm good with your suggested approach. Changes made.

PR #2286, opened 2017-09-08
travis: changes to allow ppc64le support

rgdoliveira:
Just download the APK tools if the install-alpine
script is running in a system that doesn't have APK
installed.
Also add pub key from Alpine ppc64le VM that is in
alpinebox ppc64le machine.


rgdoliveira:
I sent an email in alpine-devel that explains why this PR.

rgdoliveira:
@jirutka thanks for reviewing it. Yeah, I agree with you.
I just edited the PR to only download the APK if it is not available in the system.

jirutka:
Merged in ae05c20cbd159d5b6de295364d05a9f74a65d8d0 and dafc33a68dd51c96974df9a7358d581e9f5ccab5. Thanks!

PR #2300, opened 2017-09-11
new aport: sslh (Applicative Protocol Multiplexer)

castorinop:


tcely:
This should be closed as there is already a `testing/sslh` that is reasonable now.

andypost:
There's already 1.19c version in testing

PR #2302, opened 2017-09-13
main/mariadb: upgrade to 10.2.13, modernize

TBK:
After a lot of fiddeling and 3 hours of compiling I think it is in a state where it is ready for some feedback.

Things that I question:
- Should the cmake flags be updated?
- Related to the previous question - the official deb build uses quite a few more makedepends. Should more be added? (https://github.com/MariaDB/server/blob/10.2/debian/control)
- All of the symlinks for the mariadb-client-libs are missing. I am not sure how to fix that:
```
libmariadb.so -> libmariadb.so.3
libmariadb.so.3
libmysqlclient_r.so -> libmariadb.so.3
libmysqlclient.so -> libmariadb.so.3
```
- Should there be added a check function to run the test suite?

### DIFF
--- a/main/mariadb/APKBUILD
+++ b/main/mariadb/APKBUILD
@@ -2,27 +2,34 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
+# Contributor: TBK <alpine@jjtc.eu>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=mariadb
-pkgver=10.1.28
-pkgrel=1
+pkgver=10.2.13
+pkgrel=0
 pkgdesc="A fast SQL database server"
 url="http://www.mariadb.org"
 pkgusers="mysql"
 pkggroups="mysql"
 arch="all"
-license="GPL"
+license="GPL-2.0"
 depends="$pkgname-common"
 depends_dev="libressl-dev zlib-dev"
-makedepends="cmake libressl-dev zlib-dev readline-dev libaio-dev ncurses-dev
-	linux-headers"
+makedepends="bison cmake curl-dev libaio-dev libarchive-dev libevent-dev libressl-dev
+	libxml2-dev ncurses-dev pcre-dev readline-dev zlib-dev xz-dev linux-headers"
 install="$pkgname.pre-install"
+subpackages="$pkgname-doc $pkgname-dev $pkgname-common
+	$pkgname-client-libs:_client_libs $pkgname-libs
+	$pkgname-client $pkgname-bench $pkgname-test:mytest
+	mysql mysql-client:_compat_client mysql-bench:_compat_bench"
 source="https://downloads.mariadb.org/interstitial/mariadb-$pkgver/source/mariadb-$pkgver.tar.gz
 	$pkgname.initd
 	fix-mysql-install-db-path.patch
 	fix-ucontext-check.patch
 	ppc-remove-glibc-dep.patch
+	pcre.cmake.patch
 	"
+builddir="$srcdir/$pkgname-$pkgver"
 
 # secfixes:
 #   10.1.22-r0:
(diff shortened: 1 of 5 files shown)


main/mariadb/APKBUILD
```diff
@@ -2,9 +2,10 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
+# Contributor: TBK <alpine@jjtc.eu>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=mariadb
-pkgver=10.1.26
+pkgver=10.2.8
```
ncopa: Is 10.2.x ABI compatible with 10.1.x?

main/mariadb/APKBUILD
```diff
@@ -24,23 +25,6 @@ source="https://downloads.mariadb.org/interstitial/mariadb-$pkgver/source/mariad
...
-#     - CVE-2017-3243
-#     - CVE-2017-3244
-#     - CVE-2017-3257
-#     - CVE-2017-3258
-#     - CVE-2017-3265
-#     - CVE-2017-3291
-#     - CVE-2017-3312
-#     - CVE-2017-3317
-#     - CVE-2017-3318
-
```
ncopa: please keep the secfixes history

ncopa:
> Should the cmake flags be updated?

Why? are the current flags broken or missing anything?

> Related to the previous question - the official deb build uses quite a few more makedepends. Should more be added? (https://github.com/MariaDB/server/blob/10.2/debian/control)

Why should we add those if nobody asked for those features?

> All of the symlinks for the mariadb-client-libs are missing. I am not sure how to fix that:

They are not missing. They are provided in the -dev subpackage. the client works just fine without those symlinks.

> Should there be added a check function to run the test suite?

That would be good yes.

main/mariadb/APKBUILD
TBK: Roger.

main/mariadb/APKBUILD
TBK: From what I have been able to read it is almost. The differences are revolving around libmysqlclient.so which is now libmariadb.so - https://lists.launchpad.net/maria-discuss/msg04574.html

I will test it later with abi-compliance-checker.

main/mariadb/APKBUILD
```diff
@@ -77,43 +81,39 @@ build() {
...
-		-DWITH_EXTRA_CHARSETS=complex \
-		-DWITH_EMBEDDED_SERVER=ON \
-		-DWITH_ARCHIVE_STORAGE_ENGINE=1 \
-		-DWITH_BLACKHOLE_STORAGE_ENGINE=1 \
-		-DWITH_INNOBASE_STORAGE_ENGINE=1 \
-		-DWITH_PARTITION_STORAGE_ENGINE=1 \
-		-DPLUGIN_TOKUDB=NO \
-		-DWITHOUT_EXAMPLE_STORAGE_ENGINE=1 \
-		-DWITHOUT_FEDERATED_STORAGE_ENGINE=1 \
-		-DWITHOUT_PBXT_STORAGE_ENGINE=1 \
```
jirutka: Why have you removed these flags…?

TBK: EXAMPLE, PBXT and FEDERATED have been droped quite white a while ago.
https://mariadb.com/kb/en/library/example-storage-engine/
https://mariadb.com/kb/en/library/pbxt-storage-engine/
https://mariadb.com/kb/en/library/federated-storage-engine/

Plus the -DWITHOUT_xxxxxx_STORAGE_ENGINE flag format is no longer in use, they all start with -DPLUGIN_xxxxx=NO

TBK:
With the release of 10.2.13 in 3 days the failing test should be resolved - https://jira.mariadb.org/projects/MDEV/versions/22910

TBK:
@ncopa @jirutka The latest version (10.2.13) fixed the issue with the failing "mf_iocache" test. It build 100% sucessfully on my build box, but not on Travis. 

```
      Start 60: dbug
60/60 Test #60: dbug .............................***Failed    0.00 sec
98% tests passed, 1 tests failed out of 60
Total Test time (real) =  63.72 sec
The following tests FAILED:
	 60 - dbug (Failed)
Errors while running CTest
```

main/mariadb/APKBUILD
```diff
@@ -136,11 +152,9 @@ _client_libs() {
...
 	depends="mariadb-common"
 	mkdir -p "$subpkgdir"/usr/lib \
-		"$subpkgdir"/usr/share/mysql \
-		|| return 1
-	mv "$pkgdir"/usr/lib/libmysqlclient.so* \
-		"$subpkgdir"/usr/lib/ \
-		|| return 1
+		"$subpkgdir"/usr/share/mysql
+	mv "$pkgdir"/usr/lib/libmariadb.so* \
+		"$subpkgdir"/usr/lib/
```
ncopa: are there no `libmysqlclient.so.*` anymore? If not, this will break everyting thats linked to libmysqlclient

TBK: It has been replaced by Maria's own lib. I am pretty sure it makes a symlink for backwards compatibility.

Been a while since I worked with this PR, I still need to figure out the dbug problem.

PR #2312, opened 2017-09-15
community/go: don't build cross-compilers

Xe:
Cross-compiler subpackages like this have not been needed [since Go 1.5](https://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5).


nmeum:
The `pkgrel` needs to be increased in order to trigger a rebuild on the build server.

nmeum:
Applied! Thanks for the patch, I bumped the pkgrel for you and removed a for-loop which was no longer needed

PR #2314, opened 2017-09-15
testing/pdal: new aport

chambbj:
https://www.pdal.io
PDAL is a C++ BSD library for translating and manipulating point cloud data.


testing/libspatialite/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: Bradley J Chambers <brad.chambers@gmail.com>
+# Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+pkgname=libspatialite
+pkgver=4.3.0
+_pkgver="$pkgver"a
```
ncopa: Can we use `pkgver="4.3.0a"` instead of using `_pkgver`?

testing/libspatialite/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgrel=0
+pkgdesc="SpatiaLite is an open source library intended to extend the SQLite core to support fully fledged Spatial SQL capabilities."
```
ncopa: Can we make the pkgdesc shorter? (max ~70 chars?)

testing/libspatialite/APKBUILD
jirutka: What does "a" mean in this case - alpha or like in openssl?

chambbj: I'm not intimately familiar with the project or the naming convention, but I do not think it's `alpha` as they have explicitly used `beta` before. Regardless, if `pkgver="4.3.0a"` works and is acceptable, I'll just do that. I was likely thinking that `pkgver` didn't accept characters, based on the issue I ran into with `testing/nitro` with `dev` being part of the version number.

PR #2318, opened 2017-09-18
September Adélie merge: User applications

awilfox:
These modernise and add testing where available to user-facing applications.  It also adds two -lang packages, which I will be happy to remove if they are not desired.

### DIFF
--- a/main/aspell/APKBUILD
+++ b/main/aspell/APKBUILD
@@ -2,23 +2,24 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 pkgname=aspell
 pkgver=0.60.6.1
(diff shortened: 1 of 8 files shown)


awilfox:
It looks like the Travis builder doesn't have the version of abuild that has the OpenRC split packages, which are the only build failures.  Can Travis' abuild be bumped?

jirutka:
Travis installs Alpine from scratch with every build, so it’s quite not possible to update any package… it has always the latest packages available on the mirror.

https://github.com/alpinelinux/aports/tree/master/.travis/

ncopa:
I think this is merged. Thanks!

PR #2319, opened 2017-09-18
September Adélie merge: Desktop libraries

awilfox:
This modernises and add test suites to four libraries used for X11 desktops.

### DIFF
--- a/main/libgudev/APKBUILD
+++ b/main/libgudev/APKBUILD
@@ -1,11 +1,12 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=libgudev
 pkgver=230
-pkgrel=1
+pkgrel=2
 pkgdesc="GObject-based wrapper library for libudev"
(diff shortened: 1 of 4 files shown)


tmpfile:
You forgot to run `abuild checksum` during the modernization.
That would ensure packages with old checksums (md5 and sha256) are updated to display only the new hashes (sha512).

Actually... could we run `abuild checksum` on every single package?
That would update all APKBUILD with old hashes + show us if any packages have wrong hashes.
And there's no side-effect, because APKBUILD with correct checksums will remain that way.

Opinions?

ncopa:
I dont think we need to stress with the checksums. We can update those when convenient.

ncopa:
i updated the checksums and pushed.

awilfox:
@tmpfile actually there have been a few checksum mismatches for me, especially with some GNOME packages, that turned out to be their FTP server dying in the middle of the transfer.  So I am very glad that the checksums are there.  I wouldn't want to blanket-rewrite them, because then it could be corrupted...

PR #2320, opened 2017-09-18
September Adélie merge: Development packages

awilfox:
This add test information and modernises 16 development-related packages, and also bumps a few to use Python 3 when building instead of Python 2 where available.  This does not actually change any Python bindings or py-* packages, which may be pushed later if desired (we have about 10 of these).

### DIFF
--- a/main/cmake/APKBUILD
+++ b/main/cmake/APKBUILD
@@ -2,13 +2,15 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=cmake
 pkgver=3.9.1
-pkgrel=1
+pkgrel=2
 pkgdesc="CMake is a cross-platform open-source make system"
 url="http://www.cmake.org"
 arch="all"
 license="CMake"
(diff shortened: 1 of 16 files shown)


awilfox:
* gobject-introspection cannot be migrated to Python 3 yet because it gets confused when Python 2 and 3 are both present on the build system, and its deps still pull in Python 2.
* CMake needed `file` which I didn't realise because my Alpine test box already had that installed for development purposes.
* Likewise, my Alpine test box had `diffutils` already so I didn't realise libidn needed it.

awilfox:
It looks like libidn isn't passing because of a [missing patch in main/musl](http://git.musl-libc.org/cgit/musl/commit/?id=947d330f68c49680dcc54439f56da2a297228962).  We ship this in Adélie already; would a PR for that be welcome here?  All builds would pass then.

kaniini:
Yes.

ncopa:
merged. thanks!

PR #2321, opened 2017-09-18
September Adélie merge: Networking packages

awilfox:
Various fixes to network-related packages: add tests where available, support IPv6, and two package version bumps: main/apr and main/apr-util.

### DIFF
--- a/main/apr/APKBUILD
+++ b/main/apr/APKBUILD
@@ -11,29 +11,29 @@ depends_dev="util-linux-dev bash"
 makedepends="$depends_dev"
 subpackages="$pkgname-dev"
 source="http://www.apache.org/dist/$pkgname/$pkgname-$pkgver.tar.bz2
+	apr-1.6.2-dont-test-dlclose.patch
 	"
 
-_builddir="$srcdir"/$pkgname-$pkgver
-prepare() {
-	cd "$_builddir"
-}
-
 build() {
-	cd "$_builddir"
+	cd "$builddir"
 	./configure \
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
 		--datadir=/usr/share \
 		--enable-nonportable-atomics \
-		--with-devrandom=/dev/urandom \
-		|| return 1
-	make || return 1
+		--with-devrandom=/dev/urandom
+	make
+}
+
+check() {
+	cd "$builddir"
+	make check
 }
 
 package() {
-	cd "$_builddir"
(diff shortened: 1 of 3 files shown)


awilfox:
Removed the bind commit because it was causing Travis to fail due to large log size.  All it did was add options="!check" anyway, so there is nothing much lost.

fabled:
Cherry-picked some of these. apr build failed, probably concurrency issues (and it's a major update, so needs rebuilding dependencies?).
One mariadb test failed with:
```
      Start 21: mf_iocache
21/59 Test #21: mf_iocache .......................***Failed    0.00 sec
```

awilfox:
apr builds fine with -j8 on my Xeon and builds fine with -j4 on my PowerPC.  I didn't really think about bumping deps for it, really sorry... I know subversion and apache will need bumping.  I can find the rest and make a list tomorrow, and make a PR that bumps them all, if that helps.

Re mariadb: That's an odd test failure.  What architecture was it?

fabled:
Yes, please create separate APR update + dependency rebuild PR. The concurrent build failure happened on ou 24-core builder.

Mariadb test failure is the same 24-core x86_64 build box.

ncopa:
Can you please rebase this PR so we see whats left?
thanks!

main/apr-util/APKBUILD
```diff
@@ -30,16 +30,21 @@ build() {
...
-		--without-gdbm \
-		|| return 1
-	make || return 1
+		--without-gdbm
+	make
+}
+
+check() {
+	cd "$builddir"
+	# testxlate fails because UTF-7 is unsupported
```
jirutka: This is not very good. Can you somehow easily disable/remove just the failing tests?

awilfox:
So the MariaDB test failure is actually because Adélie uses OpenSSL and Alpine uses LibreSSL.  LibreSSL is not supported in MariaDB [until 10.2.6 or 10.3](https://github.com/MariaDB/server/commit/ccca4f43c92916c347210a7f9a8126f2aa3f6c31).

Any one of the following options must be taken:
* The patch needs to be backported to 10.1
* The port needs to be upgraded to 10.2.6+ or 10.3.0+
* The port needs to use OpenSSL instead of LibreSSL.

ncopa:
fixed with 5fdfcb1f1d2bacb0300539d319d35d3cb0032033, 6556a505d52c020d0bf835a0f1e12afceb303674 and 9f6c5c776a20c9ff350aa839795d1c2bf65fb0ac

PR #2322, opened 2017-09-18
September Adélie merge: Security packages

awilfox:
Add test suite to five packages.  Use system PCRE2 to avoid CVE-2017-11164 in main/fish.  Support cross-compilation in main/libssh2, main/sudo, and main/shadow.


awilfox:
* GnuTLS requires netstat to work, which it doesn't on hardened kernels and I have no idea how to fix it so I removed that commit entirely.

* shadow was committed improperly so I recommitted it, and it should work now.

ncopa:
fish needs bump pkgrel

main/openssh/APKBUILD
```diff
@@ -9,7 +9,7 @@ pkgdesc="Port of OpenBSD's free SSH release"
 url="http://www.openssh.org/portable.html"
 arch="all"
 license="as-is"
-options="suid"
+options="suid !check"
```
jirutka: If the provided tests are stupidly broken, then please add at least very basic check that would just run `sshd --version` or something like that, to at least verify that the binary exists and is not completely broken.

Also it’d be nice to add some brief detail into the commit message about how are the tests broken.

community/shadow/APKBUILD
```diff
@@ -3,18 +3,20 @@
...
 pkgdesc="PAM-using login and passwd utilities (usermod, useradd, ...)"
 url="http://pkg-shadow.alioth.debian.org/"
 arch="all"
 license="GPL"
 depends=""
-makedepends="linux-pam-dev"
-subpackages="$pkgname-doc $pkgname-dbg $pkgname-uidmap"
+makedepends_host="linux-pam-dev"
+subpackages="$pkgname-doc $pkgname-dbg"
+[ ${CBUILD} == ${CHOST} ] && subpackages="$subpackages $pkgname-uidmap"
```
jirutka: `==` is a (unnecessary) bashism, POSIX Command Language specifies only `=` for comparison.

Also it’s better to use `||` in such construct. Some shells don’t recognize such construct as a special case when resulting non-zero status does not mean a failure (when using `set -e`).

And also it’s better to always quote (outside defined) variables, to be defensive against errors.

```sh
[ "$CBUILD" != "$CHOST" ] || subpackages="$subpackages $pkgname-uidmap"
```

community/shadow/APKBUILD
```diff
@@ -31,6 +33,7 @@ builddir="$srcdir/shadow-$pkgver"
 
 build() {
 	cd "$builddir"
+	[ ${CBUILD} != ${CHOST} ] && extra_conf="--disable-subordinate-ids"
```
jirutka: The same as above.

jirutka: The same as above about quoting and condition.

Variable `extra_conf` will be undefined, or defined outside the script with some arbitrary content (!), when the condition is evaluated as false. That’s quite error prone. Moreover, you unnecessarily pollute global space of variables.

Please always declare variables used only in a single function as `local`, i.e. `local extra_conf` or `local extra_conf=""`.

community/shadow/APKBUILD
```diff
@@ -62,6 +66,10 @@ check() {
 package() {
 	cd "$builddir"
 
+	if [ ${CBUILD} != ${CHOST} ]; then
+		mkdir -p "$pkgdir"/usr/bin
+		touch "$pkgdir"/usr/bin/new{u,g}idmap
```
jirutka: What’s the purpose of this? Please add a comment.

main/krb5-conf/APKBUILD
```diff
@@ -13,6 +13,7 @@ arch="noarch"
 license="MIT"
 depends=""
 makedepends=""
+options="!check"
```
jirutka: Why? Upstream does not provide any tests? Please always add a short comment after the `!check` why it’s disabled.

jirutka:
@awilfox gentle reminder

awilfox:
The community/shadow cross-compilation fix is probably not worth including.  It is very hacky and needs more upstream support before it should be considered for inclusion IMO.

I haven't had enough time to see if OpenSSH test suite is better in the new release so I am not adding the !check to it in this edition of the PR.

Everything else should be good.

awilfox:
We are no longer contributing these kinds of fixes to Alpine, as it takes too much of our time away from development.  This (and any other still-open PR) may be closed.

PR #2326, opened 2017-09-18
Septermber Adélie merge: X11 packages

awilfox:
These packages are fonts/software for X11, modernised and with test suites where available.  I also am going to maintain main/twm for Alpine.


ncopa:
merged. thanks!

PR #2327, opened 2017-09-18
September Adélie merge: X11 libraries

awilfox:
Modernise and fix dependencies on a lot of X11 libraries.  Most of them need util-macros to be able to properly mark things like font encodings dirs and such so they are all bumped when util-macros is added.  Also, libepoxy has said upstream they are deprecating Python 2 support so that package now uses Python 3.


ncopa:
merged. thanks!

PR #2328, opened 2017-09-18
September Adélie merge: X11 protocol packages

awilfox:
* Some packages were marked 'all' instead of 'noarch'.  All X11 proto packages are noarch, as abuild liked to warn me when I was doing the modernisations.
* Some packages were missing -doc but still installing to /usr/share/doc, so I added -doc subpkgs to those that did that.
* A few have tests, but most don't.  I added test suites where I could and wrote options="!check" where not.


ncopa:
this is merged. thanks!

PR #2330, opened 2017-09-19
testing/msitools: new aport

misery:
Also fix libgcab as msitools requires a valid .pc file

### DIFF
--- a/testing/msitools/APKBUILD
+++ b/testing/msitools/APKBUILD
@@ -1,0 +1,46 @@
+# Maintainer: André Klitzing <aklitzing@gmail.com>
+pkgname=msitools
+pkgver=0.97
+pkgrel=0
+pkgdesc="Set of programs to inspect and build Windows Installer (.MSI) files"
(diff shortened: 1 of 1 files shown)


misery:
Well.... ci malfunction? It does not install the previous build libgcab 0.7-r1

(32/59) Installing libgcab (0.7-r0)
(33/59) Installing libgcab-dev (0.7-r0)

jirutka:
Hm, this is a second case since yesterday where it looks like mirror nl.alpinelinux.org is outdated. I'll take a look.

jirutka:
Hm, no, this is a different issue, you’ve updated libgcap in this PR. It really should pick aport built in the previous step, but does not for some reason.

misery:
Should I split the PR?

ncopa:
I have merged the libgcab already so you can probably just rebase this PR and `git push -f`. I think libtool is needed in the makedeps and one of the tests failed in my env.

PR #2331, opened 2017-09-19
testing/lttng-ust: Update lttng-ust to 2.10.0

am11:
The patch `testing/lttng-ust/musl-has-no-dlmopen.patch` is not needed
anymore.

Also added test suite.


am11:
@ncopa, lttng-ust from this PR and even the existing/previous one has dependency on `userspace-rcu-dev`. The `edge/main` has `v0.10.0-r0` of `userspace-rcu-dev`, which is compatible with both the current and previous versions of lttng-ust; while release 3.6 branch has `v0.9.3-r1` which is incompatible. Would you consider adding lttng-ust (and perhaps lttng-tools as well) to edge/main and at some point to release?

This is one of the dependency of CoreCLR's event tracing module, which requires developer to opt into testing package repo (and edge/main for `userspace-rcu-dev@0.10.0-r0`). With https://github.com/dotnet/core-setup/pull/3212 merged, the Alpine distro builds will soon be able to build via https://github.com/dotnet/source-build.

If that's acceptable, I can send a separate PR to add lttng-ust to main. :)

FYI: @janvorli

ncopa:
I am ok to move it to main or community, after a minor cleanup.

I'm adding a versioned dependency `userspace-rcu-dev>0.10`

Is 6 month support (latest stable) enough for this or do we need longer time support?
main -> 2 year support
community -> 6 month support (latest stable only)

am11:
Thanks @ncopa!

Given libcoreclrtraceptprovider's dependency on lttng-ust is not going anywhere for foreseeable future (at least there is no tracking issue in CoreCLR repo to that effect), I would say a long-term support would make sense.

PR #2332, opened 2017-09-19
testing/libupnpp: version update to 0.16.0

xsteadfastx:


nmeum:
Checksum was missing. I fixed that for you.

PR #2334, opened 2017-09-19
main/checkbashisms: enable tests

HRio:


HRio:
@ncopa seems I and @awilfox was racing a bit here ;-)

PR #2336, opened 2017-09-20
main/perl-gd: version bump, add tests

awilfox:
Old version of perl-gd uses nonstandard private GD APIs which is fixed in the new version, which is also maintained by a new CPAN Author (RURBAN instead of LDS).

It uses perl-extconfig-pkgconfig so this needs to be moved to main to satisfy the dependency.


ncopa:
merged. thanks.

PR #2339, opened 2017-09-20
main/rsync: include rrsync

HRio:
rrsync is used to restrict rsync to a subdirectory declared in .ssh/authorized_keys

rrsync put in /usr/share in order to avoid a perl dependency for the package.


ncopa:
would it be an idea to have a subpackage called `rrsync` which has perl as dep instead? that way you can `apk add rrsync`

HRio:
yes a subpackage would be an option but I usually copy it to ~/bin/ and customize. This as its conf is part of the perl script it self.

PR #2340, opened 2017-09-20
main/kamailio: Updated kamailio version. Updated APKBUILD to use GitHub kamailio sources

sergey-safarov:


clandmeter:
Hi

_gitcommit is never defined so source cannot be downloaded.
Also arch is defined multiple times, can you fix that as wel?

main/kamailio/APKBUILD
```diff
@@ -3,18 +3,17 @@
...
-_giturl="git://github.com/$pkgname/$pkgname.git"
-_gittag=HEAD
+#_gitcommit=39b9a1e0b05c05c50656588a863b564fd0708a6d
 
-
-pkgver=5.0.2
-pkgrel=2
-[ -z "${_gitcommit}" ] && _suffix="_src" || _suffix="-${_gitcommit}"
+[ ! -z "${_gitcommit}" ] && pkgver="${pkgver}.`date +%Y%m%d`"
+[ ! -z "${_gitcommit}" ] && pkgrel="${pkgrel}.${_gitcommit:0:7}"
```
jirutka: This is completely wrong, `pkgrel` is APKBUILD’s version, not the upstream’s package. Moreover it’s not even used here, so please just remove it, keep the abuild simple.

main/kamailio/APKBUILD
```diff
@@ -3,18 +3,17 @@
...
-#_gitcommit=e27802533cee
-_giturl="git://github.com/$pkgname/$pkgname.git"
-_gittag=HEAD
+#_gitcommit=39b9a1e0b05c05c50656588a863b564fd0708a6d
 
-
-pkgver=5.0.2
-pkgrel=2
-[ -z "${_gitcommit}" ] && _suffix="_src" || _suffix="-${_gitcommit}"
+[ ! -z "${_gitcommit}" ] && pkgver="${pkgver}.`date +%Y%m%d`"
```
jirutka: Backticks subshell is basically deprecated, please always use `$(...)` instead.

sergey-safarov:
@jirutka please make review of updated PR

PR #2342, opened 2017-09-20
testing/lldb: move to main

am11:


am11:
@ncopa, with this and #2341, the dependency situation for coreclr will look like this:

```dockerfile
FROM alpine:3.6
RUN cat /etc/*-release

RUN apk update
RUN apk add --no-cache \
        autoconf bash build-base clang clang-dev \
        cmake curl-dev gcc gettext-dev git icu-dev krb5-dev libtool \
        linux-headers libunwind-dev llvm make openssl python zlib-dev \
        util-linux-dev

RUN apk add --no-cache \
       -X https://dl-cdn.alpinelinux.org/alpine/edge/main \
       lldb-dev lttng-ust-dev

RUN git clone https://github.com/dotnet/coreclr -b master --single-branch \
    && cd coreclr && ./build.sh
```

lldb-dev is also used by Rust runtime APKBUILD.

martell:
@am11 just to drop you a line on this so we are not ignoring you on the other thread.
I believe this should be updated to 5.0 along with #2393 before moving it to main.

am11:
Thanks for sharing @martell! Unfortunately CoreCLR does not compile with LLVM5 as noted https://github.com/dotnet/coreclr/issues/14757#issuecomment-341284969. The fix was merged 7 days after 5.0 was released and release_50 branch hasn't picked up `rL313278` for 5.0.1 (checked https://github.com/llvm-mirror/clang/tree/release_50 and https://github.com/llvm-mirror/llvm/tree/release_50). That is not even on their 5.0.1 agenda: https://bugs.llvm.org/show_bug.cgi?id=34492. Registration on bug tracker is disabled, I can't lodge a request :(

Would it be possible to patch release_50 with `rL313278` in aports in/after PR #2393, so coreclr still compile with OOTB llvm 5.0 toolset that is going to be available in Alpine 3.7?

cc @janvorli

I will try to update this PR with lldb-dev 5 bits.

jirutka:
@am11 Is it [this patch](https://reviews.llvm.org/D33852?download=true)?

am11:
@jirutka, yup; once we get the build, we can test coreclr with [abovementioned docker](https://github.com/alpinelinux/aports/pull/2342#issuecomment-330894890).  :)

jirutka:
@am11 I’m gonna add the patch to the clang package, but not test it with some docker…

am11:
Tyvm, I will do the testing. I am trying out building lldb 5.0 atm.

martell:
@am11 there is already a bug tracking it's merging into the 5.0 branch here
https://bugs.llvm.org/show_bug.cgi?id=35080

I added my +1 in the merge request but I believe Richard has to approve it first.
I also added it in a comment into the 5.0.1 blocker thread so it is not forgotten.
https://bugs.llvm.org/show_bug.cgi?id=34492


This regression was originally introduced because I added support for selectany for the mingw platform and there were no tests that explicitly said this should be supported as non windows features so I limited it to windows only.

This patch missed the 5.0 window for 2 reasons.
1. It became apparent that it was broken for some targets, most notably `x86_64-win32-macho`
2. There were some last minute documentation requirements right as the window closed.

This should make it into 5.0.1 and is already in trunk for 6.0.


@jirutka you are probably better off getting the diff out of the actual commit and not the differential because some times subtle changes are made in the actual commit that don't need review.

https://reviews.llvm.org/rL313278 -> you can download the raw diff here.

Or alternatively get it from the git mirror which is probably much better with `format-patch` here is the commit hash `1af16988a3734bcb5dccd5b70a5b75b32402a25a`

Sorry for the confusion and I'll make sure that makes it into 5.0.1

jirutka:
> https://reviews.llvm.org/rL313278 -> you can download the raw diff here.

This diff is quite weird, it has context ∞ or it’d not a diff at all. Commit has is more helpful, thanks.

jirutka:
There’s something wrong, I’m trying to build lldb 5.0.0 and it fails:

```
...
/usr/bin/../lib/gcc/x86_64-alpine-linux-musl/6.4.0/../../../libclangSema.a(SemaExprObjC.cpp.o): In function `applyCocoaAPICheck(clang::Sema&, clang::ObjCMessageExpr const*, unsigned int, bool (*)(clang::ObjCMessageExpr const*, clang::NSAPI const&, clang::edit::Commit&)) [clone .constprop.696]':
SemaExprObjC.cpp:(.text._ZL18applyCocoaAPICheckRN5clang4SemaEPKNS_15ObjCMessageExprEjPFbS4_RKNS_5NSAPIERNS_4edit6CommitEE.constprop.696+0x13c): undefined reference to `clang::edit::rewriteObjCRedundantCallWithLiteral(clang::ObjCMessageExpr const*, clang::NSAPI const&, clang::edit::Commit&)'
/usr/bin/../lib/gcc/x86_64-alpine-linux-musl/6.4.0/../../../libclangSema.a(SemaExprObjC.cpp.o): In function `(anonymous namespace)::ARCCastChecker::VisitCallExpr(clang::CallExpr*)':
SemaExprObjC.cpp:(.text._ZN12_GLOBAL__N_114ARCCastChecker13VisitCallExprEPN5clang8CallExprE+0x12e): undefined reference to `clang::ento::coreFoundation::followsCreateRule(clang::FunctionDecl const*)'
clang-5.0: error: linker command failed with exit code 1 (use -v to see invocation)
```

I’ve tried to build it against clang both with and without the patch for rL313278, it fails in both cases.

am11:
I haven't gotten that far to be able to patch lldb5 code, I rebased my branch on top of `xentec:pr-llvm-5` and struggling to build the entire dependency graph 😿 

```ERROR: unsatisfiable constraints:
  .makedepends-clang-0:
    masked in: cache
    satisfies: world[.makedepends-clang]
>>> ERROR: clang: abuildindex failed
>>> clang: Uninstalling dependencies...
>>> ERROR: lldb: all failed
>>> lldb: Uninstalling dependencies...```

**Comment by martell** [comment]
> This diff is quite weird, it has context ∞ or it’d not a diff at all. Commit has is more helpful, thanks.
Patches are generally built with -U9999 to help with reviews :)  
Getting the actual commit rather then shrinking the context makes more sense. Glad that helps.

**Comment by jirutka** [comment]
@am11 LLVM and clang 5.0 are not merged yet. I’ve built them locally to test if everything works. And it does not, lldb fails to build and compiler-rt fails in two tests.

**Comment by am11** [comment]
With v5.0.1, I have gotten the same error as @jirutka  https://github.com/alpinelinux/aports/pull/2342#issuecomment-342011868.

I then rebased over https://github.com/alpinelinux/aports/pull/3583 branch, and built with LLDB v6.0, the following errors were thrown:

```sh
-- Looking for compression_encode_buffer in compression - not found
-- Found Doxygen: /usr/bin/doxygen (found version "1.8.13") found components:  doxygen missing components:  dot
-- Found PythonInterp: /usr/bin/python2 (found version "2.7.14")
-- Found SWIG: /usr/bin/swig (found version "3.0.12")
-- Performing Test CXX_SUPPORTS_NO_MACRO_REDEFINED
-- Performing Test CXX_SUPPORTS_NO_MACRO_REDEFINED - Success
-- Symbols (liblldb): exporting all symbols from the lldb namespace
-- Configuring done
CMake Error at cmake/modules/AddLLDB.cmake:79 (add_dependencies):
  The dependency target "clang-tablegen-targets" of target "lldbBase" does
  not exist.
  source/CMakeLists.txt:69 (add_lldb_library)


CMake Error at cmake/modules/AddLLDB.cmake:79 (add_dependencies):
  The dependency target "clang-tablegen-targets" of target "lldbBreakpoint"
  does not exist.
  source/Breakpoint/CMakeLists.txt:1 (add_lldb_library)


CMake Error at cmake/modules/AddLLDB.cmake:79 (add_dependencies):
  The dependency target "clang-tablegen-targets" of target "lldbCommands"
  does not exist.
  source/Commands/CMakeLists.txt:1 (add_lldb_library)


CMake Error at cmake/modules/AddLLDB.cmake:79 (add_dependencies):
  The dependency target "clang-tablegen-targets" of target "lldbCore" does
  not exist.
  source/Core/CMakeLists.txt:10 (add_lldb_library)


CMake Error at cmake/modules/AddLLDB.cmake:79 (add_dependencies):
  The dependency target "clang-tablegen-targets" of target
  "lldbDataFormatters" does not exist.
  source/DataFormatters/CMakeLists.txt:1 (add_lldb_library)

# dozens of these
```

Seemed like a target called `clang-tablegen-targets` was missing from clang build. Hence I rebuilt clang by adding `make clang-tablegen-targets` after the line [`make clang-tblgen`](https://github.com/xentec/aports/blob/103d4bfc6f383033e14c2069b9daf35083309572/main/clang/APKBUILD#L55):

```sh
Scanning dependencies of target ClangAttrPCHWrite
[100%] Building AttrPCHWrite.inc...
[100%] Updating AttrPCHRead.inc...
[100%] Updating AttrPCHWrite.inc...
[100%] Built target ClangAttrPCHRead
[100%] Built target ClangAttrPCHWrite
Scanning dependencies of target clang-tablegen-targets
[100%] Built target clang-tablegen-targets
Scanning dependencies of target ClangDriverOptions
[  0%] Building Options.inc...
[  0%] Built target obj.clang-tblgen
...
...
Scanning dependencies of target clang-headers
[  3%] Built target ClangSACheckers
[  4%] Copying clang's adxintrin.h...
[  4%] Copying clang's altivec.h...
...
...
[ 12%] Built target ClangAttrTemplateInstantiate
[ 12%] Built target ClangAttrParsedAttrList
[ 12%] Built target ClangAttrParsedAttrImpl
[ 14%] Built target ClangAttrPCHRead
[ 14%] Built target ClangAttrPCHWrite
[ 14%] Built target clang-tablegen-targets
Scanning dependencies of target clangBasic
[ 14%] Building CXX object lib/Basic/CMakeFiles/clangBasic.dir/Attributes.cpp.o
...
```

However, the subsequent lldb build still failed to find that target.

Looking at commit history of https://github.com/llvm-mirror/lldb, I couldn't find another way to satisfy this directive https://github.com/llvm-mirror/lldb/blob/e0293d42d90e427009f562e1e458aa1a72d4b005/cmake/modules/AddLLDB.cmake#L76-L79.

@llvm-beanz, could you please share if you have an idea how we can recover from this error? Thanks! :)

am11:
The issue was with ClangConfig.cmake not being discovered by LLDB. When pointed to `/usr/lib/cmake/clang/ClangConfig.cmake` via a patch in LLDB, ClangConfig was in turn trying to (re)discover LLVMConfig in `/usr/lib/cmake/llvm6/` directory, which again tries to find `LLVM-Config.cmake` file in a non-existent directory (resolved via CMake canonical path mechanics) `/usr/lib/cmake/llvm/` (without suffix `6`). To fix that, I made a patch in Clang, where it generates ClangConfig.cmake to the LLVM directory for install directory target, that is otherwise used only for the build tree ([L16](https://github.com/llvm-mirror/clang/blob/4e56966f84ebf321114a0269f8f3efd1027bda6a/cmake/modules/CMakeLists.txt#L16) vs. [L42](https://github.com/llvm-mirror/clang/blob/4e56966f84ebf321114a0269f8f3efd1027bda6a/cmake/modules/CMakeLists.txt#L42)).

Another patch I have cherry-picked for LLDB is https://github.com/llvm-mirror/lldb/commit/8f577000b2fe2f5bf5d08e352a2f15f9421f9c82, that Fedora and Debian are using to include missing headers in the final package. This is fixed for v6.1 in release_60 branch.

@jirutka, this is ready (pending green CI). I have rebased over @xentec's `pr-llvm-6`and then @alpinelinux' `master` branches. If you think 3583 should be merged separately, I will wait and re-rebase over alpinelinux' `master` branch, once that PR is merged.

Thanks for your patience and again apologies that it took ages. 😊

am11:
@jirutka, thanks for updating the lldb package and aligning its version with other LLVM products. I will rebase over master soon after 3.8 is shipped.

Incidentally, it seems like the build didn't get initialized for armhf architecture and strangely, the directory was renamed to `lldpb`: http://build.alpinelinux.org/buildlogs/build-edge-armhf/testing/lldpd/ (empty at the moment). Is there some configuration that we can update to fix this issue?

am11:
@jirutka, now that v3.8 is out, could we move forward with https://github.com/alpinelinux/aports/pull/3583? I will then rebase my branch over master. lldb is mostly version revving but this PR will move it to main at the same time.

janvorli:
@jirutka, @am11, I have recently tried to install the lldb package on Alpine 3.6 and my colleague on Alpine 3.7. In both cases, it would hang when launching even the simplest hello world c++ app. Does it work correctly for you? 
I am also curious why the build disables libedit support. I have tried to build lldb 5.0.2 on Alpine 3.6 x64 from sources without any patches and with libedit enabled and the line editing / history seemed to work fine (and interestingly, the lldb was able to start applications without hanging, although breakpoints didn't work).

jirutka:
@janvorli It's not surprising that lldb from edge does not work on v3.6 or v3.7. Try it on edge (i.e. all dependencies installed from edge).

janvorli:
@jirutka ah, I have incorrectly assumed that the apk would take care of that automatically when I install lldb. Is there a way to install all the needed dependencies from the right branch for lldb without having to go one by one?

am11:
I think @janvorli is spot on; we have libedit as make dependency but later we are disabling it: https://github.com/alpinelinux/aports/blob/04213a552282023ed2e893047cf1643dd9cce03f/testing/lldb/APKBUILD#L47
The implication of `LLDB_DISABLE_LIBEDIT` is that it doesn't compile with libedit headers (and disables relevant features). We should probably remove this line.

am11:
@janvorli, just wanted to give an update (in case you missed it) that lldb 11 on `edge` branch has recently moved from `testing/lldb` to `community/lldb` repository.

I have tested .NET 5 and .NET 6 applications; both shared framework as well as singlefile -based deployments; the debugging of C# code experience with SOS plugin is great, on par with Ubuntu and other distros!

Kudos to everyone who made these improvements all around. 🎉

janvorli:
That's awesome news! Many thanks to everyone involved. 
cc: @mikem8361, @tommcdon, @hoyosjs

PR #2348, opened 2017-09-21
main/clutter: add libxi-dev to depends_dev

NotKit:
libxi-dev is needed for Clutter to be built with X11 input extension support,
otherwise touch screen input won't work in apps relying on X11 backend, such
as hildon-desktop in postmarketOS.


zaoqi-unsafe:
@NotKit

Is it possible to install hildon-desktop in alpinelinux?

PR #2350, opened 2017-09-22
testing/gdal: enable curl support

chambbj:
Enable CURL support within GDAL by adding curl-dev as a make dependency.


Ikke:
Looks good. Verified that curl is detected.

PR #2353, opened 2017-09-23
non-free/adobe-flashplayer: added 32-bit support

Head-on-a-Stick:


kaniini:
this change isn't compatible with our automated package maintenance tools (abump)

Head-on-a-Stick:
Oh, OK, sorry to bother you then.

PR #2361, opened 2017-09-25
Package git-cola

Aerdan:
git-cola is a Qt5-based interface for git.

### DIFF
--- a/testing/git-cola/APKBUILD
+++ b/testing/git-cola/APKBUILD
@@ -1,0 +1,33 @@
+# Contributor: Kiyoshi Aman <kiyoshi.aman@gmail.com>
+# Maintainer: Kiyoshi Aman <kiyoshi.aman@gmail.com>
+pkgname=git-cola
+pkgver=2.11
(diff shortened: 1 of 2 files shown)


testing/git-cola/APKBUILD
```diff
@@ -0,0 +1,33 @@
+# Contributor: Kiyoshi Aman <kiyoshi.aman@gmail.com>
+# Maintainer: Kiyoshi Aman <kiyoshi.aman@gmail.com>
+pkgname=git-cola
+pkgver=2.11
+pkgrel=0
+pkgdesc="A GUI application for git built on Python & Qt5"
+url="https://git-cola.github.io"
+arch="noarch"
+license="GPL2"
```
fabaff: Should be `GPL2+`, see ", either version 2 of the License, or (at your option) any later version." from https://github.com/git-cola/git-cola/blob/master/README.md.

Aerdan: Oops, fixed.

PR #2373, opened 2017-09-26
testing/netdata: upgrade to 1.8

HRio:
Add check()

### DIFF
--- a/testing/netdata/APKBUILD
+++ b/testing/netdata/APKBUILD
@@ -11,16 +11,11 @@ depends="bash"
 makedepends="zlib-dev e2fsprogs-dev"
 pkgusers="netdata"
 pkggroups="netdata"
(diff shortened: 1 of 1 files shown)


HRio:
@clandmeter can you have a look?

testing/netdata/APKBUILD
```diff
@@ -19,7 +19,7 @@ builddir="$srcdir"/$pkgname-$pkgver
 
 prepare() {
 	default_prepare
-	update_config_guess
+	update_config_guess || true
```
fabled: You don't need this. If `update_config_guess` fails, it means the files don't need updating and the call should be removed. In that case, you could remove the whole `prepare()` block.

HRio: @fabled done, thanks.

PR #2393, opened 2017-09-28
LLVM 5.0

xentec:
While all packages compile and pass their tests, this PR may have some issues and is still WIP.
I'd like to enable `LLVM_LINK_LLVM_DYLIB` for at least `lld` but bugs like [27685](https://bugs.llvm.org/show_bug.cgi?id=27685) continue to surface somehow.


jirutka:
@Shizmob, could you please take care of this PR?

main/llvm5/llvm-fix-build-with-musl-libc.patch
```diff
@@ -0,0 +1,46 @@
...
+
+diff --git a/include/llvm/Analysis/TargetLibraryInfo.h b/include/llvm/Analysis/TargetLibraryInfo.h
+index 7becdf0..7f14427 100644
+--- a/include/llvm/Analysis/TargetLibraryInfo.h
++++ b/include/llvm/Analysis/TargetLibraryInfo.h
+@@ -18,6 +18,26 @@
+ #include "llvm/IR/Module.h"
+ #include "llvm/Pass.h"
+ 
++#undef fopen64
```
martell: We don't need this patch beyond llvm4, this was fixed upstream in https://reviews.llvm.org/rL292848

martell:
@xentec Sorry for hijacking the review.
Of the clang patches `0001,0002,0003 and 0006` can be upstreamed.
Same applies for `llvm-fix-DynamicLibrary-to-build-with-musl-libc.patch`
I don't see why we would keep maintaining them for every llvm release.
@ncopa @jirutka as the authors of these patches can you submit them to https://review.llvm.org and add me as a reviewer and I will get them merged in tree so we do not have to maintain so many patches when llvm 6.0 comes around.
@Shizmob any thoughts here?
A member of my team currently maintains a private alpine llvm-svn package and this would benefit both us and the official alpine maintainers so I would be glad to make sure they get merged.

Other then that this mostly LGTM bar the build bot timeouts

community/llvm-libunwind/APKBUILD
```diff
@@ -1,42 +1,41 @@
...
 
 build() {
-	cd "$builddir"
+	mkdir -p "$builddir/build"
+	cd "$builddir/build"
 
-	cmake \
-		-DCMAKE_BUILD_TYPE=MinSizeRel \
+	cmake .. \
 		-DCMAKE_INSTALL_PREFIX=/usr \
```
martell: add `-DCMAKE_BUILD_TYPE=Release` ?

kaniini:
Overall I like @martell's idea, but I want to hold back on LLVM5 until after 3.7 release is cut.

am11:
Could LLDB be added to the family of first-class LLVM package? I proposed moving LLDB from testing to main: https://github.com/alpinelinux/aports/pull/2342. LLDB has been very useful in debugging during the port of different .NET Core repos to Alpine.

community/llvm-libunwind/APKBUILD
xentec: Not quite. Setting build type to `Release` mode tells cmake to use `CMAKE_*_FLAGS_RELEASE` while abuild's own `CFLAGS` in `CMAKE_C*_FLAGS` are ignored.

xentec:
@martell Thanks for trying to upstream the patches! I don't have much time these days so feel free to hijack this PR as whole ;).

community/llvm-libunwind/APKBUILD
jirutka: That's why there's MinSizeRel. Why did you remove it?

xentec: Because cmake's `CMAKE_C_FLAGS_MINSIZEREL="-Os -DNDEBUG"` aren't quite the same `CFLAGS` as provided by abuild:  `CMAKE_C_FLAGS="-Os -fomit-frame-pointer"`

ncopa:
I think it would be awesome if we could squeeze in llvm5 to v3.7.

I have verified that llvm5 builds on all supported platforms.

We need an additional fix for `lvm-config --version` which currently adds a version suffix with the git hash from aports. I have a fix for that.

I think we can include this in v3.7 since it is does not have big impact. We still have llvm4 and there are only 3 packages in community that uses clang.

main/llvm5/llvm-fix-DynamicLibrary-to-build-with-musl-libc.patch
```diff
@@ -0,0 +1,33 @@
+From d12ecb83d01dcb580dd94f4d57828f33d3eb4c35 Mon Sep 17 00:00:00 2001
+From: Natanael Copa <ncopa@alpinelinux.org>
+Date: Thu, 18 Feb 2016 15:33:21 +0100
+Subject: [PATCH 3/3] Fix DynamicLibrary to build with musl libc
+
+stdin/out/err is part of the libc and not the kernel so we check for the
+specific libc that does the unexpected instead of linux.
+
+This is needed for making it build with musl libc.
```
martell: Pretty sure this won't apply since https://reviews.llvm.org/rL301153

martell:
We need to apply  https://reviews.llvm.org/rL316672 and `llvm-fix-DynamicLibrary-to-build-with-musl-libc.patch` seems to be already merged in https://reviews.llvm.org/rL304707

community/llvm-libunwind/APKBUILD
jirutka: That's true, but it's more complicated in CMake... IIRC the default profile is debug. Have you verified that it doesn't add `-g` to flags...?

ncopa: I think you are right. adding back the `-DCMAKE_BUILD_TYPE=MinSizeRel` made the `llvm5-static` size reduce to 117MB instead of gigabytes. Good catch!

ncopa:
I have fixes locally for:
- `-DCMAKE_BUILD_TYPE=MinSizeRel`
- fix for version suffix (it currently include the git hash from aports tree)
- trivial build fix for `compiler-rt`

@jirutka are you ok that I push this?

ncopa:
I though i could push my commits directly to this PR but I'm not sure how. I get permission denied.

jirutka:
If @xentec hasn’t disabled it, then it should work.

```sh
git fetch git@github.com:xentec/aports.git pr-llvm-5:pr-llvm-5
git checkout pr-llvm-5
…
git push git@github.com:xentec/aports.git pr-llvm-5:pr-llvm-5
```

xentec:
@ncopa You need to be 'Maintainer' of aports. Right now you're a 'Contributor'.

community/llvm-libunwind/APKBUILD
jirutka: That’s what I do, I drink and I know things
(except I don’t drink)

(Related to CMake profile: http://lists.alpinelinux.org/alpine-devel/5877.html)

jirutka:
@xentec No, this has nothing in common with metadata in our aports. When you open a PR, there’s a checkbox allowing the target repository’s owners to push into your branch. It’s enabled by default.

martell:
Woo, let’s get things in shape to get this shipped.
Good job @xentec

xentec:
@martell I'm still waiting for @ncopa's fixes which might help me to upgrade lldb and libc++.

jirutka:
Are you sure we want to build LLVM with -Os instead of -O2?

jirutka:
> - trivial build fix for `compiler-rt`

@ncopa What is that trivial build fix? It builds on my machine, but 2 tests fail. Is it related? Please push your changes to this PR (see https://github.com/alpinelinux/aports/pull/2393#issuecomment-341751377).

jirutka:
Well, compiler-rt should be okay now.

There’s a weird linking error in lldb build (see https://github.com/alpinelinux/aports/pull/2342#issuecomment-342011868), could someone please look at it? I’d like to move it to community or main and merge together with this. And more importantly, make sure that it’s not caused by some problem in clang or llvm package.

⚠️ Please don’t merge this yet, I wanna rebase it (squash some fixups) before merging into master.

jirutka:
BTW it’d be great to also build main/mesa against llvm5…

community/lld/APKBUILD
```diff
@@ -1,42 +1,45 @@
...
+		-DCMAKE_BUILD_TYPE=MinSizeRel \
+		-DCMAKE_C_FLAGS_MINSIZEREL_INIT="$CFLAGS" \
+		-DCMAKE_CXX_FLAGS_MINSIZEREL_INIT="$CXXFLAGS" \
+		-DCMAKE_EXE_LINKER_FLAGS_MINSIZEREL_INIT="$LDFLAGS" \
 		-DCMAKE_INSTALL_PREFIX=/usr \
-		-DCMAKE_CXX_FLAGS="$CXXFLAGS" \
-		-DCMAKE_C_FLAGS="$CFLAGS" \
+		\
 		-DCMAKE_SKIP_INSTALL_RPATH=ON \
-		-DLLVM_LINK_LLVM_DYLIB=ON
```
jirutka: @xentec Why did you remove this?

xentec: I've already described it in the OP. Have you build `lld` and run it's tests without errors?

jirutka: Ah, sorry, I've overlooked it. No, still build errors from linker. Currently I'm trying to build clang with shared libs and link lldb with clang dynamically when linking with static fails.

jbergstroem:
> @jirutka said:
> Are you sure we want to build LLVM with -Os instead of -O2?

Out of curiosity, what is the resulting size change of the installed binaries?

xentec:
@jbergstroem 

package         | `-Os`       | `-O2`
----------------|-------------|-------------
llvm            |   6 483 968 |   9 031 680
llvm-dev        |  16 527 360 |  16 605 184
llvm-libs       |  40 017 920 |  54 067 200
llvm-static     | 123 363 328 | 124 616 704
llvm-test-utils |     786 432 |     897 024

martell:
@jirutka what is left blocking this? Also `-Os` looks like a win

jirutka:
LLDB (see above)

martell:
I feel as though we could handle that as part of #2342?
Getting this merged and a day of us all testing might be a good idea instead of blocking llvm + clang on lldb?

I don't know your procedure for these things though it seems lldb will be happy with the llvm4 package if this is just merged.

xentec:
@jirutka so is this PR good to go or are we missing something?

Meanwhile I was able to dynamically link `clang` and `lld` by building them together with LLVM with minor test fails:
```
clang/test/Analysis/crash-trace.c:18:11: error: expected string not found 
clang/test/Parser/crash-report.c:11:11: error: expected string not found 
```
Maybe this will also fix `lldb` bugs, but I haven't tried it yet.

jirutka:
Reminder for myself (and maybe others): look at #1574.

jirutka:
Merged! Thanks a lot to @xentec and @martell. Now we have to somehow fix lldb.

martell:
@jirutka @xentec @ncopa https://reviews.llvm.org/rL318608
llvm 3.6 should support alpine out of the box. :)

the way we were previously detecting pie would break cross building and relied on the vendor
it now is the default for musl targets.

PR #2410, opened 2017-09-29
community/php7: Libressl is required for php7-openssl package

petk:
This patch is related to [https://bugs.php.net/bug.php?id=75219](https://bugs.php.net/bug.php?id=75219). Currently php7-openssl automatically requires only libressl2.5-libcrypto and libressl2.5-libssl which don't provide the required `/etc/ssl/openssl.cnf` file.

Thanks.


vakartel:
I think it's not a php bug but libressl's.

ncopa:
this should be resolved with #2413

petk:
Awesome. Thank you for this. :+1:

PR #2432, opened 2017-10-01
main/sdl2: Add support for DirectFB

craftyguy:


ncopa:
applied with b6809c90e9a01eb3acc7757a96a5ef8691b3824d

PR #2433, opened 2017-10-01
testing/fzf: add package

Ikke:


testing/fzf/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+	cd "$builddir"
+
+	# Just copies the target binary to $buildir/bin
+	make install
+
+	install -Dm0755 bin/fzf $pkgdir/usr/bin/fzf
+}
+
+tmux() {
+	depends="tmux bash"
```
ncopa: we need to ad tmux and bash to makedepends as a workaround to help builder figure out the build order properly. (image if something else depends on fzf-tmux before tmux or bash is built)

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+# Maintainer: Kevin Daudt <ops@ikke.info>
+pkgname=fzf
+_pkgver="0.17.0-2"
+pkgver=${_pkgver/-/.}
+pkgrel=0
+pkgdesc="A command-line fuzzy finder"
+url="https://github.com/junegunn/fzf"
+arch="all"
+license="MIT"
+depends="go"
```
ncopa: is `go` a buildtime or runtime dependency?

testing/fzf/APKBUILD
Ikke: I've added a `makedepends="$depends"`

testing/fzf/APKBUILD
Ikke: You are right, go is only a makedepends.

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+
+	# Just copies the target binary to $buildir/bin
+	make install
+
+	install -Dm0755 bin/fzf $pkgdir/usr/bin/fzf
+}
+
+tmux() {
+	depends="tmux bash"
+	makedepends="$depends"
```
ncopa: this will not work. the split function is not execute til after the package is built, so there is no way for abuild to be aware of those dependencies til after the package is built. We need to have it at the top of the build.

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
+# Contributor: Kevin Daudt <ops@ikke.info>
+# Maintainer: Kevin Daudt <ops@ikke.info>
+pkgname=fzf
+_pkgver="0.17.0-2"
+pkgver=${_pkgver/-/.}
```
jirutka:
```diff
-_pkgver="0.17.0-2"
-pkgver=${_pkgver/-/.}
+pkgver=0.17.0.2
+_pkgver="${pkgver%.*}-${pkgver##*.}"
```

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+	make test
+}
+
+package() {
+	cd "$builddir"
+
+	# Just copies the target binary to $buildir/bin
+	make install
+
+	install -Dm0755 bin/fzf $pkgdir/usr/bin/fzf
```
jirutka:
```diff
-install -Dm0755 bin/fzf $pkgdir/usr/bin/fzf
+install -Dm0755 bin/fzf "$pkgdir"/usr/bin/fzf
```

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+}
+
+tmux() {
+	depends="tmux bash"
+	makedepends="$depends"
+	description="Helper script to start fzf in a tmux pane"
+	arch="noarch"
```
jirutka: This does not work anymore.

```diff
- subpackages="$pkgname-tmux"
+ subpackages="$pkgname-tmux::noarch"

- arch="noarch"
```

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+	cd "$builddir"
+	install -Dm0755 bin/fzf-tmux $subpkgdir/usr/bin/fzf-tmux
```
jirutka:
```diff
- install -Dm0755 bin/fzf-tmux $subpkgdir/usr/bin/fzf-tmux
+ install -Dm0755 bin/fzf-tmux "$subpkgdir"/usr/bin/fzf-tmux
```

testing/fzf/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+arch="all"
+license="MIT"
+makedepends="go"
+subpackages="$pkgname-tmux"
+source="$pkgname-$_pkgver.tar.gz::https://github.com/junegunn/fzf/archive/$_pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$_pkgver"
+
+build() {
+	cd "$builddir"
+	make
```
jirutka: It fetches some dependencies from Internet using `go get`, i.e. without any proper verification. That’s forbidden. Use glide with lock file.

Ikke:
@jirutka Thanks for the review, I'll incorporate the changes.

testing/fzf/APKBUILD
Ikke: Much better way to solve it, thanks.

testing/fzf/APKBUILD
Ikke: Is there any guide how to properly do this? Or a package that shows how it's done?

testing/fzf/APKBUILD
Ikke: I'm wondering, these are just run-time dependencies. They are never needed during building, and only for this sub-package. Is it than still necessary to mark them as global dependencies (defeating the purpose, only requiring bash and tmux if you want the tmux integration)

Ikke:
I've updated the package to include the requested changes.

Ikke:
Can this be merged?

PR #2452, opened 2017-10-05
main/py-dbus: update to 1.2.4 and py3

z3ntu:
python2 -c "import dbus"
python3 -c "import dbus"
both work

Please see the comments (line 10 and 40) and give me feedback what to improve.
Also these "mv" commands used for the dev and the doc packages aren't beautiful. How should I handle that?

Travis is failing because autoconf-archive is only in edge/testing.

### DIFF
--- a/main/py-dbus/APKBUILD
+++ b/main/py-dbus/APKBUILD
@@ -1,39 +1,68 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=py-dbus
-pkgver=1.2.0
-pkgrel=2
+pkgver=1.2.4
+pkgrel=0
 pkgdesc="Python bindings for DBUS"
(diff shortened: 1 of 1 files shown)


main/py-dbus/APKBUILD
```diff
@@ -1,39 +1,68 @@
...
+}
+
+_py() {
+	local python="$1"
+	pkgdesc="$pkgdesc (for $python)"
+	depends="$depends $python"
+	# should this line be here ?
+	# install_if="$pkgname-$pkgver-r$pkgrel $python"
+	cd "$_builddir-$python"
+	PYTHON=$python ./configure \
```
ncopa: i think we should build the thing from `build()`, not during subpackage split function.

ncopa:
lets follow this up in #4846 which solves it better.

PR #2458, opened 2017-10-06
testing/wireguard: version bump to 0.0.20171005

zx2c4:
Straight-forward version bump. Release notes [from](https://lists.zx2c4.com/pipermail/wireguard/2017-October/001791.html) the mailing list:

```
  * community: code welcomed
  
  There's a small disorganized list of things to do, some of which are large
  projects, and others are just little cleanups:
  
  https://www.wireguard.com/todo/
  
  If anybody is interested in working on something, please don't hesitate to
  send an email to this list, team at wireguard.com, or me directly.
  
  * tools: simmer down silly compilers
  * tools: compile on non-Linux
  * contrib: remove worthless build artifact
  * kernel-tree: remember UAPI in patch creation
  * curve25519-neon-arm: force ARM encoding, since this is unrepresentable in Thumb
  * compat: support ptr_ring for old kernels
  * compat: conditionally redefine GENL_UNS_ADMIN_PERM
  * compat: RHEL backported netlink changes
  
  These here are all compatibility-related fixes mostly left over from churn of
  the previous snapshots, where we lost some compatibility with old kernels and
  weird toolchains. The above series of fixes brings us back up to par, and
  should make life slightly easier for a few packagers who had to work-around
  things in the last snapshot.
  
  * compat: macro rewrite netlink instead of cluttering
  * global: satisfy bitshift pedantry
  * global: use _WG prefix for include guards
  * global: add space around variable declarations
  * queueing: cleanup skb_padding
  
  Style, mostly.
  
  * Makefile: add non-verbose mode to tools
  * Makefile: clang now builds the kernel, so use scan-build
  
  One touch static analysis: `make check`.
  
  * receive: simplify message type validation
  * receive: use local keypair, not ctx keypair in error path
  * send: put keypair reference
  * receive: we're not planning on turning that into a while loop now
  * queueing: use ptr_ring instead of linked lists
  * receive: do not store endpoint in ctx
  * queueing: move from ctx to cb
  
  This is another huge change, and the main motivation for releasing this
  snapshot. We move from using a linked list-based queue to a ring buffer-based
  queue, which yields considerable performance increases. It also allows us to
  entirely rid ourselves of a memory cache object, which further increases
  performance and decreases latency. The move to a ring buffer will also make
  writing lock-less algorithms easier, which will eventually increase our
  performance on systems with extremely high core counts.
```

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20171001
-_mypkgrel=1
+_ver=0.0.20171005
+_mypkgrel=0
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


ncopa:
i rebased it and pushed. Thanks!

PR #2470, opened 2017-10-09
main/openrc: avoid modloop error with `BOOT_IMAGE=(loop)/boot/vmlinuz-hardened`

CasperVector:
cf. alpinelinux/mkinitfs#16; this can happen with grub2, when the following `grub.cfg` snippet is used:
```
menuentry "Alpine" {
	search --no-floppy --file --set=pia /iso/alpine.iso
	loopback loop ($pia)/iso/alpine.iso
	linux	(loop)/boot/vmlinuz-hardened modules=loop,squashfs,isofs,ntfs,sd-mod,usb-storage quiet
	initrd	(loop)/boot/initramfs-hardened
}
```

### DIFF
--- a/main/openrc/modloop.initd
+++ b/main/openrc/modloop.initd
@@ -10,8 +10,7 @@ depend() {
 
 # read kernel options
 init_KOPT() {
-	eval set -- $(cat /proc/cmdline 2>/dev/null)
-	for opt; do
+	for opt in $(cat /proc/cmdline 2>/dev/null); do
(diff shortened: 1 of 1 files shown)


ncopa:
It fails to build:

````
modloop.initd: FAILED
````

We also need to bump pkgrel, otherwise will not the builder pick it up (no new version = nothing new to build).

ncopa:
i fixed it for you and verified that it boots. thanks!

PR #2473, opened 2017-10-10
testing/git-crypt: new aport

errm:


jirutka:
git-crypt seems to be abandoned and the last release is from 2015-05-30. It’s a crypto stuff, so it quite matters… I’d not recommend to use it and not sure if we really want to accept such package.

errm:
Ok sounds reasonable...

PR #2488, opened 2017-10-13
testing/pavucontrol: new aport

ctrlcctrlv:
https://freedesktop.org/software/pulseaudio/pavucontrol/
A graphical interface (GTK3) for changing the volume of PulseAudio streams and other PulseAudio related tasks

This port only works on Edge, as it requires the `pulseaudio` package.

### DIFF
--- a/testing/pavucontrol/APKBUILD
+++ b/testing/pavucontrol/APKBUILD
@@ -1,0 +1,36 @@
+# Contributor: Fredrick Brennan <copypaste@kittens.ph>
+# Maintainer: Fredrick Brennan <copypaste@kittens.ph>
+pkgname=pavucontrol
+pkgver="20171007"
+pkgrel=1
+pkgdesc="A graphical interface for changing the volume of PulseAudio streams and other PulseAudio related tasks"
+url="https://freedesktop.org/software/pulseaudio/pavucontrol/"
+arch="all"
+license="GPL2+"
+depends="pulseaudio dbus gtkmm3 libsigc++ libcanberra"
+makedepends="automake autoconf gtkmm3-dev libsigc++-dev libcanberra-dev pulseaudio-dev"
+install=""
+subpackages="$pkgname-doc $pkgname-lang"
+_commit="335c26c57c18d95cc7d4ca693a75ef94fe919e1d"
+source="https://github.com/pulseaudio/$pkgname/archive/$_commit.zip"
+builddir="$srcdir/$pkgname-$_commit"
+
+build() {
+	cd "$builddir"
+	./autogen.sh
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
+		--prefix=/usr \
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var
+	make
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install
+}
+
+sha512sums="dee0fb2a7c5ea096fa7798bb9202beb76261ccb85ac25334a4fce885232e09b41bf307bf434cbf68eece8c481e5063869cdedb3a18814fa97f980dd629956d1a  $_commit.zip"


testing/pavucontrol/APKBUILD
```diff
@@ -0,0 +1,36 @@
+# Contributor: Fredrick Brennan <copypaste@kittens.ph>
+# Maintainer: Fredrick Brennan <copypaste@kittens.ph>
+pkgname=pavucontrol
+pkgver="20171007"
+pkgrel=1
+pkgdesc="A graphical interface for changing the volume of PulseAudio streams and other PulseAudio related tasks"
+url="https://freedesktop.org/software/pulseaudio/pavucontrol/"
+arch="all"
+license="GPL2+"
+depends="pulseaudio dbus gtkmm3 libsigc++ libcanberra"
```
ncopa: I dont think you need the pulseaudio, gtkmm3, libsigc++ or libcanberra. abuild should automatically detect the shared library dependencies.

testing/pavucontrol/APKBUILD
```diff
@@ -0,0 +1,36 @@
+# Contributor: Fredrick Brennan <copypaste@kittens.ph>
+# Maintainer: Fredrick Brennan <copypaste@kittens.ph>
+pkgname=pavucontrol
+pkgver="20171007"
```
ncopa: we try avoid making releases of random git snapshots. Instead, we should use releases from upstream. In this case the latest release is 3.0.

Also, if 4.0 comes out, apk will consider the number `20171007` bigger than `4` so it will not upgrade.

testing/pavucontrol/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgrel=1
+pkgdesc="A graphical interface for changing the volume of PulseAudio streams and other PulseAudio related tasks"
+url="https://freedesktop.org/software/pulseaudio/pavucontrol/"
+arch="all"
+license="GPL2+"
+depends="pulseaudio dbus gtkmm3 libsigc++ libcanberra"
+makedepends="automake autoconf gtkmm3-dev libsigc++-dev libcanberra-dev pulseaudio-dev"
+install=""
+subpackages="$pkgname-doc $pkgname-lang"
+_commit="335c26c57c18d95cc7d4ca693a75ef94fe919e1d"
```
ncopa: We should use release number instead of a random commit (which in practice means that you do the release and have to take responsibility and maintainership of it instead of letting upstream take that)

testing/pavucontrol/APKBUILD
ctrlcctrlv: I tried that first actually. The current release of `pavucontrol` is very old and no longer works correctly with the `pulseaudio` package. What should I do?

testing/pavucontrol/APKBUILD
ctrlcctrlv: You do. They are all linked, and `pavucontrol` won't start without having them. Libcanberra is used for the sound it makes when you change volume to tell you the current volume. Libsigc++ is also linked and it won't start. It uses GTKmm3 for its toolkit. It needs every library I listed to start. Yes, this could be considered "bloat". If you don't like it, don't use this package; I plan to also package `pulseaudio-mixer-cli` if this gets approved, which has fewer deps (only ncurses, pulseaudio and python3 I think).

testing/pavucontrol/APKBUILD
ctrlcctrlv: How do I do a release on a project I'm not a maintainer of?

testing/pavucontrol/APKBUILD
ctrlcctrlv: It also links against libpulse.so because it uses the PulseAudio API.

ctrlcctrlv:
@ncopa Hello? I've tried reaching you over IRC also. Can you please get back to me about this port? Looking at the `community` repo, this wouldn't even be close to the only port that's based on a commit and whose version is a date rather than a specific version. Please, if you have a chance, let me know what to do next.

testing/pavucontrol/APKBUILD
jirutka: Contact upstream and convince them to finally release a new version.

Meanwhile you can use version like `3.0_git20171007`. Or upstream tell you that the next release will increment _major_, it may be better to use `4.0_pre20171007`.

testing/pavucontrol/APKBUILD
jirutka: Well, you can’t.

testing/pavucontrol/APKBUILD
jirutka: It’s not needed to explicitly specify dependencies that are automatically detected by abuild. You can see it in the abuild log:

https://travis-ci.org/alpinelinux/aports/builds/287546330#L1329-L1350:
```
>>> pavucontrol*: Tracing dependencies...
	pulseaudio
	dbus
	gtkmm3
	libsigc++
	libcanberra
	so:libatkmm-1.6.so.1
	so:libc.musl-x86_64.so.1
	so:libcanberra-gtk3.so.0
	so:libcanberra.so.0
	so:libgcc_s.so.1
	so:libgdkmm-3.0.so.1
	so:libglib-2.0.so.0
	so:libglibmm-2.4.so.1
	so:libgobject-2.0.so.0
	so:libgtk-3.so.0
	so:libgtkmm-3.0.so.1
	so:libintl.so.8
	so:libpulse-mainloop-glib.so.0
	so:libpulse.so.0
	so:libsigc-2.0.so.0
	so:libstdc++.so.6
```

testing/pavucontrol/APKBUILD
ctrlcctrlv: I know, as referenced https://github.com/alpinelinux/aports/pull/2488#issuecomment-337126071. My port would not be the only port based on a commit.

ctrlcctrlv:
Closing -- I no longer use Alpine and have no interest in pursuing this port anymore. Way too many holes to jump through -- consider adding something like the AUR. "Contact[ing] upstream" to make them fit their package into some restrictive self-imposed mold is completely silly to me; I will not waste the time of the upstream Pulse team to get them to do something literally no other distro requires.

PR #2495, opened 2017-10-14
Add KDE Frameworks Tier 1

PureTryOut:
Adds tier 1 packages of the KDE frameworks. Originally packaged for [postmarketOS](https://postmarketos.org).

Only Kirigami2 is missing due to CMake not being able to find the header files of `qt5-qtquickcontrols2`. However, `qt5-qtquickcontrols2` does not seem to have header files as adding the `$pkgname-dev` subpackage fails.

### DIFF
--- a/testing/kcoreaddons/APKBUILD
+++ b/testing/kcoreaddons/APKBUILD
@@ -1,0 +1,33 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=kcoreaddons
+pkgver=5.39.0
+pkgrel=0
+pkgdesc='Addons to QtCore'
+arch="all"
+url='https://community.kde.org/Frameworks'
+license="LGPL-2.1"
+depends=""
+makedepends="extra-cmake-modules qt5-qttools-dev doxygen shared-mime-info"
+source="https://download.kde.org/stable/frameworks/${pkgver%.*}/${pkgname}-${pkgver}.tar.xz"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
+
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DKDE_INSTALL_LIBDIR=lib \
+		-DBUILD_QCH=ON
+	make
+}
+
+check() {
+	cd "$builddir"
+	CTEST_OUTPUT_ON_FAILURE=TRUE ctest
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="${pkgdir}" install
+}
+sha512sums="6579087b2d34ae6bc8d8254c420ae2f6f68d55cec587b5a58429bb82b9a2de111221351d58e2514705fda52c9b433bf98add574e4aba6ee0408340dc442707dd  kcoreaddons-5.39.0.tar.xz"
--- a/testing/attica-qt/APKBUILD
+++ b/testing/attica-qt/APKBUILD
@@ -1,0 +1,35 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=attica-qt
+pkgver=5.39.0
+pkgrel=0
+arch="all"
+pkgdesc='Qt5 library that implements the Open Collaboration Services API'
+url='https://community.kde.org/Frameworks'
+license="LGPL-2.1"
+depends=""
+depends_dev="qt5-qtbase-dev"
+makedepends="$depends_dev extra-cmake-modules doxygen qt5-qttools-dev"
+source="https://download.kde.org/stable/frameworks/${pkgver%.*}/attica-${pkgver}.tar.xz"
+subpackages="$pkgname-dev $pkgname-doc"
+builddir="$srcdir/${pkgname/-qt/}-$pkgver"
+
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DKDE_INSTALL_LIBDIR=lib \
+		-DBUILD_QCH=ON
+	make
+}
+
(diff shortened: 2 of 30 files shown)


kaniini:
@awilfox what is your opinion on this vs the KDE packaging in adelie?

awilfox:
* I don't understand why a lot of these packages depend on `clang` or `py3-qt`; they don't use either.
* There are definitely no packages that require Python 2, but a few have them in dependencies.
* No KF5 package has a direct requirement on Mesa.  There are a few in the _Applications_ collection that require GLU, but I haven't seen raw GLX calls anywhere in KF5.
* Tests are disabled on a large swath of these despite having perfectly valid tests.  A few of the packages do require X11 running, and I've noted that in the Adélie builds in the hopes of extending `abuild`'s check support to include virtual X11 support.  Plenty of the frameworks do not use X11 in their tests, however.
* The patching is unnecessarily complex.  If the patches were written correctly and the packages used `$builddir` instead of hardcoding `$srcdir/build` (where KDE packages, with the exception of `kactivities`, do not even require out-of-source building) then they could use default_prepare and not even have prepare phases.
* All of the kinit/etc patches are not required as long as extra-cmake-modules is patched to provide `_XOPEN_SOURCE` which is what we do.  This takes care of packages in the Applications collection as well as the Plasma collection.
* None of these packages declare `$depends_dev`, so they wouldn't be very usable to anyone who wants to pull them in alone (LXQt uses KWindowSystem and is being ported to KGlobalAccel), and it causes the `makedepends` of each one to be bloated.

I can do a more thorough review later if desired.

I do however like the patch to syntax-highlighting that adds APKBUILD support.  That's pretty nifty.

awilfox:
Additionally:

* There are packages that build QCH sets that require GraphViz (/usr/bin/dot) but do not have GraphViz in their makedepends.  This will cause the build to fail.
* `kscreenlocker` is the only package that declares a dependency of linux-pam but `plasma-workspace` is what provides the `/etc/pam.d/kde` file, which seems odd to me.  Plasma Workspace doesn't even use PAM; only `kscreenlocker` does, and that is the package that should have that file installed.  (You can use `kscreenlocker` from other desktops, if you wanted to, and it wouldn't work without that.)
* Speaking of `kscreenlocker`, it _requires_ SUID to read `/etc/shadow`, otherwise sessions will not be unlockable.
* Plasma Workspace is missing a *lot* of shell runtime deps.  It is also unstable 5.11 version which has known bugs with Wayland.
* KWin does _not_ require ConsoleKit; it is an optional, though definitely recommended, dependency of Plasma itself.  Plasma can use ConsoleKit for seat management and it can also be used to unlock the session if the session unlocker fails (I've seen that happen a few times due to `ENFILE`.), but KWin itself doesn't.
* Sonnet has no backends enabled.  This makes it entirely useless.  (In Adélie we split aspell and build hunspell in by default, since that is what upstream recommends.)

Should I just open a PR with our KF5 packaging?  I'm in the process of upgrading it from 5.38.0 to 5.39.0, and then I wouldn't have a problem with it if Alpine is ready/willing to ship them.

(Personally, I was waiting until Plasma Desktop was more heavily tested, and then we were going to offer the entire Plasma environment as one PR.  If KF5 would be useful to Alpine beyond Plasma, we can definitely provide that.)

awilfox:
I've just noticed one more pretty fatal flaw: `start_kdeinit` has its SUID bit removed in the `testing/kinit` package due to "SUID violating Alpine standards" (`options="suid"` is what to use when it is needed).  KInit can be built with `libcap` to avoid SUID, and that's what we do in Adélie; without one of these, the `kdeinit` process becomes vulnerable to the Linux OOM killer.

awilfox:
One idea I've just had is to commit all tier 1 frameworks first (as they are required to not have any dependencies on any KF5 libs).  That way they can all be built and Travis won't throw errors due to the other deps missing.

Then commit tier 2 frameworks which can only depend on tier 1, then tier 3, then Plasma which has a few tier 4 deps.  This would be four PRs, but they would be verifiable by Travis, and ensure issues are caught quickly and can therefore be resolved quickly.

PureTryOut:
I've fixed most things mentioned. Travis reports build failed due to hitting the time limit, not because a compilation failed.

I'll add the packages from the other tiers in separate PR's.

testing/attica-qt5/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=attica-qt5
```
kaniini: Why does this package name divert from upstream package name?

Do you honestly believe it will be built against Qt4?

testing/polkit-qt5/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=polkit-qt5
```
kaniini: Again, this should be `polkit-qt`, to match upstream.  It won't ever be built against Qt4.

testing/kcoreaddons/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+source="https://download.kde.org/stable/frameworks/${pkgver%.*}/${pkgname}-${pkgver}.tar.xz"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
+
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DKDE_INSTALL_LIBDIR=lib \
+		-DBUILD_QCH=ON
+		make
```
kaniini: indention level is wrong

PureTryOut:
Could you guys as the Alpine team have some internal discussion about this? On the mailing list [someone mentioned](https://lists.alpinelinux.org/alpine-devel/5882.html) that he preferred non-LTS instead.

PureTryOut:
Since this PR only seems to cause trouble, which we obviously do not want, I'll close this and let Adélie handle upstreaming of KDE stuff.

PR #2503, opened 2017-10-16
update Haxe and Neko

andyli:
Update testing/haxe to 3.4.4. Add check(), change arch to all.
Add myself as maintainer. (I'm a member of the Haxe Foundation, which is the org behind Haxe and Neko. I'm also the Haxe/Neko package maintainer of Debian/Fedora/openSUSE)

### DIFF
--- a/testing/neko/APKBUILD
+++ b/testing/neko/APKBUILD
@@ -17,6 +17,7 @@ subpackages="$pkgname-dev $pkgname-libs"
 source="$pkgname-$pkgver.tar.gz::https://github.com/HaxeFoundation/neko/archive/v${pkgver//./-}.tar.gz
 	compilation-fixes.patch
 	nojit.patch
+	mincoming-stack-boundary.patch
 	"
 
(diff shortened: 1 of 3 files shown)


kaniini:
hi,

i did these, but the neko compiler is crashing on x86.  so i am reopening this.

andyli:
I've just fixed the neko x86 segfault by disabling jit, which is optional and only enabled in x86 by default.
I'm not sure if there is any problem on archs other than x86 and x86_64, if so, please share the build log and I will look into that.

kaniini:
arm build fails due to unknown CFLAGS.

andyli:
ummm... Was that something about `-mincoming-stack-boundary`?

andyli:
Just tested with a arm alpine docker image, it was indeed `-mincoming-stack-boundary`. I've added an upstream patch to fix it. Hopefully it is now buildable in all arches.

andyli:
May I know if there is any remaining issue?

ncopa:
Seems like it builds on all arches now. Thank you!

PR #2504, opened 2017-10-16
main/bluez: add gatttool to bluez-deprecated

sharukins:
gatttool is being built but missing in package bluez-deprecated


sharukins:
seems that gatttool is not being installed on compilation..

PR #2505, opened 2017-10-16
main/vlan: check if subinterface does not already exist before creating it

Anthony25:
The vlan preup script tried to create the subinterface even if it
already exists. It does not work on dualstack (ipv4 + ipv6)
configurations, and on subinterfaces with multiple addresses setup.

Now the vlan and mvlan scripts check if the interface does not
already exist, and only if it does not, try to create it.


kaniini:
it looks fine to me, but would like @fabled and @ncopa opinion too.

main/vlan/mvlan.pre-up
```diff
@@ -22,9 +22,11 @@ if ! ip link show "$RAW_DEVICE" >/dev/null; then
 	exit 1
 fi
 
-ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan
+if ! ip link show "$IFACE" &>/dev/null; then
+	ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan
+fi
 
-if ! ip link show "$IFACE" >/dev/null; then
+if ! ip link show "$IFACE" &>/dev/null; then
```
fabled: Move to be guarded by the added if. No point to check the interface twice if it exists.

main/vlan/vlan.pre-up
```diff
@@ -36,10 +36,12 @@ if ! [ -d /proc/net/vlan ]; then
...
 
-ip link set dev "$RAW_DEVICE" up
-ip link add link "$RAW_DEVICE" name "$IFACE" type vlan id "$VLAN_ID"
+if [ ! -e "/proc/net/vlan/${IFACE}" ]; then
+	ip link set dev "$RAW_DEVICE" up
+	ip link add link "$RAW_DEVICE" name "$IFACE" type vlan id "$VLAN_ID"
+fi
 
-if ! ip link show "$IFACE" >/dev/null; then
+if [ ! -e "/proc/net/vlan/${IFACE}" ]; then
```
fabled: Same here. Move to be guarded by the above if.

Anthony25:
Is it okay this way, or do you prefer that I keep the explicit error message (`Failed to create vlan device…`) in case of failure?

fabled:
I think it would be preferrable to keep the explicit error. At least the `exit 1` should be kept to abort execution.

Anthony25:
Ok, so I used trap to deal with the `set -e` and print an explicit error.
I just have a question about the commit 8faf34e: in `mvlan.pre-up`, is `$VLAN_ID` ever defined? I just copied the error message added by this commit, and I preferred asking before changing it.

Anthony25:
@fabled: did you have the time to review the changes yet?

fabled:
Looks ok for me. @ncopa any comments?

main/vlan/mvlan.pre-up
```diff
@@ -22,9 +27,7 @@ if ! ip link show "$RAW_DEVICE" >/dev/null; then
 	exit 1
 fi
 
-ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan
-
-if ! ip link show "$IFACE" >/dev/null; then
-	echo "Failed to create vlan device $IFACE on device $RAW_DEVICE with tag $VLAN_ID"
-	exit 1
+trap "device_creation_error" ERR
+if ! ip link show "$IFACE" &>/dev/null; then
```
jirutka: `&>` is a bashism. Please replace it with `>/dev/null 2>&1`.

main/vlan/mvlan.pre-up
```diff
@@ -22,9 +27,7 @@ if ! ip link show "$RAW_DEVICE" >/dev/null; then
 	exit 1
 fi
 
-ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan
-
-if ! ip link show "$IFACE" >/dev/null; then
-	echo "Failed to create vlan device $IFACE on device $RAW_DEVICE with tag $VLAN_ID"
-	exit 1
+trap "device_creation_error" ERR
```
jirutka: Why do you use `trap` here? Why not just add `|| device_creation_error` after execution of `ip` command?

main/vlan/mvlan.pre-up
Anthony25: Fixed, thanks

main/vlan/mvlan.pre-up
Anthony25: I used trap to get something similar to the way I did in `vlan.pre-up`. If you prefer, I could add `|| device_creation_error` instead.

main/vlan/mvlan.pre-up
```diff
@@ -22,9 +27,7 @@ if ! ip link show "$RAW_DEVICE" >/dev/null; then
...
 fi
 
-ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan
-
-if ! ip link show "$IFACE" >/dev/null; then
-	echo "Failed to create vlan device $IFACE on device $RAW_DEVICE with tag $VLAN_ID"
-	exit 1
+if ! ip link show "$IFACE" >/dev/null 2>&1; then
+	ip link add link "$RAW_DEVICE" name "$IFACE" type macvlan || \
+        device_creation_error
```
jirutka: There are spaces… please use only tabs for indentation.

Anthony25:
Hi @jirutka,
Thanks, I fixed the issue.

main/vlan/vlan.pre-up
```diff
@@ -36,10 +41,8 @@ if ! [ -d /proc/net/vlan ]; then
...
 
-ip link set dev "$RAW_DEVICE" up
-ip link add link "$RAW_DEVICE" name "$IFACE" type vlan id "$VLAN_ID"
-
-if ! ip link show "$IFACE" >/dev/null; then
-	echo "Failed to create vlan device $IFACE on device $RAW_DEVICE with tag $VLAN_ID"
-	exit 1
+trap "device_creation_error" ERR
+if [ ! -e "/proc/net/vlan/${IFACE}" ]; then
+	ip link set dev "$RAW_DEVICE" up
```
ncopa: I think the `ip link set ... up` should be outside the `/proc/net/vlan/$IFACE` check, in case the interface exists but is `down`.

ncopa: So:
````sh
trap "device_creation_error" ERR
ip link set dev "$RAW_DEVICE" up
ip [ ! -e "/proc/net/vlan/${IFACE}" ]; then
￼	ip link add link "$RAW_DEVICE" name "$IFACE" type vlan id "$VLAN_ID"
fi
````

Anthony25: Yes, you're right, thanks!

Anthony25:
Sorry, I updated the checksums

PR #2506, opened 2017-10-17
Fix bugs in 2 packages by removing Unicode whitespace.

andychu:
I found these problems by parsing all APKBUILD scripts with my shell
(http://www.oilshell.org/).

The problem only occurs if 'make' fails.  Here is an excerpt:

```
$ od -c unicode-space.sh
0000000   m   a   k   e       |   | 302 240   r   e   t   u   r   n
0000020   1  \n
0000022
```

`\302 \204` is a utf-8 whitespace.  No shells accept this -- it's parsed
as part of the 'return' word, which makes it an invalid command.

```
$ busybox ash unicode-space.sh
make: *** No targets specified and no makefile found.  Stop.
unicode-space.sh: line 1:  return: not found

$ bash unicode-space.sh
make: *** No targets specified and no makefile found.  Stop.
unicode-space.sh: line 1:  return: command not found

$ dash unicode-space.sh
make: *** No targets specified and no makefile found.  Stop.
unicode-space.sh: 1: unicode-space.sh:  return: not found
```

### DIFF
--- a/community/rng-tools/APKBUILD
+++ b/community/rng-tools/APKBUILD
@@ -27,12 +27,12 @@ build() {
 		--libexecdir=/usr/lib/rng-tools \
 		--sysconfdir=/etc \
 		--disable-silent-rules
-	make || return 1
+	make
 }
 
 package() {
 	cd "$builddir"
-	make DESTDIR="${pkgdir}" install || return 1
+	make DESTDIR="${pkgdir}" install
 
 	install -m644 -D "${srcdir}/rngd.confd" "${pkgdir}/etc/conf.d/rngd" || return 1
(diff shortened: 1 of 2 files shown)


fabled:
Thanks. The preferred way is to remove `|| return 1` altogether since abuild now executes split functions with `set -e`

andychu:
OK I pushed another change that removed those parts.

I didn't remove the rest of the `|| return 1` because I don't have a good way of testing the APKBUILD.  I've run Alpine in a VM, but I'm making these edits on an Ubuntu machine.  I would guess that not many people build Alpine packages on Ubuntu, if at all?  The build environment for Alpine is Alpine I would guess :)

I also had some problems with building Alpine packages in an Alpine chroot but we can talk about that elsewhere.

PR #2513, opened 2017-10-18
add gatttool to bluez-deprecated

sharukins:
gatttool is being build but not installed. Therefore we move the binary to /usr/bin

### DIFF
--- a/main/bluez/APKBUILD
+++ b/main/bluez/APKBUILD
@@ -84,7 +84,7 @@ deprecated() {
 	pkgdesc="Deprecated bluetooth tools"
 	mkdir -p "$subpkgdir"/usr/bin
 	mv "$pkgdir"/usr/bin/ciptool \
-		"$builddir"/attrib/gatttool \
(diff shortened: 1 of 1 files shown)


nmeum:
Merged. Please squash your commits the next time, prefix your commit message with `main/bluez` and bump the pkgrel.

sharukins:
thanks and sorry, I'm pretty new to this.

nmeum:
No problem, just giving some feedback in case you want to send additional patches in the future. You might also want to check out the development documentation in the wiki if that's the case.

sharukins:
Thanks, I really appreciate the feedback. I should probably also learn how to properly use github ^^
If you allow me one question: would this patch be viable for a backport to 3.6? I would consider this a bugfix, but others might see this as adding a feature.

nmeum:
> If you allow me one question: would this patch be viable for a backport to 3.6?

Probably not, we mostly backport security fixes to the stable branches. But I guess this should be decided by the maintainer (@ncopa).

PR #2502, opened 2017-10-16
community/elasticsearch, testing/filebeat: upgrade to 5.6.3

snakeliwei:

PR #2517, opened 2017-10-19
testing/filebeat: upgrade to 5.6.3

snakeliwei:


nmeum:
Thanks! What about the elastic search upgrade you proposed in #2502?

PR #2528, opened 2017-10-20
libwebsockets rebuild

fabaff:
Depends on #2527


jirutka:
Please add these commits to #2527, so it is build together.

fabaff:
Moved to #2527.

PR #2531, opened 2017-10-21
community/cacti: upgrade to 1.1.26 + some fixes ... upgrade to 1.1.30

vakartel:
- minor fixes in APKBUILD
- remove executable bit from scripts and images
- add mariadb-client as dependency for cacti-setup
- add missing php modules to cacti-php* subpackages
- fix error in php-fpm pool definition


community/cacti/APKBUILD
```diff
@@ -50,15 +54,15 @@ package() {
 	# move variable and writable data to cacti's home
 	local dir; for dir in cache plugins resource rra scripts; do
 		mv "$pkgdir"/usr/share/webapps/$pkgname/$dir \
-			"$pkgdir"/var/lib/$pkgname/$dir || return 1
-		chown -R $pkgusers:$pkggroups "$pkgdir"/var/lib/$pkgname/$dir
+			"$pkgdir"/var/lib/$pkgname/$dir
+		chown -R $pkgname: "$pkgdir"/var/lib/$pkgname/$dir
```
ncopa: the `/var/lib/cacti` does not need to be writeable by web server (eg `www-data`)?

community/cacti/APKBUILD
```diff
@@ -50,15 +54,15 @@ package() {
...
 		ln -s /var/lib/$pkgname/$dir "$pkgdir"/usr/share/webapps/$pkgname/$dir
 	done
 
 	# move log to /var/log
 	mv "$pkgdir"/usr/share/webapps/$pkgname/log \
-		"$pkgdir"/var/log/$pkgname || return 1
-	chown -R $pkgusers:$pkggroups "$pkgdir"/var/log/$pkgname
+		"$pkgdir"/var/log/$pkgname
+	chown -R $pkgname: "$pkgdir"/var/log/$pkgname
```
ncopa: Are you sure web server does not need to have write permissions to log directory?

community/cacti/APKBUILD
vakartel: the cacti's PHP-FPM pool is configured to start as a user "cacti". So those dirs are writabe to it? and log dir too.

In apache configuration it's better to run virtual-host with suexec cacti

jirutka: I agree, but keep in mind that there are still many uneducated users who run PHP stuff with some mod_php module in Apache under the same user. That's of course insecure and silly, but unfortunately still quite common.

So, if you wanna do this change, add also a post-upgrade script printing a brief warning about this change (see e.g. nginx pkg for example).

I'd be great to also add an article to wiki describing how to use pkgs with provided php-fpm runscript (i.e. run it as normal program and set up proxy in nginx or apache) and add link to the warning message.

jirutka: Wait, this pkg does not create symlink /etc/init.d/cacti -> php-fpm and install /etc/conf.d/cacti... See e.g. nextcloud pkg for example what I mean.

vakartel: @jirutka I know it, but:::
1. cacti have 2 separate instances: cron poller process to collect data and web-served application.

2. I think it's better to include a preconfigured virtual-host template for apache web server to docs-subpackage.

3. Cacti's web installer is fair enough to describe all steps to do, so I think it's have no sense to describe it in wiki. wiki is static and it's info can be obsolete.


and about php-fpm link. php-fpm have it's own configs an starts great with /etc/init.d/php-fpm start (restart).

I remember our debate when I tried to simplify php-fpm.init.d script. And I still disagree with you. php-fpm have it's own config and can run as a single system-wide process and forks to it's preconfigured pools that runs with it's own user privileges. It's a bad idea to move service-wide initialization to system-wide level.

IMHO The only reason to do so and  make symlinks in /etc/init.d is a configless daemon that get all in conf from command line.

And in cacti's variant, if we starti its web-app from apache or lighthttpd, we don't need php-fpm at all. So php-fpm it's a very specific case.

jirutka: > cacti have 2 separate instances: cron poller process to collect data and web-served application.

IIRC cron poller process can and should be run with php-cli, not php-fpm or web server.

> I think it's better to include a preconfigured virtual-host template for apache web server to docs-subpackage.

I’m not against this.

> Cacti's web installer is fair enough to describe all install steps to do

It does not describe any Alpine-specific such as php-fpm runscript.

> php-fpm have it's own configs an starts great with /etc/init.d/php-fpm start

This config is needed and used in both cases.

> php-fpm have it's own config and can run as a single system-wide process and forks to it's preconfigured pools that runs with it's own user privileges.

Which is bad from basically all angles. It’s less secure, because you must start php-fpm with root privileges. It’s less reliable, because when some of the apps fail, it may take down the php-fpm master process and all apps with it. You cannot define service dependencies per app. You cannot start/stop/restart php-fpm apps separately using system service manager. I don’t know about single advantage, it’s just wrong. Similar to running php under web server.

> It's a bad idea to move service-wide initialization to system-wide level.

Why?

> And in cacti's variant, If we start its web-app with apache or litespeed http servers, we don't need php-fpm at all. So php-fpm is a very specific case.

That’s why it’s in a subpackage, so user can decide…

community/cacti/APKBUILD
```diff
@@ -1,41 +1,69 @@
...
 	$pkgname.php-fpm.conf
 	"
-builddir="$srcdir/$pkgname-$pkgver"
+_cactidir="$srcdir/$pkgname-$pkgver"
+_spinedir="$srcdir/$pkgname-spine-$pkgver"
+builddir=$_spinedir
 
-package() {
+prepare() {
+	default_prepare()
```
jirutka: Ehm, this is not a valid shell syntax. ;)

TBK:
Cacti 1.1.33 was released january 22nd.

PR #2533, opened 2017-10-22
main/ffmpeg: upgrade to 3.4

TBK:


ncopa:
is this ABI compatible with 3.3?

nmeum:
https://abi-laboratory.pro/tracker/timeline/ffmpeg/

TBK:
@nmeum thx for the link.

The backwards compatibility medium severity issues revolve around 3 header files (pixfmt.h, avcodec.h and avfilter.h) where a couple of enums and structs have been added new members/fields, so a few have shifted position.

Some of which should not be an issue according to the code comments:
pixfmt.h:
```
AVCOL_SPC_NB ///< Not part of ABI

....

AV_PIX_FMT_NB ///< number of pixel formats, DO NOT USE THIS if you want to link with shared libav* because the number of formats might differ between versions
```

avcodec.h:
```
    /**
     * The number of side data elements (in fact a bit more than it).
     * This is not part of the public API/ABI in the sense that it may
     * change when new side data types are added.
     * This must stay the last enum value.
     * If its value becomes huge, some code using it
     * needs to be updated as it assumes it to be smaller than other limits.
     */
AV_PKT_DATA_NB
```

It is the changes to _receive_packet_ and the removal of _send_packet_ AVCodec struct fields that breaks compatibility along with the addition of the _preinit_ field in the AVFilter struct (avfilter.h).

I guess the only action is to either rebuild all dependent packages or to make it a seperate package.

@ncopa what should be cause of action be?

ncopa:
I suppose we can just bump ffmpeg. they didnt consider the changes big enough to change the SONAME so I things break we can always blame on upstream (and rebuild whatever broke).

PR #2535, opened 2017-10-22
community/openlibm: upgrade to 0.5.5

tmpfile:


clandmeter:
Please consider adding abuild checks to aports which are missing them.
If the build script/project does not support it please disable them with:
```options="!check"```

jirutka:
… and add a comment after it like `# no tests provided`.

PR #2550, opened 2017-10-24
community/rust: upgrade to 1.19

jeizsm:


community/rust/APKBUILD
```diff
@@ -54,7 +52,7 @@ prepare() {
 	cd "$builddir"
 
 	# Remove bundled dependencies.
-	rm -Rf src/llvm/ src/jemalloc/
+	rm -Rf src/llvm/
```
jirutka: Why did you remove it? There’s still src/jemalloc in the tarball.

community/rust/alpine-allow-crt-static.patch
```diff
@@ -1,41 +0,0 @@
...
-         return crt_static;
---- a/src/test/compile-fail/crt-static-gated.rs
-+++ b/src/test/compile-fail/crt-static-gated.rs
-@@ -10,5 +10,6 @@
- 
- // compile-flags:-C target-feature=+crt-static
- // error-pattern: specifying the `crt-static` target feature is only allowed
-+// ignore-test we just want this feature and no one will stop us!
- 
- fn main() {}
```
jirutka: Are you sure that this is not needed anymore? They said that this will be always gated (i.e. allowed only in nightly).

community/rust/musl-support-dynamic-linking.patch
```diff
@@ -38,14 +36,21 @@ diff --git a/src/bootstrap/compile.rs b/src/bootstrap/compile.rs
...
 +    if target.contains("musl") {
          copy_musl_third_party_objects(build, target, &libdir);
      }
+ 
+     if build.config.sanitizers && compiler.stage != 0 && target == "x86_64-apple-darwin" {
+         // The sanitizers are only built in stage1 or above, so the dylibs will
+         // be missing in stage0 and causes panic. See the `std()` function above
+         // for reason why the sanitizers are not built in stage0.
+         copy_apple_sanitizer_dylibs(&build.native_dir(target), "osx", &libdir);
+     }
```
jirutka: Why is it here? This is relevant only on macOS…?

community/rust/APKBUILD
jeizsm: It doesn't compile without it despite of --disable-jemalloc, but I'm pretty sure jemalloc is unused

community/rust/alpine-allow-crt-static.patch
jeizsm: Yes, I'm sure. It works on stable from this version

community/rust/APKBUILD
jirutka: What is the error? If it doesn’t build without it, then how you can be sure it’s not really used?

community/rust/alpine-allow-crt-static.patch
jirutka: Interesting. Even on official build from Rust team?

community/rust/musl-support-dynamic-linking.patch
jeizsm: I just copy it from file to fix patch

community/rust/alpine-allow-crt-static.patch
jeizsm: Yes, I try it on version from rustup

community/rust/musl-support-dynamic-linking.patch
jirutka: Eh, yeah, I misunderstood it. :)

community/rust/APKBUILD
jeizsm: I don't save error, but it something about missing files. I compile hello world with builded version and stripped binary about 150kb size and I don't see jemalloc string in binary

PR #2561, opened 2017-10-25
main/bash: upgrade to 4.4.12; modernize

TBK:


main/bash/APKBUILD
```diff
@@ -30,23 +29,24 @@ while [ $_i -le $_patchlevel ]; do
...
-_builddir="$srcdir"/$pkgname-$_myver
+builddir="$srcdir"/$pkgname-$_myver
+
 prepare() {
 	local p
-	cd "$_builddir"
-	update_config_sub || return 1
+	cd "$builddir"
+	update_config_sub
 	for p in $source; do
```
nmeum: Please use `default_prepare` instead of this for-loop.

TBK:
@nmeum Done :)

nmeum:
Does this upgrade fix CVE-2016-0634? If so: please add a secfixes comment. See: https://bugs.alpinelinux.org/issues/8060

TBK:
It was fixed with this path https://ftp.gnu.org/pub/gnu/bash/bash-4.3-patches/bash43-047 (bash43-047 | 2016-10-04 13:57 | 4.3K )

I have added a secfix comment. Please check that it is correct.

nmeum:
> I have added a secfix comment. Please check that it is correct.

It wasn't correct, I fixed it upstream for you. Thanks for your patch!

PR #2566, opened 2017-10-25
main/protobuf: upgrade to 3.4.1

TBK:
ABI is not backward compatibility - https://abi-laboratory.pro/tracker/timeline/protobuf/

### DIFF
--- a/community/mumble/APKBUILD
+++ b/community/mumble/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Johannes Matheis <jomat+alpinebuild@jmt.gr>
 pkgname=mumble
 pkgver=1.2.19
-pkgrel=2
+pkgrel=3
 pkgdesc="A low-latency, high quality voice chat software"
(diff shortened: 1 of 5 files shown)


nmeum:
> ABI is not backward compatibility

In that case you should trigger a rebuild of all packages depending on protobuf by bumping their `pkgrel`.

TBK:
@nmeum Thanks for the advice. I was in doubt of the procedure :)

I will fix that straight away.

nmeum:
Sorry, for not being clear on this. When rebuild packages you should add a new commit for every rebuild. The commit message of the rebuild commits should be `${repository}/${pkgname}: rebuild against protobuf 3.4.1`

jirutka:
If you don't have a strong reason for merging this before branching v3.7, I'd postpone this after v3.7, we're in freeze state now.

ncopa:
we updated to 3.4.1. thanks!

PR #2582, opened 2017-10-26
main/vala: upgrade to 0.38.0

rgdoliveira:


rnalrd:
I think a safe approach is to rebuild all packages using vala-0.38.
Thanks

rgdoliveira:
@rnalrd thanks for the review.
I'm added package main/xfce4-vala to be rebuild. Is there any other package that should be rebuild?

rnalrd:
Search in all APKBUILDs for ”vala" and you''ll find a number of those.

Thanks!

/eo
--
Written with a tiny keyboard. Please excuse brevity and typos

Il 27 ott 2017 5:33 PM, "Roberto Oliveira" <notifications@github.com> ha
scritto:

@rnalrd <https://github.com/rnalrd> thanks for the review.
I'm added package main/xfce4-vala to be rebuild. Is there any other package
that should be rebuild?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/alpinelinux/aports/pull/2582#issuecomment-340005557>,
or mute the thread

ncopa:
IIRC think the only needed package (eg the only thing that breaks after vala upgrade) is xfce4-vala. I think the rest can be rebuilt as we do the rebuild for 3.7-stable.

rgdoliveira:
I will close it for now

PR #2584, opened 2017-10-26
community/rust: upgrade to 1.21

jeizsm:
musl-support-dynamic-linking.patch and most of musl-fix-static-linking.patch is already in rust https://github.com/rust-lang/rust/pull/40113/files


jirutka:
Great job, thank you very much! 👍

PR #2588, opened 2017-10-26
community/sphinx: upgrade to 2.3.2

TBK:


rnalrd:
We don't usually push beta releases. Any good reason why not waiting the next stable version?
Thanks

TBK:
It was to use the HTTP protocol. The development of Sphinx seems to have stagnated.

v3 is on the horizon... it will hopefully be here sooner rather then later.

I'll try to make a package for the fork which has seems be more actively developed -
https://github.com/manticoresoftware/manticore

ncopa:
I close this since no good reason to go for a beta release was given. Feel free to re-open if you disagree.

PR #2589, opened 2017-10-26
testing/haxe: enable for all arches

andyli:
Following up on https://github.com/alpinelinux/aports/pull/2503.

I'm not sure if I should use `arch="x86_64 aarch64 ppc64le"` instead of `arch="all"` since [camlp4](https://pkgs.alpinelinux.org/packages?name=camlp4&branch=&repo=&arch=&maintainer=) is only available in those arches. Just hoping the alpine build infra is smart enough to ignore arches that build deps are not available.


rnalrd:
Sorry we aren't that smart :) Please add only the supported arches, thanks!

PR #2592, opened 2017-10-26
testing/tcc: new aport

HinataYanagi:
Tiny C Compiler (http://bellard.org/tcc/) on Alpine!
tcc has not made releases for long time, though it has improved obscurely.
This package is introducing not official but the latest working tcc from http://repo.or.cz/tinycc.git .


nmeum:
Is the source your are using here for tcc also used by other distributions?

HinataYanagi:
*Yup!* 
Arch Linux also uses this source. 
See https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/tcc#n19 .

testing/tcc/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+# Maintainer: Hinata Yanagi <hinasssan@gmail.com>
+pkgname=tcc
+pkgver=2017.10.19
+pkgrel=0
+pkgdesc="Tiny C Compiler"
+url="http://repo.or.cz/tinycc.git"
+arch="all"
+license="the GNU Lesser General Public License"
+depends="musl-dev"
+makedepends="gcc texinfo"
```
nmeum: gcc shouldn't be needed in `makedepends` since it is part of `build-base`.

testing/tcc/APKBUILD
```diff
@@ -0,0 +1,34 @@
+# Contributor:
+# Maintainer: Hinata Yanagi <hinasssan@gmail.com>
+pkgname=tcc
+pkgver=2017.10.19
+pkgrel=0
+pkgdesc="Tiny C Compiler"
+url="http://repo.or.cz/tinycc.git"
+arch="all"
+license="the GNU Lesser General Public License"
+depends="musl-dev"
```
nmeum: other compilers such as gcc or pcc don't depend on musl-dev either, I would leave it up to the user to install it.

testing/tcc/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+makedepends="gcc texinfo"
+commit=fc0188ffbcf29f856caedee8cef07d88be01e439
+source="http://repo.or.cz/tinycc.git/snapshot/${commit}.tar.gz"
+subpackages="${pkgname}-doc"
+
+builddir="$srcdir/"
```
nmeum: Please set `$builddir` to a meaningful value and use that value in `build()`, `check()` and `package()` instead of `"${builddir}tinycc-"*`. This is important because function like `default_prepare` rely on `$builddir`.

nmeum:
Looking good, added some review comments. Since Arch is also using this source I would be willing to merge this as soon as the review comments are resolved.

HinataYanagi:
Make some correction. Check it out, please.

nmeum:
Merged with a few stylistic changes applied manually by myself. Besides I also squashed your commits. Sadly the test suite fails on the server even though it passed on my laptop. No idea why this is the cause, I will investigate this further. For now the test suite has been disabled.

Anyway, thanks a lot for your patch.

HinataYanagi:
Feels nice to have Tiny C Compiler on tny Alpine.
Thanks!

PR #2607, opened 2017-10-27
main/ansible: upgrade to 2.4.1.0

fabaff:


nmeum:
@fabaff could you add a check() function for the ansible aport as a part of this PR?

PR #2611, opened 2017-10-28
main/py-virtualenv: py3 support; modernize

TBK:


main/py-virtualenv/APKBUILD
```diff
@@ -8,21 +9,50 @@ pkgdesc="Virtual Python Environment builder"
...
 
 package() {
-	cd "$builddir"
-	python2 setup.py install --prefix=/usr --root="$pkgdir" || return 1
+	mkdir -p "$pkgdir"
+}
+
+_py2() {
+  replaces="$pkgname"
+  depends="${depends//py-/py2-} py2-pip"
```
jirutka: `depends` is empty, so `${depends//py-/py2-}` is useless here.

main/py-virtualenv/APKBUILD
```diff
@@ -8,21 +9,50 @@ pkgdesc="Virtual Python Environment builder"
...
+  _py python2
+}
+ 
+_py3() {
+  depends="${depends//py-/py3-}" # pip is part of the python3 pkg
```
jirutka: `depends` is empty, so this is useless here.

main/py-virtualenv/APKBUILD
```diff
@@ -8,21 +9,50 @@ pkgdesc="Virtual Python Environment builder"
...
 build() {
 	cd "$builddir"
-	python2 setup.py build || return 1
+
+	python2 setup.py build
+	python3 setup.py build
+}
+
+check() {
+  cd "$builddir"
```
jirutka: Please use only tabs for indentation (and alignment) in APKBUILDs and other shell scripts, not spaces. If you install http://editorconfig.org/ to your editor, it’ll take care of it for you.

main/py-virtualenv/APKBUILD
```diff
@@ -8,21 +9,50 @@ pkgdesc="Virtual Python Environment builder"
 url="http://www.virtualenv.org/en/latest/"
 arch="noarch"
 license="MIT"
-depends="python2 py2-pip"
-makedepends="python2-dev"
+depends=""
+makedepends="python2-dev python3-dev" # add back py-tox for check()
```
jirutka: What does `# add back py-tox for check()` mean? Dependencies for check are defined in `checkdepends`.

TBK: I figured that out after looking throught the abuild code, could not find any docs providing that info.

I will fix it in a sec.

TBK:
@jirutka I fixed the things that you mentioned :)

PR #2613, opened 2017-10-29
testing/mapnik: Added missing dependency icu-libs.

mjuenema:
A user reported the following error:

  so:libicuuc.so.59 (missing):
  required by:
  mapnik-3.0.15-r0[so:libicuuc.so.59]
  mapnik-3.0.15-r0[so:libicuuc.so.59]

libicuuc.so.59 is part of the icu-libs package.


ncopa:
Thank you for looking into this, but unfortunally the fix is wrong. `icu-libs` should have a `provides=so:libicuuc.so.59`:

````console
$ apk info --provides icu-libs
icu-libs-59.1-r1 provides:
so:libicudata.so.59=59.1
so:libicui18n.so.59=59.1
so:libicuio.so.59=59.1
so:libicutest.so.59=59.1
so:libicutu.so.59=59.1
so:libicuuc.so.59=59.1
````

It also installs just fine in my machine.

It sounds like the user installed mapnik from `edge/testing` while not adding `edge/main` repo.

The fix in this case is to:
1) test that `mapnik` is working and review that the packaging of `mapnik` is done correctly
2) move the `mapnik` package to community.

If this is done before v3.7 release, then it wil be available from there without needing to mix in `edge/testing` repo

PR #2625, opened 2017-11-01
community/erlang: Added wx package (required to run debbuger and observer)

sergey-safarov:


community/erlang/APKBUILD
```diff
@@ -58,7 +58,9 @@ subpackages="$pkgname-dev
 	     $pkgname-stdlib:stdlib
 	     $pkgname-syntax-tools:syntax_tools
 	     $pkgname-tools:tools
-	     $pkgname-xmerl:xmerl"
+	     $pkgname-xmerl:xmerl
+	     $pkgname-wx:wx"
+options="!check"
```
nmeum: Why did you disable the test suite? I would be surprised if upstream didn't have a test suite.

sergey-safarov: This done to fix [travis error](https://prnt.sc/h4yyrg).
Now abuild will fail if test not exist.

PR #2628, opened 2017-11-01
main/bash-completion: new maintainer

lramage94:
as per the [mailing list](http://lists.alpinelinux.org/alpine-devel/5898.html)


ncopa:
Thank you very much for stepping up!

Can you please add a `check` function? Anything that check that the build was ok.
If not possible, then please add a line:
````sh
options="!check"
````

nmeum:
Merged! Thanks a lot for taking over maintainership.

PR #2631, opened 2017-11-02
main/itstool: add libxml2 python module in makedepends

rgdoliveira:
Build fails with error:
configure: error: Python module libxml2 is needed to run this package

Adding py-libxml2 in makedepends to fix the build error.


rgdoliveira:
on hold. I will fix it soon...

ncopa:
I think this is wrong fix.

PR #2632, opened 2017-11-02
community/xpra: upgrade to 2.1.3

itoffshore:
community/xpra: upgrade to `2.1.3`


ncopa:
looks like it was upgraded already

PR #2641, opened 2017-11-03
testing/ostree: upgrade to 2017.13

misery:


nmeum:
This is not closely related to your changes but looking at the aport I noticed that checks are disabled. Why is this the case? On a first glance [upstream does seem to have a test suite](https://github.com/ostreedev/ostree/tree/master/tests).

Your changes look good though.

PR #2644, opened 2017-11-04
testing/libatomic_ops: new aport

TBK:


misery:
Already exists: https://git.alpinelinux.org/cgit/aports/tree/main/libatomic_ops

TBK:
Thanks, my search was apprently not sufficient :/

PR #2648, opened 2017-11-04
php5-suhosin / php5-apcu: restore to community

itoffshore:
Both of these packages were removed due to being [non core extensions](https://github.com/alpinelinux/aports/commit/71d024f91830436d4007f82ec37a25da282dbca9). I've been using them both for almost a year without any problems.

[suhosin](https://suhosin.org) improves the security of `php5`.

### DIFF
--- a/community/php5-apcu/APKBUILD
+++ b/community/php5-apcu/APKBUILD
@@ -1,0 +1,44 @@
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+# Contributor: Valery Kartel <valery.kartel@gmail.com>
+_php=php5
+pkgname=${_php}-apcu
+_pkgreal=apcu
+# release 5 is php7+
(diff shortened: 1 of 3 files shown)


jirutka:
php5 is outdated, why you don’t use php7 instead?

itoffshore:
[suhosin7](https://github.com/sektioneins/suhosin7) is alpha at the moment

andypost:
php5 has security support til Dec'18 http://php.net/supported-versions.php
It makes sense to keep it while a lot of docker images using it & a lot of environments still not possible to migrate to php7.

btw php5-apcu is very needed without it we have to stick on 3.5 alpine but somehow updates stuck in https://github.com/alpinelinux/aports/pull/2634

andypost:
Also would be great to return php5-xdebug

PR #2654, opened 2017-11-05
main/sdl2: re-add directfb support

craftyguy:
This re-adds directfb support, which is required by a program we use in pmOS for unlocking rootfs/LUKS on boot.

This also adds a simple check to the package, since no test suite exists for this app.

### DIFF
--- a/main/sdl2/APKBUILD
+++ b/main/sdl2/APKBUILD
@@ -2,13 +2,13 @@
 # Maintainer: August Klein <amatcoder@gmail.com>
 pkgname=sdl2
 pkgver=2.0.7
-pkgrel=0
+pkgrel=1
 pkgdesc="A development library designed to provide low level access to audio, keyboard, mouse, joystick and graphics"
 url="http://www.libsdl.org"
 arch="all"
 license="zlib"
 depends=""
-makedepends="alsa-lib-dev libxcursor-dev libxrandr-dev libxrender-dev mesa-dev
+makedepends="alsa-lib-dev directfb-dev libxcursor-dev libxrandr-dev libxrender-dev mesa-dev
 	libxscrnsaver-dev libxi-dev libx11-dev libxext-dev libusb-dev libxkbcommon-dev"
 subpackages="$pkgname-dev"
 source="https://www.libsdl.org/release/SDL2-$pkgver.tar.gz"
@@ -16,6 +16,8 @@ builddir="$srcdir/SDL2-$pkgver"
 
 build() {
 	cd "$builddir"
+	# NOTE: Please do not remove the --enable-video-directfb flag.
+	# Some apps, used for booting Alpine derivatives, require this to function!
 	./configure \
 		--build=$CBUILD \
 		--host=$CHOST \
@@ -26,8 +28,13 @@ build() {
 		--disable-esd \
 		--disable-nas \
 		--disable-rpath \
-		--with-x
+		--with-x \
+		--enable-video-directfb
 	make
+}
+
+check() {
+    [ "$($builddir/sdl2-config --version)" == "$pkgver" ]
 }
 
 package() {


main/sdl2/APKBUILD
```diff
@@ -2,17 +2,18 @@
...
 arch="all"
 license="zlib"
 depends=""
-makedepends="alsa-lib-dev libxcursor-dev libxrandr-dev libxrender-dev mesa-dev
+makedepends="alsa-lib-dev directfb-dev libxcursor-dev libxrandr-dev libxrender-dev mesa-dev
 	libxscrnsaver-dev libxi-dev libx11-dev libxext-dev libusb-dev libxkbcommon-dev"
 subpackages="$pkgname-dev"
 source="https://www.libsdl.org/release/SDL2-$pkgver.tar.gz"
 builddir="$srcdir/SDL2-$pkgver"
+options="!check"
```
nmeum: Please add a comment why you disabled tests. Does upstream have a test suite?

craftyguy: Travis fails without it.. the original APKBUILD in Alpine didn't include any check function.. 

I can't find anything in SDL2 to run checks/tests after building, it doesn't appear like they implemented anything.

ollieparanoid:
@craftyguy, your comment almost reads like @ncopa may have wanted to sabotage our project. But I highly doubt that 😀  I guess the reason this got removed was that he thought, there would be no use case for the flag?

Anyway, as @craftyguy wrote changing this little flag means that users with encrypted devices **couldn't boot** [postmarketOS](https://postmarketos.org/) anymore, so that would be pretty fatal (if it was used as daily driver, which is the goal we're heading for).

Is there anything we can do to ensure that this flag does not get removed by accident in the future, maybe add a comment next to the flag?

Also regarding `!check`, since they have no testsuite, we could call `sdl2-config --version` and compare that with `$pkgver`. I have read in other aports PR reviews, that something like that is preferred over using `!check`.

craftyguy:
> your comment almost reads like @ncopa may have wanted to sabotage our project.

Ugh, I didn't mean to blame anyone. I'm guessing it was an honest mistake :)


Regarding the !check thing, I can remove it from the APKBUILD, but Travis will fail here. I'm assuming that is OK since this was never set in this package in the past, and there has not been any testsuite in this project.

@AmatCoder, FYI, since you are listed as the maintainer for this package.

craftyguy:
I amended this commit to add a simple check (version compare) since no test suite exists for this app.

main/sdl2/APKBUILD
```diff
@@ -26,10 +28,15 @@ build() {
...
 		--disable-nas \
 		--disable-rpath \
-		--with-x
+		--with-x \
+		--enable-video-directfb
 	make
 }
 
+check() {
+    [ "$($builddir/sdl2-config --version)" == "$pkgver" ]
```
jirutka: This is a library, right? Then such check is quite useless, so just disable it with comment (`options="!check"  # upstream doesn't provide tests`).

ollieparanoid: Wait, I thought that would be better than no tests - at least we know that one binary in the package can be executed, so if the compiler was broken or some other fundamental error, it would get caught. But this only counts for programs, that are part of the main program we're compiling, not some program in `$pkgname-dev`?

ncopa:
directfb support was removed because it seems to be dead upstream and the only thing using it was sdl2. I thought it was leftovers from old days and wanted clean up. Sorry.

ncopa:
Now I remember why i removed it in the first place. DirectFB headers are broken and it seemed to be dead upstream so I could not find a fix.

This happens when building gource:

````
mv -f src/core/.deps/gource-quadtree.Tpo src/core/.deps/gource-quadtree.Po
In file included from /usr/include/directfb/directfb++.h:35:0,
                 from /usr/include/directfb/directfb.h:32,
                 from /usr/include/SDL2/SDL_syswm.h:83,
                 from src/core/sdlapp.cpp:31:
/usr/include/directfb/direct/Types++.h:46:6: error: template with C linkage
      template <typename _Item> class List;
      ^~~~~~~~
/usr/include/directfb/direct/Types++.h:47:6: error: template with C linkage
      template <typename _Item> class ListLocked;
      ^~~~~~~~
/usr/include/directfb/direct/Types++.h:48:6: error: template with C linkage
      template <typename _Item> class ListSimple;
      ^~~~~~~~
/usr/include/directfb/direct/Types++.h:50:6: error: template with C linkage
      template <typename _CharT> class StringBase;
      ^~~~~~~~
/usr/include/directfb/direct/Types++.h:51:6: error: template with C linkage
      template <typename _CharT> class StringsBase;
      ^~~~~~~~
mv -f src/core/.deps/gource-vbo.Tpo src/core/.deps/gource-vbo.Po
````

Other distros had killed it too:
http://pkgs.fedoraproject.org/cgit/rpms/directfb.git/tree/

ollieparanoid:
The problem is, that SDL2 only provides framebuffer access through DirectFB (SDL1 was able to use framebuffer without directfb). And while most drivers can be used over the newer DRI/DRM kernel API instead of framebuffer, we still need it to support all the phones we have.

How do we move forward with this, would it be acceptable if we can provide a patch for these headers, or do you want to remove it for good (we would need to fork it then)?

PR #2656, opened 2017-11-06
testing/phoronix-test-suite

lramage94:
https://www.phoronix-test-suite.com
The most comprehensive testing and benchmarking platform available for Linux

### DIFF
--- a/testing/phoronix-test-suite/APKBUILD
+++ b/testing/phoronix-test-suite/APKBUILD
@@ -1,0 +1,21 @@
+# Contributor: Lucas Ramage <ramage.lucas@openmailbox.org>
+# Maintainer: Lucas Ramage <ramage.lucas@openmailbox.org>
+pkgname=phoronix-test-suite
+pkgver=7.4.0
+pkgrel=0
+pkgdesc="The most comprehensive testing and benchmarking platform available for Linux"
+url="https://www.phoronix-test-suite.com"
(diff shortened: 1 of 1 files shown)


testing/phoronix-test-suite/APKBUILD
```diff
@@ -0,0 +1,21 @@
+# Contributor: Lucas Ramage <ramage.lucas@openmailbox.org>
+# Maintainer: Lucas Ramage <ramage.lucas@openmailbox.org>
+pkgname=phoronix-test-suite
+pkgver=7.4.0
+pkgrel=0
+pkgdesc="The most comprehensive testing and benchmarking platform available for Linux"
+url="https://www.phoronix-test-suite.com"
+arch="noarch"
+license="GPL-3.0"
+depends="php"
```
jirutka: There’s no package named `php`. Only `php5` (legacy) and `php7`.

lramage94: I was under the assumption that php was a virtual package, provided by either php5 or php7. This package will work with both. How should I proceed?

lramage94: Nevermind, I updated the dependencies using the ones from upstream.

https://github.com/phoronix-test-suite/phoronix-test-suite/blob/master/phoronix-test-suite

PR #2657, opened 2017-11-06
testing/proftpd: disable wtmp

xpecex:


ncopa:
Thank you! You need to bump `pkgrel` or the builder will not pick up the change.

Please also update the commit message: `testing/proftpd: ....`

jirutka:
I guess that this is related to the recent CVE…?

jirutka:
Why do you want to disable `wtmp`? For security reasons? Please write a reason to the commit message.

ncopa:
I fixed the commit message and pushed. thanks!

PR #2669, opened 2017-11-07
community/docker: upgrade to 17.10.0

TBK:


martell:
@ncopa would be great if we can get this in before 3.7
17.10 adds support for multi arch images as noted here.
https://github.com/docker/cli/pull/138#issuecomment-340615545
and will be great for cross bootstrapping alpine and some other magic tricks :)

PR #2674, opened 2017-11-07
main/smartmontools: upgrade to 6.6; modernize

TBK:


ncopa:
Thank you for reporting and submitting patch upstream. :+1:

TBK:
@ncopa No problem. Btw is there a place I can read about how to use the lua-aports tools?

Several of my PRs are stuck atm because they need to bump the dependent packages. I am not fond of having to do it manually.

PR #2675, opened 2017-11-08
community/glide: upgrade to v0.13.1

leprechau:
Updated glide to latest release and added check function to execute upstream tests.

### DIFF
--- a/community/glide/APKBUILD
+++ b/community/glide/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Aaron Hurt <ahurt@ena.com>
 # Maintainer: Aaron Hurt <ahurt@ena.com>
 pkgname="glide"
-pkgver=0.13.0
+pkgver=0.13.1
 pkgrel=0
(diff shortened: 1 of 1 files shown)


community/glide/APKBUILD
```diff
@@ -26,9 +26,15 @@ build() {
 	go build -o glide -ldflags "-X main.version=$pkgver" glide.go || return 1
```
TBK: Commit (13e67c291ac92df3239e6bcd7dc76eeae96e5d2e) by @rgdoliveira upgraded Glide to 0.13.1.

So this PR could be updated to add the check.

` || return 1` should be removed.

clandmeter:
I'm closing this as its stale. If somebody want to add a check please create a new PR.

leprechau:
Not really sure what the problem was with this PR ... did the package ever get updated?

leprechau:
Nevermind I see the referenced commit for the update.  If nobody else is concerned about the lack of tests I'm good with letting it die.

PR #2682, opened 2017-11-08
testing/grepcidr: new aport

vincentmele:
http://www.pc-tools.net/unix/grepcidr/
Filter IPv4 and IPv6 addresses matching CIDR patterns

### DIFF
--- a/testing/grepcidr/APKBUILD
+++ b/testing/grepcidr/APKBUILD
@@ -5,14 +5,13 @@ pkgver=2.0
 pkgrel=0
 pkgdesc="Filter IPv4 and IPv6 addresses matching CIDR patterns"
 url="http://www.pc-tools.net/unix/grepcidr/"
-arch="x86_64"
+arch="all"
 license="GPL3"
 depends=""
 makedepends=""
 install=""
 subpackages=""
-options="!check"
-source="http://www.pc-tools.net/files/unix/grepcidr-2.0.tar.gz"
+source="http://www.pc-tools.net/files/unix/$pkgname-$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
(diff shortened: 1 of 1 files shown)


fcolista:
Thanks for your contribution!
Before applying this patch, I've some comments:

`arch="x86_64"`
There's a strong reason why only x86_64 is enabled?
You can set "all" in order to build this package for all the architecture supported by Alpine.
See: [APKBUILD:arch](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#arch)

`options="!check"`
There's a reason why check is disabled? 

`source="http://www.pc-tools.net/files/unix/grepcidr-2.0.tar.gz"`
We'd prefer to use variables here, so when will come to bump a new version, it's enough changing the $pkgver. So, the filename should be $pkgname-$pkgver.tar.gz
See:[APKBUILD:source](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#source)

`mv "$builddir"/grepcidr "$pkgdir"/bin`
This should go to /usr/bin rather than /bin.

Thanks a lot!
.:Francesco

vincentmele:
Thank you for very much for reviewing and suggesting changes! I will make the changes and update this request.

testing/grepcidr/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	make
+}
+
+check() {
+	cd "$builddir"
+	if [ -x grepcidr ]; then echo OK; else exit 1; fi
```
jirutka: This is not an useful check. Could you run grepcidr, e.g. `./grepcidr --version`?

kaniini: Better yet, it would be good to run grepcidr against some sample data and confirm it behaves as expected.

PR #2684, opened 2017-11-08
main/fish: fixes #8112, modernize and add check

rgdoliveira:


rgdoliveira:
Tests are failing. Will investigate what is happening...

PR #2685, opened 2017-11-09
main/zsh: upgrade to 5.4.2

TBK:


main/zsh/APKBUILD
```diff
@@ -2,15 +2,16 @@
...
 pkgrel=0
 pkgdesc="A very advanced and programmable command interpreter (shell)"
 url="http://www.zsh.org/"
 arch="all"
 license="BSD"
+options="!check" # As of 5.4.2 - 3 tests fails
 makedepends="ncurses-dev"
 install="zsh.post-install zsh.post-upgrade zsh.pre-deinstall"
-source="http://www.zsh.org/pub/$pkgname-$pkgver.tar.xz
+source="http://www.zsh.org/pub/$pkgname-$pkgver.tar.gz
```
jbergstroem: strange that `.xz` is missing upstream..

TBK: I thought the same. Guess it is released by hand and somebody forgot.

jbergstroem: I emailed the authors. One minor improvement for this abuild could be switching to https which seems to be working just fine.

TBK: Ok. Which email address did you use? I need it to report the failing tests.

jbergstroem: At the [main site](http://zsh.sourceforge.net) there's a maintainers link in the footer. Not sure if that's most accurate, but I thought it was worth a shot.

jbergstroem: Sourceforge seems to have the `xz` tarball; don't know if its worth switching.

TBK:
@jbergstroem I have changed the source to SF.net since it provices the xz package and offers a faster download.

TBK:
The xz file on SF is giving abuild a stroke, so had to switch to gz.

jirutka:
No, there's no cache enabled in Travis.

jirutka:
Please don't change it to sourceforge, it's not confident and reliable source.

TBK:
@jirutka K, after your previous comment I suspected that. I have just changed it back.

jbergstroem:
Perhaps the tarballs are generated differently. The SHASUM at least seems to point at that.

PR #2695, opened 2017-11-10
main/s6: upgrade to 2.6.1.1

andypost:
It needs updated skalibs https://skarnet.org/software/s6/upgrade.html


andypost:
skalibs upgrade should went in first

PR #2707, opened 2017-11-12
 testing/litespeed: upgrade to 1.4.28

vakartel:


vakartel:
@jirutka 
I just upgrade litespeed to 1.4.28 and made a patch to build with libressl2.6 but push it to my local old 'litespeed' branch. And then tried to combine it all.

Looks awful.

I think it's better to drop this pull request and my local 'litespeed' branch and then create new fresh one and push a commit there.

PR #2709, opened 2017-11-13
main/lxc: fix running systemd based containers under OpenRC / update initd for 2.1 config

itoffshore:
* `systemd` based containers require the `/sys/fs/cgroup/systemd cgroup`
  for `lxc-console` & services inside the containers to work

* `systemd_ctr()` based on recommendations from lxc:

lxc/lxc#1704

* initd use: `lxc.uts.name` & `lxc.rootfs.path`

### DIFF
--- a/main/lxc/lxc.initd
+++ b/main/lxc/lxc.initd
@@ -3,7 +3,10 @@
 # Distributed under the terms of the GNU General Public License v2
 # $Header: /var/cvsroot/gentoo-x86/app-emulation/lxc/files/lxc.initd.2,v 1.5 2012/07/21 05:07:15 flameeyes Exp $
 
+description="Linux Containers (LXC)"
+description_reboot="Reboot containers"
 CONTAINER=${SVCNAME#*.}
+: ${lxc_group:=$LXC_GROUP}
 command="/usr/bin/lxc-start"
 pidfile="/var/run/lxc/$CONTAINER.pid"
 extra_started_commands="reboot"
@@ -51,15 +54,36 @@ checkconfig() {
 	fi
 }
 
+systemd_ctr() {
+	# required for lxc-console & services inside systemd containers
+	local cgroup=/sys/fs/cgroup/systemd cmd=$1
+	local mnt_opts='rw,nosuid,nodev,noexec,relatime,none,name=systemd'
+
+	case "$cmd" in
+		 mount) checkpath -d $cgroup
+			if ! mount | grep $cgroup 1>/dev/null; then
+				mount -t cgroup -o $mnt_opts cgroup $cgroup
+			fi
+			;;
+	       unmount) if mount | grep $cgroup 1>/dev/null; then
+				umount $cgroup
+			fi
+			;;
+	esac
+}
+
 _autostart() {
 	ebegin "$1 LXC containers"
 	shift
-	lxc-autostart --group "${LXC_GROUP:-onboot,}" "$@"
+	lxc-autostart --group "$lxc_group" "$@"
 	eend $?
 }
 
 start() {
 	checkconfig || return 1
+	if yesno "$systemd_container"; then
+		systemd_ctr mount
+	fi
 	if [ -z "$CONTAINER" ]; then
 		_autostart "Starting"
 		return
@@ -97,6 +121,8 @@ start() {
 
 stop() {
 	checkconfig || return 1
+	systemd_ctr unmount
+
 	if [ -z "$CONTAINER" ]; then
 		_autostart "Stopping" --shutdown --timeout ${LXC_TIMEOUT:-30}
 		return
--- a/main/lxc/lxc.initd
+++ b/main/lxc/lxc.initd
@@ -43,10 +43,10 @@ checkconfig() {
 	# no need to output anything, the function takes care of that.
 	[ -z "${CONFIGFILE}" ] && return 1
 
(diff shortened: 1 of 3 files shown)


main/lxc/lxc.initd
```diff
@@ -60,6 +69,9 @@ _autostart() {
 
 start() {
 	checkconfig || return 1
+	if [ -n "${systemd_container}" ]; then
```
ncopa: if you set `systemd_container=no` then will it still run systemd_ctr. Maybe do:

````sh
        if yesno "$systemd_container"; then
                systemd_ctr
        fi
````

And we should probably umount it on lxc stop?

Comment (itoffshore) [reply main/lxc/lxc.initd]: `yesno` is not available so I added a `case` statement instead

added an unmount to `stop()`

File: main/lxc/lxc.initd [hunk 2]
```diff
@@ -51,15 +51,36 @@ checkconfig() {
...
+			fi
+			;;
+	esac
+}
+
 _autostart() {
 	ebegin "$1 LXC containers"
 	shift
-	lxc-autostart --group "${LXC_GROUP:-onboot,}" "$@"
+	lxc-autostart --group "${lxc_group:-onboot,}" "$@"
```
ncopa: this change will break backwards compatibility. It would be nice to have something like:
````sh
: ${lxc_group:=$LXC_GROUP}
````
so that things will still work after upgrade in case `LXC_GROUP` was set

File: main/lxc/lxc.initd [hunk 3]
```diff
@@ -51,15 +51,36 @@ checkconfig() {
...
 	eend $?
 }
 
 start() {
 	checkconfig || return 1
+	case "$systemd_container" in
+		y*|Y*) systemd_ctr mount;;
+	esac
```
ncopa: I prefer using openrc's `yesno` for this. That way things will work even with things like `systemd_container=1`.

````sh
if yesno "$systemd_container"; then
    systemd_ctr mount
fi

Comment (itoffshore) [reply main/lxc/lxc.initd]: `yesno` returns zero for `Y` or `y` so `confd` changed to `ON`

adding the descriptions prevents `rc-service lxc describe` putting warnings in the logs:
```
Nov 16 23:06:00 xxxx daemon.warn /etc/init.d/lxc[28565]: No description for lxc
Nov 16 23:06:00 xxxx daemon.warn /etc/init.d/lxc[28566]: ^[[1m^[[36mreboot^[[m: no description
```

File: main/lxc/lxc.initd [hunk 4]
```diff
@@ -43,31 +46,52 @@ checkconfig() {
...
+		 mount) checkpath -d $cgroup
+			if ! mount | grep $cgroup 1>/dev/null; then
+				mount -t cgroup -o $mnt_opts cgroup $cgroup
+			fi
+			;;
+	       unmount) if mount | grep $cgroup 1>/dev/null; then
+				umount $cgroup
+			fi
+			;;
+	esac
```
jirutka: Please use tabs for indentation, not spaces.

If you use vim, install [vim-editorconfig](https://pkgs.alpinelinux.org/package/edge/community/x86_64/vim-editorconfig). There’s editorconfig plugin for basically all relevant editors.

main/lxc/lxc.initd
jirutka: > `yesno` returns zero for `Y` or `y` so `confd` changed to `ON`

Yes, it returns 0 (aka true) for `Y`, `y`, `yes` and similar, 1 (aka false) for anything else. I really don’t understand why you used `ON` because of this.

itoffshore: it [does not return true](http://linuxhowtos.org/manpages/8/openrc-run.htm) for `Y` or `y`:
```
yesno value
    If value matches YES, TRUE, ON or 1 regardless of case then we return 0, otherwise 1. 
```
I had the config file as `Y` before & it didn't work - which is why I set it to `ON`

main/lxc/lxc.initd
itoffshore: it's tabs with a few spaces so the case statements line up - rather than only tabs & being uneven

itoffshore: with only tabs:
```
        case "$cmd" in
                mount) checkpath -d $cgroup
                        if ! mount | grep $cgroup 1>/dev/null; then
                                mount -t cgroup -o $mnt_opts cgroup $cgroup
                        fi
                        ;;
                unmount) if mount | grep $cgroup 1>/dev/null; then
                                umount $cgroup
                        fi
                        ;;
        esac
```
as it currently is with mount / unmount level with each other:
```
        case "$cmd" in
                 mount) checkpath -d $cgroup
                        if ! mount | grep $cgroup 1>/dev/null; then
                                mount -t cgroup -o $mnt_opts cgroup $cgroup
                        fi
                        ;;
               unmount) if mount | grep $cgroup 1>/dev/null; then
                                umount $cgroup
                        fi
                        ;;
        esac
```

jirutka: That’s wrong alignment. Simple test: change tab width in your editor and look how it gets broken.

When you use tabs (as we do in shell scripts, e.g. to make `<<-EOF` usable), it’s better to resign to alignment with spaces and stick only to tabs (at least in community/team project), because most people can’t (or don’t care to) do it right. And so it quickly becomes total mess. Actually that’s the main reason why most langs/projects prefer spaces for both indentation and alignment.

PR #2715, opened 2017-11-13
community/wrk: upgrade to 4.0.2

itoffshore:
https://github.com/wg/wrk/blob/master/CHANGES

### DIFF
--- a/community/wrk/APKBUILD
+++ b/community/wrk/APKBUILD
@@ -2,8 +2,8 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=wrk
-pkgver=4.0.1
-pkgrel=3
+pkgver=4.0.2
+pkgrel=0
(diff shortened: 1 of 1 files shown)


community/wrk/APKBUILD
```diff
@@ -2,27 +2,30 @@
...
 pkgdesc="wrk is a modern HTTP benchmarking tool"
 url="https://github.com/wg/wrk"
 # requires -march=i586 to build on x86
 arch="x86_64"
 license="ASL 2.0"
 depends=""
 depends_dev=""
-makedepends="libressl-dev"
-install=""
+options="!check"
```
jirutka: **Why** it's disabled?

itoffshore: I tried all of the following:
```
make test
make tests
make check
make checks
```
& ran `grep` in `src` looking for `test` & `check` but didn't find anything

jirutka: Then add a comment like "upstream doesn't provide tests" after the option=.

itoffshore: comment added in https://github.com/alpinelinux/aports/pull/2730

PR #2718, opened 2017-11-14
testing/libneo4j-client: update to version 2.2.0

cleishm:


testing/libneo4j-client/APKBUILD
```diff
@@ -21,17 +21,42 @@ build() {
...
     mkdir -p "$pkgdir"/usr/share/doc/libneo4j-client
     mv doc/html/* "$pkgdir"/usr/share/doc/libneo4j-client/
 }
 
+doc() {
+    depends="$depends_doc"
+    pkgdesc="$pkgdesc (documentation)"
+    install_if="docs $pkgname=$pkgver-r$pkgrel"
+
+    if [ "X$subpkgname" = "Xlibneo4j-client-doc" ]; then
```
jirutka: Oh my… please, this `x` non-sense in `[` is a workaround for a bug in one shell implementation that has been fixed several **decades** ago! It’s not needed anymore for very long time.

testing/libneo4j-client/APKBUILD
```diff
@@ -21,17 +21,42 @@ build() {
...
+        mv "$pkgdir"/usr/share/man "$subpkgdir"/usr/share/
+        find "$subpkgdir"/usr/share/man/ -type f | xargs gzip -9
+    else
+        die "Unknown doc package $pkgname"
+    fi
+
+    # remove if empty, ignore error (not empty)
+    rmdir "$pkgdir"/usr/share "$pkgdir"/usr || true 2>/dev/null
+    return 0
+}
```
jirutka: This is all unnecessarily complicated. We usually don’t make multiple `-doc` subpackages from a single abuild. Why do you need it?

testing/libneo4j-client/APKBUILD
cleishm: Old habits, happy to remove.

testing/libneo4j-client/APKBUILD
cleishm: It's building two different binary packages - a library and a shell client. The library has documentation file (HTML) and the shell has a man page. So I put them in separate doc packages.

Happy to do so differently if you can advise your preference.

PR #2725, opened 2017-11-15
testing/py-astroid: Add missing dependencies

NHellFire:
Packaged all missing dependencies: https://github.com/PyCQA/astroid/blob/master/astroid/__pkginfo__.py#L23

### DIFF
--- a/testing/py-astroid/APKBUILD
+++ b/testing/py-astroid/APKBUILD
@@ -8,6 +8,7 @@ pkgdesc="A new abstract syntax tree from Python's ast"
 url="http://www.astroid.org/"
 arch="noarch"
 license="LGPL2+"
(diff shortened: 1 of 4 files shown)


jirutka:
Tbh, I’d prefer to remove py2 subpackage, i.e. keep it only for Python 3. Do you really need Legacy Python?

NHellFire:
Astroid is a dependency of pylint. Both already had python 2 packages.

On 29 Nov 2017 21:27, "Jakub Jirutka" <notifications@github.com> wrote:

Tbh, I’d prefer to remove py2 subpackage, i.e. keep it only for Python 3.
Do you really need Legacy Python?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/alpinelinux/aports/pull/2725#issuecomment-348001671>,
or mute the thread

PR #2726, opened 2017-11-15
testing/radare2: upgrade to 2.0.1

vakartel:


testing/radare2/APKBUILD
```diff
@@ -4,12 +4,13 @@
...
 pkgname=radare2
-pkgver=1.6.0
-pkgrel=3
+pkgver=2.0.1
+pkgrel=0
 pkgdesc="An opensource, crossplatform reverse engineering framework"
 url="http://www.radare.org"
 arch="all !aarch64"
 license="GPLv3"
+options="!check"
```
jirutka: **Why** it's disabled? Upstream does not provide any tests? Always write a reason as a comment after options=.

ncopa:
I added a comment and pushed. thanks!

PR #2729, opened 2017-11-15
3.6 stable - pull request for "radicale 2.1.8"

74cmonty:
Hello!
Hereby I open a pull request (as instructed by Sören Tempel) for "radicale 2.1.8".
The URL for this application is: http://radicale.org
The URL for the source code is: https://github.com/Kozea/Radicale/archive/2.1.8.tar.gz

THX


ncopa:
looks like the radicale package is in testing.

I think we should:
1) update `testing/radicale` in git master branch to 2.1.8.
2) verify that it works.
3) move it to `community` repo

After that we can consider backporting it to v3.6, however, v3.7 will be out in a few weeks so I don't know if its worth backporting it to v3.6

ncopa:
i pushed radicale 2.1.8 to edge/testing 4c35fcd940ba935930508237bd939e0e8ed4bf85

can you please verify that it works?

74cmonty:
Hello!
The package is working if the init-script will be modified. I have openend a [bug](https://bugs.alpinelinux.org/issues/8196) for this issue.
This issue is caused by parameter _start_stop_daemon_args="--user radicale"_.

However, in my understanding the init-script is launching radicale in **foreground** (= not as daemon).
I could change paramter in init-script to _command_args="--daemon"_ but then Radicale fails to start.

For security reasons the daemon should run as user **radicale** (and not root), right? Currently all files in storage _/var/lib/radicale/collections_ belong to root. 
If yes, what configuration file must be modified? 

I assume umask 0027 should be set before starting the daemon, too.

THX

jirutka:
It should be launched in foreground, that's the init system's job to demonize it.

jirutka:
This PR doesn't make any sense, it's from our 3.6-stable to our master!?

PR #2743, opened 2017-11-17
Update samba and tevent

leo-unglaub:
It's important to update tevent first, because samba depends on the new version.


leo-unglaub:
As I stayed before, the build works fine. Travis just does not keep the build order correctly, that's why the PR is marked as broken.

jirutka:
I’ve merged it in 1ed89e8289a0f2e434b9c3d23b1a36abd4a78c9b and 23cdeab27bff0c0e17ebce1ce3dc793360e327cd. I’ve also added CVEs into secfixes by mistake, these are for 4.7.3.

PR #2744, opened 2017-11-17
base/transmission: fix previously invalid URL, run tests

Xe:


ncopa:
I fixed commit message and bumped pkgrel. thanks!

PR #2755, opened 2017-11-17
testing/dnsdist: new aport

vincentmele:
testing/dnsdist: new aport

https://dnsdist.org
dnsdist is a highly DNS-, DoS-, and abuse-aware loadbalancer.

### DIFF
--- a/testing/dnsdist/APKBUILD
+++ b/testing/dnsdist/APKBUILD
@@ -1,0 +1,54 @@
+# Contributor: Vince Mele <vmele@inoc.com>
+# Maintainer: Vince Mele <vmele@inoc.com>
+pkgname=dnsdist
+pkgver=1.2.0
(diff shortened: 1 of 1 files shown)


jirutka:
Actually if it doesn't really need root privileges, then it definitely should not be started under root as default, but its separate user.

tcely:
`dnsdist` has `--uid` and `--gid` flags for dropping privileges.
 
Examples are at: https://github.com/tcely/dockerhub-powerdns

tcely:
@vincentmele Please merge the suggested changes. https://github.com/vincentmele/aports/pull/1

tcely:
This can be closed as it is now at least a version behind.

ncopa:
We follow this up in #3321 
Thanks!

PR #2756, opened 2017-11-17
main/tig: upgrade to 2.3.0

fabaff:


main/tig/APKBUILD
```diff
@@ -1,16 +1,17 @@
...
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=tig
-pkgver=2.2
+pkgver=2.3.0
 pkgrel=0
 pkgdesc="Text-mode interface for the git revision control system"
 url="http://jonas.nitro.dk/tig"
 arch="all"
 license="GPL2+"
+options="!check"
```
jirutka: Why have you disabled check? There are tests in the tig tarball.

fabaff: Because like on Fedora they are failing.

jirutka: Then write a comment (`options="!check"  # tests are broken`) and explanation to the commit message.

PR #2773, opened 2017-11-19
main/openvswitch: fix ovs-modules initd

itoffshore:
* `openvswitch` module cannot be removed without a reboot so do not unload it


main/openvswitch/ovs-modules.initd
```diff
@@ -3,23 +3,18 @@
...
 	eend $?
 }
 
 stop() {
-        ebegin "Unloading Openvswitch kernel modules"
-        for mod in $MODULES; do
-		rmmod $mod
-	done
-        eend $?
+	# modules cannot be removed with ovs running
```
jirutka: Please write what you stated in PR description here, i.e. cannot be removed without restart. That’s important difference.

itoffshore: changed

PR #2775, opened 2017-11-20
community/wxgtk: fix bug and modernize

tmpfile:


tmpfile:
check fails.

Error loading shared library libwx_gtk2u_core-3.0.so.0: No such file or directory (needed by ./minimal)

Error loading shared library libwx_baseu-3.0.so.0: No such file or directory (needed by ./minimal)

I already tried adding LD_LIBRARY_PATH="$builddir"

PR #2778, opened 2017-11-20
testing/py-coverage: upgrade to 4.4.2

vakartel:
add python3 bindings
remove -tools subpackage


fcolista:
Hi Valerie, there's a reason why the check is disabled here?
Thanks.
.: Francesco

vakartel:
@fcolista 
I have no idea why I disabled it :)
I added a commit with check()

PR #2779, opened 2017-11-21
testing/lxd: upgrade to 2.20

vifino:


testing/lxd/APKBUILD
```diff
@@ -1,7 +1,8 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
+# Contributor: Adrian Pistol <vifino@tty.sh>
```
rgdoliveira: I think the "Contributor" field is for who contributed the package to Alpine

vifino: Not for those who contributed to the package?

jirutka: Yes, who created the abuild or significantly improved. Package bump doesn't fit this category.

vifino: Ah. I did contribute more than just this to the LXD abuild, fixing it up, but I'll recommit without that line.

jirutka: However, the definition is currently very fuzzy and it's actually not used anywhere, so we're thinking about redefining it.

vifino:
Alright. Change made.

PR #2782, opened 2017-11-21
main/gnumeric: add py-libxml2 as makedepends

rgdoliveira:
Add py-libxml2 to fix build error:
File "/usr/bin/itstool", line 27, in <module>
import libxml2
ImportError: No module named libxml2


fabled:
`itstool` should depend on it instead. And it in fact does do so, this looks some other issue in the builder.

ncopa:
I think this fix is wrong. itstool seems to have the proper deps. Might be it was built with "wrong" python earlier due to autodetection.

PR #2783, opened 2017-11-21
testing/mdevd: new aport

jirutka:
@skarnet How do you recommend to run it with OpenRC when I’m not much into adding dependency on execline (the package is quite huge)?


skarnet:
If you won't use execline's `pipeline` command, then you need to live with the shell that spawns the pipeline. It won't take many resources while it lives, but at shutdown time you'll need to make sure that OpenRC actually kills the pipeline and not only the shell.

jirutka:
Superseded by 5333ec4645d3267e2a9169d5b16a562055e23987 and 24cff32974dee2579d7e785d450ed180a4ef0171.

PR #2787, opened 2017-11-22
main/perl-mailtools: point sources to new url

mepholic:
the source tarball for this package seems to no longer exist at the old
location. the url has been updated to point to the correct location.

I tested building this on my desktop, which is running edge.
the checksums didn't change, so we know this is still the original source file.
it might be a good idea to backport this to 3.6.

### DIFF
--- a/main/perl-mailtools/APKBUILD
+++ b/main/perl-mailtools/APKBUILD
@@ -4,7 +4,7 @@
 pkgname=perl-mailtools
 _pkgreal=MailTools
 pkgver=2.18
-pkgrel=1
+pkgrel=2
 pkgdesc="Various e-mail related modules"
 url="http://search.cpan.org/dist/MailTools/"
 arch="noarch"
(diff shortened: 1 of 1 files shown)


main/perl-mailtools/APKBUILD
```diff
@@ -32,6 +32,11 @@ build() {
 	make && make test
 }
 
+check() {
+	cd "$_builddir"
+	make test || return 1
```
TBK: The ` || return 1` parts are no longer needed and can be removed.

main/perl-mailtools/APKBUILD
```diff
@@ -32,6 +32,11 @@ build() {
...
+}
+
 package() {
 	cd "$_builddir"
 	make DESTDIR="$pkgdir" install || return 1
```
TBK: `|| return 1`

main/perl-mailtools/APKBUILD
```diff
@@ -32,6 +32,11 @@ build() {
 	make && make test
```
TBK: `&& make test` should be removed since it is done in the `check()` function

main/perl-mailtools/APKBUILD
```diff
@@ -16,7 +16,7 @@ provides="perl-mail-tools=$pkgver"
 depends="$cpandepends"
 makedepends="perl-dev $cpanmakedepends"
 subpackages="$pkgname-doc"
-source="http://search.cpan.org/CPAN/authors/id/M/MA/MARKOV/$_pkgreal-$pkgver.tar.gz"
+source="http://cpan.metacpan.org/authors/id/M/MA/MARKOV/$_pkgreal-$pkgver.tar.gz"
 
 _builddir="$srcdir/$_pkgreal-$pkgver"
```
TBK: Please rename all occurrences of `_builddir` to `builddir`

mepholic:
@TBK I've made the changes you've requested.

PR #2793, opened 2017-11-23
testing/py-incremental: new aport

n3mur1t0r:
https://github.com/twisted/incremental
A library for versioning your Python projects.

Add py-incremental runtime dependency to py-twisted, fix #7157


n3mur1t0r:
Sorry, I know the build fails because the package I added doesn't yet exist in the repository. However, once it will have been built it shouldn't fail.

Should I remove the dependency from twisted for now, so that the check succeeds?

TBK:
New aports start their life in testing.

n3mur1t0r:
done :)

jirutka:
Please read https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python. If you don’t necessarily need support for Legacy Python, then build the new modules only for Python 3.
Use only tabs for indentation (and alignment), don’t mix it with spaces.

PR #2823, opened 2017-11-27
main/bash: move development data to -dev and some cleanups

vakartel:
add check() function
change pre/post-* script's interpreter to /bin/bash

### DIFF
--- a/main/bash/bash.post-install
+++ b/main/bash/bash.post-install
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/bin/sh
 add-shell '/bin/bash'
 exit 0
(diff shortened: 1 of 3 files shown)


main/bash/bash.post-install
```diff
@@ -1,3 +1,3 @@
-#!/bin/sh
+#!/bin/bash
 add-shell '/bin/bash'
 exit 0
```
ncopa: why? I cannot see any good reason to why use bash for this?

Please revert this back to `#!/bin/sh`

main/bash/bash.pre-deinstall
```diff
@@ -1,3 +1,3 @@
-#!/bin/sh
+#!/bin/bash
```
ncopa: please revert this back to `#!/bin/sh`.

vakartel: I did so, to remove package dependency to /bin/sh. But it really have no much sense, so I revert it.

PR #2849, opened 2017-11-29
testing/py-daemonize: new aport

z3ntu:
https://github.com/thesharp/daemonize
Library to enable your code run as a daemon process on Unix-like systems.

### DIFF
--- a/testing/py-daemonize/APKBUILD
+++ b/testing/py-daemonize/APKBUILD
@@ -1,3 +1,4 @@
+# Maintainer: Luca Weiss <luca (at) z3ntu (dot) xyz>
 pkgname=py-daemonize
 _pkgname=daemonize
 pkgver=2.4.7
--- a/testing/py-daemonize/APKBUILD
+++ b/testing/py-daemonize/APKBUILD
@@ -1,7 +1,7 @@
 pkgname=py-daemonize
 _pkgname=daemonize
 pkgver=2.4.7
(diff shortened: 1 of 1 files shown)


fcolista:
Hi, thanks for your contribution!
Can you please check this points?
> +pkgrel=1
pkgrel should starts from 0.

> +options="!check"
Why check is disabled here? I've tried to add check() and it works without issue.
As last point:
There's no maintainer/contributor.
Even if it is not mandatory for packages in testing, If you want to move this in the next future to the community repository a maintainer is needed.
Once this will be fixed, we will be happy to merge it.
Thanks again!
.: Francesco

z3ntu:
> pkgrel should starts from 0.
Done (I'm getting confused between Arch Linux and Alpine 😄)
> Why check is disabled here? I've tried to add check() and it works without issue.
Running `python3 setup.py test` doesn't actually run any tests (`Ran 0 tests in 0.000s`)
> There's no maintainer/contributor.
Should I add myself as contributor?

rnalrd:
> Should I add myself as contributor?

Yes please, or as a maintainer if you can.

z3ntu:
@rnalrd According to the wiki page being a maintainer means:
* fix bugs in package
* keep it updated. This includes follow upstream announcemt list, rss feed etc and provide updates to Alpine Linux in timely fashion.

I can do that :)

PR #2762, opened 2017-11-19
testing/gdal: Support PostgreSQL and SQLite drivers

mroutis:
Enable support for the following drivers:

- PostgreSQL
- SQLite

PR #2850, opened 2017-11-29
testing/gdal: add sqlite-dev to support MBTiles driver

klinden:
Add sqlite-dev to the list of make dependencies so that GDAL supports the MBTiles driver.


Ikke:
Hello klinden,

Thanks for your contribution. I was about to suggest adding a check function, but I noticed you already did that.

The reason this one failed is because you have to call the program in the correct directory, like this:

```bash
check() {
        cd "$builddir"
        apps/gdal-config --version
}
```

klinden:
@Ikke Thanks for your help with that. I updated the check and it should be working now.

mroutis:
@klinden, I have this pull request that solves that issue too: https://github.com/alpinelinux/aports/pull/2762

I just need the PostgreSQL driver, but added the rest of the drivers.

How about merging it in one pull request?

I'll add the test check that you added!

klinden:
@mroutis Fine by me. It looks like what you have will accomplish what I need.

PR #2866, opened 2017-11-30
main/grub: move bash-completion data to subpackage and more

vakartel:
- move /etc/grub.d/README to -doc subpackage
- point localedir to /usr/share/grub, so we don't need to create /usr/share/locale empty unused directory in package tree and have no abuild warnings about it

### DIFF
--- a/main/grub/APKBUILD
+++ b/main/grub/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Timo Teräs <timo.teras@iki.fi>
 pkgname=grub
 pkgver=2.02
-pkgrel=3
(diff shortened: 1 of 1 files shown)


vakartel:
TravisCI stuck on build:
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over). The job has been terminated

kaniini:
:-1: i don't think it is a good idea to move /etc/grub.d/README to -doc subpackage

vakartel:
I see no sence to keep this file in. But ok, I reverted it.

PR #2873, opened 2017-12-01
community/php7: upgrade to 7.2.4

TBK:
This PR does
- upgrades PHP next version which has backward incompatible changes
- creates `community/php7-mcrypt`  as this extension no longer shipped
- removes `testing/php7-libsodium` as now it shipped with main package
- moves `testing/php7-cassandra` to unmaintained because it has no PHP 7.2 compatibility
- bumps all php package versions

PR #2895, opened 2017-12-03
community/phpmyadmin: upgrade to 4.7.6

andypost:
Since 4.7.4 it should be compatible with PHP 7.2 https://github.com/phpmyadmin/phpmyadmin/pull/13414


andypost:
Prepare for https://github.com/alpinelinux/aports/pull/2873

PS: There's no way to run tests because they require mysql server

PR #2896, opened 2017-12-03
testing/composer: upgrade to 1.5.5

andypost:


andypost:
Looks no way to pass `diagnose` because a user has no composer keys so kinda `composer -V` should work in `check()`

andypost:
required packages added because 
```
  [Composer\Exception\NoSslException]                                                                                                                                                                               
  The openssl extension is required for SSL/TLS protection but is not available. If you can not enable the openssl extension, you can disable this error, at your own risk, by setting the 'disable-tls' option to  
   true. 
```
and
```
The zlib extension is not loaded, this can slow down Composer a lot.
If possible, enable it or recompile php with --with-zlib
```

skwashd:
@andypost thanks for the patch. I have been traveling over the weekend haven’t had a chance to work on the upgrade. I am planning on switching the dependency to php 7 as 5.6 goes EOL in 2018.

andypost:
@skwashd upgrade of dependencies should go as separate patch & soon #2873 we switch to 7.2

PR #2898, opened 2017-12-04
testing/sssd: new aport

vincentmele:
https://pagure.io/SSSD/sssd
System Security Services Daemon
SSSD provides a set of daemons to manage access to remote directories and authentication mechanisms such as LDAP, Kerberos or FreeIPA. It provides an NSS and PAM interface toward the system and a pluggable backend system to connect to multiple different account sources.


vincentmele:
FYI: init.d script does not work (yet). Working on it.

vincentmele:
init.d script fixed and working. Still no valid checks yet.

ncopa:
Can you please merge the commits into 2? one with commit message `testing/ding-libs: new aport` and one with `testing/sssd: new aport`

testing/sssd/APKBUILD
```diff
@@ -0,0 +1,105 @@
...
+	c-ares \
+	curl \
+	cyrus-sasl-gssapi \
+	ding-libs \
+	http-parser \
+	jansson \
+	libnl \
+	libunistring \
+	nss \
+	"
```
ncopa: I am not sure how many of those runtime depends you need to specify. `abuild` will autodetect shared libraries for you, but if you for example need `/usr/bin/curl` you need to specify `curl` in `depends`.

testing/ding-libs/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+# Maintainer: Vince Mele <vmele@inoc.com>
+pkgname=ding-libs
+pkgver=0.6.1
+pkgrel=0
+pkgdesc="'D is not GLib' utility libraries"
+url="https://pagure.io/SSSD/ding-libs/"
+arch="all"
+license="GPL3 LGPL3"
+depends=""
+makedepends="icu-dev gettext-dev doxygen"
```
HRio: @vincentmele test built this without doxygen with same? results as without it, unused make dependency?

testing/sssd/APKBUILD
HRio: @ncopa it would seem arch-linux also has curl as a runtime dependency, that dependency was added in arch for 1.15.3 
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/sssd

it would seem the rest of the depends list quite closely matches that of the arch packaging.

HRio:
@vincentmele thanks! really nice of you to take the time to package this for Alpine. Did you see the request from @ncopa to squash it into two commits in total (one per package)?

testing/sssd/APKBUILD
```diff
@@ -0,0 +1,114 @@
...
+	python2-dev \
+	python3-dev \
+	nss-dev \
+	openldap-dev \
+	krb5-dev \
+        keyutils-dev \
+	check \
+	"
+install=""
+subpackages=""
```
HRio: subpackages="$pkgname-openrc"

PR #2899, opened 2017-12-05
testing/kubernetes: update to version 1.8.5

mati865:
Changed source download location because release 1.8.4 on GitHub [is broken](https://github.com/kubernetes/kubernetes/issues/56346).

I don't see the need for `build-with-debug-info.patch` so I removed it from `source` but left patch untouched in case it will be required in the future.

### DIFF
--- a/testing/kubernetes/APKBUILD
+++ b/testing/kubernetes/APKBUILD
@@ -1,24 +1,29 @@
+# Contributor: Mateusz Mikuła <matti@marinelayer.io>
 # Contributor: Francesco Colista <fcolista@alpinelinux.org>
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=kubernetes
-pkgver=1.7.4
+pkgver=1.8.5
 pkgrel=0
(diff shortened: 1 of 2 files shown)


testing/kubernetes/APKBUILD
```diff
@@ -1,24 +1,29 @@
...
-source="$pkgname-$pkgver.tar.gz::https://github.com/$pkgname/$pkgname/archive/v$pkgver.tar.gz
-no-gnu-grep.patch
-make-e2e_node-run-over-distro-bins.patch
-make-test-cmd-run-over-hyperkube-based-kubectl.patch
-build-with-debug-info.patch
-fix-support-for-ppc64le.patch
-remove-apiserver-add-kube-prefix-for-hyperkube.patch"
+makedepends="bash coreutils findutils go linux-headers rsync"
+source="$pkgname-$pkgver.tar.gz::https://dl.k8s.io/v1.8.4/kubernetes-src.tar.gz
+	no-gnu-grep.patch
```
mati865: This patch can be removed if GNU grep is used during the build.

PR #2901, opened 2017-12-06
main/nodejs: upgrade to 8.9.2

timbru31:


main/nodejs/APKBUILD
```diff
@@ -5,16 +5,10 @@
 # Contributor: Tadahisa Kamijo <kamijin@live.jp>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 #
-# secfixes:
-#   6.11.1-r0:
-#     - CVE-2017-1000381
-#   6.11.5-r0:
-#     - CVE-2017-14919
-#
```
jirutka: Why did you remove secfixes?

timbru31:
IMO they are no longer relevant, since they are for v6, but Node.js LTS is already at v8.
If you want, I can restore it.

jirutka:
They are relevant, we don't remove old records for secfixes.

timbru31:
Restored.

PR #2910, opened 2017-12-09
community/mediainfo: upgrade to 17.10

dawidd6:
It needs first #2909 to be merged


TBK:
If you combine the two PRs then both will succede.

PR #2912, opened 2017-12-10
libressl: fix cross compile build

strfry:
LibreSSL depends on the autotools suite.
When cross-building with scripts/bootstrap.sh, libressl fails to build because it expects these packets to be installed on the target architecture, where they're not available yet.
I moved them from makedepends_host to makedepends_build, so it uses the versions for the host architecture.


kaniini:
it looks good to me

PR #2922, opened 2017-12-12
main/uwsgi: add php plugin support

MorphBonehunter:
There are 2 package requests open and i thould i could fullfill them:
https://bugs.alpinelinux.org/issues/6602
https://bugs.alpinelinux.org/issues/7061
The changes extend the uwsgi with two new packages:
* uwsgi-php5
* uwsgi-php7

### DIFF
--- a/main/uwsgi/APKBUILD
+++ b/main/uwsgi/APKBUILD
@@ -19,17 +19,24 @@ makedepends="
 	geoip-dev
 	jansson-dev
 	libcap-dev
+	libedit-dev
 	linux-headers
 	linux-pam-dev
 	libxml2-dev
 	libxslt-dev
 	lua5.2-dev
+	ncurses-dev
 	paxmark
 	pcre-dev
+	php5-dev
+	php5-embed
+	php7-dev
+	php7-embed
 	postgresql-dev
 	python
 	python2-dev
 	python3-dev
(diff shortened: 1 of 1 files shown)


jirutka:
Please fix the commit message (amend and force push), uwsgi is in main, not testing.

MorphBonehunter:
sorry...first contribution...to much copy & paste from contribution guide...

rnalrd:
pkgrel needs +1

rnalrd:
I actually I took care of it, also slightly modified commit message so that bugs got resolved automatically

Thanks!

MorphBonehunter:
So this means that uwsgi will never get php support because of the do live in different repos?
So this should definitely mentioned in the contribution guide (can not find this somewhere).

jirutka:
> So this means that uwsgi will never get php support because of the do live in different repos?

Yeah, it looks like that. We may consider moving uwsgi to the community repository.

> So this should definitely mentioned in the contribution guide (can not find this somewhere).

Agree. You’d also know it from CI, if it didn’t failed due to missing `check()` function before even trying to build it. I’ve already reverted this bad decision in d903a94387dd251b043b8b50c57095d0ef498f38, partially as a reaction to this incident.

PR #2931, opened 2017-12-14
main/qemu: upgrade to 2.11.0

andypost:


jirutka:
There’s a problem with patch:

```
>>> qemu: 0001-linux-user-fix-build-with-musl-on-ppc64le.patch
patching file linux-user/host/ppc64/hostdep.h
can't find file to patch at input line 17
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|--- a/user-exec.c
|+++ a/user-exec.c
--------------------------
File to patch: 
```

nmeum:
Ping?

nmeum:
There is also a pending patch by daniel on patchwork which doesn't seem to have that problem https://patchwork.alpinelinux.org/patch/3794/

andypost:
Already updated

PR #2936, opened 2017-12-14
community/php7: build with IMAGETYPE_SWC const

bborysenko:
Fix for [redmine-8299](https://bugs.alpinelinux.org/issues/8299).

```
php > echo IMAGETYPE_SWC;
PHP Notice:  Use of undefined constant IMAGETYPE_SWC - assumed 'IMAGETYPE_SWC' in php shell code on line 1
IMAGETYPE_SWC
```

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -93,6 +93,7 @@ source="http://php.net/distributions/$_pkgreal-$pkgver.tar.bz2
 	php7-fpm-version-suffix.patch
 	allow-build-recode-and-imap-together.patch
 	fix-tests-devserver.patch
+	ext-standard-image.patch
 	"
 builddir="$srcdir/$_pkgreal-$pkgver"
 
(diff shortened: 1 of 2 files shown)


vakartel:
With this patch all php sapi binaries would be depend on zlib library. So it can be done simplier and without patch. Just make zlib extension static in APKBUILD.

`--- a/APKBUILD	2017-11-30 13:40:17.734184877 +0200
+++ b/APKBUILD	2017-12-15 10:15:42.313314466 +0200
@@ -158,7 +158,6 @@
 	xmlwriter
 	xsl
 	zip
-	zlib
 	"
 for _ext in $_extensions; do
 	case "$_ext" in
@@ -325,7 +324,7 @@
 		--with-xsl=shared \
 		--enable-zip=shared \
 			--with-libzip=/usr \
-		--with-zlib=shared \
+		--with-zlib \
 		--with-zlib-dir=/usr \
 		$@
 	make
`

bborysenko:
Is there is a reason why php built with dynamic zlib? If not, @vakartel could you please update packages in alpine repo?

vakartel:
The reason is to modularize php as possible. To not to install extra unneeded dependencies.

But I really confused, because you point to the bad build behavior, when core part lacks functionality when some extra parts are built dynamically.

I'll try to fix it ASAP.

andypost:
Faced with that one day yep better to bundle "core extensions" statically

vakartel:
I think to move somehow this constant's initialization and the corresponded function into zlib extension.

vakartel:
I saw many distros have zlib static in php. So I made the same
#2951

jirutka:
Do I understand correctly that this is not needed after merging #2951?

andypost:
Yes, static compile of zlib solves that
Ref https://bugs.php.net/bug.php?id=34338

PR #2937, opened 2017-12-15
main/newsboat: new aport (replaces newsbeuter)

mcrute:
The newsbeuter project is [abandoned](https://github.com/akrennmair/newsbeuter/blob/master/README.md) and has been forked to newsboat which is maintained. This adds newsboat as a replacement for newsbeuter. There were a few Alpine patches for newsbeuter that were already in the upstream newsboat so those have been removed.

I didn't see any clear documentation on replacement packages in the wiki so I didn't modify the newsbeuter package.


main/newsboat/APKBUILD
```diff
@@ -0,0 +1,59 @@
...
+    sqlite-dev
+    stfl-dev
+    "
+subpackages="$pkgname-doc $pkgname-lang"
+source="https://github.com/newsboat/newsboat/archive/r$pkgver.tar.gz
+	newsbeuter-2.10.1-ncurses6.patch
+	musl_const.patch
+	"
+
+_builddir="$srcdir/$pkgname-r$pkgver"
```
rnalrd: Please use "builddir"

main/newsboat/APKBUILD
```diff
@@ -0,0 +1,59 @@
...
+
+prepare() {
```
rnalrd: This is now a default action&function for abuild. It can be removed entirely.

mcrute:
Updated as requested. On the provides/replaces is there a wiki page that I missed that specifies the correct behavior for replacing packages per Alpine policy? I didn't see one and would like to add that to the docs for the next contributor.

rnalrd:
Sorry one more change which should I have tell you before:  "provides=" needs to have package version specified, otherwise is considered a virtual package:

provides="newsbeuter=$pkgver"

There's no specific wiki page about package replacing policy. We have only the generic wiki page for [APKBUILD](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference)

mcrute:
Updated the `musl_const.patch` file in preparation to upstream it. `newsbeuter-2.10.1-ncurses6.patch` has already been accepted upstream and should be in their next release.

mcrute:
Updated per comments from @jirutka in the diff.

mcrute:
Is there anything else I need to fix, or is this ready for merge now?

As a side note all of our local patches have now been accepted upstream so once upstream does another release we'll be able to remove all of them.

PR #2945, opened 2017-12-20
testing/bird: upgrade to 2.0.0

Fusl:


Fusl:
Note: The subpackage was removed because bird and bird6 have been merged into a single binary with version 2.0.0

PR #2952, opened 2017-12-22

zx2c4:
```

== Changes ==

  * keygen-html: remove prebuilt file
  
  This follows our mailing list discussion.
  
  * wg-quick: add the "Table" config option
  
  In collaboration with Luis Ressel, wg-quick(8) grew an option! We generally
  do not like to add things to wg-quick or allow feature-creep, but this was
  basic enough and mostly involves disabling functionality. Specifically,
  wg-quick now accepts a Table= parameter with these semantics:
  
    ~ Table=auto (default) selects the current behaviour
    ~ Table=off disables creation of routes from allowed ips altogether
    ~ All other values are passed through to "ip route add"'s table option
  
  This should enable people to do basic policy routing. It also matches the
  functionality provided by LEDE/OpenWRT's uci config as well as NixOS's
  networking configuration.
  
  * wg-quick: dumber matching for default routes
  
  Efficiency.
  
  * crypto: compile on UML
  
  UML allows you to compile a Linux Kernel as a standalone ELF binary that runs
  within normal Linux. WireGuard can now be compiled as a normal Linux program,
  runnable on Linux, which is useful for the test suite... and other things.
  
  * compat: kernels < 3.13 modified genl_ops
  
  This fixes a rather important bug with 3.10, 3.11, and 3.12 kernels, where in
  some cases, gcc failed to de-constify a struct that was marked as const when
  it should not have been on on these older kernels, triggering an oops at
  module insertion time.
```

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20171211
-_mypkgrel=0
+_ver=0.0.20171221
+_mypkgrel=1
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


ncopa:
i think this was applied with 7c3dcfd5450d7d57e495a57dd6673fc41924440c

PR #2957, opened 2017-12-22
main/vim: upgrade to 8.0.1422

mcrute:


jirutka:
This aport has been already upgraded to 8.0.1424.

mcrute:
Ah, looks like @ncopa got it. Thanks

PR #2961, opened 2017-12-24
main/aws-ena-driver: new aport

mcrute:
Add kernel drivers for AWS EC2 Elastic Network Interface which are maintained out-of-tree by Amazon. These drivers are required to bring-up Alpine on any of the new 5 series instances that AWS recently released as well as the i3 series of instances. I'm only shipping a hardened and a vanilla flavor because virthardened lacks nvme support which is another requirement to bring up Alpine on the newer hardware so I don't see a point shipping the driver if the instance bring-up will still fail.

I've tested this on the 5 series instances, the i3 instances and a few of the older series of instances as well to confirm that it works.


jirutka:
All new aports must go to the testing repository first.

mcrute:
Ah, yeah sorry about that. I forgot about this one. Moved to testing.

PR #2962, opened 2017-12-24
main/tiny-ec2-bootstrap: new aport

mcrute:
This project supports bootstrapping an EC2 instance in a similar manner to Ubuntu's cloud-init but doesn't have any dependencies outside of Busybox, whereas the former pulls in Python and a lot of other libraries.


kaniini:
Since you're upstream for this software, I assume the packaged version already works and is acceptable for inclusion in main?

mcrute:
Yes, it works and I've tested it on quite a few instances and types. It's maintained and the stability should be suitable for main.

jirutka:
All new packages must go to the testing repository first. After it passes all builders, we can move it to upper repository.

mcrute:
Moved to testing, also corrected to be a `noarch` build as there's only scripts in the final package. What's the process for moving this to main? I just want to make sure I'm doing the right things.

PR #2964, opened 2017-12-25
community/wine: upgrade to 3.0

martell:
preparing for 3.0 release


martell:
@jirutka I think this should be in good shape to merge.
Will do a followup in the coming week or so with arm and aarch64 support when I get the chance.

PR #2967, opened 2017-12-25
main/gnupg: enable tofu

tcely:
When sqlite-dev isn't present tofu gets disabled.


ncopa:
When this kind of changes are made we need to bump pkgrel. Otherwise the builders will not pick it up and if thei did you would not be able to know if `gnupg-2.2.3-r0` has support for tofu as it would depend on when it was built.

I think we should also make configure to error if sqlite is missing instead of only depend on the right packages and hope that gnupg finds them. (I just noticed that gnupg does not abort with error with `--enable-tofu` if sqlite is missing. it just builds it without tofu support. oh well...)

PR #2971, opened 2017-12-27
main/llvm5: upgrade to 5.0.1

am11:
This fixes a __declspec issue https://bugs.llvm.org/show_bug.cgi?id=35080.

/cc @jirutka


martell:
@am11 Don't forget to update the other llvm components.

community/lld
main/compiler-rt
and 
community/llvm-libunwind

I have changes for aports/testing/libc++ which can be kept separate for now.
That is still on version 4.0 actually much like lldb.

PR #2973, opened 2017-12-27
testing/aws-ena-driver-[vanilla/hardened]: new aport

mcrute:
Add kernel drivers for AWS EC2 Elastic Network Interface which are maintained out-of-tree by Amazon. These drivers are required to bring-up Alpine on any of the new 5 series instances that AWS recently released as well as the i3 series of instances. I'm only shipping a hardened and a vanilla flavor because virthardened lacks nvme support which is another requirement to bring up Alpine on the newer hardware so I don't see a point shipping the driver if the instance bring-up will still fail.

I've tested this on the 5 series instances, the i3 instances and a few of the older series of instances as well to confirm that it works.

This is related to work I've been doing to build a functional [Alpine EC2 AMI](https://github.com/mcrute/alpine-ec2-ami) and I've been testing it out of a private repository. It works when installed on an instance but I didn't have many other good examples of kernel modules that needed more than just the kernel headers to build so I did what I think makes sense and will update if you have feedback.

Ideally I'd like this to be a single multi-flavor package that builds subpackages for the kernel variants but installing both kernel dev packages at the same time seemed to confuse the busybox trigger. If there's a better way to do that I'd be happy to refactor the package.

Supersedes #2961


kaniini:
I merged this by mistake while I was playing with the aports, sorry.

This needs to be adjusted to take kernel version and ABI into it's version.  Right now we don't have the greatest approach for this, but you could look at the wireguard module packages as an example.

mcrute:
Updated to co-version with the kernel as wireguard does. I opted not to duplicate the kernel version in my APKBUILD file as that seemed a bit error prone and did not source the kernel APKBUILD because I don't really want to pull in unrelated functionality accidentally, thus the eval/sed approach.

Aside from the flavor the APKBUILD files for the two packages are identical so rather than duplicate them I sourced one into the other. I'm happy to go back to duplicating them if that's the preferred way to handle flavors it just struck me as a recipe for accidental divergence over time.

Let me know if there's any other changes required to merge this.

PR #2960, opened 2017-12-23
main/boost: upgrade to 1.66.0

andypost:

PR #2984, opened 2017-12-29
testing/boost-build: upgrade to 1.66.0

vakartel:


testing/boost-build/APKBUILD
```diff
@@ -1,38 +1,35 @@
...
 url="http://boost.org/build"
 arch="all"
+options="!check"
 license="custom:boost"
 depends="boost-dev"
 makedepends=""
 subpackages="$pkgname-doc"
 source="$pkgname-$pkgver.tar.gz::https://github.com/boostorg/${pkgname#*-}/archive/${pkgname%%-*}-$pkgver.tar.gz
-	$pkgname-$pkgver-LICENSE.txt::http://boost.org/LICENSE_1_0.txt"
+	$pkgname-LICENSE.txt::http://boost.org/LICENSE_1_0.txt"
```
jirutka: Why did you change this?

vakartel: Because license file not assigned to package version and not changes between versions.

andypost:
Related https://github.com/alpinelinux/aports/pull/2960

testing/boost-build/APKBUILD
jirutka: That does not matter. What matters is that source names must be unique among all aports in all versions. If Boost change this file or release new version and you keep the current name, it will fail on builders because sources are cached and identified only by name.

PR #2988, opened 2017-12-29
testing/ratpoison: new aport

duncanguthrie:


testing/ratpoison/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: Duncan Guthrie <dguthrie@posteo.net>
+# Maintainer: Duncan Guthrie <dguthrie@posteo.net>
+pkgname=ratpoison
+pkgver=1.4.9
+pkgrel=0
+pkgdesc="ratpoison: Say good-bye to the rodent"
```
ncopa: The description does not tell anything about what this is

testing/ratpoison/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+url="http://ratpoison.nongnu.org"
+arch="all"
+license="GPL2"
```
ncopa: We recently changed to the SPDX format of specifying licenses

testing/ratpoison/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+depends="libx11"
```
ncopa: abuild will automatically add a dependency for libx11 so this is not needed

testing/ratpoison/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+makedepends="libx11-dev"
+install=""
+subpackages="$pkgname-dev $pkgname-doc"
+source="http://download.savannah.nongnu.org/releases/$pkgname/$pkgname-$pkgver.tar.xz"
+builddir="$srcdir/$pkgname-$pkgver"
+options="!check"
```
ncopa: We normally add a check function to do a minimal verification that software runs. It may be as simple as just print the version number or help. If this is not possible, then it would be nice to have a comment on why the check was disabled.

ncopa:
i fixed it for you. thanks!

PR #2408, opened 2017-09-29
testing/netdata: improve shutdown

HRio:

PR #2999, opened 2017-12-30
testing/netdata: improve shutdown

HRio:
Make sure it actually restarts (entually) even if a plugin does not stop.


HRio:
@rgdoliveira please check

testing/netdata/netdata.initd
```diff
@@ -1,7 +1,13 @@
 #!/sbin/openrc-run
 
 command=/usr/sbin/netdata
-command_args=""
 pidfile=/run/netdata.pid
+command_args="-P ${pidfile}"
 name="netdata"
 description="Real-time performance monitoring"
+retry="TERM/60/KILL/5"
```
ncopa: I'm not sure I like to kill it with sigkill. Would be better to try fix the real issue and report upstream.

HRio: on 1.8 is was this plugin: https://github.com/alpinelinux/aports/pull/2408/files#diff-c819e8a3e0dc7999b6670a318b0c2c01R48

that one seems to give me grief also in 1.9, could be `-hardened` related, as I see problems with disk io starts also when running `-hardened` kernel.

HRio: after testing some more with both `-vanilla` and `-virthardended` is seems we can drop the retry line with 1.9.

PR #3004, opened 2017-12-31
main/doxygen: upgrade to 1.8.14

HRio:


HRio:
@jirutka texlive needed by check(), but texlive is only in testing, can texlive be moved from testing you think?

jirutka:
I prefer to not, texlive package is not a good shape. doxygen does not allow to disable tests that require TeX?

HRio:
OK, I will look into that.

PR #3005, opened 2018-01-01
testing/nheko: new package

Aerdan:
`nheko` is a Qt5-based Matrix client which provides a clean, straightforward interface.


kaniini:
tar.bz2 should probably be tar.gz.

PureTryOut:
Honestly, no it shouldn't. Not in this case anyway. The archive used is manually made by the developer to include the submodules as well. The other archives, as well as the archives provided by Github automatically, are lacking this. There just is no .tar.gz archive that includes all the code required.

Aerdan:
What happened was, the developer wrote a script to produce a release-grade tarball (which includes the submodules), but had neglected to pass `-j` (perhaps assuming that their `tar` would have automatically called `bzip2` for a `.tar.bz2` filename). I pointed out that the script was producing a raw TAR archive, and that issue has since been fixed.

PR #3008, opened 2018-01-03
testing/s3cmd: new aport

gena01:


testing/s3cmd/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+# Maintainer: Gennady Feldman <gena01@gmail.com>
+pkgname=s3cmd
+pkgver=2.0.1
+pkgrel=0
+pkgdesc="Command line tool for managing Amazon S3 and CloudFront services"
+url="https://github.com/s3tools/s3cmd"
+arch="noarch"
+license="GPL-2.0"
+depends="python2 py2-dateutil py2-magic"
+makedepends="python2-dev py-setuptools"
```
ncopa: Can we use python3 instead?

> Python 3+ is also supported starting with S3cmd version 2.

https://github.com/s3tools/s3cmd

gena01:
@ncopa updated. Care to do another round/review?

ncopa:
Do we need support python2 at all?

gena01:
@ncopa I am not sure about python2 since this is my first python package. How are others doing it?

ncopa:
Many other python modules that previously only had support for python2 added python3 support in similar way that you propose. However, we want to remove python2 at some point (far) in future, so I am thinking that we don't add more python2 modules at this point unless it is actually needed.

Also, this looks like a command line tool and user will likely not be interested in what language it is implemented in so we just make sure the runtime dependencies are met. In this case we can chose either python2 or python3 and I am suggesting we chose python3 since we can.

gena01:
@ncopa works for me. updated.

ncopa:
I pushed with this diff:
````diff
diff --git a/testing/s3cmd/APKBUILD b/testing/s3cmd/APKBUILD
index 89ad9ff2d7..0ab0063893 100644
--- a/testing/s3cmd/APKBUILD
+++ b/testing/s3cmd/APKBUILD
@@ -8,7 +8,7 @@ url="https://github.com/s3tools/s3cmd"
 arch="noarch"
 license="GPL-2.0"
 depends="python3 py3-dateutil py3-magic"
-makedepends="python3-dev py-setuptools"
+makedepends="python3-dev py3-setuptools"
 subpackages="$pkgname-doc"
 source="$pkgname-$pkgver.tar.gz::https://github.com/s3tools/s3cmd/archive/v$pkgver.tar.gz"
````

PR #3010, opened 2018-01-03
testing/gdal: enable sqlite support

chambbj:
Enable SQLite support within GDAL by adding sqlite-dev as a make dependency.


fabled:
I think #2762 supersedes this in features.

chambbj:
@fabled The key here was to add `sqlite-dev`, so #2762 works in that regard, assuming they rebase on top of master, bringing the GDAL version to 2.2.3 (the original PR was still on 2.2.2).

mroutis:
oh, 💩 , missed the version number on the merge @chambbj , I'll correct it

PR #3034, opened 2018-01-07
testing/influxdb: upgrade to 1.4.2

stwa:
* bump package version
* improve APKBUILD


testing/influxdb/APKBUILD
```diff
@@ -1,74 +1,60 @@
...
+	mkdir -p ${builddir%/*}
+	mv "$srcdir"/$pkgname-$pkgver "$builddir"/ || return 1
+	default_prepare
 }
 
 build() {
 	export GOPATH="$srcdir"
 	cd "$builddir"
-	python2 build.py --no-get || return 1
+	go get github.com/sparrc/gdm
```
nmeum: Using `go get` is generally a bad idea since it simply fetches the HEAD of the git repository and therefore makes the build irreproducable unless they have a `vendor/` directory in `$builddir`. Please consider packing `gdm` instead (if it isn't already).

clandmeter:
I made some changes to move fetching of sources in prepare and replace go install with build. We don't want to use install because it can fetch updates sources from git repos. I also bumped the version while at it.

stwa:
Thanks for your help! I checked your changes, looks good.

PR #3035, opened 2018-01-07
testing/crda: fix homepage url and source from http:// to https://

stwa:


testing/crda/APKBUILD
```diff
@@ -4,13 +4,13 @@ pkgname=crda
 pkgver=3.18
 pkgrel=0
 pkgdesc="Linux Central Regulatory Domain Agent"
-url="http://www.linuxwireless.org/en/developers/Regulatory/CRDA/"
+url="https://wireless.wiki.kernel.org/en/developers/Regulatory/CRDA"
 arch="all"
 license="GPL"
 depends="wireless-regdb"
-makedepends="linux-headers libnl-dev py2-m2crypto libgcrypt-dev python"
+makedepends="linux-headers libnl-dev py-m2crypto libgcrypt-dev python2"
```
ncopa: I think we should have `py2-m2crypto` and `python2` in the makedepends.

PR #3041, opened 2018-01-08
main/lxc: make initd $logdir global

itoffshore:
* don't hard code `$logdir` path


main/lxc/lxc.initd
```diff
@@ -12,6 +12,8 @@ CONTAINER=${SVCNAME#*.}
 : ${lxc_group:=$LXC_GROUP}
 : ${systemd_container:=no}
 
+LOGDIR=/var/log/lxc
```
jirutka: Please use lower case for variable name.

itoffshore:
changed & rebased

main/lxc/lxc.initd
```diff
@@ -12,6 +12,8 @@ CONTAINER=${SVCNAME#*.}
 : ${lxc_group:=$LXC_GROUP}
 : ${systemd_container:=no}
 
+logdir=/var/log/lxc
```
jirutka: If you want to be able to redefine this variable in the corresponding config file, then this will not help… (see variables above)

main/lxc/lxc.initd
```diff
@@ -96,7 +98,7 @@ start() {
 		return
 	fi
 
-	rm -f /var/log/lxc/${CONTAINER}.log
+	rm -f $logdir/${CONTAINER}.log
```
jirutka: `$logdir` definitely should be quoted.

main/lxc/lxc.initd
itoffshore: I'm not trying to define `$logdir` in any config file (it would not make much sense to do that)

this patch makes the log directory configurable in the `initd` with one change at the top of the script not 3 changes half way down it

main/lxc/lxc.initd
itoffshore: quoted & rebased

main/lxc/lxc.initd
jirutka: > this patch makes the log directory **configurable** in the `initd`

Users should not modify files inside `/etc/init.d`, that’s why there’s `/etc/conf.d` (each runscript implicitly sources corresponding config file in `/etc/conf.d`). If you want to make it configurable, define it in a way that allows changing it from conf.d file.

itoffshore: all of the containers are going to be in the same path depending on the value of `lxc.lxcpath`.

the majority of the containers will probably be started with `autoboot` & are not even going to use the `$logdir` value & will instead log into the container directory.

containers started with an `initd` symlinked to the main script will use the `$logdir` setting - & they will also be in the `lxc.lxcpath` - so it makes no sense to have `$logdir` set in multiple `confd` files - when all you need is a single setting in the main `initd`.

jirutka: `/etc/init.d/lxc.foo` sources both `/etc/conf.d/lxc.foo` and `/etc/conf.d/lxc`.

itoffshore: ok - moved `$logdir` to the `confd`

main/lxc/lxc.initd
```diff
@@ -96,7 +97,7 @@ start() {
 		return
 	fi
 
-	rm -f /var/log/lxc/${CONTAINER}.log
+	rm -f "$logdir"/${CONTAINER}.log
```
jirutka: Why does it remove the previous log file anyway?

itoffshore: if a container starts successfully nothing is logged

itoffshore:
why is this still not pulled ?

PR #3044, opened 2018-01-09
main/mkinitfs: skip apk hooks

HRio:
Skip pre/post apk hooks on diskless initramfs installation.


HRio:
@fabled please review, thanks.

fabled:
Looks good. I would like to confirm it has been tested. And I'm also wondering if should add versioned dependency to `apk-tools>=2.9.1`.

HRio:
Yes, tested on my cubox, that now can use etckeeper and pass boot, by skipping its hooks.

PR #3045, opened 2018-01-09
testing/mbuffer: initial package commit based on version 20171011

Fusl:


kaniini:
please flatten

Fusl:
@kaniini done

testing/mbuffer/APKBUILD
```diff
@@ -0,0 +1,39 @@
+# Contributor: Katie Holly <holly@fuslvz.ws>
+# Maintainer: Katie Holly <holly@fuslvz.ws>
+pkgname=mbuffer
+pkgver=20171011
+pkgrel=0
+pkgdesc="mbuffer buffers I/O operations and displays the throughput rate"
+url="http://www.maier-komor.de/mbuffer.html"
+arch="all"
+license="GPL-3.0"
```
ncopa: We are moving towards [SPDX](https://spdx.org/licenses/) so it should be `license="GPL-3.0-only"`.

testing/mbuffer/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+depends=""
```
ncopa: This row could be removed (but it does not really hurt to keep it either)

testing/mbuffer/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+makedepends="linux-headers libressl-dev"
+install=""
```
ncopa: This could be removed too.

testing/mbuffer/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+subpackages="$pkgname-doc"
+source="
+	http://www.maier-komor.de/software/mbuffer/mbuffer-$pkgver.tgz
+	"
```
ncopa: When there is only a single url we normally have it on a single line:
`source="http://www.maier-komor.de/software/mbuffer/mbuffer-$pkgver.tgz"`

testing/mbuffer/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+
+builddir="$srcdir"/mbuffer-$pkgver
+
+prepare() {
+	cd "$builddir"
+	default_prepare
+}
```
ncopa: The entire `prepare()` function can be removed since the default is to run `default_prepare`

PR #3054, opened 2018-01-10
testing/readline: cleanup APKBUILD, adding inputrc.

gena01:


martell:
hey @gena01, did you notice that the readline-static package is missing from 3.7 ?
I'm wondering since you have been working on this aport.
It seems available on 3.6

martell:
after some digging I see that it was accidentally removed in a previous commit when it was supposed to move to the -dev package.
could you add the following to the -dev package function
`mv "$pkgdir"/usr/lib/*.a "$subpkgdir"/usr/lib/`
so that we may have access to these files again? :)

gena01:
@martell i am not sure about 3.6 it's there in current (before my change). Look at package contents on pkgs.alpinelinux site

/cc @nmeum

gena01:
@fabled sorry, updated commit and PR title.

PR #3057, opened 2018-01-11
testing/rsyslog: add -elasticsearch subpackage

rgerhards:
Elasticsearch is very popular with rsyslog. This PR provides the
necessary feature to make rsyslog talk directly to it.

### DIFF
--- a/main/rsyslog/APKBUILD
+++ b/main/rsyslog/APKBUILD
@@ -12,10 +12,10 @@ arch="all"
 license="GPL-3.0 LGPL-3.0"
 options="!check"
 makedepends="zlib-dev gnutls-dev mariadb-dev postgresql-dev net-snmp-dev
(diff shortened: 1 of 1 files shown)


main/rsyslog/APKBUILD
```diff
@@ -104,6 +104,15 @@ snmp() {
 		"$subpkgdir"/usr/lib/rsyslog/
 }
 
+elasticsearch() {
+	pkgdesc="rsyslog elasticsearch support"
+	makedepends="curl"
```
jirutka: `makedepends` must not be declared inside a split function.

rgerhards: thx -- I thought I had read this would be the correct way to do it. Now I know better :-) Will correct soon, I have seen one other thing also glitched through.

PR #3066, opened 2018-01-16
main/lz4: upgrade to 1.8.1.2

itoffshore:
main/lz4: upgrade to `1.8.1.2`

### DIFF
--- a/main/lz4/APKBUILD
+++ b/main/lz4/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=lz4
-pkgver=1.8.1
+pkgver=1.8.1.2
 pkgrel=0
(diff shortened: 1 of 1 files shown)


main/lz4/APKBUILD
```diff
@@ -1,13 +1,14 @@
...
 pkgname=lz4
-pkgver=1.8.1
+pkgver=1.8.1.2
 pkgrel=0
 pkgdesc="LZ4 is lossless compression algorithm with fast decoder @ multiple GB/s per core."
 url="https://github.com/lz4/lz4"
 arch="all"
 license="BSD"
 checkdepends="diffutils"
+options="!check" # tests take too long for Travis
```
nmeum: Hm, that doesn't necessarily mean that we can't run them on the builders. Is it possible to run a subset of the test suite?

itoffshore: rebased with a shorter test.

I run the full test whenever I update the package

nmeum:
Committed, decided not to change the `check()` function for now, even if the long running tests don't work on travis they *should* work fine on the official builders.

PR #3071, opened 2018-01-17
main/xen: XPTI xsa254

HRio:
Add Xen page-table isolation (XPTI) for XEN 4.9.1

More info: http://xenbits.xen.org/xsa/xsa254/README.pti


HRio:
@ncopa I have build tested and loaded the result on a Alpine 3.7 based dom0. The only test system I have available tonight is AMD based.. but have booted guests with `xpti = 1`

Note:
https://github.com/alpinelinux/aports/commit/1f6e64f5d412c09c7e07b45b3d263c2d7eafc594#diff-d21f431da8c1324fae72aa90228787e6 is missing in master, its only in alpine-3.7. So I have tested without that as this is a master build.

HRio:
@kaniini  can you please have a look?

HRio:
@ncopa will send a PR for the patch missing in master, but is in alpine-3.7 when I have tested it:
https://github.com/HRio/aports/commit/3d69027c17be0fca9ed2491d0a80ae2e4e045842

PR #3074, opened 2018-01-18
main/xen: XSA-254 XPTI

HRio:
Add Xen page-table isolation (XPTI) for XEN 4.7.3

More info: http://xenbits.xen.org/xsa/xsa254/README.pti


HRio:
@ncopa @kaniini tested on a Alpine 3.5 domU with Intel Processor. 16 domUs both 32 and 64-bit.

HRio:
**Note** for the Alpine **3.5** branch

PR #3075, opened 2018-01-19

zx2c4:
* receive: treat packet checking as irrelevant for timers

Small simplification to the state machine, as discussed with Mathias
Hall-Andersen.

* socket: check for null socket before fishing out sport
* wg-quick: ifnames have max len of 15
* tools: plug memleak in config error path

Important bug fixes.

* external-tests: add python implementation

Piotr Lizonczyk has contributed a test vector written in Python.

* poly1305: remove indirect calls

From Samuel Neves, we now are in a better position to mitigate speculative
execution attacks.

* curve25519: modularize implementation
* curve25519: import 32-bit fiat-crypto implementation
* curve25519: import 64-bit hacl-star implementation
* curve25519: resolve symbol clash between fe types
* curve25519: wire up new impls and remove donna
* tools: import new curve25519 implementations
* contrib: keygen-html: update curve25519 implementation

Two of our Curve25519 implementations now use formally verified C. Read this
mailing list post for more information:
https://lists.zx2c4.com/pipermail/wireguard/2018-January/002304.html

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20171221
-_mypkgrel=0
+_ver=0.0.20180118
+_mypkgrel=1
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


nmeum:
Please use a separate commit for each package upgrade and change the commit message to `${repository}/${pkgname}: upgrade to ${pkgver}`.

zx2c4:
This is how every other version bump for WireGuard has gone, and each of them have been accepted exactly as such.

Could you point me to a recent policy change I might not be aware of?

nmeum:
> Could you point me to a recent policy change I might not be aware of?

Well…we don't really have a properly documented policy but in general changes to different packages should be done in different commits. I guess in this particular case you could also do them in one commit.

Different developers handle it differently I guess. Anyway we are currently working on a properly documented policy to clearly document how stuff like this should be done.

zx2c4:
>  I guess in this particular case you could also do them in one commit.

Ah, okay, so I just pushed new commits for no reason then. I'll change it back and push again one single unified commit.

nmeum:
> Ah, okay, so I just pushed new commits for no reason then. I'll change it back and push again one single unified commit.

Sorry, already pushed your changes.

zx2c4:
Strange paper shuffling we've got going here. Okay then. Closing this PR.

PR #3076, opened 2018-01-19
main/cryptsetup: upgrade to 2.0.0

nmeum:
This is a new major version of cryptsetup with also comes with an soname bump. Among the new features of this major release is a new LUKS2 container format. mkinitfs needs a very simple patch to support that format which is included in this commit, if there are no objections I will also push this changes to the mkinitfs repository.

This includes:

* An upgrade of cryptsetup to 2.0.0 including a fix for https://gitlab.com/cryptsetup/cryptsetup/issues/362
* A rebuild of all packages using libcryptsetup
* An upgrade of `testing/argon2` **including the removal of libargon2** (I added a -libs subpackage instead)
    * I didn't manage to find a package which depends on libargon2 so this shouldn't be a problem
* A commit moving `testing/argon2` to `main/`
* A rebuild of `main/cryptsetup` against the system argon2 library
* A backport of the patch for mkinitfs

I tested this on a computer using the LUKS1 format and on one using the LUKS2 format, both seem to work fine. With LUKS2 a warning is omitted when initially creating the container because the APKBUILD doesn't create the lockdir and it is instead created by cryptsetup itself.


nmeum:
The build fails because mkinitfs is still build against cryptsetup 1.7.5 which doesn't define `CRYPT_LUKS`.

main/cryptsetup/APKBUILD
```diff
@@ -43,7 +49,14 @@ package() {
...
 	mkdir -p "$pkgdir"/usr/share/doc/$pkgname/
-	install -m644 README TODO FAQ "$pkgdir"/usr/share/doc/$pkgname/
+	install -m644 README TODO FAQ NEWS docs/v${pkgver}-ReleaseNotes \
+		"$pkgdir"/usr/share/doc/$pkgname/
+
+	# From v2.0.0-ReleaseNotes:
+	#   If your distro does not support tmpfiles.d directory, you
+	#   have to create locking directory (/run/lock/cryptsetup) in
+	#   cryptsetup package (or init scripts).
+	mkdir -p "$pkgdir"/"$_lockdir"
```
jirutka: `/run` is tmpfs, it’s wiped on each restart, so it does not make sense and should not be created by the package, but runscript.

nmeum: Ah, indeed it is actually not even necessary to create the directory in the package since it will in fact be created by cryptsetup. It will just emit a warning when creating the directory (see https://gitlab.com/cryptsetup/cryptsetup/issues/362).

I would suggest simply not creating the directory.

nmeum: I removed the creation of the locking directory from the APKBUILD.

nmeum:
Been using this the last few days on two computers (one of them with LUKS1 and one with LUKS2) and haven't noticed any problems so far. If there aren't any further comments I will assume lazy consent and will push this myself.

PR #3079, opened 2018-01-19
main/tinc: upgrade to 1.0.33

adamruzicka:


main/tinc/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=tinc
-pkgver=1.0.32
-pkgrel=2
+pkgver=1.0.33
+pkgrel=1
```
nmeum: I you bump `$pkgver` then you should set `$pkgrel` to 0 instead of 1. But just in case you want to submit further upgrades keep that in mind ;)

Thanks for you contribution BTW. :)

adamruzicka: Oh, off-by-one error, thank you for fixing that for me. I'll keep that in mind for the future :)

PR #2951, opened 2017-12-22
community/php7: make zlib extension builtin

vakartel:
php lacks some functionality when zlib extension built dynamic
https://bugs.alpinelinux.org/issues/8299

PR #1572, opened 2017-05-25
community/php7 / community/php5 Add php virtual packages

skwashd:
This allows other packages to depend on php5 or php7 via the php or php-[extension] virtual packages. A good example of this is the composer package. It will run with PHP 5.6 or 7.x. This virtual package will allow me to define either as the dependency via the `php`, `php-json` dependencies instead of hardcoding `php5` and `php5-json` packages.

PR #3081, opened 2018-01-20
Composer updates

skwashd:
* Upgrade to 1.6.2
* Switch to PHP 7 before it goes EOL in December 2018
* Move to community


andypost:
Please split upgrade to php7 because `php7-zlib` removed in https://github.com/alpinelinux/aports/pull/2951

skwashd:
I accidentally deleted the branch in my repo while doing some housekeeping. I have resubmitted this as #3092. 

@andypost I removed the `php7-zlib` dependency. I don't want to maintain this in community with it having a hard dependency on PHP 5. I'm very close to not having to deal with PHP 5 in production for any of my stuff.

If there were PHP virtual packages that were provided by the PHP 5.6 and 7.x packages I would depend on them. I proposed this under #1572 but it stalled.

PR #3088, opened 2018-01-20
community/php7: security upgrade to 7.0.27

andypost:
Only 3.5 needs upgrade
Ref http://php.net/archive/2018.php#id2018-01-04-1

PS: patch accepted upstream


andypost:
And again 7.0 in 3.5 missed upgrade https://github.com/alpinelinux/aports/pull/3500

PR #3090, opened 2018-01-21
testing/go-gdm: new aport

stwa:


stwa:
* How can I name this? gdm clashes with Gnome Display Manager
* The provided Godeps file is too old to work with the latest version of godep in alpine, that's why there is a custom Godeps.json

PR #3096, opened 2018-01-22
testing/eigen: Add signature file

hmenke:
The Eigen 3 library can be found automatically by build systems like CMake.  All of those systems look for a directory containing the file `signature_of_eigen3_matrix_library`.  Without this file auto-discovery will fail.

### DIFF
--- a/testing/eigen/APKBUILD
+++ b/testing/eigen/APKBUILD
@@ -27,6 +27,7 @@ package() {
         mkdir -p "$pkgdir"/usr/include/eigen3
         cp -r "$builddir"/Eigen "$pkgdir"/usr/include/eigen3
         cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
+        cp "$builddir"/signature_of_eigen3_matrix_library "$pkgdir"/usr/include/eigen3
 }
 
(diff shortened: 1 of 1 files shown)


hmenke:
Are there any requirements that I failed to meet, as this hasn't been merged or even reviewed yet?

nmeum:
I honestly don't understand why this aport doesn't use eigens cmake build system which would probably install that file by default.

hmenke:
Eigen is a header-only library.  The CMake build system only has to be used for building the tests.  And no, the CMake build system does not install `FindEigen3.cmake`.  I don't think `FindEigen3.cmake` should be part of the Alpine package because then CMake would need to be a dependency for which there is no reason.  Developers can simply include `FindEigen3.cmake` in their own projects and are then able to find Eigen, given that the signature file is present.

nmeum:
You misunderstood my comment, according to the `INSTALL` file in the tarball eigen has two installation methods: Installing without using CMake and Installing using CMake. I would suggest switching to the latter installation method since that method hopefully takes care of installing other files such as the signature file automatically.

hmenke:
Actually installing with CMake is only necessary when you want to run the testsuite.  Otherwise you just copy the required header files.  From Eigen's Doxygen documentation http://eigen.tuxfamily.org/dox/GettingStarted.html#title0

> # How to "install" Eigen?
> In order to use Eigen, you just need to download and extract Eigen's source code (see the wiki for download instructions). In fact, the header files in the Eigen subdirectory are the only files required to compile programs using Eigen. The header files are the same for all platforms. It is not necessary to use CMake or install anything.

ncopa:
fixed with b946d428588c8f17e32c5dfaad43199a9bc0c6dc

PR #3102, opened 2018-01-22
community/cassandra-cpp-driver: Upgrade to 2.8.0

tcely:


jirutka:
Please bacport the patch for libressl into this version.

community/cassandra-cpp-driver/APKBUILD
```diff
@@ -2,18 +2,31 @@
...
-pkgver=2.7.0
-pkgrel=1
+pkgver=2.8.0
+pkgrel=0
 pkgdesc="Cassandra CPP Driver"
 url="http://datastax.github.io/cpp-driver/"
 arch="x86_64 x86 ppc64le"
 license="Apache-2.0"
-depends=
+depends="libuv libressl2.6-libcrypto libressl2.6-libssl"
```
fabled: abuild will automatically pick up these and add dependency via the `so:` prefixed package names. Please leave `depends` empty.

community/cassandra-cpp-driver/APKBUILD
```diff
@@ -2,18 +2,31 @@
...
 makedepends="cmake make libressl-dev libuv-dev"
 subpackages="$pkgname-dev"
-source="$_pkgreal-$pkgver.tar.gz::https://github.com/datastax/cpp-driver/archive/$pkgver.tar.gz"
+source="$_pkgreal-$pkgver.tar.gz::https://github.com/datastax/cpp-driver/archive/$pkgver.tar.gz
+        b0830f3168b73be821e564b66d7550d24bec823f.patch"
 builddir="$srcdir/$_pkgreal-$pkgver"
 
+prepare() {
```
fabled: If `prepare` only patches, you can leave the whole `prepare` block out, as abuild then uses `default_prepare` which does the same thing.

tcely: @fabled Where might I find the default_* functions defined?

tcely:
Done

PR #3104, opened 2018-01-22
community/cassandra-cpp-driver: add tests step

tcely:


tcely:
It looks like the upstream tests are broken. I don't want to take the time to fix those.

PR #3107, opened 2018-01-22
testing/botan: new aport

tcely:
https://botan.randombit.net/
Crypto and TLS for C++11

### DIFF
--- a/testing/botan/APKBUILD
+++ b/testing/botan/APKBUILD
@@ -1,0 +1,44 @@
+# Contributor: tcely <tcely@users.noreply.github.com>
+# Maintainer:
+pkgname=botan
+_pkgreal=Botan
+pkgver=2.4.0
+pkgrel=0
+pkgdesc="Crypto and TLS for C++11"
+url="https://botan.randombit.net/"
+arch="x86_64 x86"
+license="BSD-2-Clause"
+depends=""
+depends_dev="python2 boost-dev libressl-dev sqlite-dev bzip2-dev zlib-dev"
+makedepends="$depends_dev"
+subpackages="$pkgname-dev $pkgname-doc"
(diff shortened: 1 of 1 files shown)


testing/botan/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+pkgname=botan
+_pkgreal=Botan
+pkgver=2.4.0
+pkgrel=0
+pkgdesc="Crypto and TLS for C++11"
+url="https://botan.randombit.net/"
+arch="x86_64 x86"
+license="BSD-2-Clause"
+depends=""
+depends_dev="python boost-dev libressl-dev sqlite-dev bzip2-dev zlib-dev"
```
jirutka: There's no package named `python`, correct name is `python3` or `python2` for Legacy Python. The same for executable.

testing/botan/APKBUILD
```diff
@@ -0,0 +1,44 @@
...
+		--with-sqlite3 \
+		--with-bzip2 \
+		--with-zlib
+	make
+}
+
+check() {
+	cd "$builddir"
+
+	LD_LIBRARY_PATH=. python src/python/botan2.py
```
jirutka: Please fix `python` to `python2`.

testing/botan/APKBUILD
```diff
@@ -0,0 +1,44 @@
+# Contributor:
+# Maintainer:
```
jirutka: Please fill your name and email at least to `Contributor`.

testing/botan/APKBUILD
tcely: Sure. Why prefer one symbolic link over the other?

jirutka: `python` may be changed to python3 in the future. `python2` makes is more clear that it uses Legacy Python.

tcely:
Done

PR #3109, opened 2018-01-23
main/dhcp: Upgrade to 4.4.1

tcely:


main/dhcp/APKBUILD
```diff
@@ -67,13 +72,15 @@ package() {
...
 	pkgdesc="ISC dhcp client"
+	depends="bash"
 	install -d -o dhcp -g dhcp "$subpkgdir"/var/lib/dhcp
 	install -d -o dhcp -g dhcp "$subpkgdir"/var/run/dhcp
 	install -d "$subpkgdir"/usr/sbin
 	install -d "$subpkgdir"/sbin
 	install -d "$subpkgdir"/etc/dhcp
 	mv "$pkgdir"/usr/sbin/dhclient "$subpkgdir"/usr/sbin/
 	mv "$pkgdir"/sbin/dhclient-script "$subpkgdir"/sbin/
+	ln -s ../sbin/dhclient-script "$subpkgdir"/etc/dhclient-script
```
HRio: should we no install the file in /etc/ instead of having a symlink? this will break generation of .apk-new files if the user has modified the script.

Also should the script not be in the /etc/dhcp directory?

tcely: The setup here is the `dhclient` binary living in /sbin, it has a configuration file in `/etc/dhcp` and that configuration specifies where the `dhclient-script` script is located.

The long standing tradition for where that script lives is `/etc/dhclient-script` but for whatever reason it lives in `/sbin` on Alpine. If it was modified that's where the modified file would be.

Now these are the options I see:
- move the script from `/sbin` to `/etc/dhcp` in `pre-install`
- update the configured location for the script
- leave things as they were installed and let the user configure everything

I chose that last option. By adding a link from the traditional location to the installed script we prevent surprising failures when copying the example configuration or an old configuration into `/etc/dhcp` for example.

HRio: it seems to live here `/sbin/dhclient-script` in Debian as well, so its probably no intended to be modified at all. Hence the symlink should not be needed at all then, if where not for the example :smille: in /etc/dhcp/dhclient.conf.example... 

Also in Alpine is seems to currently live in `/sbin` https://pkgs.alpinelinux.org/contents?branch=edge&name=dhclient&arch=x86_64&repo=main

So yes...you are right the symlink is probably a good idea. 

What do you think should the example be patched to point to `/sbin/dhclient-script` perhaps?

tcely: My opinion is that patching the example makes little sense. I don't see a strong argument for why the script belongs in `/sbin` as opposed to say `/usr/bin` but I don't care about this enough to change it.

I know my current configuration uses the `/etc/dhclient-script` location and so have most of the configurations I've seen others use.

HRio: that it is in `/sbin` and not `/usr/sbin` at least in Debian is I guess historical for thin client setups where /usr was mounted via NFS later during boot.

main/dhcp/APKBUILD
```diff
@@ -11,28 +11,20 @@ depends=
 pkgusers="dhcp"
 pkggroups="dhcp"
 makedepends="perl linux-headers file"
-install="dhcp.pre-install dhcp.pre-upgrade dhcp.post-upgrade"
+install="$pkgname.pre-install $pkgname.pre-upgrade $pkgname.post-upgrade"
 subpackages="$pkgname-doc $pkgname-dev $pkgname-dbg dhclient dhcrelay"
-source="http://ftp.isc.org/isc/dhcp/${_realver}/$pkgname-$_realver.tar.gz
+source="http://ftp.isc.org/isc/$pkgname/${_realver}/$pkgname-$_realver.tar.gz
+	dhclient-script-fix-bare-ip.patch
```
ncopa: I generally prefer to avoid using `$pkgname` in `source` because you may want copy paste the url to browser to find new/latest version. The `$pkgname` makes that harder and it does not really provide any benefit using a variable there.

At times we also have two versions of the same package (for example php5 and php7) and when that happens you cannot use pkgname variable in `builddir` and `source`

main/dhcp/dhclient-script.patch
```diff
@@ -1,30 +0,0 @@
---- a/client/scripts/linux
-+++ b/client/scripts/linux
-@@ -1,4 +1,4 @@
--#!/bin/bash
-+#!/bin/sh
```
ncopa: this patch needs to either be rebased or we need to add bash as a dependency for dhclient. I would prefer not depend on bash.

main/dhcp/APKBUILD
tcely: There were already variables in use in source, so copy / paste directly won't work anyway.

`(. APKBUILD; echo "$source")`

main/dhcp/dhclient-script.patch
tcely: That patch is gone. The `dhclient` package does depend on `bash` and rewriting the script so that it doesn't is a bit unnatural.

You wouldn't so the same thing with a `python` script, would you?

PR #3111, opened 2018-01-23
main/dwm: Fix emoji titlebar crash

orsonteodoro:
Test case: Open Firefox and go to https://getemoji.com/

### DIFF
--- a/main/dwm/00-no-emoji-title-crash.patch
+++ b/main/dwm/00-no-emoji-title-crash.patch
@@ -1,0 +1,13 @@
+diff -urp dwm-6.1.orig/drw.c dwm-6.1/drw.c
+--- dwm-6.1.orig/drw.c	2015-11-08 14:39:37.000000000 -0800
++++ dwm-6.1/drw.c	2016-12-05 17:55:37.082900924 -0800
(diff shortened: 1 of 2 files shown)


nmeum:
Hi,

Thanks for your contribution. I didn't look at your patch in depth so far but I have also experienced this problem with dwm. Did you suggest this patch upstream?

orsonteodoro:
The fix was recommended by Gaetan Bisson from Arch Linux. https://lists.suckless.org/dev/1610/30720.html on Oct 25, 2016 on their dev mailing list.

nmeum:
So, I have been using this patch with my fork of dwm for the past month and it works fine as is. However, it doesn't render emojis properly it just prevents a crash and I don't know if it breaks any other unicode characters. I would actually prefer a proper upstream fix for this issue instead of committing a patch downstream.

BTW:  Gaetan Bisson himself stated that he didn't understand why dwm crashes in the first places and why his patch fixes this problem.

nmeum:
I am closing this for two reason:

1. This doesn't seem to be a proper fix for the issue
2. This should be fixed upstream, not downstream

PR #3119, opened 2018-01-23
testing/rabbitmq-server: updating to 3.7.2

nathanejohnson:


TBK:
Seem like it is missing a build dependency: elixir

PR #3127, opened 2018-01-25
main/redis: upgrade to 4.0.8

TBK:


ncopa:
I have noticed that the test suite leaves a lot of redis-server processes. I have mulitple times needed to `killall -9 redis-server` on the build servers. Do you think you could have a look at that and maybe report it upstream?

TBK:
Sure, I will look into the issue.

fabled:
I fixed the testsuite zombie processes too. It needed `procps` in `checkdepends`.

TBK:
@fabled Thanks 👍

PR #3130, opened 2018-01-25
main/xen: add basic xen-pci rc script

radhus:
Adds a basic xen-pci release OpenRC script that will run before
xendomains and release PCI devices from dom0, so they can
be pci "passthroughed" to e.g. network or storage domain domUs.


radhus:
As typical with xen, takes too long for Travis to be happy with it.

```

The job has been terminated
```

HRio:
@jirutka I have tested this rc script, works fine! thanks @radhus

PR #3135, opened 2018-01-27
testing/oh-my-zsh: new package

orsonteodoro:
Add powerline-extra-symbols font (https://github.com/ryanoasis/powerline-extra-symbols) as a dependency for oh-my-zsh to properly show arrows for the agnoster theme.

Add oh-my-zsh to allow zsh users to use themes (https://github.com/robbyrussell/oh-my-zsh/wiki/Themes) and plugins (https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins).


orsonteodoro:
I couldn't revive my fork so the updated version is at #3238 .

PR #3138, opened 2018-01-27
testing/megatools: revision bump to rebuild

mepholic:
This is currently built against libressl 2.5.
A user in IRC was having issues installing this package on a
system with libressl 2.6.


TBK:
The default summery format for pkgrel bumps is like this: "testing/megatools: rebuild against libressl 2.6.3"

PR #3140, opened 2018-01-27
main/build-base: configure scripts often expect file to work

tcely:


ncopa:
I think the reason was MIPS? i think we could maybe mention it in the commit message that it is for MIPS.

tcely:
@ncopa What's this about MIPS?

PR #3154, opened 2018-01-28
main/xen: fix iscsi for tgtd target

HRio:
tun-1 is the disk if the iscsi target is tgtd


HRio:
Build failed due to
```

The job has been terminated
```
@kaniini this is the fix for the problem I mentioned on IRC the other day...

HRio:
@kaniini bump, can you have a look?

PR #3155, opened 2018-01-28
testing/plzip: Upgrade to 1.6

tcely:


testing/plzip/APKBUILD
```diff
@@ -13,6 +13,7 @@ makedepends="lzlib"
 install=""
 subpackages="${pkgname}-doc"
 source="http://download.savannah.gnu.org/releases/lzip/${pkgname}/${pkgname}-${pkgver}.tar.gz"
+options="!check"
```
nmeum: The tarball of this aport contains a `testsuite/` directory, I didn't look at further but I do therefore assume that upstream has a test suite. Could you elaborate further why you disabled check?

tcely: I didn't find any notes / logs explaining why it shouldn't be enabled so I added the check function.

PR #3159, opened 2018-01-30
testing/ycmd: new aport

orsonteodoro:
This is a code completion server used for YouCompleteMe used in vim and emacs-ycmd.  It provides IntelliSense, code correction, documentation to open source text editors that the big IDE editors have.  Languages supported C, C++, C#, Objective C, Python, JavaScript (Node.JS), Rust, Go, Java.  TypeScript support enabled.

rust-src ( #3174 ) is an optional dependency that never gets pulled but up to the user to decide if they want rust support. Already merged.

For Python code completion support, you must pull the py-jedi ( #3425 ) package.

For Java (and maybe Ruby) support you need openjdk8.

Check dependencies not merged but required for Travis CI to evaluate:
* py-pyhamchrest ( #3475 ).
* py-jedi ( #3425  )
* py2-configparser ( #3467 ) -- this was not listed in checkdepends because it gets indirectly pulled by py-flake8
* typescript ( #3466 )

### DIFF
--- a/testing/py-ycmd/APKBUILD
+++ b/testing/py-ycmd/APKBUILD
@@ -1,0 +1,457 @@
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=py-ycmd
+_pkgname=ycmd
+pkgver=31_git20180224
+pkgrel=0
+pkgdesc="A code-completion & code-comprehension server"
+url="https://github.com/Valloric/ycmd"
+arch="all"
+_e_java=
+_e_c=
+_e_go=
+_e_js=
+_e_cs=
+_e_rust=
+_e_typescript=
+_arch_makedepends=
+_arch_configure=
+case "$CARCH" in
+	aarch64*|ppc64le|arm*|s390x|x86|x86_64)
+		_e_java=1
+		_e_c=1
+		_e_go=1
+		_e_js=1
+		_arch_makedepends="clang-dev clang-libs go nodejs nodejs-npm"
+		_arch_configure="$_arch_configure --system-libclang --clang-completer "
+		_arch_configure+="--go-completer --js-completer --java-completer" ;;
+esac
+case "$CARCH" in
+	x86|x86_64)
+		_e_cs=1
+		_arch_makedepends="$_arch_makedepends mono-dev"
+		_arch_configure="$_arch_configure --cs-completer" ;;
+esac
+case "$CARCH" in
+	x86_64)
+		_e_rust=1
+		_e_typescript=1
+		_arch_makedepends="$_arch_makedepends rust cargo"
+		_arch_configure="$_arch_configure --rust-completer" ;;
+esac
+license="GPL-3.0-or-later Apache-2.0 MIT Python-2.0 GPL-2.1-or-later MPL-2.0 "
+license+="BSD-3-Clause BSD-2-Clause ZPL-2.1 ISC MS-PL EPL-1.0 CPL-1.0 W3C Apache-1.1 "
+license+="MPL-1.1 Plexus ICU custom"
+checkdepends="py-pyhamcrest py-webtest py-flake8 py-enum34 py-pycodestyle py-nose "
+checkdepends+="py-mock py-psutil py-nose-exclude py-requests py-urllib3 py-future "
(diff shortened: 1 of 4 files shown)


testing/py-ycmd/APKBUILD
```diff
@@ -0,0 +1,390 @@
...
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=py-ycmd
+_pkgname=ycmd
+pkgver=0_git20180224
+pkgrel=0
+pkgdesc="A code-completion & code-comprehension server"
+url="https://github.com/Valloric/ycmd"
+arch="aarch64 armhf ppc64le s390x x86 x86_64"
+install="$pkgname.post-install $pkgname.post-upgrade"
+_e_java= ; _e_c= ; _e_go= ; _e_js= ; _e_cs= ; _e_rust= ; _arch_makedepends= ;
```
kaniini: Please put each of these on a separate line, as well as in the `case` blocks.

testing/py-ycmd/APKBUILD
```diff
@@ -0,0 +1,390 @@
...
+	aarch64*|ppc64le|arm*|s390x|x86|x86_64) _e_java= ; _e_c=1 ; _e_go=1 ; _e_js=1 ; _arch_makedepends="clang-dev clang-libs go nodejs nodejs-npm" ;;
+esac
+case "$CARCH" in
+	x86|x86_64) _e_cs=1 ; _arch_makedepends="$_arch_makedepends mono-dev" ;;
+esac
+case "$CARCH" in
+	x86_64)	_e_rust=1 ; _arch_makedepends="$_arch_makedepends rust cargo" ;;
+esac
+license="GPL-3.0-or-later Apache-2.0 MIT Python-2.0 GPL-2.1-or-later MPL-2.0 BSD-3-Clause BSD-2-Clause ZPL-2.1 ISC MS-PL EPL-1.0 CPL-1.0 W3C Apache-1.1 MPL-1.1 Plexus ICU custom"
+checkdepends="py-pyhamcrest py-webtest py-flake8 py-enum34 py-pycodestyle py-nose py-mock py-psutil py-nose-exclude py-requests py-urllib3 py-future py2-unittest2 !py2-bottle !py3-bottle"
```
kaniini: Why are you conflicting on packages in `$checkdepends`?

testing/py-ycmd/APKBUILD
```diff
@@ -0,0 +1,390 @@
...
+
+_pybuild() {
+	local python="$1"
+	cp -a "$builddir" "${builddir}_$python"
+	cd "${builddir}_$python"
+	#[[ "$python" == "python3" ]] && sed -i -e "s|python3|python|g" cpp/ycm/CMakeLists.txt
+	sed -i -e "s|EXTERNAL_LIBCLANG_PATH \${TEMP}|EXTERNAL_LIBCLANG_PATH $(ls /usr/lib/libclang.so.*)|g" cpp/ycm/CMakeLists.txt
+	find "$srcdir" -type d -print0 | xargs -0 chmod 755
+	local arch_configure=""
+	[[ -n "$_e_c"    && "$_e_c" == "1" ]]    && arch_configure="$arch_configure --system-libclang --clang-completer"
```
kaniini: Instead of all of this, why not just build `$_arch_configure` as you enable the options outside `_pybuild()`?

testing/py-ycmd/py-ycmd.post-install
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
kaniini: Lets put this message as `README.Alpine` in `/usr/share/doc/ycmd`.  Then if people install the documentation hooks for apk it will display it, instead of cluttering install/update logs.

testing/py-ycmd/APKBUILD
orsonteodoro: The tests requires the internal bottle.  If you use the external py-bottle, it will trigger Shutdown_test FAILs.  The checkdepends stay up when there are some .makedepends-$pkgname and keep them up with the -K switch.  https://github.com/Valloric/ycmd/issues/670

testing/py-ycmd/APKBUILD
orsonteodoro: That is a valid suggestion.  I did it this way because old habit or because other package systems set up configure options in a configure() phase but APKBUILD does it in the build phase.  It would put the configuration in global scope so before all phases at the very top near lines 12-20.

PR #3162, opened 2018-01-30
testing/flycheck: new package

orsonteodoro:
Flycheck adds syntax checking to emacs.  It depends on #3161  .


orsonteodoro:

PR #3163, opened 2018-01-30
testing/company-mode: new package

orsonteodoro:
This add in-buffer completion to emacs.  It requires #3161 .


orsonteodoro:

PR #3164, opened 2018-01-30
testing/s-el: new package

orsonteodoro:
Adding dependency package for emacs-ycmd.  It requires #3161 .

s.el is a string manipulation library for emacs.  More info can be found at https://github.com/magnars/s.el .


orsonteodoro:

PR #3165, opened 2018-01-30
testing/f-el: new package

orsonteodoro:
Add dependency for emacs-ycmd.  It requires #3161 .

f.el is a package for working with files and directories in emacs.  More information can be found at https://github.com/rejeep/f.el .


orsonteodoro:

PR #3166, opened 2018-01-30
testing/request-el: new package

orsonteodoro:
Adding dependency for emacs-ycmd. It requires #3161 .

request.el is for http request for emacs.  More information can be found at https://github.com/tkf/emacs-request .  ycmd uses http as a IPC communication with client plugins.


orsonteodoro:

PR #3167, opened 2018-01-30
testing/dash-el: new package

orsonteodoro:
Adding dependency for emacs-ycmd.  It requires #3161 .

dash.el is a list (data structure) library used in emacs programs.  More information can be found at https://github.com/magnars/dash.el .


orsonteodoro:

PR #3168, opened 2018-01-30
testing/deferred-el: new package

orsonteodoro:
Adding emacs-ycmd dependency.  It requires #3161 .

deferred.el is a asynchronous task programming package used for emacs programming.  More information can be found at https://github.com/kiwanami/emacs-deferred .


orsonteodoro:

PR #3169, opened 2018-01-30
testing/popup-el: new package

orsonteodoro:
Adding dependency for emacs-ycmd.  It requires #3161 .

This is required for popup UI widget capability for menus in emacs-ycmd and other emacs programs.  More information can be found at https://github.com/auto-complete/popup-el .


orsonteodoro:

PR #3172, opened 2018-01-30
testing/bear: new aport

orsonteodoro:
Add a compile_commands.json generator for ycmd that hints how to compile the source code file for c/c++/objc so that ycmd's FixIt feature works.

### DIFF
--- a/testing/bear/APKBUILD
+++ b/testing/bear/APKBUILD
@@ -1,0 +1,43 @@
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+
+pkgname=bear
+_pkgname=Bear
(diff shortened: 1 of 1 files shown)


testing/bear/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgname=bear
+_pkgname=Bear
+pkgver=2.3.11
+pkgrel=0
+pkgdesc="Build EAR - Bear is a tool that generates a compilation database for clang tooling."
+url="https://github.com/rizsotto/Bear"
+arch="noarch"
+license="GPL-3.0"
+makedepends="cmake py3-lit"
+options="check !archcheck"
```
kaniini: This legitimately contains arch-specific binaries (`libear.so`), so it should be `arch="all"` and `!checkroot` should be removed.

PR #3174, opened 2018-01-30
community/rust: add $pkgname-src submodule for ycmd and racerd code completion support

orsonteodoro:
This downloads the source code so that code completion could work properly with ycmd with the rust programming language.

### DIFF
--- a/community/rust/APKBUILD
+++ b/community/rust/APKBUILD
@@ -6,11 +6,11 @@ pkgname=rust
 pkgver=1.23.0
 _llvmver=4
 _bootver=1.22.0
-pkgrel=0
+pkgrel=1
 pkgdesc="The Rust Programming Language (compiler)"
 url="http://www.rust-lang.org"
 arch="x86_64"
-license="Apache-2.0 BSD ISC MIT"
+license="Apache-2.0 BSD ISC MIT GPL-3.0-or-later GPL-3.0-with-GCC-exception LGPL-3.0 CC-BY-SA-3.0 OFL-1.1"
 
 # gcc is needed at runtime just for linking. Someday rustc might invoke
 # the linker directly, and then we'll only need binutils.
@@ -26,7 +26,7 @@ makedepends="rust>=$_bootver cargo
 	python2 tar zlib-dev"
 
 subpackages="$pkgname-dbg $pkgname-stdlib
-	$pkgname-gdb::noarch $pkgname-lldb::noarch $pkgname-doc"
+	$pkgname-gdb::noarch $pkgname-lldb::noarch $pkgname-doc $pkgname-src"
 source="https://static.rust-lang.org/dist/rustc-$pkgver-src.tar.gz
 	musl-fix-static-linking.patch
 	musl-fix-linux_musl_base.patch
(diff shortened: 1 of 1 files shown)


jirutka:
Please add it as a subpackage to community/rust.

jirutka:
I said “add it as a **subpackage** to community/**rust**”, not “move testing/rust-src into community”.

orsonteodoro:
We just delete this pull request and we just change the rust APKBUILD.  That is what you meant.  I'm okay with that.

orsonteodoro:
If we create a subpackage $pkgname-src, it may pull in compile time cruft and from the output it was 303.6MB.  The testing/rust-src had archive of 67.0 MB.  I will extract the sources before building the actual rust source (./x.py build) code which increase the build time requirements to duplicate the 67.0MB src folder.

kaniini:
please flatten

community/rust/APKBUILD
```diff
@@ -161,6 +164,17 @@ lldb() {
 	_mv "$pkgdir"/$_sharedir/etc/lldb_*.py $_sharedir/etc/
 }
 
+src() {
+	pkgdesc="Rust source code for ycmd and racerd code completion support"
```
jirutka: Please change it to `pkgdesc="$pkgdesc (source code)"`.

community/rust/APKBUILD
```diff
@@ -161,6 +164,17 @@ lldb() {
...
+	depends="$pkgname"
+	license="$license OFL-1.1 GPL-3.0-or-later GPL-3.0-with-GCC-exception CC-BY-SA-3.0 LGPL-3.0"
```
jirutka: Where is this list of licenses from?

orsonteodoro: They are in the src folder.  You can just grep for them.  We could put them in a doc subpackage.

community/rust/APKBUILD
```diff
@@ -161,6 +164,17 @@ lldb() {
...
+
+	mkdir -p "$subpkgdir"/usr/share/rust
+	cd "$subpkgdir"
+
+	_mv "$srcdir"/src usr/share/rust/
```
jirutka: Please install sources into `/usr/src/rust` instead of `/usr/share/rust`.

orsonteodoro: The final result is /usr/share/rust/src.  I will unload the contents of src in there (/usr/src/rust).

community/rust/APKBUILD
jirutka: Then please add them to top-level `license=`. I guess that licenses should be the same for both source code and binary built from the source code.

jirutka: > They are in the src folder. You can just grep for them.

Could you please describe how exactly did you extracted this licenses list? I don’t see them.

Also:
[rust README.md](https://github.com/rust-lang/rust#license):
> Rust is primarily distributed under the terms of both the MIT license and the Apache License (Version 2.0), with portions covered by various BSD-like licenses.

[Fedora rust.spec](https://src.fedoraproject.org/rpms/rust/blob/f27/f/rust.spec#_53-54):
> License:        (ASL 2.0 or MIT) and (BSD and ISC and MIT)
>  ^ written as: (rust itself) and (bundled libraries)

There are many bundled build-only dependencies, so maybe this explains the extra licenses you found?

jirutka: Meanwhile I merged addition of -src subpackage with your original suggestion of licenses.

PR #3177, opened 2018-01-30
community/php7-xdebug: upgrade to 2.6.0

andypost:
Very valuable upgrade to unblock php 7.2 https://xdebug.org/#2018_01_29


f3l1x:
Good job. 👍

PR #3178, opened 2018-01-30
testing/capnproto-c++: new aport

florolf:
Cap'n Proto is a binary serialization framework in the spirit of Google
Protobuf.

Signed-off-by: Florian Larysch <fl@n621.de>

### DIFF
--- a/testing/capnproto-c++/APKBUILD
+++ b/testing/capnproto-c++/APKBUILD
@@ -1,0 +1,38 @@
+# Contributor: Florian Larysch <fl@n621.de>
+# Maintainer: Florian Larysch <fl@n621.de>
+pkgname=capnproto-c++
+pkgver=0.6.1
+pkgrel=0
+pkgdesc="Cap'n Proto serialization/RPC system"
+url="https://capnproto.org"
+arch="all"
+license="MIT"
+depends=""
+makedepends="linux-headers"
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::https://capnproto.org/$pkgname-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure \
+		--build=$CBUILD \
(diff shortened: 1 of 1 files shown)


florolf:
Hey, is there anything I still can/have to do to get this merged?

FWIW, the Cap'n Proto library is pretty stable, so this PR is still up to date.

tcely:
You could add a `check` function to run the tests or use `options="!check"` and a comment as to why checks aren't being run.

florolf:
Good point, thanks. I'll do that.

Nebukadneza:
Hi,

sorry to bug everyone … but is there a reason we cannot merge this? As @florolf probably cannot care about this PR anymore, I would gladly continue to add any suggestions / fix anything that needs to be fixed to get this package in.

Having this Cap’n Proto binding in Alpine would be a big benefit for many: As it’s a more lightweight alternative to Google-Protobuf, it’s used by many projects — plus the Python lib depends on `capnproto-c++`, so this package also enables people to install the python-lib via `pip`, enabling even more projects to run hassle-free on alpine!

Thanks for your support and understanding ^_^.
Best Regards,
-Dario

tcely:
@florolf I have no more remarks to add here. I don't control when these things are merged and am in the same situation with a few of my own.

I view this as possibly the largest problem `aports` faces at the moment. It would be nice to have a more efficient process to merge these pull requests so that the backlog doesn't sit above 200+ all the time.

@ncopa If this is something the project would like to address I do have a few ideas.

nmeum:
> @ncopa @nmeum If this is something the project would like to address I do have a few ideas.

This is indeed something we would like to address. If you have ideas regarding this we can discuss those on the alpine-devel mailing list. Just send an email to that mailing list.

Nebukadneza:
Hi everyone,

sorry to derail the (undoubtfully important) project / process discussion, @ncopa @nmeum  … but is there anything i could do to get this in? ^_^

Thanks’n’Sorry,
-Dario

PR #3180, opened 2018-01-30
community/repmgr [BUMP] to v4.0.2

blaggacao:


blaggacao:
@jirutka Is that about how a version bump it should be done?

PR #3181, opened 2018-01-30
community/repmgr bump to v4.0.2

blaggacao:
@jirutka Is that about how a version bump it should be done?

### DIFF
--- a/community/repmgr/APKBUILD
+++ b/community/repmgr/APKBUILD
@@ -1,16 +1,16 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=repmgr
-pkgver=3.3.2
-pkgrel=3
+pkgver=4.0.2
+pkgrel=0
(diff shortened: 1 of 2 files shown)


blaggacao:
building the documentation seems to be now an extra step in version 4, so I removed it, as we would need extra dependencies... Not sure if that's the idea, but definitely reduces the package size, which is good in my personal opinion.

blaggacao:
@jirutka See how the sample config, specially logging, has changed: https://github.com/2ndQuadrant/repmgr/blob/master/repmgr.conf.sample

community/repmgr/APKBUILD
```diff
@@ -31,15 +31,12 @@ check() {
 package() {
 	cd "$builddir"
 
-	make USE_PGXS=1 DESTDIR="$pkgdir" install
+	./configure && make USE_PGXS=1 DESTDIR="$pkgdir" install
```
ncopa: do you really need to re-run configure before installing it? that does not seems right

blaggacao: I din't (don't) fully understand the build process. But I strongly guess it's not necesary.. :+1:

PR #3182, opened 2018-02-01
community/bats: Update to git-snapshot of successor-project

Nebukadneza:
sstephenson/bats has been abandoned for quite some time. The designated
successor-fork seems to be bats-core/bats-core. Unfortunately, bats-core
has not released a new version yet. To still provide the bugfixes in
bats-core to everyone, we switch community/bats to bats-core, and build
a git-snapshot.
Since bats also works way nicer with `tput` from ncurses, we now include
ncurses as a dependency. Note that this does not adversely affect
non-terminal runs, such as CI systems.


mbland:
We're stable now: https://github.com/bats-core/bats-core/releases/tag/v1.0.0

HRio:
@Nebukadneza can you update this PR for bats v1.0.1, I can then commit it after we are done with Alpine v3.8.

Nebukadneza:
Hi,

sorry for the delay, i was a little busy. I’ve now updated the APKBUILD file to package the latest release, v1.0.1.

I hope this is working for everyone — please tell me if i’ve made any mistake here ^_^.
Thanks!

community/bats/APKBUILD
```diff
@@ -1,20 +1,20 @@
 # Maintainer: Jose-Luis Rivas <ghostbar@riseup.net>
 pkgname=bats
-pkgver=0.4.0
-pkgrel=2
+pkgver=1.0.1
+pkgrel=1
```
HRio: pkgrel starts at 0

community/bats/APKBUILD
```diff
@@ -31,6 +31,4 @@ package() {
 			"$pkgdir"/usr/share/man/man$i/bats.$i
 	done
```
HRio: Consider switching to the `./install.sh` script so the `/usr/bin/bats` wrapper is installed correctly (a symlink does not work anymore)

e.g.
```bash
package() {
        cd "$builddir"
        ./install.sh "$pkgdir"/usr
}
```

I you like you can patch the script to install in `usr/libexec/bats/` instead of `usr/libexec/`

Or fix the package() phase to not create a symlink but instead install the `/usr/bin/bats` wrapper

community/bats/APKBUILD
```diff
@@ -1,20 +1,20 @@
...
-	0001-test-bats-Add-fake-tput-to-fix-Alpine-tests.patch"
-builddir="$srcdir/$pkgname-$pkgver"
+source="$pkgname-$pkgver.zip::https://github.com/bats-core/bats-core/archive/v$pkgver.zip"
+builddir="$srcdir/bats-core-$pkgver"
 
 check() {
 	cd "$builddir"
-	bash -c 'bin/bats --tap test'
+	export BATS_ROOT="$builddir"
+	bash -c 'libexec/bats --tap test'
```
HRio: this change should not be needed if the wrapper is used.

Nebukadneza:
Thank you for your comments and code-review, @HRio ! I must admit i didn’t look closely at install.sh upstream changes, and whether the current package() can be shortened. You are right ….

OTOH, changing libexec/<files> to libexec/bats/<files> was a bit bigger patch which i didn’t feel was quite the right thing to do. I hope that’s fine with you?

Thanks for your feedback ^_^.

HRio:
@Nebukadneza lest leave the libexec stuff directly under `/usr/libexec` for now.

@mbland would be really nice as its not just one single libexec file from bats if you could consider creating a bats subdir in the future, like for instance git and others are doing.

mbland:
@HRio Ask and ye shall receive: bats-core/bats-core#105

Nebukadneza:
Hi,

And thanks for all your support. Sorry to say now, I'm away with essentially no internet for ~2 weeks. Hope I'm not delaying anyone to much.

Thanks and best,
Dario

mbland:
bats-core/bats-core#105 is merged; I'll see about cutting bats-core v1.0.2 either later this afternoon or early tomorrow.

Point taken, @Nebukadneza. @HRio is there a chance you could pick this up after I tag bats-core v1.0.2? Though I suppose there's no rush; I'm just excited to see this coming together.

HRio:
@mbland Alpine 3.8.0 is at rc5, hence a 3.8.0 release is imminent. The original plan was to commit this right after the release. I will start reg. testing the packages using bats for check() right now with bats-core 1.0.1 to see how we are doing anyhow,

HRio:
@Nebukadneza 

From commit msg:
> Since bats also works way nicer with `tput` from ncurses, we now include
ncurses as a dependency. Note that this does not adversely affect
non-terminal runs, such as CI systems.

Should we really pull in ncurses? is not done by the APKBUILD now... drop this paragraph from the commit msg?

mbland:
Just pushed [v1.0.2](https://github.com/bats-core/bats-core/releases/tag/v1.0.2) with the improved `libexec/bats-core` layout and updated `install.sh` from bats-core/bats-core#105, FYI.

Nebukadneza:
Hi All,

sorry for the delay, and sorry for forgetting the commit message!
@mbland Thanks for https://github.com/bats-core/bats-core/pull/105 ! It works out of the box, and like a charm!

I have also removed the ncurses paragraph — as you are right: In alpine, minimalism is key, and depending on a relatively large package like ncurses is not following that idea. I verified that pretty-formatting on TTYs work after separately installing ncurses though.

As for bats-doc, I took the liberty of adding `man` as a dependency though: Since it only contains 2 man-pages, it would really be useless without man.

Hoping this is good now ^_^. Sorry again!
Best,
-Dario

HRio:
@Nebukadneza & @mbland big thanks!  

I will be committing later today (evening CET).

mbland:
Thanks, @HRio!

Out of curiosity, what's the protocol for creating packages for new versions when they're available? Should we keep a checklist to remind ourselves to submit a PR? https://wiki.alpinelinux.org/wiki/Alpine_Linux:Contribute has links to how to create patches using Git, which leaves me thinking that opening a new PR is all that's required.

HRio:
@mbland If you submit at PR with a new version I will commit after checking it. Just ping me after opening the PR. Thanks!

PR #3194, opened 2018-02-03
main/gc: upgrade to 7.6.4; modernize

TBK:
100% backwards compatible - https://abi-laboratory.pro/tracker/timeline/bdwgc/


nmeum:
Hey, thanks for your patch. Just a minor suggest: Could you split the modernization changes and the upgrade into two seperate commits? :)

PR #2722, opened 2017-11-14
main/syslog-ng: upgrade to 3.12.1

andypost:

PR #3206, opened 2018-02-04
main/syslog-ng: upgrade to 3.13.2; modernize

adrianguenter:
Supersedes and fixes #2722

5 of 69 tests fail, one less than the current version (3.9.1).

The new [`getent` module](https://github.com/balabit/syslog-ng/commits/master/modules/getent) is disabled during packaging as it [references two functions](https://github.com/balabit/syslog-ng/commit/b5f554808df8a51e2dfdba005e84c8753d0a2f27#commitcomment-20731217) that aren't currently available in musl libc.

The init script now starts syslog-ng with `--process-mode=background`, as the default process mode for syslog-ng starts its own supervisor which causes problems with `start-stop-daemon`.

The template system has been modified to allow for more than one source definition by moving the `s_all` source block definition to the `syslog-ng-source.std` file.

The post-install script logic has been fixed. It would previously fail if any of the log files it attempts to create already exist.


andypost:
Another upgrade https://github.com/alpinelinux/aports/pull/5153

PR #3207, opened 2018-02-05
community/ocaml: add gcc dependency

ganwell:
Based on this issue: https://bugs.alpinelinux.org/issues/8472

Since ocaml packages use ocamlopt to compile and therefore ocaml is _not_ a run-time dependency [1], I think it is ok the add this dependency.

[1] https://pkgs.alpinelinux.org/package/edge/community/x86_64/unison

### DIFF
--- a/community/ocaml/APKBUILD
+++ b/community/ocaml/APKBUILD
@@ -2,12 +2,12 @@
 # Maintainer: Borys Zhukov <mp5@mp5.im>
 pkgname=ocaml
 pkgver=4.04.2
-pkgrel=1
+pkgrel=2
 pkgdesc="Main implementation of the Caml programming language"
 url="http://ocaml.org/"
 arch="all !x86 !armhf !s390x"
 license="LGPL-2.0"
-makedepends="ncurses-dev zlib-dev gdbm-dev"
+makedepends="ncurses-dev zlib-dev gdbm-dev gcc libc-dev"
 depends="ncurses-dev"
 options="textrels"
 subpackages="$pkgname-doc"


jpdeplaix:
Thanks !! (I didn't know there was a github repo)

Hum, what about adding ```musl-dev``` (or ```libc-dev```) as requested in the issue ? Adding only gcc does not help if we can't link as well.

jpdeplaix:
Mmmh, just checked but gcc should have this dependency. Otherwise this fails as well:
```
FROM alpine:3.7
RUN apk add --no-cache gcc
RUN echo 'int main() { printf("Hello World !\n"); }' > test.c
RUN gcc test.c
```

I'll do a PR.

ganwell:
@jpdeplaix I added musl-dev

jpdeplaix:
What is the difference between musl-dev and libc-dev ? Does ```libc-dev``` refers to the glibc ?

ganwell:
I did a quick search in the repo its about 50%/50% "gcc libc-dev"/"gcc musl-dev". I guess if alpine ever switches libc again, libc-dev is better.

jpdeplaix:
Ok, awesome. I just did a follow-up PR for gcc. Thanks a lot.

jpdeplaix:
See also:
* https://github.com/ocaml/opam-repository/issues/11241
* https://github.com/alpinelinux/aports/pull/3209

fabled:
> What is the difference between musl-dev and libc-dev ? Does libc-dev refers to the glibc ?

`musl-dev` is headers from musl.
`libc-dev` is intended to be used by applications, it'll depend on used c-library header package and potentially ship few compat headers of it's own (e.g. for musl it ships bsd cdefs, queue and tree headers). If we ever build other c-library variants it would depend on the correct one (previously we built against uclibc...).
`build-base` depends on `libc-dev` and default set of utilities to build source like `make` etc.

ocaml should likely depend on `gcc` and `libc-dev`.

PR #3208, opened 2018-02-05
Fix linking with gcc by adding a missing dependency

jpdeplaix:
Follow-up PR on #3207 
This PR fixes the following failure over the most basic hello world using C:
```
FROM alpine:3.7
RUN apk add --no-cache gcc
RUN echo 'int main() { printf("Hello World !\n"); }' > test.c
RUN gcc test.c
```

Failure message:
```
/usr/lib/gcc/x86_64-alpine-linux-musl/6.4.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find Scrt1.o: No such file or directory
/usr/lib/gcc/x86_64-alpine-linux-musl/6.4.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find crti.o: No such file or directory
/usr/lib/gcc/x86_64-alpine-linux-musl/6.4.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -lssp_nonshared
The command '/bin/sh -c gcc test.c' returned a non-zero code: 1
```


fabled:
Please add `build-base` in your docker file. `gcc` is used without c-library to build kernel and other standalone components. It needs to work without c-library headers.

PR #3219, opened 2018-02-06
allow build to finish after proc_nice test has permission warnings

mksully22:
2 minor test warnings on tests that issue nice() command with invalid strings/variables. Allow build to continue by adding ppc64le so allow_fail $CARCH list.


ncopa:
I modified the commit message a bit. Added `community/php7` prefix.

Thanks!

PR #3230, opened 2018-02-07
community/php7-apcu: upgrade to 5.1.9

andypost:
Ref https://pecl.php.net/package-changelog.php?package=APCu&release=5.1.9


andypost:
Better to commit it after https://github.com/alpinelinux/aports/pull/3222

PR #3233, opened 2018-02-08
main/git: Fix installation of protocol backends

brotbert:
A trailing backslash lead to a non-properly executed `mv`-command, so all backends (`daemon`, `http-backend`, `shell`) were not available in the package.


jirutka:
Why did you closed this PR?

brotbert:
@jirutka I've closed it, though of the trailing backslash the sub-package `git-daemon` has included the needed protocol backends.

PR #3134, opened 2018-01-27
testing/keepassxc: add aport

orsonteodoro:
This is the forked version of the keychain password manager KeePassX with additional features and continued development.  Additional details can be found at https://keepassxc.org/ .

KeePassX latest release was 2.0.3 in Sep 2016.  KeePassXC latest release is 2.2.4 in Dec 2017.

PR #3236, opened 2018-02-08
testing/keepassxc: new aport (edited/fixed)

orsonteodoro:
I couldn't revive the deleted fork so this is recreated.  The old comments are at #3134

### DIFF
--- a/testing/keepassxc/APKBUILD
+++ b/testing/keepassxc/APKBUILD
@@ -1,0 +1,47 @@
+# Contributor: Leonardo Arena <rnalrd@alpinelinux.org>
+# Contributor: August Klein <amatcoder@gmail.com>
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Contributor: Max Rees <maxcrees@me.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+
+pkgname=keepassxc
+pkgver=2.2.4
+pkgrel=0
+pkgdesc="KeePassXC is a cross-platform community-driven port of the Windows \
+application Keepass Password Safe."
+url="https://keepassxc.org/"
+arch="all"
+license="GPL-3.0 BSD-3-Clause CC0-1.0 LGPL-2.0-only LGPL-2.1-only \
+	LGPL-3.0-or-later Nokia-Qt-exception-1.1 MIT BSL-1.0"
+depends="icu-libs"
+makedepends="cmake qt5-qtbase-dev qt5-qttools-dev libxtst-dev \
+	qt5-x11extras-dev libgcrypt-dev libgpg-error-dev argon2-dev zlib-dev"
+source="$pkgname-$pkgver.tar.gz::https://github.com/keepassxreboot/$pkgname/archive/$pkgver.tar.gz"
+builddir="$srcdir"/$pkgname-$pkgver
+subpackages="$pkgname-doc"
+
+build() {
+	cd "$builddir"
+	cmake -DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=/usr/lib \
+		-DCMAKE_BUILD_TYPE=Release .
+	make
+}
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install
+	install -d "$pkgdir"/usr/share/doc/$pkgname/ \
+		   "$pkgdir"/usr/share/licenses/$pkgname/
+	install -t "$pkgdir"/usr/share/doc/$pkgname/ \
+		   "$builddir"/CHANGELOG "$builddir"/README.md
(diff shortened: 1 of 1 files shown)


testing/keepassxc/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=keepassxc
+pkgver=2.2.4
+pkgrel=0
+pkgdesc="KeePassXC is a cross-platform community-driven port of the Windows application Keepass Password Safe."
+url="https://keepassxc.org/"
+arch="all"
+license="GPL-2.0 GPL-3.0 BSD-3-Clause CC0-1.0 LGPL-2.1 LGPL-3.0 Nokia-Qt-exception-1.1 MIT"
+_license_build="$license BSL-1.0"
```
jirutka: This variable is not used anywhere, why is it here?

testing/keepassxc/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+# Contributor: August Klein <amatcoder@gmail.com>
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=keepassxc
+pkgver=2.2.4
+pkgrel=0
+pkgdesc="KeePassXC is a cross-platform community-driven port of the Windows application Keepass Password Safe."
+url="https://keepassxc.org/"
+arch="all"
+license="GPL-2.0 GPL-3.0 BSD-3-Clause CC0-1.0 LGPL-2.1 LGPL-3.0 Nokia-Qt-exception-1.1 MIT"
```
jirutka: Please specify exact variant of GPL licenses – `GPL-x.y-only`, or `GPL-x.y-or-later`.

jirutka:
Please add fixed to this PR (branch), **do not close it and create new PR**.

sroracle:
I made a [similar aport for Adélie](https://code.foxkit.us/sroracle/packages/commit/00c1f17da34cfd220d56355f26490e4c7959daa8) and it passed its test suite there, so maybe add a `check()` for `make test`.

testing/keepassxc/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Contributor: Max Rees <maxcrees@me.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=keepassxc
+pkgver=2.2.4
+pkgrel=0
+pkgdesc="KeePassXC is a cross-platform community-driven port of the Windows application Keepass Password Safe."
+url="https://keepassxc.org/"
+arch="all"
+license="GPL-2.0 GPL-3.0 BSD-3-Clause CC0-1.0 LGPL-2.1-only LGPL-3.0-only Nokia-Qt-exception-1.1 MIT BSL-1.0"
```
kaniini: Please use `GPL-2.0+` or `GPL-2.0-or-later` (your preference) for both the GPL2 and GPL3 ones, unless they are GPL-2.0-only and GPL-3.0-only, in which case we cannot accept the package at all as the GPL2 and GPL3 are incompatible with each other when non-upgradeable.

kaniini:
licenses is still wrong

testing/keepassxc/APKBUILD
orsonteodoro: The LICENSE.txt (https://github.com/keepassxreboot/keepassxc/blob/148dbc5d41b8f7828c6ae730a92f5d97c13454a7/COPYING) says the art assets keepassxc.png is LGPL-2 (implying only), share/icons/database/* LGPL-2.1 (implying only), share/icons/application/* is LGPL-3.0+.  The code assets are src/streams/qtiocompressor* (case insensitive (LGPL-2 [imply only] or GPL-3; src/gui/KMessageWidget (LGPL 2.1 [implying only with SPDX v2]).

orsonteodoro: The main project is GPL-2 or GPL-3 without plus.

kaniini: In this case, we will take the manual and main.cpp under `GPL-3.0-only`, so the `GPL-2.0` can be removed.

We can take qtiocompressor under `GPL-3.0-or-later`.  The `LGPL-3.0` should become `LGPL-3.0+` or `LGPL-3.0-or-later`.

kaniini: Please update $license accordingly, and I will merge this with the differences in the adelie package and commit it for you.

sroracle:
Can probably be bumped to 2.3.1 if we're still merging this, but needs to add `makedepends="argon2-dev"` and `subpackages="$pkgname-doc"`

PR #3247, opened 2018-02-09
community/neovim: ignore st terminfo cursor shape

ganwell:
Fixes a crash that started to appear with the ncurses 6.0_p20171125-r0 update.

It would be nice if this could be backported to 3.7 since it fixes a crash.

Fixes: https://bugs.alpinelinux.org/issues/8469


ganwell:
The upstream issue is: neovim/neovim#7641

PR #3250, opened 2018-02-09
testing/bfs: new aport

dai9ah:


testing/bfs/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+source="$pkgname-$pkgver.tar.gz::https://github.com/tavianator/$pkgname/archive/$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	make -C "$builddir"
+}
+
+check() {
+	cd "$builddir"
+	./bfs -nohidden > /dev/null
```
tavianator: Why not `make -C "$builddir" check`?

jirutka: The contributor was probably too lazy and the committer didn’t check it during code-review. 😞
Could you please create a PR changing this to run `make check`?

tavianator: Done: https://github.com/alpinelinux/aports/pull/3596

PR #3252, opened 2018-02-09
main/linux-vanilla: renable Netronome nics after driver rename

mmlb:
The Netronome nic driver was renamed in v4.11, the kernel jump from 4.9
to 4.14 in 9b052dbbcfdc3645493c91f1a80012ba3cdb0d16 correctly drops the
old nic config, but does not replace it with the new one. This commit
adds the driver back.


mmlb:
boot/nic tested on x86_64

mmlb:
ping @ncopa. FYI this PR has been tested in the x86_64 config, the travis failure is a timeout during kernel compile. Its only reintroducing a NIC driver that was previously enabled/built but was accidentally dropped when the config option changed. I'd really like to be able to stop building this in ci and lean on the alpine pkg/mirror.

mmlb:
@jirutka whats the procedure re: ci-malfunction tag?

ncopa:
I merged the patches and updated the commit message a bit. Thanks!

PR #3257, opened 2018-02-10
testing/[various]: fixes on s390x

tmh1999:
need py-numpy   : py-cpuinfo, websockify, py-gnuplot, py-scipy,
                  py-fonttools, py-matplotlib, py-opengl
need openblas   : harminv
no jit support  : texlive
no arch support : tcc, nim, perl-sys-syscall

radare2         : need disabling debugger
thermald        : seems to be intel-only thing
lazperf         : io tests hang


tmh1999:
git commit message was formatted alright. github makes it weird.

tmh1999:
The CI failure has nothing to do with this patch.

PR #3263, opened 2018-02-10
community/roundcubemail: fetch jsdeps

HRio:
From 1.3.x and forward jsdeps need to be fetched

https://github.com/roundcube/roundcubemail/wiki/Build-from-source#3-install-javascript-dependencies

### DIFF
--- a/community/roundcubemail/APKBUILD
+++ b/community/roundcubemail/APKBUILD
@@ -3,17 +3,18 @@
 _php=php7
 pkgname=roundcubemail
 pkgver=1.3.4
-pkgrel=1
+pkgrel=2
 pkgdesc="A PHP web-based mail client"
 url="http://www.roundcube.net"
 arch="noarch"
-license="GPL-3.0-or-later"
+license="GPL-3.0-or-later LGPL-3.0-or-later MIT"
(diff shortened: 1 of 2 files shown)


HRio:
@jirutka can you have a look on this change? thanks.

rnalrd:
@HRio we allow only abuild to fetch sources. The proper fix is to package any additional dependency and add it to roundcubemail APKBUILD "depends".

Since roundcubemail is in "community" repo these new dependencies need to be committed in the same repo and not into "testing".

Thanks

HRio:
fixed in 91498e87a2d6a142d639fabf5d7f9f583483a041 by @jirutka big thanks!

@jirutka can you cherry-pick your big roundcube improvements (or at least this one) to v3.8 so we can have a working roundcube package in an alpine release again?

jirutka:
@HRio, I've backported 91498e87a2d6a142d639fabf5d7f9f583483a041 (fix JS deps) to 3.8-stable and 3.7-stable. I don't think that the rest of the changes are suitable for backporting into stable branches.

HRio:
@jirutka very good, thanks!

PR #3287, opened 2018-02-13
testing/libzookeeper: new aport

gilek:
https://zookeeper.apache.org
The Apache ZooKeeper system for distributed coordination is a high-performance service for building distributed applications.

### DIFF
--- a/testing/libzookeeper/APKBUILD
+++ b/testing/libzookeeper/APKBUILD
@@ -11,7 +11,7 @@ depends=""
 makedepends=""
 install=""
 subpackages="$pkgname-dev"
-source="ftp://ftp.task.gda.pl/pub/www/apache/dist/zookeeper/zookeeper-$pkgver/zookeeper-$pkgver.tar.gz"
+source="http://www-eu.apache.org/dist/zookeeper/zookeeper-$pkgver/zookeeper-$pkgver.tar.gz"
 builddir="$srcdir/zookeeper-$pkgver/src/c"
 
 build() {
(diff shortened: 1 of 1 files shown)


testing/libzookeeper/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgrel=0
+pkgdesc="The Apache ZooKeeper system for distributed coordination is a high-performance service for building distributed applications."
+url="https://zookeeper.apache.org"
+arch="all"
+license="Apache License"
+depends=""
+makedepends=""
+install=""
+subpackages="$pkgname-dev"
+source="ftp://ftp.task.gda.pl/pub/www/apache/dist/zookeeper/zookeeper-$pkgver/zookeeper-$pkgver.tar.gz"
```
Ikke: Wouldn't it be better to use an [HTTP](https://www.apache.org/dyn/closer.cgi#http) mirror here? For example travis does not allow downloads via FTP.

testing/libzookeeper/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+	./configure --prefix=/usr
+	make
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install
+}
+
+dev() {
```
Ikke: Isn't this handled by `default_dev`? Or am I missing something here.

jirutka: Yes, this should be handled by default dev…

testing/libzookeeper/APKBUILD
jirutka: Definitely, FTP should not be used in abuilds.

PR #3296, opened 2018-02-13
main/pcre2: upgrade to 10.31, include license

dai9ah:


main/pcre2/APKBUILD
```diff
@@ -1,18 +1,18 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=pcre2
-pkgver=10.30
+pkgver=10.31
 pkgrel=0
 pkgdesc="Perl-compatible regular expression library"
 url="http://pcre.sourceforge.net/"
 arch="all"
 license="BSD"
```
nmeum: What kind of BSD license are they using? Alpine uses [SPDIX license names](https://spdx.org/licenses/) these days. It would imho be better to use the proper BSD license name for this license and not install the LICENSE to `/usr/share/licenses`.

dai9ah: Thank you for the heads-up. Also, is the above method preferred in packages that don't include a license file in docs? It would be useful to know from now on.

nmeum:
Build fails on the x86_64 builder, the issue might be related to the fact that the builder uses linux-hardened instead of linux-vanilla. The buildlog can be found here: http://build.alpinelinux.org/buildlogs/build-edge-x86_64/main/pcre2/pcre2-10.31-r0.log

I will probably have to revert the upgrade unless you have any idea what might be causing this.

jirutka:
pcre2 has JIT, so it doesn't work with PaX mprotect - it must be disabled using paxmark for the pcre2 binary.

However, I've already handled this for pcre2, so I wonder what has been changed.

PR #3298, opened 2018-02-13
main/lcms: add extra check to detect powerpc endianess

mksully22:
for powerpc cases add additional check to determine if running on big or little endian machine


mksully22:
will resubmit after Travis fixes mpfr3-3.1.5-r1: BAD archive errors

jirutka:
It seems that this has been already fixed in 3714e669c9b3dad8b85d2d148c2ee414d466eb4d, 738cd2ce25f290bf22f7f04943aeae4501b06bf4, 415b2c8973df03018b74edb31173caf8c0f86b36.

PR #3299, opened 2018-02-13
main/memcached: Upgrade to 1.5.6

tcely:


main/memcached/APKBUILD
```diff
@@ -16,13 +16,23 @@ source="https://www.memcached.org/files/${pkgname}-${pkgver}.tar.gz
 builddir="${srcdir}/${pkgname}-${pkgver}"
 
 build() {
+	# extstore is broken on 32-bit
+	# https://github.com/memcached/memcached/issues/319
+	local _extstore=''
+	case "$(uname -m)" in
```
fabled: `uname` should not be used as it's "running system". Use instead `$CARCH`. See e.g. `main/x264` on how to use it. However, I am not sure if we want to keep this kind of feature in disparity between arches. Is there use case for this, or is this just 'let's enable things that exist'?

tcely:
The major feature of 1.5.4 is `extstore` so I'd like to enable this on as many
platforms as we know it works.

https://github.com/memcached/memcached/wiki/ReleaseNotes154

That said, 1.5.5 is out now so I'll be updating this pull request shortly for that.

main/libmemcached/APKBUILD
```diff
@@ -1,49 +1,45 @@
...
+		--enable-memaslap
+	make
 }
 
-package() {
-	cd "$_builddir"
-	make DESTDIR="$pkgdir" install || return 1
+check() {
+	# Tests are broken as of 1.0.18
 	return 0
```
jirutka: If tests are broken, then disable check adding `options="!check"  # tests are broken` at top-level.

main/memcached/APKBUILD
```diff
@@ -1,28 +1,39 @@
...
-subpackages="$pkgname-dev $pkgname-doc"
+depends_dev="${pkgname}=${pkgver}-r${pkgrel}"
+makedepends="${depends_dev} cyrus-sasl-dev libevent-dev libseccomp-dev linux-headers"
+install="${pkgname}.pre-install"
+subpackages="${pkgname}-dev ${pkgname}-doc"
 source="https://www.memcached.org/files/${pkgname}-${pkgver}.tar.gz
-	$pkgname.confd
-	$pkgname.initd"
+	${pkgname}.confd
+	${pkgname}.initd"
```
jirutka: Please stop adding unnecessary curly braces around all variables.

main/libmemcached/APKBUILD
tcely: Is there some difference that's important about how `check` is disabled?

jirutka: Of course there is, that’s why we have `options` – we can easily analyse what abuilds don’t have check enabled or implemented.

tcely:
I added the patch from https://github.com/memcached/memcached/pull/349 mentioned in https://github.com/alpinelinux/aports/pull/3439#issuecomment-369223202 to this PR>

tcely:
Switching to `https` for the source again makes sense to me.

Did we ever figure out what caused the problem reported in https://github.com/alpinelinux/aports/pull/3445?

main/memcached/APKBUILD
```diff
@@ -1,29 +1,41 @@
...
 	$pkgname.confd
 	$pkgname.initd"
-builddir="${srcdir}/${pkgname}-${pkgver}"
+builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
+	# extstore is broken on 32-bit
+	# https://github.com/memcached/memcached/issues/319
+	local _extstore=''
+	case "$CARCH" in
```
kaniini: Use `[ "$CARCH" != "x86" ] && _extstore="--enable-extstore"` instead.

main/libmemcached/APKBUILD
```diff
@@ -1,49 +1,43 @@
...
-}
+builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
-	cd "$_builddir"
+	cd "$builddir"
 	./configure \
-		--build=$CBUILD \
-		--host=$CHOST \
+		--build="$CBUILD" \
```
kaniini: why the quoting for `--build` and `--host`?

main/memcached/APKBUILD
jirutka: No, `[ "$CARCH" = "x86" ] || _extstore="--enable-extstore"` is better. Your variant would be interpreted as a failure on some shells (with `set -e` enabled).

jirutka:
Please do not mix unrelated changes in single commit, especially unrelated changes in two aports. There should be _at least_ two commits, one for `main/memcached` and one for `main/libmemcached` cleanup.

main/memcached/APKBUILD
tcely: The `case` should be easiest to read and maintain going forward. What's the argument against using case here?

main/libmemcached/APKBUILD
tcely: These should be quoted all the time.
The only case where the current usage isn't a problem is when `CBUILD` and `CHOST` are unmodified by word splitting.

main/memcached/APKBUILD
jirutka: I agree that `case` is a bit better in this case. My reaction was to @kaniini whose suggestion is actually worse, at least in terms of compatibility (and in this case it could be written in more compatible way that does not feel unnaturally).

I just don’t see how totally unnecessary left parenthesis (`  (x86) _extstore=…`) improves readability; it was the first time ever I’ve seen left parenthesis in `case` and I even wouldn’t know that it’s a valid shell syntax if I did not write a shell parser some time ago (and so studied POSIX Shell Command Language in detail).

main/libmemcached/APKBUILD
jirutka: I agree that it’s better to always quote all variables defined outside of the APKBUILD itself (i.e. author can’t be sure that it will not contain any spaces).

main/memcached/APKBUILD
tcely: I had the opposite reaction that you seem to be expressing. I always thought `foo)` was very odd looking and was quite happy to find out that `(foo)` was valid syntax too.

Do you know of any other usages of the `foo)` syntax other than `case` in `C` based languages?

andypost:
Btw maybe patch for secomp is not needed? Report https://bugs.alpinelinux.org/issues/8608 said it required

kaniini:
the patch for seccomp is required

andypost:
@kaniini Is it possible to backport to 3.7? right now  3.7 version does not work

kaniini:
Yes, I can do it shortly.

PR #3302, opened 2018-02-13
testing/tcsh: fix out of memory crash

judgegrubb:
Patch is from: https://github.com/openwrt/packages/blob/master/utils/tcsh/patches/001-sysmalloc.patch

Work done by [Milan P. Stanić](https://bugs.alpinelinux.org/users/2974).


judgegrubb:
Builds fine on my edge build. Seems like some package is damaged or missing in the Travis CI?

``ERROR: mpfr3-3.1.5-r1: BAD archive``

clandmeter:
should be fixed now.

clandmeter:
You did not bump pkgrel so this package will not be build on our builders. Please increase it.

clandmeter:
I just enabled testsuites on this package but it seems to fail many tests. Could you perhaps report it upstream and see what is going wrong?

[testsuite.log](http://tpaste.us/L6R4)

judgegrubb:
Alright, I bumped pkgrel. I submitted a bug report to tcsh-bugs@mx.gw.com which according to [their github repo](https://github.com/tcsh-org/tcsh) may be the best place to reach the developers. Unfortunately I can't seem to access [tcsh.org](http://www.tcsh.org) or [bugs.gw.com/](http://bugs.gw.com/) so I'm not crazy optimistic about a response.

PR #3304, opened 2018-02-13
main/autoconf: backport patch for runstatedir support

adrianguenter:
Autoconf release 2.70 appears to be stalled indefinitely, and with Alpine using `/run` instead of `/var/run`, having access to `--runstatedir=/run` for `configure` would be very nice indeed.

[Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=759647) has backported it, and given that it is small and unobtrusive in nature I was hoping we could do the same.

Thanks for considering!

### DIFF
--- a/main/autoconf/APKBUILD
+++ b/main/autoconf/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=autoconf
 pkgver=2.69
-pkgrel=1
+pkgrel=2
 pkgdesc="A GNU tool for automatically configuring source code"
 arch="noarch"
 license="GPL-2.0-or-later"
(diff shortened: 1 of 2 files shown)


adrianguenter:
Travis build failed with `ERROR: mpfr3-3.1.5-r1: BAD archive` during deps

tcely:
Doesn't the symbolic link from `/var/run` to `/run` help with this?

adrianguenter:
@tcely some packages rely on being able to specify `localstatedir` and
`runstatedir` independently. Like if `localstatedir` is `/var/package`, you
wouldn't necessarily want run state data in `/var/package/run` or `/run`;
you'd probably want to be able to specify `/run/package`.

adrianguenter:
@tcely some packages rely on being able to specify `localstatedir` and
`runstatedir` independently. Like if `localstatedir` is `/var/package`, you
wouldn't necessarily want run state data in `/var/package/run` or `/run`;
you'd probably want to be able to specify `/run/package`.

PR #3314, opened 2018-02-14
main/pgpool: Upgrade to 3.7.2

tcely:


main/pgpool/APKBUILD
```diff
@@ -24,12 +27,15 @@ prepare() {
...
+		--build="$CBUILD" \
+		--host="$CHOST" \
 		--prefix=/usr \
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
-		--infodir=/usr/share/info
+		--infodir=/usr/share/info \
+		--with-memcached=/usr \
+		--with-openssl \
+		--with-pam
```
ncopa: Why do we need PAM support? Do we really want that? Can you please include a good reason to why we should add PAM support in the commit message if you still think this is a good idea.

We normally try avoid PAM if possible, and it may be difficult to remove it once its added, so we just need to be sure that this is really needed.

tcely: Removed `pam` from configure options.

fabled:
Pgpool has been updated to 3.7.2 in commit 48e8f4d2a7b37e0594bf7dfcd0416eaa0e31d102. If the other changes are required, please open new PR and provide requested explanation for them.

PR #3315, opened 2018-02-14
 community/unibilium: upgrade to 2.0.0, rebuilds

ganwell:
I was not sure if is better to keep upgrade/bumps in one PR or not. I can split into 4 PRs if needed.


ganwell:
Fixes: https://bugs.alpinelinux.org/issues/8469 (again, this time at the source)

ganwell:
Vis doesn't fail in my edge-build-docker :confused:

ganwell:
`Purging unibilium-dev (2.0.0-r0)`

I double checked. I build vis against new unibilium and libtermkey and the checks don't fail.

PR #3317, opened 2018-02-15
testing/nim: use libucontext for coros

ganwell:


ganwell:
I just realized that libucontext 0.1 doesn't build on x86, maybe we should wait til 0.2.

kaniini:
0.1.1 builds on x86.

ganwell:
Great. Then this could be merged. I have only tested x86_64, but for some reason the build-server didn't run the checks [1] last time. So if it doesn't run the checks again, we won't see if coros work on the other ARCHs.

[1] http://build.alpinelinux.org/buildlogs/build-edge-x86_64/testing/nim/nim-0.17.2-r1.log (there should be a lot of "PASS:")

kaniini:
i guess algitbot is asleep.  i merged this.  https://git.alpinelinux.org/cgit/aports/commit/?id=d5700e15606bd8cc9c882ca4aeb4227a560f6776

PR #3326, opened 2018-02-16
testing/lmms: new aport

orsonteodoro:
LMMS is a free alternative to FruityLoops / FL Studio.  It can be used to produce music.

ALSA, PulseAudio [for casual users that want share audio across IP address, etc], Jack [for low latency for professional producers] were enabled.

The plugins (representing instruments, sound effects) were subpackaged, so there could be a warning about them when you start up the program.  If you don't like warnings just add them all.

Some subpackages were not enabled because there was missing packages like fluidsynth.  The APKBUILD can be modified if you need them.  A note was left in the APKBUILD for those interested in accessing those plugins.  Make a feature pull request to enable them after this pull request is merge.

QT5 is the default but QT4 might be considered for low resource builds.

More information:  https://lmms.io/
Source Code: https://github.com/LMMS/lmms
Main License: GPL-2.0-or-later
Other Licenses: GPL-2.0-only LGPL-2.1-or-later LGPL-2.1-only MIT CC-BY-3.0

### DIFF
--- a/testing/lmms/APKBUILD
+++ b/testing/lmms/APKBUILD
@@ -1,0 +1,613 @@
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=lmms
+pkgver=1.2.0_rc5
+_pkgver=${pkgver//_/-}
+pkgrel=0
+pkgdesc="Cross-platform music production software"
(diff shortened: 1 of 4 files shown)


kaniini:
Note that this software cannot touch any dependencies that are GPL-3.0-or-later because of the GPL-2.0-only declaration.  Whoever sponsors this package needs to check this explicitly.

PR #3327, opened 2018-02-16
testing/bird: upgrade to 2.0.1

Fusl:


nmeum:
It would be neat if you could add a comment to the patch file explaining why this patch is necessary.

Fusl:
@nmeum I tracked down the issue which turned out to be the difference between what glibc and musl-libc return on an I/O error. glibc returns `Input/output error` while musl-libc returns `I/O error`. The (now patched) test checks the return text and length which obviously do not match on musl-libc.

PR #3329, opened 2018-02-16
main/gmime: upgrade to 3.2.0 and modernize

rgdoliveira:


nmeum:
According to the checkapk output the soname changed:

```
…
-usr/lib/libgmime-2.6.a
-usr/lib/libgmime-2.6.so
+usr/lib/libgmime-3.0.a
+usr/lib/libgmime-3.0.so
…
```

You should rebuild all packages depending on gmime by bumping their pkgrel.

PR #3337, opened 2018-02-17
community/py3-qt5: subpackage qtwebkit, qtwebkitengine, qtwebchannel

orsonteodoro:
This will generate Python bindings for qtwebchannel, qtwebengine, qt5webkit.  py3-qt5-qtwebengine requires qt5-webengine-dev (#3338 ) which is not in aports yet.  The py3-qt5-qtwebengine subpackage is used by qutebrowser (not in aports yet).

### DIFF
--- a/community/py3-qt5/APKBUILD
+++ b/community/py3-qt5/APKBUILD
@@ -1,18 +1,20 @@
 # Contributor: Francesco Colista <fcolista@alpinelinux.org>
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=py3-qt5
 _pkgname=PyQt5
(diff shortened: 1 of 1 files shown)


nmeum:
Could you elaborate why you believe that the split is necessary? Can't we just make qutebrowser depend on py3-qt5?

orsonteodoro:
The reason why is that if we eliminate the subpackages (py3-qt5-qtwebkit py3-qt5--qtwebengine) then it will pull both qt5-webengine and qt5-qtwebkit large packages.  People just want one HTML rendering engine not both.  With the subpackage, it will pull the corresponding rendering engine.  For build time, it needs to pull both, but it is not necessary when resolving minimal dependencies.  Actually, py3-qt5-qtwebkit doesn't work for qutebrowser because it requires a special fork (or version) which seems like is not distributed on Alpine.

PR #3344, opened 2018-02-18
testing/fzf: Update and move to community

Ikke:
Update fzf to latest version and move it to community after positive feedback from multiple users.


community/fzf/APKBUILD
```diff
@@ -48,5 +47,13 @@ tmux() {
 	install -Dm0755 bin/fzf-tmux "$subpkgdir"/usr/bin/fzf-tmux
 }
 
-sha512sums="94695831b5c2460f516c982bcd006b4e380509dc77ca9d8785424029f1a5fc25c1150b91c0add6804fcc48d7880a63b94b81b63e56ce817a4837bb659884676e  fzf-0.17.0-2.tar.gz
+extras() {
+        description="additional scripts like shell completion and keybindings"
```
nmeum: For shell completions we usually add separated sub packages, e.g. `$pkgname-{fish,bash,zh}-completion` subpackages.

Take a look at `main/docker` if you need an example for this.

Ikke: Ah ok, I found the `-extras` from a different package (can't recall which), but I will adjust it. Thanks

nmeum: Well, looks like it. Would be nice if you could provide a new PRs introducing a completion subpackage.

Ikke: I will

PR #3347, opened 2018-02-18
testing/wireguard: upgrade to 0.0.20180218

zx2c4:
```

  * keygen-html: fix up copyright
  
  Copy and paste errors.
  
  * tools: do not collide types with libc clashes
  * tools: FreeBSD doesn't have EAI_NODATA
  * tools: fixup errno handling
  * tools: endian.h is not portable
  
  Fixes compilation and correctness several places.
  
  * tools: allow in-line comments
  
  You can now put a # comment anywhere in a line, in which case, it extends
  until the end of the line.
  
  * wg-quick: match from beginning rather than shift right
  
  This raises the proper error when providing interface names that are too long.
  
  * qemu: add support for powerpc
  
  Now that we have known PPC users, it's probably a good thing to ensure we
  don't introduce bugs, so PPC has been added to our CI on build.wireguard.com.
  
  * poly1305: fix up selftest counter
  
  Make sure we're using the right array length in the debug-mode-only
  self-tests.
  
  * netns: replace n0 ip with ip0, per custom
  
  Fixes up console output consistency.
  
  * qemu: more granular memleak detection
  
  This avoids us getting memory leak errors due to upstream's power management
  drivers leaking, or the like, when we're only interested in WireGuard memory
  leaks.
  
  * socket: free skb if there isn't an endpoint
  
  Fixes a memory leak.
  
  * allowedips: indicate to clang-analyzer that trie is non-null
  
  Hopefully future versions are slightly smarter...
  
  * blake2s: use union instead of casting
  
  Similarly fixes a clang-analyzer issue, as well as ensuring alignment.
  
  * tools: normalize strncpy/snprintf usage
  
  Correctness.
  
  * contrib: add embeddable wireguard library
  
  See: https://lists.zx2c4.com/pipermail/wireguard/2018-February/002387.html
  And: https://git.zx2c4.com/WireGuard/tree/contrib/examples/embeddable-wg-library/README
```

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20180202
-_mypkgrel=2
+_ver=0.0.20180218
+_mypkgrel=1
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


ncopa:
fixed with 1a1f1ce27697d3e65358cecee720aad2bea56d1c

PR #3351, opened 2018-02-18
testing/ddgr: new aport

dai9ah:


testing/ddgr/APKBUILD
```diff
@@ -0,0 +1,57 @@
+# Contributor: dai9ah <dai9ah@protonmail.com>
+# Maintainer:
+pkgname=ddgr
+pkgver=1.2
+pkgrel=0
+pkgdesc="DuckDuckGo from the terminal"
+url="https://github.com/jarun/googler"
```
awilfox: This seems like the wrong URL (it should be /ddgr not /googler).

dai9ah: Fixed URL.

PR #3358, opened 2018-02-19
Virtual OpenRC keymaps service

nmeum:
I used to have the following problem: I use a keymap which isn't properly parsed by busyboxs loadkmap applet. I therefore always used the loadkeys utility from the kbd package to set my keymap.

Unfortunately the keymaps services provided by the openrc aport only supports the busybox loadkmap applet. I therefore always manually changed the OpenRC keymaps services to use loadkeys instead. However, since there might be other people with this problem or people who prefer loadkeys over loadkmap I decided to implement a different solution which is purposed in this PR.

This solution makes the keymaps services a virtual services and let's main/kbd ship an OpenRC service `/etc/init.d/loadkeys` which provides this virtual service. Since it is not legal to have a virtual and real service with the same name I also renamed the keymaps service provided by the openrc aport to `/etc/init.d/loadkmap`.

This is currently WIP because I didn't test it enough so far but before I do that I would like to get some comments whether or not this change is likely to be merged.


nmeum:
I also considered moving the loadkmap service from the openrc aport to the busybox-initscripts aport since it no longer replaces an existing openrc upstream service, any opinions on that?

kaniini:
Sounds good, but make sure `replaces` is set correctly.

nmeum:
> Sounds good, but make sure replaces is set correctly.

Could you elaborate why this is necessary?

kaniini:
It is not necessarily guaranteed that the transactions will be processed in the correct order (upgrade openrc, upgrade busybox-initscripts).

Using `replaces` or declaring an explicit dependency on the new openrc version from busybox-initscripts are necessary to ensure the files wind up with the right owning package.

nmeum:
I moved the loadkmap OpenRC service to the busybox-initscripts and added some code the post-upgrade script to rename the old keymaps service to loadkmap. I also rebased the changes and tested this with my kdb setup (it seems to work fine so far).

Apart from the comments I added this should be good to merge, it would probably be a good idea though if the person who merges this tests this before doing so.

main/busybox-initscripts/APKBUILD
```diff
@@ -12,12 +12,12 @@ makedepends=""
 install="$pkgname.post-install $pkgname.post-upgrade"
 subpackages=""
 options="!check"
-replaces="alpine-baselayout openntpd"
+replaces="alpine-baselayout openntpd openrc"
```
nmeum: @kaniini that's good enough, right?

main/busybox-initscripts/busybox-initscripts.post-install
```diff
@@ -9,18 +9,29 @@ if [ -L /etc/runlevels/boot/mdev ] && ! [ -L /etc/runlevels/sysinit/mdev ]; then
...
-		continue
-	fi
-	if [ -d "$link" ] && ! [ -d "${link}d" ]; then
-		mv "$link" "${link}d"
-	fi
-done
+rename_service cron crond
+
+# keymaps was renamed to loadkmap
+rename_service keymaps loadkmap
```
nmeum: Do we also want to output some kind of message if the service was renamed? Users still need to adjust their /etc/conf.d/loadkmaps configuration file afterwards and if they made changes to /etc/init.d/keymaps they might need to remove that file manually.

main/busybox-initscripts/APKBUILD
kaniini: honestly i think i would prefer to see `depends="openrc>=[...]"` than using `replaces`.

it will be more robust.  the replaces then, should be unnecessary as openrc upgrade will always come before busybox-initscripts.

nmeum: Thanks for your input, done.

main/busybox-initscripts/busybox-initscripts.post-install
kaniini: Yes, I think it is a good idea.

nmeum:
The post-upgrade script now emits a message informing the user that the service has been renamed and suggesting additional steps that he might want to take.

kaniini:
looks good to me, go ahead and push it.  or do you want me to?

nmeum:
Did you test it as well? If so I would prefer if you push it. In case something breaks I am not the only one to blame then :wink:

nmeum:
Actually, it just came to my mind that we also need to update the setup-keymaps script of alpine-conf. I will provide a patch in a second.

nmeum:
Rebased, added a patch for `main/alpine-conf` and made it depend on `openrc>=0.24.1-r6`. Should (finally) be good to merge now!

kaniini:
Okay, going to give it a go on my laptop and if it's all good, I'll push it.

nmeum:
This was merged but somehow @algitbot didn't notice it, closing it.

BTW: Thanks for your feedback @kaniini!

PR #3360, opened 2018-02-19
main/fsarchiver: upgrade to 0.8.4, fix license

dai9ah:
~~added support for zstd compression~~
package built without zstd support, until further notice.

### DIFF
--- a/main/fsarchiver/APKBUILD
+++ b/main/fsarchiver/APKBUILD
@@ -1,14 +1,16 @@
 # Contributor: Bartłomiej Piotrowski <bpiotrowski@alpinelinux.org>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=fsarchiver
-pkgver=0.8.3
-pkgrel=1
+pkgver=0.8.4
+pkgrel=0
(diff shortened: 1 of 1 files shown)


nmeum:
zstd is currently in `community/` packages from `main/` can't depend on packages from `community/`.  If you really need to depend on zstd you either need to move zstd from `community/` to `main/` or move fsarchiver from `main/` to `community/`. I would personally prefer the later.

dai9ah:
@nmeum since zstd isn't a hard dependency, fsarchiver 0.8.4 can be successfully built without zstd support. is this preferable?

nmeum:
It didn't depend on zstd before, so yeah that's probably preferable. Anyway, @jirutka is the maintainer of this aport he might have an opinion on this.

jirutka:
I dunno, I don’t use this package myself, so don’t know if/how useful is support for zstd in fsarchiver.

nmeum:
> I dunno, I don’t use this package myself, so don’t know if/how useful is support for zstd in fsarchiver.

Well…in that case, not adding zstd support for now seems to be a good idea. Merged it, thanks for the patch @dai9ah!

PR #3157, opened 2018-01-29
testing/rakudo: new aport

CurtTilmes:
Supersedes PR https://github.com/alpinelinux/aports/pull/3042

Updated to latest release, 2018.01.

Adding 3 new aports: MoarVM, NQP, and Rakudo. Rakudo is dependent on NQP which is dependent on MoarVM, so they need to go together.

PR #3362, opened 2018-02-20
testing/rakudo: new aport

CurtTilmes:
Supersedes PR #3157 from last month that was never merged which superseded PR #3042 from the previous month which was never merged)
(Does anyone actually look at these or act on them?  Am I doing something incorrectly you could suggest I do differently? It would be really nice to get these into testing at least.)

Updated to latest release, 2018.02.

Adding 3 new aports: MoarVM, NQP, and Rakudo. Rakudo is dependent on NQP which is dependent on MoarVM, so they need to go together.


testing/moarvm/APKBUILD
```diff
@@ -0,0 +1,55 @@
...
+depends=""
+makedepends="perl gcc musl-dev make"
+install=""
+subpackages="$pkgname-dev $pkgname-doc"
+source="${pkgname}-${pkgver}.tar.gz::http://moarvm.org/releases/MoarVM-$pkgver.tar.gz"
+builddir="$srcdir"/MoarVM-"$pkgver"
+
+build() {
+	cd "$builddir"
+        unset CPPFLAGS CFLAGS
```
awilfox: Mixture of tabs and spaces.  Generally Alpine prefers tabs.

testing/nqp/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+# Maintainer: Curt Tilmes <Curt.Tilmes@nasa.gov>
+pkgname=nqp
+pkgver=2018.02
+pkgrel=0
+pkgdesc="Not Quite Perl"
+url="https://github.com/perl6/nqp"
+arch="all"
+options="!archcheck" # Arch dependencies are embedded
+license="Artistic-2.0"
+depends="moarvm"
```
awilfox: If the package uses a shared library from `moarvm`, declaring this is unnecessary.  `abuild` will automatically declare all shared library dependencies.

testing/nqp/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+makedepends="perl-utils gcc musl-dev make moarvm-dev"
```
awilfox: As with `moarvm`, adding `gcc`, `musl-dev`, and `make` is unnecessary here.

testing/nqp/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+	for file in CREDITS LICENSE README.pod VERSION docs; do
+                cp -r $file "$subpkgdir"/usr/share/doc/"$pkgname"
+        done
+}
+
+examples() {
+	cd "$builddir"
+        mkdir -p "$subpkgdir"/usr/share/doc/"$pkgname"
+        for file in examples; do
+                cp -r $file "$subpkgdir"/usr/share/doc/"$pkgname"
```
awilfox: Typically, this is installed to `"$subpkgdir"/usr/share/doc/"$pkgname"/examples`, and is installed with -doc.  However, if this is a large set of examples it may be better to use a separate subpackage.

wictory:
Hi!

I built the port. There was some tests that didn't go through for rakudo, so I disabled them for now. However I noticed this:
```
# apk add rakudo-dev
(1/3) Installing moarvm (2018.02-r0)
(2/3) Installing nqp (2018.02-r0)
(3/3) Installing rakudo-dev (2018.02-r0)
Executing busybox-1.27.2-r8.trigger
OK: 518 MiB in 134 packages
alpinevm:/etc/apk# perl6-m
Unhandled exception: While looking for '/home/wictory/src/outside/aports/testing/rakudo/src/rakudo-2018.02/perl6.moarvm': no such file or directory
```

wictory:
The failing tests are:

`t/04-nativecall/08-callbacks.t .................................. Failed 8/8 subtests`
and
`t/04-nativecall/21-callback-other-thread.t ...................... Failed 9/9 subtests`

CurtTilmes:
Thanks for building it.  Which architecture did the tests fail on?

wictory:
x86_64

CurtTilmes:
I fixed the embedded path problem.  How do I express that "rakudo" is a dependency of "rakudo-dev"?

wictory:
I ran `strace` on `t/04-nativecall/08-callbacks.t`, and this is end of the output:
```
open("/home/wictory/src/outside/aports/testing/rakudo/src/rakudo-2018.02/lib08-callbacks.so", O_RDONLY|O_CLOEXEC) = 29
fcntl(29, F_SETFD, FD_CLOEXEC)          = 0
fstat(29, {st_mode=S_IFREG|0755, st_size=9232, ...}) = 0
read(29, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0 \6\0\0\0\0\0\0"..., 960) = 960
mmap(NULL, 2105344, PROT_READ|PROT_EXEC, MAP_PRIVATE, 29, 0) = 0x751116f83000
mmap(0x751117183000, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED, 29, 0) = 0x751117183000
close(29)                               = 0
mprotect(0x751117183000, 4096, PROT_READ) = 0
mmap(NULL, 40, PROT_READ|PROT_WRITE|PROT_EXEC, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = -1 EPERM (Operation not permitted)
+++ killed by SIGKILL +++
```

The other test is multi-threaded, so it has a more complicated trace, but it's probably the same cause. I guess the problem here is that `moarvm` tries to allocate memory with both write and execute permissions.

CurtTilmes:
How much memory does your VM have allocated?  Make sure it has at least 2GB.

wictory:
The VM has 2G, I don't think it's the problem, since PaX enforces W^X permission on every page.

CurtTilmes:
I'm not quite sure how to handle this -- I think I could run paxctl in the build process, before and after the "make check" to allow the tests to succeed.

CurtTilmes:
No, actually the test itself builds the shared library that needs the PaX flags set, so the paxctl would need to be patched into that test.

wictory:
I believe that it should be fixed in moarvm itself. It's not a good thing to allocate memory with both write and execute permission, and there is probably a way to avoid it, i.e first allocate with write and read access, and then change the permissions to read and execute access.

awilfox:
> I fixed the embedded path problem. How do I express that "rakudo" is a dependency of "rakudo-dev"?

You set `depends=` inside the `dev` function.  This is typically not necessary because most `dev` packages include bare `.so` files that symlink to `.so.$SOVER`, which `abuild` then marks the package containing the real `.so.$SOVER` file as a dependency of `dev`.  This isn't one of those, so you'll have to declare it manually.

> It's not a good thing to allocate memory with both write and execute permission

That's not how JITs work.  Most packages, like `main/python3` and `community/openjdk8`, automatically find the `paxmark` utility and call it during the build process.  Perhaps add `paxmark` to `makedepends` in `moarvm`?

CurtTilmes:
Per @awilfox suggestion, I added `paxmark` dependency and called that on `moar` similar to many other alpine packages.  I built a PaX enabled system and the tests succeeded after adding that.

wictory:
With the `paxmark` call added, the tests are now passing for me too. 

However, when looking for culprits for W^X violation I found [this line](https://github.com/MoarVM/dyncall/blob/1dabd7c2fd438920502c85eaf9ca4212d722c4b5/dyncallback/dyncall_alloc_wx_mmap.c#L33) in the 3rdparty dyncall library. It looks like the W^X violation could be prevented in this case. The allocation is made in [this function](https://github.com/MoarVM/dyncall/blob/1dabd7c2fd438920502c85eaf9ca4212d722c4b5/dyncallback/dyncall_callback_x64.c#L44), which has a separate implementation for every architecture, and I see no reason that all RWX permissions should be present at the same time. This issue concerns all users of dyncall on all platforms actually implementing execute permissions, unfortunately it becomes apparent on systems where W^X is enforced.

Concerning this port I think it should be merged, if the alpine people agrees, to give people a chance to run perl6 on alpine. I got interested in this PR because I wanted to scratch my own itch. Thanks @CurtTilmes for the work and thanks @awilfox for the feedback!

PR #3364, opened 2018-02-20
main/rrdtool: upgrade to 1.7.0

TBK:
As of 1.6.0 librrd is fully thread-safe and librrd_th was removed.

https://oss.oetiker.ch/rrdtool/pub/CHANGES

I was wondering if a py3 subpackage would be appropiate since:
`* all new and much more complete python bindings working on both python 2.6+ and 3.3+`


kaniini:
it fails CI.  please fix, also definitely please add py3 package :)

TBK:
@kaniini Fixed the build issue.

I am not sure on how to add py3 in the best way, do you have an example?

Would copying the py2 subpkg content be an acceptable solution?

ncopa:
do we still need the python2 support? What is needed to be able to switch from python2 to python3 only?

PR #3372, opened 2018-02-21
main/procps: install some utilities to /usr/bin

awilfox:
This should fix [Alpine #7737](https://bugs.alpinelinux.org/issues/7737).

### DIFF
--- a/main/procps/APKBUILD
+++ b/main/procps/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=procps
 pkgver=3.3.12
-pkgrel=4
+pkgrel=5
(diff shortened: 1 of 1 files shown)


kaniini:
I merged this.

main/procps/APKBUILD
```diff
@@ -60,6 +60,12 @@ package() {
 	# These binaries are identical.
 	ln -sf pgrep "$pkgdir"/bin/pkill
 
+	# Override BusyBox symlinks
+	mkdir -p "$pkgdir"/usr/bin
+	for i in free pgrep pkill pmap pwdx top uptime; do
```
jirutka: Just for the next time, `i` should be declared as `local`.

awilfox: Thank you, I completely forgot about this.  I will add that to my notes.

PR #3384, opened 2018-02-23
main/postfix: upgrade to 3.3.0

andypost:
Ref http://www.postfix.org/announcements/postfix-3.3.0.html


main/postfix/APKBUILD
```diff
@@ -93,6 +92,11 @@ build() {
 	make OPT="$CFLAGS"
 }
 
+check() {
+	cd "$builddir"
+	make tests
```
andypost: somehow tests fail, so just update in this PR

PR #3396, opened 2018-02-23
main/alpine-baselayout: need /dev/console on s390x

tmh1999:
for some reasons tty are not recognized at the moment


tmh1999:
cc @ncopa

PR #3401, opened 2018-02-24
testing/keybase-client: upgrade to 1.0.44

dai9ah:


testing/keybase-client/APKBUILD
```diff
@@ -1,18 +1,14 @@
...
 pkgname=keybase-client
-pkgver=1.0.42
+pkgver=1.0.44
 _ver=${pkgver/_p/-}
 pkgrel=0
 pkgdesc="Keybase command line client"
 url="https://github.com/keybase/client"
 arch="all"
 license="BSD-3-Clause"
-depends=""
```
nmeum: Please don't remove variables set to an empty string, we have no clear policy on this and in the long run someone probably just readds them.

dai9ah: Fixed. Thank you for the clarification.

PR #3407, opened 2018-02-25
main/icu: upgrade to 60.2

andypost:
this upgrade needs to rebuild a lot of deps


nmeum:
> this upgrade needs to rebuild a lot of deps

In that case please include commits to rebuild the affected dependencies in this PR by bumping their pkgrel.

andypost:
@nmeum as I see some packages depends on subpackages of `icu-dev` or `icu-libs` what's not clear for me the `icu-doc` dependency
maybe @ncopa can point why docs kinda "orphaned" in dependencies https://pkgs.alpinelinux.org/package/edge/main/x86_64/icu-doc

kaniini:
I sorted this, thanks for your contribution

PR #3409, opened 2018-02-25
testing/android-tools: upgrade to 8.1.0_p7

nmeum:
This upgrade includes a ruby script for generating ninja build file
instead of manually creating a Makefile. The ruby script is taken from
the Arch Linux PKGBUILD.

Patches have also been taken from Arch Linux but needed some
modifications.


clandmeter:
@nmeum files are uploaded but it seems build is missing perl and go.

nmeum:
> @nmeum files are uploaded but it seems build is missing perl and go.

Thanks and yeah that could be the case. I didn't do an `abuild rootbld` yet, doing one right now ;)

PR #3432, opened 2018-02-27
community/qt5-qtbase: Allow more ssl API use (libressl compat patch)

jku:
Current libressl actually supports SSL_CTRL_SET_CURVES: don't prevent
using it with libressl.


jku:
oops didn't update the checksums

PR #3433, opened 2018-02-27
community/qt5-qtbase: Allow more ssl API use (libressl compat patch)

jku:
Current libressl actually supports SSL_CTRL_SET_CURVES: don't prevent
using it with libressl.

---

Full disclosure: I'm not on alpine at the moment and cannot test this on git master. The changes have been tested on 3.7.0 however: setting the elliptic curves works with libressl.


kaniini:
This patch seems to do the opposite of what you describe, it disables sections of code when libressl definitions are present.

jku:
Are you sure you didn't misread the quite unpleasant patch-of-patch format?

The original libressl-compat.patch disables the SSL_CTRL_SET_CURVES part when LIBRESSL_VERSION_NUMBER is defined. My change removes this part of the patch.

kaniini:
Yeah, I got confused.  It looks good.

PR #3434, opened 2018-02-27
main/gconf: add orbit2 to runtime deps

PureTryOut:
If gconf is compiled with orbit2, it's required as a runtime dependency as well. When a package depends on the orbit2 part of gconf, and orbit2 itself isn't actually installed, you'll get errors like `gconf/gconf-value.h: file not found` (while the file is actually located on the system in the proper location). This PR fixes this.

Thanks @abucodonosor for finding the cause and this fix!

### DIFF
--- a/main/gconf/APKBUILD
+++ b/main/gconf/APKBUILD
@@ -2,12 +2,12 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=gconf
 pkgver=3.2.6
-pkgrel=2
+pkgrel=3
 pkgdesc="GNOME configuration system"
 url="http://projects.gnome.org/gconf"
 arch="all"
 license="GPL-2.0+"
-depends=
+depends="orbit2"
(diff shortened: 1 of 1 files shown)


kaniini:
@awilfox can you fact check this?

PureTryOut:
Doing some more debugging and things are bit strange. Closing this for now till I'm absolutely sure

PureTryOut:
Ok, seeing we're probably the only one ever with this problem, I've decided to just add it as a dep to our own package. Just act like you never seen this PR :grin:

awilfox:
Do you have a test case isolated?  If so, and you can reproduce on vanilla Alpine edge, I think this should be merged.

Whether or not any Alpine packages actually use ORBit2 support in gconf, this is still technically a correctness issue and custom code compiled by Alpine users may use it.

kaniini:
We found problem, gconf-dev needs orbit2-dev now for some reason.

PureTryOut:
Ooh, well glad it was found anyway then!

PR #3438, opened 2018-02-28
testing/cloudi: upgrade to 1.7.3

okeuday:
* Enable armhf after disabling Haskell support on armhf
  (enabling Haskell support for armhf requires other testing based on
   https://github.com/alpinelinux/aports/commit/a55ed2043ebe1432fe30400d2aee3b080da32062#diff-b024f920925e868b70a04b5359308a8e )
* Changed the download URL to point to osdn.net instead of sourceforge.net
  because SourceForge has had an outage for more than 78 hours


okeuday:
@clandmeter Thanks!

PR #3442, opened 2018-03-01
testing/py-beautifulsoup4: fix doc

orsonteodoro:
Split command into two.

It could be a x86 problem.  I am on x86_64.

Addresses: http://build.alpinelinux.org/buildlogs/build-edge-x86/testing/py-beautifulsoup4/py-beautifulsoup4-4.6.0-r1.log


andypost:
Still makes sense but needs rebase on master

PR #3447, opened 2018-03-02
main/py-future: update to 0.16.0, fixes

orsonteodoro:
Added check()
Added patches to disable tests.
Split dir for python2 and python3 to silence build python3 error
Added checkdepends=, depends=
Updated subpackage=
filter depends= per python implementation
remove old re-enable-python3-5.patch

It uses community/py-argparse (available on 3.7 and edge) and testing/py2-unittest2 (available only on edge).


andypost:
Still makes sense to update & fix deps

PR #3457, opened 2018-03-02
community/php5: upgrade to 5.6.34

vakartel:
zlib extension now builtin and provided by php5-common


jirutka:
> zlib extension now builtin and provided by php5-common

Then you have to add `provides="$pkgname-zlib"` for backward compatibility.

PR #3462, opened 2018-03-02
main/py-exifread: add python3, clarify license, modernize

ghost:


fabled:
usr/bin/EXIF.py is shipped iin both python 2 and 3 version. That should be in the parent package I think.

PR #3466, opened 2018-03-03
testing/typescript: new aport

orsonteodoro:
Add optional dependency for testing/py-ycmd.

TypeScript is a superset of JavaScript for large scale JavaScript applications.

### DIFF
--- a/testing/typescript/APKBUILD
+++ b/testing/typescript/APKBUILD
@@ -1,0 +1,50 @@
+# Contributor: Orson Teodoro <orsonteodoro@hotmail.com>
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+
+pkgname=typescript
+_pkgname=TypeScript
+pkgver=2.7.2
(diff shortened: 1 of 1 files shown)


testing/typescript/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+arch="noarch"
+license="Apache-2.0 MIT ISC BSD-2-Clause BSD-3-Clause CC0-1.0 WTFPL"
+depends="nodejs-npm"
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/Microsoft/$_pkgname/archive/v$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
+	cd "$builddir"
+	npm install
```
jirutka: This is not acceptable. `npm install` does not guarantee that the same versions of dependencies will be installed on each run.

PR #3473, opened 2018-03-03
testing/tcc: updated source

HinataYanagi:
Surprisingly, tinycc has made a major release for the first time after the original developer Bellard declared giving up working on tinycc.
As tinycc has not progressed its version number for 5 years long, I adopted a date of the latest snapshot as its pkgver.
However, this update confirms that the version of tinycc is still in progress.
Forgive pkgver regression, or find some settlement.
Additionally, Arch has switched its tcc source to nongnu.org from repo.or.cz (see the [diff](https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/tcc&id=1f223217ce2913cf471c6022151021b94b64efec)).
So we'd better follow it.

### DIFF
--- a/testing/tcc/APKBUILD
+++ b/testing/tcc/APKBUILD
@@ -2,38 +2,37 @@
 # Contributor: Hinata Yanagi <hinasssan@gmail.com>
 # Maintainer: Hinata Yanagi <hinasssan@gmail.com>
 pkgname=tcc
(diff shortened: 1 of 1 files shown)


ncopa:
I reverted the tabs to whitespace changes and updated the commit message. Thanks!

PR #3478, opened 2018-03-04
main/fish: added dependency mdocml-apropos

BernhardGruen:
@kaniini I just added the missing dependency mentioned in https://bugs.alpinelinux.org/issues/8587

### DIFF
--- a/main/fish/APKBUILD
+++ b/main/fish/APKBUILD
@@ -2,12 +2,12 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=fish
 pkgver=2.7.1
-pkgrel=0
+pkgrel=1
 pkgdesc="Modern interactive commandline shell"
 url="http://www.fishshell.com"
 arch="all"
 license="GPL-2.0"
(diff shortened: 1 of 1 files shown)


xentec:
I think the more correct way would be to check for `apropos` existence in `__fish_describe_command` and return early on `false`. Something like this:
```diff
diff --git a/usr/share/fish/functions/__fish_describe_command.fish b/usr/share/fish/functions/__fish_describe_command.fish
index df2efc8..a0e47e4 100644
--- a/usr/share/fish/functions/__fish_describe_command.fish
+++ b/usr/share/fish/functions/__fish_describe_command.fish
@@ -3,6 +3,7 @@
 #
 
 function __fish_describe_command -d "Command used to find descriptions for commands"
+    type -q apropos; or return
     apropos $argv ^/dev/null | awk -v FS=" +- +" '{
                split($1, names, ", ");
                for (name in names)
```
My reasoning for this is simple: if `apropos` needs to be pulled as a dependency, then there are no docs available anyway making `apropos` useless.

PR #3479, opened 2018-03-04
linux-virt kernel does no longer depend on linux-firmware

BernhardGruen:
As written in issue https://bugs.alpinelinux.org/issues/8481 the unnecessary dependency to linux-firmware for the linux-virt kernel increases a virtual installation a lot. Also, this dependency is not there for linux-virthardened.


BernhardGruen:
Travis CI failed due to timeout.

PR #3480, opened 2018-03-04
community/namecoin: patched to build with boost 1.66.0

BernhardGruen:
@itoffshore 
After upgrading boost from 1.62.0 to 1.66.0 the build of namecoin will fail. This patch will solve it.
The patch does not change the behavior if still using boost 1.62.0

### DIFF
--- a/community/namecoin/APKBUILD
+++ b/community/namecoin/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=namecoin
 pkgver=0.13.0_rc1
 _ver=${pkgver/_/}
-pkgrel=9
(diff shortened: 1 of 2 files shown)


andypost:
Depends on boost upgrade https://github.com/alpinelinux/aports/pull/3481

BernhardGruen:
@andypost It does not strictly depend on the boost upgrade. This patch is _needed_ if the boost upgrade is done.

andypost:
Yes, patch is required for https://github.com/alpinelinux/aports/pull/2960

PR #3488, opened 2018-03-05
testing/linux-ck: new aport

orsonteodoro:
This package will add a new kernel.  This is contains Con Kolivas patchset for improved desktop interactivity with the MuQSS scheduler.

* Enabled BFQ IO scheduler by default via CONFIG_SCSI_MQ_DEFAULT.  This had to be manually set but was the default set by Kolivas. (http://ck.kolivas.org/patches/4.0/4.15/4.15-ck1/patches/0015-Enable-BFQ-io-scheduler-by-default.patch).  After testing bfq is not the default.  You need to manually set it by echo bfq > /sys/block/sda/queue/scheduler or some other method.  (This will be set in the zen kernel package I am working on will set bfq-sq to default.)
* CONFIG_BFQ_GROUP_IOSCHED was only enabled on s390x, x86_64, ppc64le or ones on servers/datacenters.
* Enabled MuQSS CPU Scheduler for all ARCHs.
* Set Timer frequency at 100 Hz.  This is the default set by Kolivas for MuQSS. (http://ck.kolivas.org/patches/4.0/4.15/4.15-ck1/patches/0012-Reinstate-default-Hz-of-100-in-combination-with-MuQS.patch)
* Set Preemptible Kernel (Low-Latency Desktop) for all ARCHS except for s390x and ppc64le which is set to No Forced Preemption (Server).  If any objections we could change it.
* All other settings are default like vanilla.  Specifically all settings are using the vanilla kernel's package configs then updated with make oldconfig.
* x86_64 has CONFIG_PAGE_TABLE_ISOLATION=y
* CONFIG_FORCE_IRQ_THREADING was enabled for all archs except for s390x.
* Setted CONFIG_RQ_MC (CPU scheduler runqueue sharing that was the default set by Kolivas) is set for all ARCHes having CONFIG_SCHED_MC (multicore support) or set to CONFIG_RQ_NONE for those without multicore support.  You can change this as a kernel parameter, so if you have hyperthreading, you can set rqshare=smt.

Additional patches will be uploaded:
* Kconfig CPU scheduler runqueue sharing for arm, aarch64, power64le, s390.
* Kconfig default option for BFQ
* Added 60-schedulers.rules README.Alpine to /usr/share/doc/$pkgname to be copied to /etc/udev/rules.d to force bfq.  The reason is because file collisions in other kernel packages.

On boot it will show "NOHZ: local_softirq_pending."  It's due to DYNTICKS and the 100 Hz and RCU set CONFIG_RCU_EXPERT to unset, CONFIG_PREEMPT=y, CONFIG_PREEMPT_RCU=y.  When I used the 1000 Hz, it wasn't a problem.  Some programs process slowly, but it is smooth interactively under high CPU usage (Two kernel compiles at the same time + YouTube video).  I will leave this 100 Hz and then create the Zen kernel which uses the essential settings from the Liquorix kernel (or binary release of Zen) which sets the Hz at 1000 and uses MuQSS.


orsonteodoro:
Closing for same reason: https://github.com/alpinelinux/aports/pull/3565#issuecomment-373613918 and https://github.com/alpinelinux/aports/pull/3561#issuecomment-373614000 .

PR #3489, opened 2018-03-05
testing/wireguard: upgrade to 0.0.20180304

zx2c4:
```
  * queueing: skb_reset: mark as xnet
  
  This allows cgroups to classify packets.
  
  * contrib: embedded-wg-library: add ability to add and del interfaces
  * contrib: embedded-wg-library: add key generation functions
  
  The embeddable library gains a few extra tricks, for people implementing
  plugins for various network managers.
  
  * crypto: read only after init
  * allowedips: fix comment style
  * messages: MESSAGE_TOTAL is unused
  * global: in gnu code, use un-underscored asm
  * noise: fix function prototype
  
  Small cleanups.
  
  * compat: workaround netlink refcount bug
  
  An upstream refcounting bug meant that in certain situations it became
  impossible to unload the module. So, we work around it in the compat code. The
  problem has been fixed in 4.16.
  
  * contrib: keygen-html: rewrite in pure javascript
  * Revert "contrib: keygen-html: rewrite in pure javascript"
  
  We nearly moved away from emscripten'ing the fiat32 code, but the resultant
  floating point javascript was just too terrifying.
  
  * Kconfig: require DST_CACHE explicitly
  
  Required for certain frankenkernels.
  
  * compat: use correct -include path
  
  Fixes certain out-of-tree build systems.
  
  * noise: align static_identity keys
  
  Gives us better alignment of private keys.
  
  * wg-quick: if resolvconf/interface-order exists, use it
  * wg-quick: if resolvconf/run/iface exists, use it
  
  Better compatibility with Debian's resolvconf.
  
  * contrib: add extract-handshakes kprobe example
  
  Small utility for extracting ephemeral key data from the kernel's memory. More
  information can be found here:
  https://lists.zx2c4.com/pipermail/wireguard/2018-February/002439.html
```

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20180218
-_mypkgrel=1
+_ver=0.0.20180304
+_mypkgrel=2
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


ncopa:
fixed with 66eefc32e96002c49637bb8db4e4f91dc4d23dad

PR #3490, opened 2018-03-05
main/gtk+3.0 Enable Wayland support

PureTryOut:
Allows GTK3.0 apps to be ran in a Wayland-only environment.


andypost:
Next one is upgrade https://github.com/alpinelinux/aports/pull/4040

PR #3491, opened 2018-03-05
main/x265: upgrade to 2.7, clarify license

ghost:


dai9ah:
The following dependent packages build successfully against libx265.so.151:
- ffmpeg-libs
- gst-plugins-bad
- handbrake
- handbrake-gtk
- vlc-plugins-codec
- x265-dev
- xpra

PR #3502, opened 2018-03-05
main/dovecot: security upgrade to 2.2.34

andypost:
https://www.dovecot.org/list/dovecot-news/2018-February/000370.html


andypost:
@ncopa Looks some components also needs upgrade but I assume it as backportable commit

PR #3507, opened 2018-03-06
main/clamav: security upgrade to 0.99.4

andypost:
http://blog.clamav.net/2018/03/clamav-0994-has-been-released.html


andypost:
I bet it needs backport to 3.7

PR #3511, opened 2018-03-06
community/docker-edge: new aport providing docker edge releases (currently 18.02.0)

BernhardGruen:
Pack, ship and run any application as a lightweight container
http://www.docker.io

Heavily based on aport community/docker

### DIFF
--- a/community/docker-edge/APKBUILD
+++ b/community/docker-edge/APKBUILD
@@ -1,0 +1,244 @@
+# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
+# Contributor: Bernhard J. M. Gruen <bernhard.gruen@googlemail.com>
+pkgname=docker-edge
+_pkgname=docker
+pkgver=18.02.0
+_ver=${pkgver/_/-}-ce
+pkgrel=0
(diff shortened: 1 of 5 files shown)


BernhardGruen:
Currently I don't check for an already installed docker apk. There should be some form of provide and conflict management. Don't know yet what would be the best way.

ncopa:
Do we want both `docker` and `docker-edge` packages? Can't we just upgrade `community/docker` to 18.02.0? People who wants stable will use stable alpine branch (currently v3.7) and people who wants latest and greatest will use alpine edge.

BernhardGruen:
The current stable version of Alpine (3.7) also had an edge release in it (17.10.0). Therefore I thought it would be nice to surely have both to choose from. At the moment one has to use a combination of edge and stable to get a stable docker release (17.12.1 vs. 17.10.0).

ncopa:
I have pushed the docker stable to alpine 3.7.

BernhardGruen:
As agreed Alpine Linux Edge versions can contain Docker Edge releases without fear. Stable Alpine Linux versions get the most current Docker CE stable version.

PR #3512, opened 2018-03-06
main/memcached: upgrade to 1.5.6, 

BernhardGruen:
disabled seccomp as it does not work when enabled
No connection with `telnet localhost 11211` is possible if compiled with seccomp

This was reported in https://bugs.alpinelinux.org/issues/8608


tcely:
https://github.com/alpinelinux/aports/pull/3299 includes these changes and a patch for seccomp

kaniini:
Done.

PR #3518, opened 2018-03-06
main/busybox: upgrade to 1.28.1

nmeum:
This is an upgrade to a new busybox major version.


main/busybox/busyboxconfig-extras
```diff
@@ -376,6 +372,14 @@ CONFIG_DEFAULT_SETFONT_DIR=""
 # CONFIG_FEATURE_START_STOP_DAEMON_FANCY is not set
 # CONFIG_WHICH is not set
 
+#
+# klibc-utils
```
nmeum: I invoked `make menuconfig` for both config files to update them for the new release. Busybox has introduced these 3 new applets in 1.28 which were enabled by `make menuconfig`, if we don't want these we should disable them.

main/busybox/APKBUILD
```diff
@@ -19,28 +19,21 @@ subpackages="$pkgname-static $pkgname-suid $pkgname-extras"
...
+	0008-fbsplash-support-image-and-bar-alignment-and-positio.patch
+	0009-depmod-support-generating-kmod-binary-index-files.patch
+	0010-Add-flag-for-not-following-symlinks-when-recursing.patch
+	0011-sysklogd-add-Z-option-to-adjust-message-timezones.patch
+	0012-udhcpc-Don-t-background-if-n-is-given.patch
+	0013-testsuite-fix-cpio-tests.patch
+	0014-miscutils-microcom-Fixed-segfault.patch
+	0015-ash-introduce-a-config-option-to-search-current-dire.patch
 
 	top-buffer-overflow.patch
```
nmeum: @kaniini you seem to have added this patch, does it indeed fix a buffer overflow (as the name suggests)? And if so: Did you report it upstream? Because this wasn't fixed in 1.28.

kaniini: Yes, it fixes a buffer overflow, but it was a quick and dirty fix I made.  I will send upstream.

kaniini: Sent upstream. http://lists.busybox.net/pipermail/busybox/2018-March/086284.html

main/busybox/busyboxconfig-extras
nmeum: Decided against enabling these by default.

nmeum:
This should actually be good to merge now. I will use this new busybox version on my laptop for the next few days in the meantime it would be great if someone could take a quick look to ensure that i didn't break any of our custom patches since some of them needed minor modifications for the upgrade.

main/busybox/APKBUILD
nmeum: Since [this has now been fixed upstream](https://git.busybox.net/busybox/commit/procps/top.c?id=86107d6dc7b6a182e01fb8296cb4e1680963184d) I decided to include that patch instead.

PR #3523, opened 2018-03-06
Added package perl-xml-tidy

sergey-safarov:


sergey-safarov:
Depend of PR https://github.com/alpinelinux/aports/pull/3524

PR #3525, opened 2018-03-06
testing/ntpsec: new aport

tcely:
https://www.ntpsec.org/
NTP reference implementation, refactored for security


tcely:
Requested in https://bugs.alpinelinux.org/issues/8573

PR #3534, opened 2018-03-07
testing/log4cplus: upgrade to 1.2.1

tcely:


testing/log4cplus/APKBUILD
```diff
@@ -42,4 +42,5 @@ package() {
 	make DESTDIR="$pkgdir" install
 }
 
-sha512sums="7a78c231415cd040eaaf9e2e921b3f177f0336e3fb3e4c5282395d248631e77c0cff9b96f74870e6f96423a097d08f4bc90da764724af9b6660560663d81e988  log4cplus-1.2.1-rc2.tar.gz"
+sha512sums="6996378061e591e8e71f9aaa9851ab7c0be6fbac56ffb30d60279538ad204f96ded9c8eb1c3f09317b208326b13c3cf89d9dac659e1b8edfbd4cd3d18d5bc6ef  log4cplus-1.2.1.tar.gz"
+_fingerprint='E406 292F 7D08 BBB0 0846  1314 04B8 9D51 DFE5 A215'
```
nmeum: What's the purpose of that variable?

tcely: It's the fingerprint of the key that signs the source release.
See https://github.com/log4cplus/log4cplus/blob/1.2.x/ChangeLog

nmeum: IIRC abuild doesn't compare fingerprints currently and (to my knowledge) we don't include fingerprints in other APKBUILDs either. I would there fore prefer it if you remove that variable for consistency.

I do agree that it would be nice to add pgp signature verification to abuild in the future but as long as this feature isn't implemented it doesn't make sense to include the public key fingerprint in the APKBUILD.

tcely: `abuild` doesn't do anything with `url` either, it's entirely informational.

Would you prefer to store this in a `FINGERPRINT` file instead?

nmeum: No, I would prefer not storing this information at all since we don't have any specified standard way of storing it.

And BTW, abuild does do something with the value of the ‘url’ variable, it's stored in the .apk metadata.

PR #3539, opened 2018-03-07
testing/py-m2crypto: fix for failing import call in python2

abbierwolf:
Fixed a bug preventing wokd from starting with an import failure.
Upgrade version to 0.29.0 and applied a patch found specific for alpine linux
which was not merged into the master before the release of 0.29.0.
For more info see https://gitlab.com/m2crypto/m2crypto/merge_requests/139


jirutka:
I looked into https://gitlab.com/m2crypto/m2crypto/merge_requests/139, but your patch here is much larger than the one in the MR. Why’s that?

PR #3561, opened 2018-03-09
testing/linux-pf: new aport

orsonteodoro:
This is a new kernel package designed for desktop.

This is post-factum's patchset.  http://pfactum.github.io/pf-kernel/

Patches merged and kept:
* PDS 0.98k CPU scheduler
* UKSM (Ultra Kernel Samepage Merging)
* GCC Patches
* BFQ patches
* Linux 4.15.6 (https://github.com/torvalds/linux/compare/v4.15...1a7aef62b47b00630e62a268d647f54ec93fb38c.diff) kernel update.
* add a sysctl to disable unprivileged user namespace unsharing -- ( https://patchwork.ozlabs.org/patch/639869/ ). The Zen Team added on top of it a Kconfig menu item.
* block: kyber: fix domain token leak during requeue (https://github.com/torvalds/linux/commit/ba989a01469d027861e55c8f1121edadef757797#diff-9d4eb3d858b49f81d5dac58a6a3c94e0)
* blk-mq: don't call io sched's .requeue_request when requeueing rq to ->dispatch (https://github.com/torvalds/linux/commit/105976f517791aed3b11f8f53b308a2069d42055#diff-fd6afa3ce27ac227b816e885bb5e3e44)
* block, bfq: limit sectors served with interactive weight raising (https://github.com/torvalds/linux/commit/8a8747dc01cee6f92a52c03ba686e9f60cb46c87#diff-702dca4797788860b69f4ff4e34e5695)
* drm/i915/edp: Only use the alternate fixed mode if it's asked
 for (https://bugs.freedesktop.org/attachment.cgi?id=135277)
* hwmon: (dell-smm) Disable fan support for Dell Inspiron 7720 (https://patchwork.kernel.org/patch/10187575/)
* platform/x86: intel_int0002_vgpio: Remove IRQF_NO_THREAD
 irq flag (https://patchwork.kernel.org/patch/10070625/)
* dell-wmi: support Dell Vostro 3360 (https://lkml.org/lkml/2017/11/20/744)

---

* Setted CONFIG_SCHED_PDS to yes.  The schedulers's formal name is "Priority and Deadline based Skiplist multiple queue Scheduler."
* Setted 1000Hz as recommended by PDS author Alfred Chen, except for s390 and ppc64le since they are servers. http://cchalpha.blogspot.com/2017/09/pds-098-release.html?showComment=1506777152198#c8911106156952036130
(Info about PDS is found in the same link.)
* Setted CONFIG_UKSM set to yes.
* Added 35 new subpackages for GraySky2 GCC patches, supporting "SKYLAKE BROADWELL HASWELL IVYBRIDGE SANDYBRIDGE SILVERMONT WESTMERE NEHALEM CORE2 ATOM PSC ZEN EXCAVATOR STEAMROLLER PILEDRIVER BULLDOZER JAGUAR BOBCAT BARCELONA K10 K8SSE3 K8 VIAC7 VIAC3_2 CYRIXIII K7 K6 PENTIUM4 PENTIUMM PENTIUMIII PENTIUMII 686 586MMX 586TSC 586" microarchitectures and generated config flags for these.  What it does is set the -march= to the above microarchitecture and enables some other optimization per microarchitecture.

Copied from the linux-zen ( #3565 ) package:

The other possibility is that we we could drop the 35 packages, the GraySky2's Gcc patch, the -O3 patches and stick to -Os.

or

We can just drop/bury the 35 subpackages and only distribute the generic builds but still keep -O3. If we bury it, the user can decide to optimize it for their architecture at home.

or

We can support a few of the 35 subpackages. Eliminate the x86 subpackages.

Reason why this is a concern is that it takes about 1-3 hours to build each subpackage. So 35 hours-105 hours to compile.  22 compiled for x86_64 and 13 built for the x86_64.

* Setted CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE=y for all ARCHs which will set -O2 compiler optimization
* BFQ set as the default scheduler
* Setted CONFIG_SCSI_MQ_DEFAULT=y (SCSI: use blk-mq I/O path) by default 
* Setted CONFIG_BFQ_GROUP_IOSCHED=y for x86_64, ppc64le, s390x for servers and datacenters that use cgroups.
* It also contains 60-schedulers.rules to force BFQ by default.  This is installed at /usr/share/doc/$pkgname because of file name and path collision with another package.  Instructions to install are in README.Alpine.

Wikipedia mentions the security risks of same page merging.  (https://en.wikipedia.org/wiki/Kernel_same-page_merging).  The person can decide to not install this package.


kaniini:
Hello,

Thank you for your contribution. Unfortunately, we are going to have to reject the contribution because adding additional kernel variants is something we do not wish to have the QA overhead of doing right now. We may add a desktop-optimized variant in the future.

You might consider hosting these kernels on your own APK repository instead.

I will be closing both the linux-zen and linux-pf PRs for this reason.

Thank you for your understanding!

PR #3564, opened 2018-03-09
community/beets: upgrade to Python3

prspkt:


nmeum:
Regarding mutagen: Does upstream still support python2? Because if so: We could provide both a -py2 and a -py3 subpackage. See: https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python#Multiversion_package_with_executables

ncopa:
we dont want add python2 support unless there is some package actually using it.

We also need to update `testing/libgpod` to use `python3` while at it.

ncopa:
looks like libgpod is python2 only and abandoned upstream. I think we should remove libgpod from testing...

PR #3568, opened 2018-03-10
Upgrade OCaml to 4.06, fix and improve ocaml packages etc.

jirutka:


avsm:
this is fantastic, thank you @jirutka!

andypost:
4.07 is out

PR #3569, opened 2018-03-10
community/elixir: upgrade to 1.6.3

gmile:


gmile:
Looks like a travis hiccup? This is what I'm seeing from travis logs:

> (1/19) Installing g++ (6.4.0-r7)
> ERROR: g++-6.4.0-r7: IO ERROR

nmeum:
Tests fail on x86: http://build.alpinelinux.org/buildlogs/build-edge-x86/community/elixir/elixir-1.6.3-r0.log do happen to have any idea how that test failure could be fixed?

gmile:
@nmeum can you please point me to the specific line with failure here http://build.alpinelinux.org/buildlogs/build-edge-x86/community/elixir/elixir-1.6.3-r0.log?

I searched for the `failures` in the log, all seems good. Apart from a couple of warning I don't see a failure. Sorry if I am missing something.

nmeum:
The test seems to have passed now. The old log was overwritten should have copied the corresponding lines. Might have been a race condition.

PR #3570, opened 2018-03-10
main/e2fsprogs: upgrade to 1.44.0, clarify license

prspkt:


main/e2fsprogs/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=e2fsprogs
-pkgver=1.43.9
+pkgver=1.44.0
 pkgrel=0
 pkgdesc="Standard Ext2/3/4 filesystem utilities"
 url="http://e2fsprogs.sourceforge.net"
 arch="all"
-license="GPL LGPL MIT"
+license="GPL-2.0-or-later"
```
nmeum: Are you sure about that? Arch Linux claims it is licensed under `GPL, LGPL, MIT` while Gentoo claims that it is licensed under `GPL-2 BSD`.

prspkt: Yes so, based on the license notice, `e2fsprogs` itself is made available under
the GPL-2.0-or-later, with the exception of the following libs:

`lib/ext2fs` and `lib/e2p` are `GPL-2.0-or-later`
`lib/uuid` is `BSD-3-Clause`
`lib/et` and `lib/ss` are `MIT`
`lib/uuid` is `BSD-3-Clause`

nmeum: That's a good question. @ncopa you are the maintainer of this aport. What's your opinion on this?

ncopa: Looks like fedora has main package as [GPLv2](https://src.fedoraproject.org/rpms/e2fsprogs/blob/master/f/e2fsprogs.spec#_7), [libs](https://src.fedoraproject.org/rpms/e2fsprogs/blob/master/f/e2fsprogs.spec#_49) subpackage as GPLv2 and LGPL, [libcom_err](https://src.fedoraproject.org/rpms/e2fsprogs/blob/master/f/e2fsprogs.spec#_94) and [libss](https://src.fedoraproject.org/rpms/e2fsprogs/blob/master/f/e2fsprogs.spec#_119) as MIT. I think we should do something similar.

prspkt: Amended licenses! There's still a tiny bit of ambiguity regarding sub-packages distributed under the LGPL-2.0, since they don't clearly specify if they're LGPL-2.0-(only/or-later).

prspkt:
`rm-ext2_ext_attr_entry_check.patch` resolves building btrfs-progs against e2fsprogs 1.44.0
https://github.com/kdave/btrfs-progs/issues/110

PR #3572, opened 2018-03-10
upgrades for various Xorg utilities

prspkt:


nmeum:
Thanks for your patch, applied it with some minor modifications. For instance sometimes the prepare function only consisted of a `default_prepare`-call which makes it unnecessary to define it.

PR #3576, opened 2018-03-10
community/wine: fix crypto emulation

martell:
seems we do not detect gnutls with pkgconfig correctly.
https://bugs.winehq.org/show_bug.cgi?id=44699

### DIFF
--- a/community/wine/APKBUILD
+++ b/community/wine/APKBUILD
@@ -7,7 +7,7 @@
 pkgname=wine
 pkgver=3.0.4
 _pkgver=${pkgver/_/-}
-pkgrel=0
+pkgrel=1
 pkgdesc="A compatibility layer for running Windows programs"
 url="https://www.winehq.org"
 arch="x86 x86_64"
@@ -25,6 +25,7 @@ makedepends="fontconfig-dev openldap-dev libxslt-dev libxxf86dga-dev
 	"
 source="https://dl.winehq.org/$pkgname/source/${pkgver%.[1-9]}/$pkgname-$_pkgver.tar.xz
 	no-pie.patch
+	wine-dlclose-bug41703.patch
 	winhlp32-flex.patch
 	"
 builddir="$srcdir/$pkgname-$_pkgver"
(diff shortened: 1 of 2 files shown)


martell:
ping?

martell:
ping @nmemu, @ncopa :)

martell:
This is not the correct fix. will update shortly

ncopa:
whats the status here?

martell:
@ncopa sorry for the delay.
Should be good now :)

martell:
ping @ncopa :)

nmeum:
You took the patch from [this upstream bug](https://bugs.winehq.org/show_bug.cgi?id=41703), right? If so: It would be nice to include a link to the bug report in the patch file. Besides did upstream really merge this patch as is? Because they did seem to have some comments on the proposed patch.

martell:
I can’t see it right now on mobile but I believe I made the patch filename
match the big report name or number.
There were several iterations but this is the last version and it seems to
have not gotten a reply yet.

I believe the first iteration of the patch was not accepted because it
would then only work for musl targets, I believe this has been fixed in
this version but I don’t believe that matters to us when our target will
always be musl libc.

I can add a comment to it or the commit to explain further if needed?

martell:
ping

martell:
@nmeum @ncopa not to be a little buggy but would it be possible to merge this?
I am hoping to make a release of a mingw-w64 clang toolchain using alpine as the build image and wine is used for running testcases.

andypost:
3.0.2 PR filed https://github.com/alpinelinux/aports/pull/5157

nmeum:
> 3.0.2 PR filed #5157 

Does 3.0.2 already fix this issues?

andypost:
I did merge 3.0.3 already

martell:
3.0.2 and 3.0.3 do not fix this issue.
dlclose does not behave the same on musl as other c libraries.
The patch for this comes from 
https://bugs.winehq.org/show_bug.cgi?id=41703#c28
It was created by xw897002528 on 2017-07-29 04:23:37 CDT but was never reviewed.

andypost:
The problem is that nobody did review of last patch https://bugs.winehq.org/attachment.cgi?id=58819&action=diff

martell:
> You took the patch from this upstream bug, right? If so: It would be nice to include a link to the bug report in the patch file. Besides did upstream really merge this patch as is? Because they did seem to have some comments on the proposed patch.

> There were several iterations but this is the last version and it seems to
have not gotten a reply yet.

> I personally have the following issues with this PR:
> It is unclear to me whether this patch was merged upstream as is because as I understand the comments in the issue tracker upstream had some issues with this patch. Correct me if I am wrong about this but in any case it would be nice to know if upstream merged this.
> The patch file itself doesn't contain any comments where it was taken from or what issues it attempts to fix. If it was in fact merged upstream I would prefer to include a git-format(1) patch.

>3.0.2 and 3.0.3 do not fix this issue.
> dlclose does not behave the same on musl as other c libraries.
> The patch for this comes from
> https://bugs.winehq.org/show_bug.cgi?id=41703#c28
> It was created by xw897002528 on 2017-07-29 04:23:37 CDT but was never reviewed.

Upstream wine devs seems to have no interest in reviewing this patch
I already named the patch `community/wine/wine-dlclose-bug41703.patch` 
Which should be very obvious as to where it came from and what it is.

I added more info to the top of the diff including this information in a fixup just now.
Is there any reason why this can't be merged?

martell:
need to merge #5633 for the buildbot to pass

martell:
updated now that @andypost merged #5633 
I believe all concerns are addressed @nmeum

martell:
ping @ncopa @nmeum

PR #3580, opened 2018-03-11
main/flite: upgrade to 2.1, modernize

prspkt:


nmeum:
Shouldn't the version tag in the pkgconfig file be updated as well? CC: @kaniini

prspkt:
Indeed it should!
Amended.

PR #3582, opened 2018-03-11
main/sic: apply two official patches

Mark-Weston:
The first patch is lineprint.patch (https://tools.suckless.org/sic/patches/lineprint).
This patch allow to store and reprint the current inserting line.
This is done by appending a suffix character ('\') to the input.
The line is (obviously) not sent.
This is useful, for example, when receiving data from the server in the middle of a sentence you’re writing.

The second patch is hidecommand.patch (https://tools.suckless.org/sic/patches/hidecommand).
This patch lets you hide IRC messages such as JOIN/QUIT.
You can use the sic command ‘h’ to specify what messages are hidden.
Example: :h JOIN,QUIT
Now all JOIN and QUIT messages will not be displayed.
The delimiter is unimportant (use what you like or nothing at all)
Use the command without arguments to stop hiding any messages.

### DIFF
--- a/main/sic/APKBUILD
+++ b/main/sic/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=sic
 pkgver=1.2
-pkgrel=0
+pkgrel=1
 pkgdesc="an extremely simple IRC client"
 url="http://tools.suckless.org/sic"
 arch="all"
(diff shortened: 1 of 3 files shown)


nmeum:
I would actually prefer not to extend the functionally of this aport with unofficial patches. We don't do this for other software from suckless.org either. I would argue that you should build your own version of this aport if you need these features.

Mark-Weston:
But these patches are from suckless website

nmeum:
> But these patches are from suckless website, so they count as official, right?

Those are patches supplied by users of sic. "Offical" patches are only those applied in the git repository https://git.suckless.org/sic/

> IMHO sic without these patches is simply unusable.

upstream seem to see this differently.

prspkt:
I believe out of tree patches for suckless ports are meant to be applied à la carte by users, not package maintainers.

nmeum:
I am sorry but I am closing this, due to the reason I outlined above.

PR #3584, opened 2018-03-11
Enable KEXEC

jrb0001:


nmeum:
Hey, thanks for your patch. I have one suggestion: Could you change the commit message to `main/linux-vanilla: enable kexec`?

Besides I would personally prefer it if new kernel features are enabled for all flavors (linux-vanilla, linux-hardened). However, I am not the maintainer of the kernel packages so this is not really up to me to decide. I will assign this to the maintainer.

jrb0001:
linux-hardened definitely shouldn't get kexec enabled because it allows to replace the currently running kernel which should be impossible in a high security environment. I can check what would be needed to get it into the virtualisation configurations though.

jrb0001:
Rebased to latest kernel version.

ncopa:
I wonder if you could explain *why* we need this. I am very sceptic to this since it is a way to modify the running kernel. It lets you bypass things like secure boot and signed kernel modules. I am also afraid that this may undermine other future kernel hardening attempts from upstream, which may want try make it harder to inject rootkits. This seems to sort of be the opposite, a syscall for installing rootkits.

jrb0001:
There is an open bug for it since two months: https://bugs.alpinelinux.org/issues/8400. If you don't want to enable these configuration options, you should drop the `kexec-tools` package immediately because it is unusable anyway.

ncopa:
Thanks for the URL to our bugtracker. Would have been nice to have had that reference in commit message.

People can still build their own kernels. I see no problem in us provide `kexec-tools` package to be used with homebuilt kernels. I do se a problem with enabling KEXEC for everyone though. We will not enable KEXEC in our default kernel.

PR #3585, opened 2018-03-11
community/fzf: shell subpackages

Ikke:
Split shell scripts in separate subpackages

The first commit makes everything consistently use tabs. The second commit fixes a build issue where the tests failed due to new checks added in Go 1.10

### DIFF
--- a/community/fzf/APKBUILD
+++ b/community/fzf/APKBUILD
@@ -48,11 +48,11 @@ tmux() {
 }
 
 extras() {
-        description="additional scripts like shell completion and keybindings"
+	description="additional scripts like shell completion and keybindings"
 
-        cd "$builddir"
-        install -dm0755 "$subpkgdir"/usr/share/fzf
-        install -m0644 shell/* "$subpkgdir"/usr/share/fzf
+	cd "$builddir"
+	install -dm0755 "$subpkgdir"/usr/share/fzf
+	install -m0644 shell/*.$1 "$subpkgdir"/usr/share/fzf
 }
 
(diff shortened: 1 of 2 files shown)


community/fzf/APKBUILD
```diff
@@ -47,13 +48,18 @@ tmux() {
 	install -Dm0755 bin/fzf-tmux "$subpkgdir"/usr/bin/fzf-tmux
 }
 
-extras() {
-        description="additional scripts like shell completion and keybindings"
+_do_shell() {
+	description="additional scripts for $1 like shell completion and keybindings"
```
nmeum: You probably want to use `pkgdesc` instead of `description` here.

Besides: You want to add an `install_if` to make sure that the subpackage is installed automatically if the corresponding shell is installed. The completion subpackages of the `community/docker` subpackage provide such an `install_if` rule.

Ikke: fixed the description and added the install_if directives.

community/fzf/APKBUILD
```diff
@@ -41,19 +47,38 @@ package() {
...
+	install -Dm0644 shell/completion.zsh "$subpkgdir"/usr/share/zsh/site-functions/_$pkgname
+	install -Dm0644 shell/key-bindings.zsh "$subpkgdir"/usr/share/$pkgname/key-bindings.zsh
+}
+
+fishcomp() {
+	pkgdesc="keybindings for fish"
+	install_if="$pkgname=$pkgver-r$pkgrel fish"
+
+	cd "$builddir"
+	install -Dm0644 shell/key-bindings.zsh "$subpkgdir"/usr/share/fzf/keybindings.fish
```
nmeum: I wouldn't ship the keybinding in a completion package since it is not a shell completion :wink: I would instead suggest installing all keybindgs from the `shell/` directory to `/usr/share/doc/$pkgname/examples` or something like that and make them part of the -doc subpackage.

nmeum:
Merged, thanks for your work on this!

PR #3590, opened 2018-03-12
testing/freeimage: Allow ppc64le to pickup Makefile defined CFLAGS

mksully22:
ppc64le needs -fPIC compiler option. Unsetting the compiler flags environment variables for ppc64le allows them to be setup by the freeimage makefile.

### DIFF
--- a/testing/freeimage/APKBUILD
+++ b/testing/freeimage/APKBUILD
@@ -2,10 +2,10 @@
 # Maintainer: Taner Tas <taner76@gmail.com>
 pkgname=freeimage
 pkgver=3.17.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Open Source library project for developers who would like to support popular graphics image formats."
 url="http://freeimage.sourceforge.net/"
-arch="!ppc64le"
+arch="all"
 license="GPL-2.0-only FreeImage"
 _distname="FreeImage"
 makedepends="dos2unix"
@@ -25,6 +25,9 @@ prepare() {
 }
 
 build() {
+	if [ "$CARCH" == "ppc64le" ]; then
+                unset CFLAGS CPPFLAGS CXXFLAGS
+        fi
 	cd "$builddir"
 	make
 }


testing/freeimage/APKBUILD
```diff
@@ -25,6 +25,9 @@ prepare() {
 }
 
 build() {
+	if [ "$CARCH" == "ppc64le" ]; then
+                unset CFLAGS CPPFLAGS CXXFLAGS
+        fi
```
ncopa: `checkbashisms` reports this:

````
possible bashism in APKBUILD line 28 (should be 'b = a'):
        if [ "$CARCH" == "ppc64le" ]; then
````

ncopa: I also wonder if we should unset the flags for all archs?

ncopa: I want avoid arch specific conditionals if possible.

What do you think about this patch instead:
````diff
diff --git a/Makefile.gnu b/Makefile.gnu
index 92f6358..293ee97 100644
--- a/Makefile.gnu
+++ b/Makefile.gnu
@@ -15,7 +15,8 @@ LIBRARIES = -lstdc++
 
 MODULES = $(SRCS:.c=.o)
 MODULES := $(MODULES:.cpp=.o)
-CFLAGS ?= -O3 -fPIC -fexceptions -fvisibility=hidden
+CFLAGS ?= -O3
+CFLAGS += -fPIC -fexceptions -fvisibility=hidden
 # OpenJPEG
 CFLAGS += -DOPJ_STATIC
 # LibRaw
@@ -23,7 +24,8 @@ CFLAGS += -DNO_LCMS
 # LibJXR
 CFLAGS += -DDISABLE_PERF_MEASUREMENT -D__ANSI__
 CFLAGS += $(INCLUDE)
-CXXFLAGS ?= -O3 -fPIC -fexceptions -fvisibility=hidden -Wno-ctor-dtor-privacy
+CXXFLAGS ?= -O3
+CXXFLAGS += -fPIC -fexceptions -fvisibility=hidden -Wno-ctor-dtor-privacy
 # LibJXR
 CXXFLAGS += -D__ANSI__
 CXXFLAGS += $(INCLUDE)
````

mksully22: Yes, that looks better and built fine on my ppc64le and x86_64 machines. I'll submit a new pull request. Thanks ncopa

PR #3592, opened 2018-03-12
testing/fs-uae: build ppc64le with jit disable and no altivec

mksully22:


testing/fs-uae/APKBUILD
```diff
@@ -16,13 +16,19 @@ build() {
 	export CC=clang
 	export CXX=clang++
 	unset CFLAGS CXXFLAGS
+	if [ "$CARCH" == "ppc64le" ]; then
+		jitoptions="--disable-jit" 
+		export CFLAGS="-g -O2 -U__ALTIVEC__"
+		export CXXFLAGS="-g -O2 -U__ALTIVEC__"
```
ncopa: why replacing the system configured CFLAGS? Eg, why conditionally replace the `-Os` with `-O2`?

What about something like this:
````sh
export CFLAGS="$CFLAGS -U__ALTIVEC__"
````

File: testing/fs-uae/APKBUILD [hunk 2]
```diff
@@ -16,13 +16,19 @@ build() {
 	export CC=clang
 	export CXX=clang++
 	unset CFLAGS CXXFLAGS
+	if [ "$CARCH" == "ppc64le" ]; then
```
ncopa: should avoid the bash specific `==`. Eg. `if [ "$CARCH" = "ppc64le" ]`.
But we may want enable this for the other disabled arches as well, so I'd prefer something like:
````sh
case "$CARCH" in
ppc64le) ....
esac
````

Because then its easier to add the other arches: `ppc64le|armhf|aarch64) ...`

testing/fs-uae/APKBUILD
mksully22: The original APKBUILD explicitly "unset CFLAGS CXXFLAGS" and let the fs-uae configure and makefile creation code arrive at the desired settings. If we leave from the original APKBUILD the "unset CFLAGS CXXFLAGS" statement alone then export CFLAGS="$CFLAGS -U__ALTIVEC__" just resolves to CFLAGs =-U__ALTIVEC__ 

In APKBUILD the patch explicityly added the 'export CFLAGS="-g -O2 -U__ALTIVEC__"' because  on prior builds for the ppc64le architecture the derived values in the Makefile were CFLAGS = "-g -O2" and  and CXXFLAGS="-g O2". I wanted to keep that the same and just add the -U__ALTIVEC__ statement.

Only the CPPFLAGS use tte -Os option which shouldn't be affected by the proposed changes.
None of the source built using CPPFLAGS involve altivec usage so there was no need to modify.

Making the change you suggested to eliminate the bashism, how do these changes to APKBUILD look instead?

--- APKBUILD.orig
+++ APKBUILD
@@ -2,10 +2,10 @@
 # Maintainer: Taner Tas <taner76@gmail.com>
 pkgname=fs-uae
 pkgver=2.8.3
-pkgrel=1
+pkgrel=2
 pkgdesc="Amiga emulator."
 url="https://fs-uae.net/"
-arch="all !ppc64le !armhf !aarch64"  # --enable-jit not supported for these arches
+arch="all !armhf !aarch64"  # --enable-jit not supported for these arches
 license="GPL-2.0-only"
 makedepends="zip clang-dev sdl2-dev libpng-dev libx11-dev libxi-dev openal-soft-dev linux-headers libmpeg2-dev glib-dev"
 subpackages="$pkgname-lang"
@@ -16,6 +16,14 @@
        export CC=clang
        export CXX=clang++
        unset CFLAGS CXXFLAGS
+       case "$CARCH" in
+               ppc64le)
+                       jitoptions="--disable-jit"
+                       export CFLAGS="-g -O2 -U__ALTIVEC__"
+                       export CXXFLAGS="-g -O2 -U__ALTIVEC__"
+               ;;
+       esac
+
        cd "$builddir"
        ./configure \
                --build=$CBUILD \
@@ -23,6 +31,7 @@
                --sysconfdir=/etc \
                --mandir=/usr/share/man \
                --localstatedir=/var \
+               $jitoptions \
                --enable-sdl2
        make
 }

mksully22: Hmm the suggested patch got mangled...I just copied from the clipboard so I must be missing a trick. Here are the key parts in a (slightly) more readable form. What do you think?

@@ -16,6 +16,14 @@
        export CC=clang
        export CXX=clang++
        unset CFLAGS CXXFLAGS
\+      case "$CARCH" in
\+               ppc64le)
\+                       jitoptions="--disable-jit"
\+                       export CFLAGS="-g -O2 -U__ALTIVEC__"
\+                       export CXXFLAGS="-g -O2 -U__ALTIVEC__"
\+               ;;
\+       esac
\+
        cd "$builddir"
        ./configure \
                --build=$CBUILD \
@@ -23,6 +31,7 @@
                --sysconfdir=/etc \
                --mandir=/usr/share/man \
                --localstatedir=/var \
\+               $jitoptions \
                --enable-sdl2
        make

PR #3593, opened 2018-03-12
testing/ocaml-camomile: workaround ppc64le segfault while building

mksully22:
build on ppc64le segvs when (cd _build/default/Camomile && ./tools/camomilelocaledef.exe --file locales/zh__PINYIN.txt locales) per https://github.com/yoriyuki/Camomile/issues/39

This patch works around the issue by increasing the hard stack limit to unlimited and soft stack limit to 65536 prior to kicking off the build on ppc64le.

### DIFF
--- a/testing/ocaml-camomile/APKBUILD
+++ b/testing/ocaml-camomile/APKBUILD
@@ -3,12 +3,12 @@
 pkgname=ocaml-camomile
 _pkgname=camomile
 pkgver=0.8.7
-pkgrel=1
+pkgrel=2
 pkgdesc="A Unicode library for OCaml"
 url="https://github.com/yoriyuki/Camomile"
 # x86, armhf, s390x: limited by ocaml aport
-# ppc64le: https://github.com/yoriyuki/Camomile/issues/39
-arch="all !x86 !armhf !s390x !ppc64le"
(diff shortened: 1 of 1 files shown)


jirutka:
I know about this workaround. The problem is that we don't have ulimit on Alpine and even if we had, it would probably not work on builders due to LXC security restrictions.

ncopa:
We have ulimit in alpine. its an internal ash command. It seems like the workaround works in LXC too.

ncopa:
I am merging this wit this minor change:
````diff
diff --git a/testing/ocaml-camomile/APKBUILD b/testing/ocaml-camomile/APKBUILD
index 82226caaee..d8f03279a1 100644
--- a/testing/ocaml-camomile/APKBUILD
+++ b/testing/ocaml-camomile/APKBUILD
@@ -17,7 +17,7 @@ source="$pkgname-$pkgver.tar.gz::https://github.com/yoriyuki/$_pkgname/archive/r
 builddir="$srcdir/Camomile-rel-$pkgver"
 
 build() {
-        if [ "$CARCH" == "ppc64le" ]; then
+        if [ "$CARCH" = "ppc64le" ]; then
                 ulimit -Hs unlimited
                 ulimit -Ss 65536
         fi
````

PR #3596, opened 2018-03-12
testing/bfs: Run the upstream testsuite

tavianator:


tavianator:
Is there a way to get the build to run as a non-root user?  The tests don't expect to be run as root.

nmeum:
The tests are run in a fakeroot, this can be disabled using:

```
options="!checkroot"
```

tavianator:
@nmeum Thanks!  I'll make the tests work with non-coreutils in a future release.

nmeum:
>  I'll make the tests work with non-coreutils in a future release.

Actually it's absolutely fine to have coreutils as checkdepends.

tavianator:
Yeah but portability is nice, and not too difficult in this case.  [These commits](https://github.com/tavianator/bfs/compare/c3b43840431b059266df6c8fcc051598e38457ef...3edbc453a86f10ddcc9fc5d2535aff07b40d2f8e) fix the non-coreutils build.  They'll make it into the next release.

In the meantime, I think this PR is good to go as-is.

PR #2930, opened 2017-12-14
community/erlang: upgrade to 20.2

gmile:

PR #3625, opened 2018-03-14
community/erlang: upgrade to 20.3

gmile:


community/erlang/0011-fix-libressl-build.patch
```diff
@@ -1,108 +0,0 @@
-From b5b6c7695c5377b5aa474d8620df7c3e9a1ba629 Mon Sep 17 00:00:00 2001
```
gmile: The patch is no longer needed, see https://github.com/alpinelinux/aports/pull/2930#issuecomment-358628891

PR #3630, opened 2018-03-15
main/dmenu: upgrade to 4.8

prspkt:


main/dmenu/APKBUILD
```diff
@@ -1,26 +1,25 @@
...
 install=""
-options="!check"
+options="!check" # no upstream/available testsuite
 subpackages="$pkgname-doc"
 source="http://dl.suckless.org/tools/$pkgname-$pkgver.tar.gz"
-builddir="$srcdir"/$pkgname-$pkgver
+builddir="$srcdir/$pkgname-$pkgver"
 
 prepare() {
-	default_prepare
```
jirutka: Why did you remove `default_prepare`? This should be always called in custom `prepare` function.

jirutka:
This aport has been already upgraded in a1516c7dd1ac1568a2aa90ea5c1bf86207847e7f.

PR #3639, opened 2018-03-15
main/iso-codes: upgrade to 3.79

prspkt:


main/iso-codes/APKBUILD
```diff
@@ -1,34 +1,32 @@
...
-pkgrel=1
-pkgdesc="Lists of the country, language, and currency names"
-url="http://pkg-isocodes.alioth.debian.org/"
+pkgver=3.79
+pkgrel=0
+pkgdesc="ISO codes and their translations"
+url="http://pkg-isocodes.alioth.debian.org"
 arch="noarch"
-license="LGPL-2.0-or-later"
+license="LGPL-2.1-or-later"
```
nmeum: I can't find an `or-later' clause in the source here either on a first sight.

nmeum: So, the files in `bin/` are in fact licensed under `LGPL-2.1-or-later` the `*.po` files only state that they are released `under the same license as the iso-codes package.` which makes me wonder if this means `LGPL-2.1` or `LGPL-2.1-or-later` :thinking:

prspkt: Same. The included license indicates the `*-or-later' clause :

    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Lesser General Public
    License as published by the Free Software Foundation; either
    version 2.1 of the License, or (at your option) any later version.

PR #3640, opened 2018-03-15
testing/wireguard-tools: depends on bash

neuschaefer:
wg-quick is explicitly a bash script, so it requires bash to be installed.

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,11 +3,12 @@
 
 pkgname=wireguard-tools
 pkgver=0.0.20180304
-pkgrel=0
+pkgrel=1
 pkgdesc="Next generation secure network tunnel: userspace tools"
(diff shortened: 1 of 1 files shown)


nmeum:
One could argue that this is only an optional dependency but since the package only ships two executable and one of them is a bash script having a dependency on bash sounds like a good idea to me. Merging this right away.

zx2c4:
Sorry I wasn't here earlier -- I've been traveling. Probably would have been nice to be notified about this commit via an @ too. Oh well.

Anyway, there are users of Alpine+WireGuard who don't need or use wg-quick and thus don't need or use bash. This therefore should not be a required dependency of the package. Please revert this commit.

neuschaefer:
Sorry about this.

(Perhaps the non-dependency on bash should also be documented in wireguard-tools/APKBUILD for extra clarity?)

PR #3642, opened 2018-03-15
testing/gnome-calculator: new aport

orsonteodoro:
Better looking basic calculator with better programming mode

License: GPL 3.0 or later


ncopa:
We normally don't add licenses that is not provided from upstream. Those license files is a problem they need to solve upstream.

PR #3648, opened 2018-03-16
main/hdparm: upgrade to 9.55, clarify license

prspkt:
Changed installation directory from `/sbin` to `/usr/bin`

### DIFF
--- a/main/hdparm/APKBUILD
+++ b/main/hdparm/APKBUILD
@@ -1,16 +1,21 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=hdparm
-pkgver=9.52
+pkgver=9.55
 pkgrel=0
-pkgdesc="A shell utility for manipulating Linux IDE drive/driver parameters"
-url="https://sourceforge.net/projects/hdparm/"
+pkgdesc="Utility for manipulating IDE device parameters"
+url="https://sourceforge.net/projects/hdparm"
 arch="all"
-license="BSD"
+license="BSD GPL-2.0"
 makedepends="linux-headers"
 options="!check"
 subpackages="$pkgname-doc"
 source="http://downloads.sourceforge.net/sourceforge/$pkgname/$pkgname-$pkgver.tar.gz"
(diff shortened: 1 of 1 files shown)


main/hdparm/APKBUILD
```diff
@@ -1,16 +1,20 @@
...
+url="https://sourceforge.net/projects/hdparm"
 arch="all"
-license="BSD"
+license="BSD GPL-2.0"
 makedepends="linux-headers"
 options="!check"
 subpackages="$pkgname-doc"
-source="http://downloads.sourceforge.net/sourceforge/$pkgname/$pkgname-$pkgver.tar.gz"
-builddir="$srcdir"/$pkgname-$pkgver
+source="$url/files/$pkgname/$pkgname-$pkgver.tar.gz"
```
ncopa: I don't like when I cannot copy paste `source`.

main/hdparm/APKBUILD
```diff
@@ -20,11 +24,10 @@ build() {
 
 package() {
 	cd "$builddir"
-	mkdir -p "$pkgdir"/usr "$pkgdir"/sbin
-	make DESTDIR="$pkgdir" install
-	install -m755 contrib/idectl "$pkgdir"/sbin
-	install -m755 contrib/ultrabayd "$pkgdir"/sbin
-	install -D -m 644 LICENSE.TXT $pkgdir/usr/share/licenses/hdparm/LICENSE.TXT
+	make DESTDIR="$pkgdir" sbindir=/usr/bin install
```
ncopa: I see that you move `/sbin/hdparm` to `/usr/bin/hdparm`. It would be nice with some sort of explanation why this is needed in the commit message. And if this is what we really want to do then we should also patch busybox hdparm to the same location.

prspkt: @ncopa Sure, source url restored.
Regarding the `/sbin/hdparm` to `/usr/bin/hdparm` move, i thought it is more consistent, because `hdparm` is not an integral part of the base system.
This of course can definitely be reverted.

PR #3653, opened 2018-03-16
main/linux-firmware: Fix name of subfolder packages

neuschaefer:
Currently, the packages are named, for example:

	linux-firmware-3com-20180110-r0 - firmware files for linux ( folder)

This patch fixes the code in folder() that generates the names. Now, it's:

	linux-firmware-3com-20180110-r1 - firmware files for linux (3com folder)

Fixes: 0f1c436cdb ("main/linux-firmware: lowercase subpackages / use local")


neuschaefer:
The build currently fails due to a different issue: https://bugs.alpinelinux.org/issues/8663

PR #3656, opened 2018-03-16
main/xen: upgrade to 4.8.3

HRio:
https://www.xenproject.org/downloads/xen-archives/xen-project-48-series/xen-483.html

Also remove some patches that has been dropped from APKBUILD (earlier) but
still included as files.


HRio:
@rnalrd xen 4.8.3 for Alpine 3.6

PR #3657, opened 2018-03-16
main/libvorbis: upgrade to 1.3.6, enable tests

prspkt:


main/libvorbis/APKBUILD
```diff
@@ -1,28 +1,31 @@
...
-url="https://xiph.org/vorbis/"
+url="https://xiph.org/vorbis"
 arch="all"
-options="!check" # Test suite doesn't compile.
 license="BSD-3-Clause"
 subpackages="$pkgname-dev $pkgname-doc"
 makedepends="libogg-dev"
-source="http://downloads.xiph.org/releases/vorbis/$pkgname-$pkgver.tar.gz
+source="http://downloads.xiph.org/releases/vorbis/$pkgname-$pkgver.tar.xz
 	CVE-2017-14160.patch
```
nmeum: CVE-2017-14160 wasn't fixed upstream? :flushed:

prspkt: Sorry. Patch removed.

nmeum: That was more of a question than a statement because the patch still applies cleanly and the libvorbis changelog doesn't mention CVE-2017-14160…

prspkt: Yes, `CVE-2017-14160.patch` still applies cleanly, since none of  the changes it introduces are present upstream `psy.c`. I believe we should keep it.

nmeum: Hm. It would good to know why upstream didn't fix it and why upstream didn't apply the patch we are currently applying downstream :thinking:

prspkt: Agreed. Submitted https://github.com/xiph/vorbis/issues/39

PR #3668, opened 2018-03-17
main/linux-vanilla: add perf subpackage

orsonteodoro:
Homepage: https://perf.wiki.kernel.org/index.php/Main_Page

Perf is a performance analysis tool.  I tried to create a separate package but requires parts of the kernel sources and the config files.

openjdk8 is in community branch.
numactl-dev is in testing branch.


kaniini:
the dependencies need to be cleaned up before we could accept this

ncopa:
I think we want build the userspace tools in separate aport. I have been working on a `linux-tools` aport some time ago. I pushed it (59050d431b87d3fdd23c30cfe2cbd48819dd09af) but realized afterwards that it needs to be merged with `testing/cpupower`.

PR #3669, opened 2018-03-17
main/varnish: upgrade to 6.0.0

tcely:
Removed all the old patches.
`musl-includes.patch` in https://github.com/alpinelinux/aports/pull/3659 was overly complicated.

Added `check` function, but disabled it because many tests failed.


andypost:
I used a bit different approach https://github.com/alpinelinux/aports/pull/3659

andypost:
Why all patches removed? 
I removed only python patch & added on more hunk for `include/vsb.h` and that's all changes
btw "stack overflow" patch still applies with offset

tcely:
@andypost The `musl-includes.patch` is a mess. It was adding headers to source files that didn't belong there.

As for the `calloc` patch, that was just changing from static to dynamic memory allocation. If this is actually still useful, I would appreciate an explanation from anyone who knows why.

andypost:
@tcely thanks for explanation, I did check the code and all your points makes sense, so I closed my PR in favour of your one. I also not sure about `calloc`

PR #3686, opened 2018-03-19
testing/hylafaxplus: fix ppc64le build break by controlling make parallelism

mksully22:
ppc64le needs to make -j1 to avoid race conditions that cause errors when trying to create files in directories before they are fully created. Current build error:
/bin/bash: sman.apps/cqtest.1m: No such file or directory

Note: make parallelism was removed in https://github.com/alpinelinux/aports/commit/234bc57e60982d09a11caf752e31bb1c57522ca9#diff-10964c35ee611a48b0cf7ba679362ecc so I have added it back in just for the ppc64le case.

### DIFF
--- a/testing/hylafaxplus/APKBUILD
+++ b/testing/hylafaxplus/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=hylafaxplus
 _pkgname=hylafax
 pkgver=5.5.9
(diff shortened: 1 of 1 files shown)


fcolista:
Committed with some cosmetic fixes, there were two spaces.

PR #3687, opened 2018-03-20
testing/gflags: new aport

wenerme:

PR #3688, opened 2018-03-20
testing/grpc: new aport

wenerme:
options="!check" because test is way too complex to run, required a lot different languages, tools.

this package depends on gflags https://github.com/alpinelinux/aports/pull/3687

grpc-cli: most time, grpc-cli is not needed, grpc-cli is a util for debug grpc service.

### DIFF
--- a/testing/grpc/APKBUILD
+++ b/testing/grpc/APKBUILD
@@ -1,0 +1,47 @@
+# Contributor: wener <wenermail@gmail.com>
+# Maintainer: wener <wenermail@gmail.com>
+pkgname=grpc
+pkgver=1.10.0
+pkgrel=0
+pkgdesc="The C based gRPC (C++, Python, Ruby, Objective-C, PHP, C#) "
+url="https://grpc.io/"
+giturl="https://github.com/grpc/grpc"
+arch="all"
+license="Apache2"
+depends="protobuf"
+depends_dev="protobuf-dev"
+makedepends="autoconf automake libtool libstdc++ protobuf-dev c-ares-dev openssl-dev gflags-dev gtest-dev zlib-dev yaml-dev"
+subpackages="$pkgname-dev $pkgname-cli"
+_googletest_rev=ec44c6c1675c25b9827aacd08c02433cccde7780
+source="$pkgname-$pkgver.tar.gz::https://github.com/grpc/grpc/archive/v$pkgver.tar.gz
+		googletest-$_googletest_rev.tar.gz::https://github.com/google/googletest/archive/$_googletest_rev.tar.gz
+		01-chttp2-maybe-uninitialized.patch"
+builddir="$srcdir/$pkgname-$pkgver"
+options="!check"
+
+unpack() {
+	default_unpack
+	cd "$srcdir"
+	rm -rf $pkgname-$pkgver/third_party/googletest
+	ln -sfn $PWD/googletest-$_googletest_rev/ $pkgname-$pkgver/third_party/googletest
+}
+
+build() {
+	cd "$builddir"
+	make -j$(nproc)
+}
+
+package() {
+	cd "$builddir"
+	make install prefix="$pkgdir/usr" -j$(nproc)
+	rm -f "$pkgdir/usr/share/grpc/roots.pem"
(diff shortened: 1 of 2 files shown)


jirutka:
⚠️ grpc must be very thoroughly reviewed if it does not bundle some libs or CA certificates as grpc-ruby do (see https://github.com/jirutka/user-aports/tree/v3.7/user/ruby-grpc). ⚠️

wenerme:
@jirutka Why grpc have to bundle libs or CA certificates ?

jirutka:
@jirutka Why grpc have to bundle libs or CA certificates ?

It does not _have to_ bundle, it definitely _should not_ bundle them, but they still do, because Google… 😕

wenerme:
I found `/usr/share/grpc/roots.pem` in final apk, should this included in package ?

jirutka:
> I found `/usr/share/grpc/roots.pem` in final apk, should this included in package ?

No, should not be.

wenerme:
DONE, ca has been removed

testing/grpc/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+unpack() {
+	default_unpack
+	cd "$srcdir"
+	rm -rf $pkgname-$pkgver/third_party/googletest
+	ln -sfn $PWD/googletest-$_googletest_rev/ $pkgname-$pkgver/third_party/googletest
+}
+
+build() {
+	cd "$builddir"
+	make -j$(nproc)
```
ncopa: you dont need to (and should not) set `-j$(nproc)` because we set this in `/etc/abuild.conf`:
````
export JOBS=20
export MAKEFLAGS="-j$JOBS"
````

File: testing/grpc/APKBUILD [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+
+package() {
+	cd "$builddir"
+	make install prefix="$pkgdir/usr" -j$(nproc)
+	rm -f "$pkgdir/usr/share/grpc/roots.pem"
+}
+
+cli() {
+	cd "$builddir"
+	make install-grpc-cli prefix="$subpkgdir/usr" -j$(nproc)
```
ncopa: same here. don't set `-j$(nproc)`

testing/grpc/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+pkgver=1.10.0
+pkgrel=0
+pkgdesc="The C based gRPC (C++, Python, Ruby, Objective-C, PHP, C#) "
+url="https://grpc.io/"
+giturl="https://github.com/grpc/grpc"
+arch="all"
+license="Apache2"
+depends="protobuf"
+depends_dev="protobuf-dev"
+makedepends="autoconf automake libtool libstdc++ protobuf-dev c-ares-dev openssl-dev gflags-dev gtest-dev zlib-dev yaml-dev"
```
ncopa: We should use `libressl-dev` instead of `openssl-dev`

ncopa:
it fails to build with libressl 2.6:
````
src/core/lib/security/credentials/jwt/jwt_verifier.cc: In function 'EVP_PKEY* pkey_from_jwk(const grpc_json*, const char*)':
src/core/lib/security/credentials/jwt/jwt_verifier.cc:526:47: error: 'RSA_set0_key' was not declared in this scope
   if (!RSA_set0_key(rsa, tmp_n, tmp_e, nullptr)) {
                                               ^
````

but I think it will build just fine with libressl 2.7, which I am working on now

wenerme:
So, should I just wait 2.7 to fix the pr

ncopa:
yes, lets wait til after libressl-2.7 (may take a few days).

I can confirm that it builds with libressl 2.7

andypost:
Looks 2.7 stabilized

andypost:
Btw 1.10 is herehttps://github.com/alpinelinux/aports/blob/master/testing/grpc/APKBUILD 

And 1.11 releasedhttps://github.com/grpc/grpc/releases/tag/v1.11.0

wenerme:
I close this pr now , creata another version upgrade pr.

PR #3700, opened 2018-03-21
main/libuv: upgrade to 1.19.2

TBK:
100% backwards compatibility - https://abi-laboratory.pro/tracker/timeline/libuv/


TBK:
`Assertion failed in test/test-connect-unspecified.c on line 56: uv_tcp_connect(&connect6, &socket6, (const struct sockaddr*) &addr6, connect_6) == 0` fails since there is no IPv6 support on CI. I have added a patch.

PR #3701, opened 2018-03-21
testing/liboggz: new aport

rgdoliveira:
Oggz provides a simple programming interface for reading and
writing Ogg files and streams


rgdoliveira:
All tests passed on ppc64le but in x86_64 got this segfault. Need to investigate...

PR #3715, opened 2018-03-21
community/cassandra-cpp-driver: upgrade to 2.8.1

andypost:
Patch removed cos it was accepted


andypost:
Also it may help to https://github.com/alpinelinux/aports/pull/2873

PR #3716, opened 2018-03-22
libcouchbase, perl-couchbase & php7-couchbase: upgrade, rebuild & modernize

TBK:
This PR should only be accepted after https://github.com/alpinelinux/aports/pull/2873


TBK:
CI fails to use the libcouchbase version it just build and therefore the php7-couchbase build fails.

PR #3722, opened 2018-03-22
main/iniparser: upgrade to 4.1

TBK:


main/iniparser/APKBUILD
```diff
@@ -1,52 +1,38 @@
...
-	cd "$_builddir"
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
+build() {
+	cd "$builddir"
+#	make CC="${CC:-gcc}" CFLAGS="${CFLAGS} -fPIC" LDFLAGS="$LDFLAGS" \
+#		libiniparser.so
```
jirutka: Why is this here?

TBK: Not sure, I forgot to investigate.

PR #3725, opened 2018-03-22
community/docker: upgrade to 18.03.0

mmonnier:


TBK:
The normal wording for version bumps is "community/docker: upgrade to 18.03.0"

dawidd6:
I think you should also change the commit message.

BernhardGruen:
Looks good for me.
Just a remark: this is a stable version that can be backported to 3.7

PR #3728, opened 2018-03-22
main/unbound: upgrade to 1.7.0

tcely:


tcely:
> unbound-dev: libevent-dev needs to be rebuilt

nmeum:
> unbound-dev: libevent-dev needs to be rebuilt

Why?

tcely:
@nmeum It looks like something went wrong with tracing dependencies.

https://github.com/alpinelinux/abuild/blob/master/abuild.in#L1214

`libevent-dev` should have been found from what I can see for `pc:libevent` so that the else block shouldn't be emitting that message.

tcely:
See https://github.com/alpinelinux/aports/pull/3814

PR #3733, opened 2018-03-23
main/busybox: fix EXTRACT_UNSAFE_SYMLINKS from upstream

tmh1999:


nmeum:
This has been fixed by upgrading busybox in #3799

PR #3735, opened 2018-03-23
testing/bird: upgrade to 2.0.2

Fusl:
```
Version 2.0.2 (2018-03-22)
  o Source-specific routing support for Linux kernel and Babel
  o BGP: New option 'disable after cease'
  o Filter: Allow silent filter execution
  o Filter: Fixed stack overflow in BGP mask expressions.  <----
  o Several bugfixes

  Notes:

  Syntax prefix:netmask for IPv4 prefixes was dropped. Just use prefix/pxlen.
```

### DIFF
--- a/testing/bird/APKBUILD
+++ b/testing/bird/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Katie Holly <holly@fuslvz.ws>
 # Maintainer: Francesco Zanini <francesco@zanini.me>
 pkgname=bird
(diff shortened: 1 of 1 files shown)


Ikke:
Package builds successfully locally, but fails on Travis because it cannot fetch from FTP.

andypost:
It was already commited https://github.com/alpinelinux/aports/commit/c13b5c993b1105a6b11b6e37ed9a53c67e5b6b81

PR #3751, opened 2018-03-23
main/talloc: upgrade to 2.1.12

prspkt:


main/talloc/APKBUILD
```diff
@@ -35,8 +35,7 @@ build() {
 }
 
 check() {
-	cd "$builddir"
-	make check
+	make -C "$builddir" check
```
nmeum: I would prefer not to commit such stylistic changes since we don't have any clear policy on this and it might just be the case that the next person who touches this packages prefers an explicit cd and changes this back.

I manually changed this in your previous PRs and will change it for this one as well, but just keep this in mind for further contributions. Thanks!

PR #3756, opened 2018-03-24
community/tinc-pre: fix race in basic test that is failing on ppc64le

mksully22:
if multiple rx packets are received in by sptps_receive_data(&s, bufp, len) invoked in sptps_test.c,  the loop only processes the first one and breaks leaving the second unprocessed. This creates a potential race condition that could result in failures/hangs in sptps-basic.test


mksully22:
will have to investigate further why this is causing a failure on travis.

PR #3758, opened 2018-03-24
testing/tlp: bump to 1.1

sroracle:
* Bump from 0.9 (released 18 Aug 2016) to 1.1 (released 24 Jan 2018)
* Add `!check`: no test suite
* Correct license
* Retab bashcomp

### DIFF
--- a/testing/tlp/APKBUILD
+++ b/testing/tlp/APKBUILD
@@ -2,14 +2,15 @@
 # Maintainer: Ivan Tham <pickfire@riseup.net>
 pkgname=tlp
 _pkgname=TLP
-pkgver=0.9
-pkgrel=1
+pkgver=1.1
+pkgrel=0
 pkgdesc="Linux Advanced Power Management"
 url="http://linrunner.de/en/tlp/tlp.html"
 arch="noarch"
-license="GPL-2.0"
+license="GPL-2.0+"
+options="!check"  # No test suite.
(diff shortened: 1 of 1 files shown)


testing/tlp/APKBUILD
```diff
@@ -2,14 +2,15 @@
...
 _pkgname=TLP
-pkgver=0.9
-pkgrel=1
+pkgver=1.1
+pkgrel=0
 pkgdesc="Linux Advanced Power Management"
 url="http://linrunner.de/en/tlp/tlp.html"
 arch="noarch"
-license="GPL-2.0"
+license="GPL-2.0+"
```
nmeum: We are using [SPDX license names](https://spdx.org/licenses/) nowadays. Please change this to `GPL-2.0-or-later`

sroracle:
I believe it’s been discussed on IRC that the “+” notation is an acceptable
holdover from SPDX 2 for the sake of conciseness.

>

sroracle:
Sorry, it was discussed on the ML, not on IRC.[[1]](http://lists.alpinelinux.org/alpine-devel/6053.html)

testing/tlp/APKBUILD
andypost: From ML
> After discussing with jirutka, we came to the conclusion that SPDX 2
shorthand identifiers are fine as long as they are not vague. For
example "GPL-2.0+" is equally valid to "GPL-2.0-or-later". This
resolves the main gripe that Adelie has with SPDX 3.

PR #3759, opened 2018-03-24
main/upower: move /rules.d to /lib/udev/rules.d

sroracle:


nmeum:
Is there any kind of udev upstream documentation regarding the difference between `/etc/udev/rules.d` and `/lib/udev/rules.d`?

sroracle:
* "debian puts most of the rules in `/lib/udev/rules.d`" [[1]](https://wiki.debian.org/udev)
* "`/lib/udev/rules.d/` - Rules in this directory are installed by certain packages, they generally should not be changed by users;" [[2]](https://wiki.gentoo.org/wiki/Udev)
* "the local administration directory `/etc/udev/rules.d` ... Files in `/etc` have the highest priority" (`/etc` should be used by local administrators to override distribution defaults) [[3]](https://mirrors.edge.kernel.org/pub/linux/utils/kernel/hotplug/udev/udev.html)

nmeum:
Ok, thanks for researching this. In that case we certainly want to use `/lib/udev/rules.d`

PR #3760, opened 2018-03-24
testing/libupnp: upgrade to 1.6.25

prspkt:


testing/libupnp/APKBUILD
```diff
@@ -1,30 +1,37 @@
...
+builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
 	cd "$builddir"
-	./configure --prefix=/usr
-	make || return 1
+	./configure \
+		--prefix=/usr \
+		--enable-ipv6 \
+		--disable-static
```
jirutka: Why `--disable-static`?

prspkt: My mistake. Restored.
Is it OK to keep `--enable-ipv6`?

jirutka: Yes, of course, IPv6 should be enabled. ;)

PR #3761, opened 2018-03-25
community/xkcdpass: upgrade to 1.16.3

tcely:


fabaff:
Duplicate of #3739.

nmeum:
I personally prefer using `sed(1)` instead of `patch(1)` in this case. But that's just my personal opinion I guess…  

If we decide to stick with `sed(1)` though we should definitely invoke `default_prepare` in `prepare`.

tcely:
@nmeum I believe the installed script is looking for `python` not `python3` currently.

fabaff:
Using `sed` is often less maintenance as you don't need to re-write the patch if changes happen.

fabaff:
> I believe the installed script is looking for python not python3 currently.

It's not the script itself, it's the test.

tcely:
@fabaff It's both. This isn't a big deal as the wrapper `/usr/bin/xkcdpass` does use `python3` as it should.

```sh
$ apk -v info -ea | grep xkcd
py3-xkcdpass-1.16.3-r0
xkcdpass-1.16.3-r0
$  head -n 4 /usr/lib/python3.6/site-packages/xkcdpass/xkcd_password.py
#!/usr/bin/env python
# encoding: utf-8

from __future__ import print_function
```

PR #3768, opened 2018-03-25
community/vis: upgrade to 0.5

prspkt:
Testing succeeds locally, but fails consistently for both vis 0.5 and vis 0.4.

### DIFF
--- a/community/vis/APKBUILD
+++ b/community/vis/APKBUILD
@@ -1,10 +1,10 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Sören Tempel <soeren+alpine@soeren-tempel.net>
 pkgname=vis
-pkgver=0.4
-_testver=0.2
-pkgrel=1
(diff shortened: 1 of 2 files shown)


jirutka:
> Testing succeeds locally, but fails consistently for both vis 0.5 and vis 0.4.

What do you mean by that? Tests for both 0.4 and 0.5 succeed locally, but fail on Travis?

Ikke:
I can confirm, the checks succeed for me locally as well. It seems something [crashes](https://travis-ci.org/alpinelinux/aports/builds/358198444#L1320) in travis, because there is a core dump.

prspkt:
@jirutka While travis runs the test suite, there's some sort of segfault and a subsequent core dump. This does not occur locally. Based on the [log](https://api.travis-ci.org/v3/job/358198445/log.txt), this does not happen for all tests.

nmeum:
The `structural-regex/commands/filter-and-edit` test also fails when building the package using `abuild rootbld` otherwise I can confirm that all tests build locally.

PR #3771, opened 2018-03-25
main/openvswitch: upgrade to 2.9.0

itoffshore:
https://github.com/openvswitch/ovs/blob/master/NEWS


main/openvswitch/debian.readme.patch
```diff
@@ -0,0 +1,119 @@
+--- openvswitch-2.9.0/debian/openvswitch-switch.README.Debian
++++ openvswitch-2.9.0/debian/openvswitch-switch.README.Debian.new
```
jirutka: It’s quite weird to have `README.Alpine` in file `README.Debian`…

itoffshore:
```
        install -m644 debian/openvswitch-switch.README.Debian \
                "$pkgdir"/usr/share/doc/openvswitch/README.alpine
```

jirutka: It’d be better to just remove README.Debian and add README.alpine to the abuild as a file, not patch.

itoffshore: fixed & rebased

jirutka: I’m very sorry for the mess, I’ve overlooked that this README file is much longer and only small part is changed to fit Alpine. 😿  Could you please revert this change and keep it as a patch? You can you `git reflog` to find SHA of the previous commit and restore it.

itoffshore: reverting the sha reverted the whole PR

PR #3780, opened 2018-03-26
community/zstd: upgrade to 1.3.4, clarify license

prspkt:


community/zstd/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
 pkgname=zstd
-pkgver=1.3.3
-pkgrel=1
+pkgver=1.3.4
+pkgrel=0
 pkgdesc="Zstandard - Fast real-time compression algorithm"
 url="http://www.zstd.net"
 arch="all !armhf"
-license="BSD"
+license="BSD-3-Clause OR GPL-2.0-or-later"
```
nmeum: We don't really support the `OR` notation yet. This will also cause a warning when https://github.com/alpinelinux/abuild/pull/36 is merged. I would either pick one or mention both. In the future we definitely need better SPDX support in abuild to be able to express something like this.

prspkt: Noted. In this case i think mentioning both is correct and doesn't hurt compliance.

PR #3788, opened 2018-03-27
testing/openimageio: upgrade to 1.8.9

andypost:
Required for #2960


andypost:
this require fresh boost to build

prspkt:
@andypost The build phase will succeed if you add `-DSTOP_ON_WARNING=0` to the template configuration flags. It will however fail later during the python subpkg.

andypost:
@prspkt good idea but maybe better to wait for boost update?

PR #3789, opened 2018-03-27
main/f2fs-tools: upgrade to 1.10.0, clarify license, modernize

prspkt:


nmeum:
Could you split the upgrade and the modernization into two separate commits? Makes it easier to review this.

prspkt:
@nmeum Yes certainly.

main/f2fs-tools/APKBUILD
```diff
@@ -36,34 +36,23 @@ snapshot() {
...
-	done
-	libtoolize --force && aclocal && autoheader && autoconf \
-		&& automake --add-missing
-}
-
 build() {
-	cd "$_builddir"
+	cd "$builddir"
+	
+	./autogen.sh
```
nmeum: Could you move the `autogen.sh` invocation to the prepare stage by adding a `prepare()` function the APKBUILD? Also make sure to call default_prepare and make sure you change the directory afterwards.

prspkt: Sure.

prspkt:
@nmeum The last commit removes snapshotting since upstream provides versioned release packages.

PR #3791, opened 2018-03-27
main/libatomic_ops: upgrade to 7.6.4

prspkt:


main/libatomic_ops/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
 pkgname=libatomic_ops
-pkgver=7.6.2
+pkgver=7.6.4
 pkgrel=0
-pkgdesc="Provides semi-portable access to hardware provided atomic memory operations"
+pkgdesc="Semi-portable access to hardware provided atomic memory operations"
 arch="all"
 url="https://github.com/ivmai/libatomic_ops"
-license="GPL-2.0"
+license="GPL-2.0-or-later"
```
TBK: To my understanding the license is restricted to 2.0 - https://github.com/ivmai/libatomic_ops/blob/master/COPYING

prspkt: Based on the license headers included, i think the GPL parts fall under the *-or-later variant.
See [here](https://github.com/ivmai/libatomic_ops/blob/master/src/atomic_ops_malloc.c#L6).

PR #3797, opened 2018-03-28
main/json-c: upgrade to 0.13.1

nmeum:
This upgrade includes an soname bump.


nmeum:
Looks like travis isn't able to download the freeradius tarball currently.

jirutka:
Please rebase it against the current master (should fix freeradius build).

nmeum:
Rebased, but the freeradius tarball still cannot be fetched.

andypost:
@nmeum http://freeradius.org/releases/ tells about mirror on github - it fails because ftp used which is not supported by travis

PR #3800, opened 2018-03-28
community/farbfeld: add missing dependency (file)

Mark-Weston:
/usr/bin/2ff shell script depends on file command in line 18
file command, albeit part of POSIX, is not present in busybox


nmeum:
You are right, I missed this because I already have file installed. Thanks for supplying a patch.

community/farbfeld/APKBUILD
```diff
@@ -7,7 +7,7 @@ pkgdesc="Lossless image format which is easy to parse, pipe and compress"
 url="https://tools.suckless.org/farbfeld/"
 arch="all"
 license="ISC"
-depends=""
+depends="file"
```
nmeum: Just a note in case you want to submit more patches in the future: You need to bump the value of the `pkgrel` variable, otherwise changes like this one won't have an effect. I fixed this for you when merging this change.

PR #3801, opened 2018-03-28
testing/sxhkd: upgrade to 0.5.9, clarify license

prspkt:


nmeum:
Thanks, do you actually use sxhkd? If so: You might want to claim maintainership and move it to community.

prspkt:
I don't use it but i can certainly maintain it. I will assume [bspwm](https://github.com/alpinelinux/aports/pull/3802) as well.

PR #3805, opened 2018-03-28
main/fuse: modernize; move /etc/udev/rules.d to /lib

sroracle:


nmeum:
Thanks for patch, will definitely merge it. But please split this into two commits:

1. One commit for modernizing the APKBUILD
2. One commit for changing the path

sroracle:
Changes reorganized as requested.

PR #3834, opened 2018-03-30
community/mbedtls: upgrade to 2.11.0

prspkt:


community/mbedtls/APKBUILD
```diff
@@ -23,10 +23,9 @@ prepare() {
 	default_prepare
 
 	# Enable flags for non-embedded systems.
-	sed -i \
+	sed -i include/mbedtls/config.h \
 		-e 's|//\(#define MBEDTLS_THREADING_C\)|\1|' \
-		-e 's|//\(#define MBEDTLS_THREADING_PTHREAD\)|\1|' \
-		"$builddir"/include/mbedtls/config.h
+		-e 's|//\(#define MBEDTLS_THREADING_PTHREAD\)|\1|'
```
jirutka: Why you keep changing order of arguments for `sed`? o.O

prspkt: Sorry, it looks a bit prettier to me -)
Will revert.

jirutka: I agree that it looks better, but I’m afraid that it may break compatibility with other sed implementations. Actually I wonder how it even works, because `-i` accepts optional argument, but not the input file path, but suffix for backup file. Input file(s) should be at the end of command line:

Busybox sed:
```
sed [-i[SFX]] [-nrE] [-f FILE]... [-e CMD]... [FILE]...
```

GNU sed:
```
gsed [OPTION]... {script-only-if-no-other-script} [input-file]...
```

fabled:
Recompile is needed for `community/julia`, `testing/hiawatha`, and potentially `testing/td-agent-bit`. Could you add these?

prspkt:
`community/julia` fails testing on both mbedtls current (2.6.1) and 2.8.0.
`testing/td-agent-bit` fails building therefore i cannot include.

jirutka:
> `community/julia` fails testing on both mbedtls current (2.6.1) and 2.8.0.

No, julia tests currently does not fail, so check should not be disabled.

PR #3839, opened 2018-03-30
testing/ffcall: upgrade to 2.1

prspkt:


testing/ffcall/APKBUILD
```diff
@@ -36,10 +36,7 @@ check() {
 
 package() {
 	cd "$builddir"
-	mkdir -p "$pkgdir"/usr/share/man \
```
andypost: why this line removed?

prspkt: `make install` already creates the required directory structure `/usr/{lib,include,share}` in the destination directory. The line above simply creates the same dir again.

PR #3841, opened 2018-03-30
main/ninja: Move gtest to checkdepends

neuschaefer:
There is a rather long cyclic dependency chain:

```
cmake
`-> makedepends jsoncpp-dev
    `-> makedepends meson, which depends on ninja
        `-> makedepends gtest
	    `-> makedepends cmake
```

Fortunately, this cycle can easily be broken by declaring ninja's dependency on
gtest as checkdepends instead of makedepends.


tmpfile:
Speaking of ninja there's a new implementation out there that might be worth checking out.
https://github.com/michaelforney/samurai implements the ninja build language with a focus on simplicity, speed, and portability.
It's used in the https://github.com/michaelforney/oasis distribution and works as a drop-in replacement there.

neuschaefer:
3k lines (samurai) instead of 15k lines (ninja). Neat!

TBK:
I have created a package for samurai - https://github.com/alpinelinux/aports/pull/3895

neuschaefer:
This patch is somewhat obsoleted by commit 27a7784306 ("main/cmake: use bundled jsoncpp to avoid circular dependency")

neuschaefer:
Thanks, @jirutka!

PR #3842, opened 2018-03-30
testing/wireguard-tools: Fix examples installation path

neuschaefer:
WireGuard's examples directory should be installed to
/usr/share/doc/wireguard-tools/examples, not /usr/share/doc/examples.
It seems that this was intended in the APKBUILD, but the code that copies the
examples directory relies on $_name, which is not set. Use $pkgname instead.


neuschaefer:
FWIW, the build failure looks like an error on Travis-CI's part.

PR #3844, opened 2018-03-30
community/php7: security upgrade to 7.1.16 (3.6)

andypost:
Ref http://php.net/ChangeLog-7.php#7.1.16


andypost:
Commited as https://git.alpinelinux.org/cgit/aports/commit/?id=3b0037e5

PR #3848, opened 2018-03-31
testing/bc: new aport

gavinhoward:
https://github.com/gavinhoward/bc
POSIX bc with GNU extensions

### DIFF
--- a/testing/bc/APKBUILD
+++ b/testing/bc/APKBUILD
@@ -1,0 +1,34 @@
+# Contributor: Gavin D. Howard <yzena.tech@gmail.com>
+# Maintainer: Gavin D. Howard <yzena.tech@gmail.com>
+pkgname=bc
+pkgver=0.5
+pkgrel=0
+pkgdesc="POSIX bc with GNU extensions"
+url="https://github.com/gavinhoward/bc"
+arch="all"
+license="0BSD"
+depends=""
+makedepends=""
+install=""
+subpackages=""
+source="https://github.com/gavinhoward/bc/releases/download/$pkgver/$pkgname-$pkgver.tar.xz"
+builddir="$srcdir/"$pkgname-$pkgver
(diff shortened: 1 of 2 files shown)


rgdoliveira:
Alpine already has a package called [bc](https://pkgs.alpinelinux.org/package/edge/main/x86_64/bc), maybe you can rename yours? Guess @jirutka and @ncopa can have a better idea.

andypost:
This way will lead to kinda MacOS has `sed` & `gnu-sed`

jirutka:
Can it be used as a drop-in replacement for [GNU bc](https://www.gnu.org/software/bc/bc.html)? So we can rename the current `bc` package (GNU bc) to `gnu-bc` and package this one as `bc`?

/cc @kaniini, @ncopa

gavinhoward:
Yes, it is a drop-in replacement. It does not have fancy math algorithms, so it's slower, but it has the same exact output. It can even run the timeconst.bc script in the Linux kernel build.

gavinhoward:
I am also open to renaming this package.

jirutka:
> …this is inconsistent with other GNU software we package, e.g. GNU sed isn't packaged as gnu-sed either.

That’s just because we don’t have any other `sed` implementation except Busybox `sed`, isn’t it? I’d personally prefer non-GNU as default where possible.

gavinhoward:
I could rename this package now, and we can move the problem of figuring out which bc should be the default to later.

How should I name this package?

ncopa:
I would not mind rename gnu bc to `gnu-bc` and use a BSD licensed `bc` as the default.

> @gavinhoward How should I name this package?

Maybe call it `gavinhoward-bc` or `howard-bc` for now?

gavinhoward:
Package has been renamed to `howard-bc`.

PR #3862, opened 2018-04-01
Upgrade s6 & related packages

andypost:
Order of build
- skalibs
- s6
- s6-rc
- s6-linux-utils
- s6-linux-init


andypost:
reordered dependencies in a way they should be build

PR #3881, opened 2018-04-02
main/sqlite: upgrade to 3.23.0, clarify license

prspkt:


andypost:
Probably this update should bump some packages which has sqlite statically build-in

PR #3884, opened 2018-04-02
main/scons: upgrade to 3.0.1

prspkt:


ncopa:
I like this. have you checked that /usr/bin/scons has `#!/usr/bin/python3` shebang and not just `python`?

prspkt:
@ncopa Excellent! We now correctly invoke python3.

ncopa:
I had to revert this because serf doesnt build with scons with python3.

PR #3886, opened 2018-04-03
main/openrc: upgrade to 0.36

jirutka:
There’s one important change:

https://github.com/OpenRC/openrc/blob/master/NEWS.md:

> In this version, the cgroups mounting logic has been moved from the sysfs service to the cgroups service. This was done so cgroups can be mounted inside an lxc/lxd container without using the other parts of the sysfs service.
>
> ?As a result of this change, if you are upgrading, you need to add cgroups to your sysinit runlevel by running the following command as root:
>
>     # rc-update add cgroups sysinit

I’m not sure what does it affect. Where is default set of services populated?


ncopa:
I think this is something that is sort of expected to be mounted, like `/sys` itself. It seems like openrc itself can set limits on services it manages, so openrc itself may need it.

Default services are set up on first boot in initramfs-init: https://git.alpinelinux.org/cgit/mkinitfs/tree/initramfs-init.in#n506 and then copied over with setup-disk.

One way to do it would be to have a pre-install/update script that checks if sysfs service is started and if it is, create a symlink in /etc/runlevels/sysinit.

I don't think many things will break if its not missing, but it is difficult to know for sure.

ncopa:
looks like `community/cgmanager/cgmanager.initd` also needs cgroups.

clandmeter:
Are we going to do this for 3.8? 
Looks like we are many releases behind.

jirutka:
> It seems like openrc itself can set limits on services it manages, so openrc itself may need it.

Cgroups support is optionally, OpenRC can work even without it.

> Are we going to do this for 3.8?

I’d like to, ’cause we’re many releases behind.

> ps, why is only the subject of this PR 0.36?

??

clandmeter:
@jirutka your patches are for openrc 0.35.5 while the subject says 0.36 which is the latest version.

jirutka:
> your patches are for openrc 0.35.5 while the subject says 0.36 which is the latest version.

Huh, I have no idea why I did that. It’s most likely a mistake.

jirutka:
Aha, 0.36 has not been released yet. But still I don’t know why I wrote 0.36 into the PR’s title.

clandmeter:
hmm, i think because its listed in https://github.com/OpenRC/openrc/blob/master/NEWS.md

PR #3887, opened 2018-04-03
Add pcl aport

chambbj:


testing/pcl/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+pkgname=pcl
+pkgver=1.8.1
+pkgrel=0
+pkgdesc="Point Cloud Library (PCL)"
+url="https://github.com/PointCloudLibrary/pcl"
+arch="all"
+license="BSD"
+options="!check"
+depends=""
+makedepends="cmake eigen-dev boost-dev flann-dev"
```
BernhardGruen: It seems this package needs `flann-dev` which is not yet in aports.

testing/pcl/APKBUILD
```diff
@@ -0,0 +1,47 @@
+# Contributor: Bradley J Chambers <brad.chambers@gmail.com>
+# Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+pkgname=pcl
+pkgver=1.8.1
+pkgrel=0
+pkgdesc="Point Cloud Library (PCL)"
+url="https://github.com/PointCloudLibrary/pcl"
+arch="all"
+license="BSD"
```
BernhardGruen: please see https://spdx.org/licenses/

testing/pcl/APKBUILD
clandmeter: To make travis happy you can combine commits into a single PR.

chambbj: @clandmeter only problem is that the pcl build takes quite a while and will timeout if combined. This is how I originally had it structured.

chambbj:
@clandmeter As discussed on IRC today, tests have been enabled (with some exceptions, see comments). This does build with all enabled tests passing locally (running `abuild rootbld`).

PR #3889, opened 2018-04-03
main/sdl2: fix DirectFB renderer regression

michitux:
The update to 2.0.8 broke the build of the DirectFB renderer such that
it cannot be used anymore, see
https://bugzilla.libsdl.org/show_bug.cgi?id=4123. This fixes the
regression by re-inserting the affected line in the configure script.

Having a working DirectFB renderer in SDL2 is crucial for postmarketOS
as it is used for unlocking full disk encryption and displaying the
charging animation on almost all devices, see also postmarketOS/osk-sdl#80.

### DIFF
--- a/main/sdl2/APKBUILD
+++ b/main/sdl2/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: August Klein <amatcoder@gmail.com>
 pkgname=sdl2
 pkgver=2.0.8
-pkgrel=0
+pkgrel=1
 pkgdesc="A development library designed to provide low level access to audio, keyboard, mouse, joystick and graphics"
 url="http://www.libsdl.org"
 arch="all"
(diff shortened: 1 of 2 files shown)


ollieparanoid:
@ncopa, @kaniini, @jirutka: Would it be possible to fast track this one, as it fixes upstream breakage that causes major issues in pmOS?

If it is not possible, just tell us and we will carry the aport in our own aports (that's not the best solution as we need to sync it with Alpine's version, which is why we're asking here first).

ollieparanoid:
Okay, seems like you guys are busy. Then we'll carry our own version and find a good solution in the long run for this (so this can be closed in my opinion).

jirutka:
Sorry for delay, I’ve just merged it.

ollieparanoid:
Thanks!

PR #2908, opened 2017-12-08
main/openssh: add check()

andypost:

PR #3890, opened 2018-04-03
main/openssh: upgrade to 7.7_p1

andypost:
Ref http://lists.mindrot.org/pipermail/openssh-unix-dev/2018-April/036789.html


andypost:
Tests still fail https://github.com/alpinelinux/aports/pull/2908

PR #3891, opened 2018-04-03
main/alsa-lib: upgrade to 1.1.6

prspkt:


fabled:
Looks otherwise OK, but I am not sure if we want to change to a mirror service. Is this to workaround Travis limitations?
@ncopa Any thoughts?

prspkt:
@fabled yes, the mirror url is due to travis failing to connect. instead of completely removing the original (ftp) url we could keep it commented out.

fabled:
I dunno how to properly solve this. I think at least for the time being we'd like to keep the upstream FTP or upstream HTTP(s) if available. Using random mirrors is not the preferred solution. This might make Travis fail, but our builders will manage it. Could you revert the to the original URL for time being? Thanks.

PR #3892, opened 2018-04-04
community/aircrack-ng: update apkbuild, clarify license

TBK:


community/aircrack-ng/APKBUILD
```diff
@@ -22,25 +23,21 @@ prepare() {
...
-	make check
-}
-
 build() {
 	cd "$builddir"
 	./configure \
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
-		sqlite=true experimental=true \
```
jirutka: Why did you remove sqlite=true?

jirutka: Okay, then document it in the commit message.

TBK: @jirutka Done.

PR #3895, opened 2018-04-04
testing/samurai: new aport

TBK:


neuschaefer:
Could this have something like `replaces="ninja"` to indicate that you can install samurai instead of ninja? But then this package would actually have to provide a symlink `/usr/bin/ninja -> samu`.

Maybe the symlink could be in a separate package? Or dynamically create as with busybox's symlinks?

The point is: I think it would be cool if samurai could be installed instead of ninja, but existing and future build dependencies should IMHO still list ninja, because that's how drop-in replacements work.

TBK:
@neuschaefer I have incorporated your suggestions.

testing/samurai/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: TBK <alpine@jjtc.eu>
+# Maintainer: TBK <alpine@jjtc.eu>
+pkgname=samurai
+pkgver=0.3
+pkgrel=0
+pkgdesc="Ninja-compatible build tool written in C99 with focus on simplicity, speed and portability."
+url="https://github.com/michaelforney/samurai"
+arch="all"
+license="GPL-2.0"
```
neuschaefer: The [LICENSE](https://github.com/michaelforney/samurai/blob/master/LICENSE) file lists a few BSD/MIT-like licenses (I haven't checked exactly) and an Apache licence 2.0, but no GPL.

testing/samurai/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+
+build() {
+	cd "$builddir"
+	make
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" PREFIX=/usr install
+	ln -s $pkgdir/usr/bin/samu $pkgdir/usr/bin/ninja
```
neuschaefer: Does this work? It looks like this would create a link with an absolute target of `$pkgdir/usr/bin/samu`, when either `/usr/bin/samu` (absolute link without the pkgdir prefix) or a relative link to `ninja` would be more appropriate.

neuschaefer:
Thanks for you work!

TBK:
@neuschaefer Thanks for the corrections 👍 I rushed it a bit, it has been fixed now.

testing/samurai/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+# Maintainer: TBK <alpine@jjtc.eu>
+pkgname=samurai
+pkgver=0.3
+pkgrel=0
+pkgdesc="Ninja-compatible build tool written in C99 with focus on simplicity, speed and portability."
+url="https://github.com/michaelforney/samurai"
+arch="all"
+license="Apache-2.0 MIT"
+options="!check" # No test suite
+provides="ninja"
```
clandmeter: You probably don't want this as it will add a second package to aports that provides ninja and i bet that the original ninja package has a higher version number. When apk finds multiple provides it will choose the one with the highest version number. Instead change it to replaces so you are able to install this instead of original ninja.

TBK: Thanks for the info. I have changed it to `replaces="ninja"`.

PR #3900, opened 2018-04-04
main/xen: upgrade to 4.7.5

HRio:
While at it, drop unused patches and patch files

rombios-no-pie.patch: not needed due to upstream:
 b704b1a09b rombios: prevent building with PIC/PIE


HRio:
@rnalrd XEN-4.7.5 for Alpine v3.5

PR #3903, opened 2018-04-04
Revert "testing/wireguard-tools: depends on bash"

zx2c4:
This reverts commit 548c24ea1e19b3739a24556248ebbde4f300e14a, which was
snuck in while I was on vacation.

There are users of Alpine+WireGuard who don't need or use
wg-quick and thus don't need or use bash. This therefore should not be a
required dependency of the package. This also breaks the pkgrel
consistency with the kernel packages. So, we revert.

Signed-off-by: Jason A. Donenfeld <Jason@zx2c4.com>
Fixes: #3640

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,12 +3,11 @@
 
 pkgname=wireguard-tools
 pkgver=0.0.20180304
-pkgrel=2
+pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
 url='https://www.wireguard.com'
 license="GPL-2.0"
-depends="bash"
 makedepends="libmnl-dev"
 subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch"
 options="!check"


neuschaefer:
Regarding `pkgrel` consistency: AFAICS, the `wireguard-{hardened,vanilla}` packages have a way to deal with pkgrel bumps the -tools package, the `_toolsrel` variable, but I see a few issues:

- in wireguard-vanilla it seems to be broken since commit 39b040c682 ("[kmodule]-vanilla: clean up, streamline, modernize") by @xentec, resulting in `install_if = wireguard-tools=0.0.20180304-r linux-vanilla=4.14.32-r0` in the .PKGINFO file for wireguard-vanilla
- the requirement to update `_toolsrel` isn't documented in wireguard-tools/APKBUILD, which makes it easy to forget about it

zx2c4:
Probably best to first commit this revert and then fix the other stuff in a followup

zx2c4:
No, extracting wg-quick into a subpackage isn't okay with me. If downstreams are going to be ripping apart the tools, then as upstream I'll probably be incentivised to make that much harder for downstreams to do. Ripping it apart is going to result in all sorts of confusing bug reports that we have to deal with upstream. Not up for it. Don't do that.

I'd appreciate it if you could merge this commit, thereby putting things back the way they were before, and then if you want, you can open a new issue to bikeshed about whatever, and maybe that time the actual maintainers of the package can chime in on it, instead of sliding this through underneath the door as before.

ncopa:
fixed with d2466648e56e5512ef0bdc87216d6b212ed45906 as requested.

But I agree with @jirutka here. This is still "ripping apart the tools" but the only difference is that error message is more misleading. If bash is missing you will still get a `file not found` error. User will think that `wg-quick` is missing while it is interpreter and will get confused when the `/usr/bin/wg-quick` is actually there. Moreover if user coincidentally/temporary has other tool using bash `wg-quick` will work as long that tool is installed. When that other tool is removed `wg-quick` will suddenly break. User may not even know that bash was ever installed.

I think we have 2 options here:
- make wireguard-tools depend on bash
- split out the tools that needs bash to a separate package

zx2c4:
Thanks for merging this. Feel free to open another PR choosing one of
the other strategies, and this time we can actually discuss it before
merging.

PR #3912, opened 2018-04-05
main/pgpool: openssl & memcached support

tcely:


jirutka:
I don’t agree with adding dependency for memcached into pgpool. I guess that most users don’t use it and they should not be forced to install libmemcached-libs and all its dependencies.

```
>>> pgpool*: Tracing dependencies...
	so:libc.musl-x86_64.so.1
	so:libcrypto.so.42
	so:libmemcached.so.11
	so:libpq.so.5
	so:libssl.so.44
```

If it can’t be extracted into a separate so library linked by pgpool, then the other option is to build two variants of pgpool (see e.g. dnsmasq for example).

ncopa:
can you please merge the commits into a single commit? optionally have the libressl support in separate commit.

ncopa:
oh, it seems like it also needs to be rebased

jirutka:
I don't want this to be merged as it now. I'll comment it later.

J0WI:
I added OpenSSL without memcached in #6660.

PR #3913, opened 2018-04-05
main/unbound: remove ldns-dev

tcely:


tcely:
@fabled it hasn't been used since 1.4.22

> separate ldns into core ldns inside ldns/ subdirectory. No more configure --with-ldns is needed and unbound does not rely on libldns.

PR #3917, opened 2018-04-05
community/pytest: fix dependencies

TBK:


community/pytest/APKBUILD
```diff
@@ -35,12 +35,12 @@ package() {
...
 _py2() {
 	replaces="$pkgname"
-	depends="py2-py"
+	depends="py2-py py2-six py2-attrs py2-setuptools py2-more-itertools py2-pluggy"
 	_py python2
 }
 
 _py3() {
-	depends="py3-py"
+	depends="py3-py py3-six py3-attrs py3-setuptools py3-more-itertools py3-pluggy"
```
jirutka: `setuptools` is built-in in python3, are you sure that `py3-setuptools` is needed (i.e. it needs newer version than provided by python3)?

PR #3927, opened 2018-04-05
community/iotop: upgrade to python3

prspkt:
`0001-Fix-build-error-with-Python-3-caused-by-itervalues-i.patch` is sourced from [upstream](http://repo.or.cz/w/iotop.git/commit/99c8d7cedce81f17b851954d94bfa73787300599) and enables building with python3.


BernhardGruen:
One question:
why do you remove the (dirty hack) patch that moves iotop to `sbin` instead of `bin`?

fabled:
Merged in 930dd957fc4c3895ad00712eb58c11fa6f743def and bd5a7ed62e67690455769710c3069e75beb44ecd by @ncopa. Thanks.

PR #3932, opened 2018-04-06
community/cargo: upgrade to 0.26.0

tcely:


tcely:
@ncopa Please let me know if this looks good to you for the libressl 2.7 changes.

community/cargo/APKBUILD
```diff
@@ -1,77 +1,50 @@
...
-	scp "$tarname.tar.gz" dev.alpinelinux.org:/archive/$pkgname/
+_exit_handler() {
+	abuild-apk del --wait 30 --purge ".bootstrap-$pkgname"
 }
 
-prepare() {
-	mv "$srcdir"/rust-openssl-openssl-v$_crate_openssl_ver "$builddir"/openssl
-	default_prepare
-}
+trap _exit_handler EXIT
```
jirutka: This definitely should not be in APKBUILD.

jirutka:
Sorry, but this is completely wrong. Moreover, the plan is to merge cargo into the rust package, to follow upstream’s intention and obstacles (missing Cargo.lock in cargo).

tcely:
@jirutka Just a heads up. When you do get rid of `cargo` and `rust` no longer depends on `cargo` it will no longer depend on `rust`, which I believe will prevent `rust` from being installed.

jirutka:
I’ve upgraded cargo in f5a1d6dda0910e46cfa06bf159712cf4a6f91ecd and kept it as a separate package for now (just b/c I don’t wanna waste any more time with it).

tcely:
@jirutka So after calling this "completely wrong" and not expanding on why, your changes are substantially similar and you're stopping at `0.25.0` now?

jirutka:
They are far from “substantially similar”, please read the diff more properly. I picked 0.25.0 to align it with the current rust in aports (0.24.0).

I’m sorry for not expanding “why”, I’m quite busy and Rust packaging already cost me too much time and nerves.

PR #3947, opened 2018-04-06
main/alpine: upgrade to 2.21.99, modernize

TBK:


fabled:
We generally prefer releases over git snapshots. Apparently upstream deleted all releases, so I've moved alpine to unmaintained for now (see 8b0d7dd4e894d2fafd4ff8bf46cadf7e40134717). I wonder if there's any fork that would do proper releases. If so, could we use one like that?

TBK:
I will try to send the dev an email.

PR #3960, opened 2018-04-07
main/libmicrohttpd: upgrade to 0.9.59 and add check

rgdoliveira:


fabled:
Works for me too. It seems to bind to loopback address (127.0.0.1); maybe this is not setup in travis?

PR #3967, opened 2018-04-08
testing/meek: new aport

kpcyrd:
https://trac.torproject.org/projects/tor/wiki/doc/meek
A pluggable transport proxy written in Go


testing/meek/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+	export GOPATH="$srcdir"
+	go build -v -o meek-client/meek-client ./meek-client/...
+	go build -v -o meek-server/meek-server ./meek-server/...
+}
+
+package() {
+	cd "$builddir"
+	install -Dm 755 meek-client/meek-client "$pkgdir/usr/bin/meek-client"
+	install -Dm 755 meek-server/meek-server "$pkgdir/usr/bin/meek-server"
+	setcap 'cap_net_bind_service=+ep' "$pkgdir/usr/bin/meek-server"
```
BernhardGruen: Don't really know if this is results in a change in the final package. You may have to set this in a post-install/post-upgrade function.

testing/meek/APKBUILD
```diff
@@ -0,0 +1,52 @@
+# Contributor: kpcyrd <git@rxv.cc>
+# Maintainer: kpcyrd <git@rxv.cc>
+pkgname=meek
+pkgver=0.30
+pkgrel=0
+pkgdesc="A pluggable transport proxy written in Go"
+url="https://trac.torproject.org/projects/tor/wiki/doc/meek"
+arch="all"
+license="CC0 1.0 Universal"
```
BernhardGruen: License format does not conform to https://spdx.org/licenses/
I think it should be "CC0-1.0".

testing/meek/APKBUILD
kpcyrd: Those are preserved by the package. There are a few packages in aports that use setcap, all of them set it in APKBUILD instead of a post-install/post-upgrade.

testing/meek/APKBUILD
kpcyrd: Thanks, done

testing/meek/APKBUILD
clandmeter: You need to depend on libcap to be able to do this.

testing/meek/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+builddir="$srcdir/$pkgname"
+
+prepare() {
+	mkdir -p ${builddir%/*}
+	mv "$srcdir/$pkgname-$pkgver" "$builddir/"
+	cd "$builddir"
+
+	export GOPATH="$srcdir"
+	(cd meek-client && go get -v -d)
+	(cd meek-server && go get -v -d)
```
clandmeter: We dont want to run go get in aports as we it doesn't version the deps and fails to create reproducible builds. Instead you should use something like glide to generate a versioned dependency list (lock/yaml file) which is included in the aport.

testing/meek/APKBUILD
kpcyrd: Done

testing/meek/APKBUILD
kpcyrd: Do you have an example in aports? I'm not meek upstream, glide works by vendoring all dependencies into the repo.

clandmeter:
glide example: https://github.com/alpinelinux/aports/tree/master/community/easypki

run glide_init on each new release to create a new lock file which holds possible new deps and lock them.

kpcyrd:
@clandmeter thanks, I've introduced glide along with some other things from easypki, this makes the build a lot more pleasant.

kpcyrd:
@clandmeter I've split the package into two packages so the server and the client can be installed individually, could you please verify this change didn't introduce any new issues? Everything addressed in this issue should be resolved by now.

kpcyrd:
@clandmeter it seems there are no leftover files because the build only uses `$startdir/src` which happens to be `$srcdir`. Changing this to `GOPATH="$srcdir"` causes problems for glide, apparently, because the source is expected to be located somewhere inside `$GOPATH/src`. I think I can solve this, but it would increase the complexity of `prepare()`. Should I do it regardless?

kpcyrd:
@clandmeter It seems I ran into https://github.com/golang/go/issues/14566, I have a working patch that would set `GOPATH="$srcdir"`. If this works for you I would go ahead and apply it (needs a `abuild glide_init` to rebuild the glide files):
```
diff --git a/testing/meek/APKBUILD b/testing/meek/APKBUILD
index 33ecc6dabb..6b29a8524c 100644
--- a/testing/meek/APKBUILD
+++ b/testing/meek/APKBUILD
@@ -19,10 +19,12 @@ source="$pkgname-$pkgver.tar.gz::https://gitweb.torproject.org/pluggable-transpo
 	glide.lock
 	glide.yaml
 	"
+builddir="$srcdir/src/$pkgname-$pkgver"
 
 prepare() {
-	export GOPATH="$startdir"
+	export GOPATH="$srcdir"
+	mkdir -p "$srcdir/src"
+	mv "$srcdir/$pkgname-$pkgver" "$builddir"
 	cd "$builddir"
 	cp "$srcdir"/glide.yaml "$srcdir"/glide.lock .
 	glide install --skip-test
@@ -30,7 +32,7 @@ prepare() {
 }
 
 build() {
-	export GOPATH="$startdir"
+	export GOPATH="$srcdir"
 	cd "$builddir"
 	go build -v -ldflags "-s -w" \
 		-o meek-client/meek-client ./meek-client/...
@@ -70,7 +72,7 @@ server() {
 glide_init() {
 	abuild clean deps unpack prepare
 	cd "$builddir"
-	export GOPATH="$startdir"
+	export GOPATH="$srcdir"
 	rm -f glide.yaml glide.lock
 	glide init --non-interactive
 	glide update
```

clandmeter:
check https://github.com/golang/go/wiki/GOPATH#directory-layout

and i think i have changed it to $srcdir like here:

https://github.com/alpinelinux/aports/blob/master/community/gitea/APKBUILD

kpcyrd:
@clandmeter I've copied some things from that apkbuild and go is now using `$srcdir` as `$GOPATH`.

kpcyrd:
oh well, I've noticed I forgot to remove the binaries from the main `meek` package. I've started to prepare a followup PR that removes the binaries from the `meek` package since they belong into the `meek-client` and `meek-server` package, but that seems to be harder than expected:

Since the main package doesn't contain any binaries, this warning appears:
```
>>> WARNING: meek*: No arch specific binaries found so arch should probably be set to "noarch"
```
Which makes sense, so I set `arch="noarch"` for `meek` and `arch="all"` in `client()` and `server()`, this causes this error though:
```
>>> ERROR: meek-client*: Split function set arch="all" for meek-client, use subpackages=pkg:split:arch format instead
```
I'm having trouble reading that error, can somebody provide a pointer?

Sorry for the hassle.

PR #3976, opened 2018-04-09
main/alpine-mirrors: add tuna and aliyun mirror

wenerme:
All mirrors is very slow or unable to access from China, aliyun and tuna mian a very good mirror of alpine, this will easy the pain to install alpine from China.

### DIFF
--- a/main/alpine-mirrors/mirrors.yaml
+++ b/main/alpine-mirrors/mirrors.yaml
@@ -141,3 +141,15 @@
   - https://mirror.neostrada.nl/alpine/
   - ftp://mirror.neostrada.nl/alpine/
   - rsync://mirror.neostrada.nl/alpine/
+- name: Tsinghua Open Source Mirror
+  location: China, Beijing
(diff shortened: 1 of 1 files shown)


wenerme:
should i increase pkgrel or version ?

tcely:
@wenerme The `git` history suggests updating `version` is what has been done in the past. When the question is answered by a project member, please would you add a comment in the `APKBUILD` file answering this question?

clandmeter:
We are planning to use a different source for mirrors so please hold this.
Are you the sysop for this mirror?

wenerme:
No, I'm not the sysop of these mirrors, but I use these mirror all the time.

clandmeter:
@wenerme From next release v3.8 and edge we will start to use https://mirrors.alpinelinux.org/mirrors.txt as source. For mirrors to be added a sysop needs to send us an email to alpine-mirrors@alpinelinux.org with mirror details.

We need to have a way to contact with them to be able to add them, so if you want please ask them to do so.

PR #3977, opened 2018-04-09
unmaintained/mongo-php7-driver: remove

TBK:
unmaintained/mongo-php7-driver is an older version of testing/php7-mongodb and can therefore be removed.


andypost:
It looks like a duplicate not older one

jirutka:
unmaintained abuilds are not built at all, so its removal cannot cause any breakage...

andypost:
We moved it to unmaintained and that broke build of php package

PR #3978, opened 2018-04-09
main/musl: fix complex math functions

awilfox:


fabled:
Applied with followup commit adding additional cherry picked fixes.

PR #3982, opened 2018-04-10
testing/usbip-utils: move to testing, upgrade to 4.16

wenerme:


HRio:
@fabled and @wenerme should this not have been 4.14 as that is the linux version in edge?

wenerme:
@HRio I think the user space tools is not tied to kernels

PR #3996, opened 2018-04-11
testing/php7-protobuf: new aport

TBK:


andypost:
Looks good to go! Interesting how many arches it supports

PR #4006, opened 2018-04-13
main/freetds: upgrade to 1.00.87

prspkt:


andypost:
next upgrade https://github.com/alpinelinux/aports/pull/5526

PR #4009, opened 2018-04-13
scripts/mkimg: use -rational-rock to avoid uid pollution

radhus:
ISO image should have files owned by root, not leaking the builder uid.


radhus:
@ncopa is this something that can be merged or do you want any tests in your build environment? I have other PR(s) that will build on this later.

PR #4026, opened 2018-04-14
main/xinit: fix default program execution

nmeum:
The exec(1) man page states the following:

> If command is specified, exec shall not return to the shell; rather
> the exit status of the process shall be the exit status of the program
> implementing command, which overlaid the shell.

Thus the shell will not continue execution after `exec startxfce4`
failed. To fix this we now check if the program exists and if that's the
case execute it using exec.


main/xinit/xinitrc
```diff
@@ -45,11 +45,8 @@ if [ -d /etc/X11/xinit/xinitrc.d ] ; then
...
-# Fall back to fluxbox
-exec startfluxbox
-
-# Start only terminal as last resort
-exec xfce4-terminal || exec mrxvt || exec urxvt || exec st
+for p in startxfce4 startfluxbox xfce4-terminal \
+		mrxvt urxvt st xterm; do
+	command -v "$p" >/dev/null && exec "$p"
+done
+unset p
```
nmeum: This is never reached unless now suitable program to execute was found. So we could remove this line.

main/xinit/xinitrc
```diff
@@ -45,11 +45,8 @@ if [ -d /etc/X11/xinit/xinitrc.d ] ; then
...
-# If not present, try the system default
-exec startxfce4
-
-# Fall back to fluxbox
-exec startfluxbox
-
-# Start only terminal as last resort
-exec xfce4-terminal || exec mrxvt || exec urxvt || exec st
+for p in startxfce4 startfluxbox xfce4-terminal \
+		mrxvt urxvt st xterm; do
```
nmeum: I also added xterm to the list of programs.

main/xinit/xinitrc
ncopa: Yeah, we can remove the `unset p`.

nmeum: Done

ncopa:
checksum  error

PR #4027, opened 2018-04-14
main/aspell-fr: modernize

TBK:


main/aspell-fr/APKBUILD
```diff
@@ -1,28 +1,28 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=aspell-fr
-pkgver=0.50_p3
-_ver=${pkgver/_p/-}
```
ncopa: im not sure i like this change.

main/aspell-fr/APKBUILD
```diff
@@ -1,28 +1,28 @@
...
+pkgver=0.50
 pkgrel=1
+_lang=${pkgname##*-}
+_name=${pkgname%%-*}-$_lang-$pkgver-3
```
ncopa: is the last `-3` a fixed number or does it have a meaning (eg like patch level?) if its patch level then we should use `pkgver=0.50_p3`

ncopa:
i think the same question above applies to `aspell-ru`, `aspell-uk`, `aspell-de` and `aspell-en` too.

PR #4042, opened 2018-04-15
community/youtube-dl: upgrade to 2018.04.16, improve abuild

prspkt:


community/youtube-dl/APKBUILD
```diff
@@ -33,7 +33,6 @@ package() {
 
 zshcomp() {
 	pkgdesc="Zsh completions for $pkgname"
-	depends=""
```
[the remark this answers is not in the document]
prspkt: @clandmeter because they are empty. Should they be restored?

community/youtube-dl/APKBUILD
```diff
@@ -20,7 +20,7 @@ source="https://github.com/rg3/$pkgname/releases/download/$pkgver/$pkgname-$pkgv
 builddir="$srcdir/$pkgname"
 
 prepare() {
-	cd "$builddir"
+	default_prepare
 	sed -i setup.py \
```
clandmeter: Yes we need default_prepare, but im not sure I like having to depend on a function to change directory.
Instead I would use full path to setup.py or change directory to it.

prspkt: All set.

PR #4047, opened 2018-04-16
testing/liquibase: upgrade to 3.6.1

CurtTilmes:
Also changed arch to noarch

### DIFF
--- a/testing/liquibase/APKBUILD
+++ b/testing/liquibase/APKBUILD
@@ -1,18 +1,23 @@
 # Contributor: Curt Tilmes <Curt.Tilmes@nasa.gov>
 # Maintainer: Curt Tilmes <Curt.Tilmes@nasa.gov>
 pkgname=liquibase
-pkgver=3.5.3
+pkgver=3.6.1
 pkgrel=0
 pkgdesc="Source Control for your Database"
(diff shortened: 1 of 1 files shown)


testing/liquibase/APKBUILD
```diff
@@ -1,19 +1,24 @@
 # Contributor: Curt Tilmes <Curt.Tilmes@nasa.gov>
 # Maintainer: Curt Tilmes <Curt.Tilmes@nasa.gov>
 pkgname=liquibase
-pkgver=3.5.3
+pkgver=3.6.1
 pkgrel=0
 pkgdesc="Source Control for your Database"
 url="http://www.liquibase.org/"
-arch="all"
+arch="noarch"
```
andypost: Is this really "noarch"? https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#arch
Because looks it download binary files

CurtTilmes: I haven't tested on other arches, but my understanding is that it is java byte-code that is executed by the VM on any architecture the same way.

andypost:
Commited already https://github.com/alpinelinux/aports/commit/d7be1948dd82f80bffe049114b89059c06d2a15b

PR #4049, opened 2018-04-16
 testing/botan: upgrade to 2.6.0

tcely:


testing/botan/add-armv8l-alias.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/src/build-data/arch/arm64.txt b/src/build-data/arch/arm64.txt
+index d4781c8..79535dd 100644
+--- a/src/build-data/arch/arm64.txt
++++ b/src/build-data/arch/arm64.txt
+@@ -6,6 +6,7 @@ family arm
+ <aliases>
+ aarch64
+ armv8
++armv8l
```
clandmeter: did you submit this upstream?

tcely: No, I haven't. Does adding that alias fix that build?

I'm also not sure if this should target arm 6 or 8. Thoughts?

clandmeter: Nice, that seems to work.
We are using a aarch64 machine which has 32bit support so the processor shows armv8l. This means you need to add armv8l to the arm32.txt so the build tools will correctly detect we are building for arm32 instead of arm64.

tcely: I'm not sure patching for that would be a good plan. Adding the CPU flag to specifically select arm_32 would be my preference.

clandmeter: No thats just a workaround. Upstream should fix this and add armv8l to arm32.txt as it is arm32 after all.

clandmeter:
@tcely thanks for helping upstream sort this out!

tcely:
@clandmeter You're welcome.

PR #4050, opened 2018-04-16
community/hugo: upgrade to 0.39

tboerger:
Updated to latest version, requires the `dep` package within `community` which have been moved my @clandmeter.

### DIFF
--- a/community/hugo/APKBUILD
+++ b/community/hugo/APKBUILD
@@ -1,40 +1,41 @@
 # Contributor: Thomas Boerger <thomas@webhippie.de>
 # Maintainer: Thomas Boerger <thomas@webhippie.de>
 pkgname=hugo
-pkgver=0.30.2
+pkgver=0.39
 pkgrel=0
 pkgdesc="A Fast and Flexible Static Site Generator built with love in GoLang"
 url="http://gohugo.io/"
 arch="all"
 license="Apache-2.0"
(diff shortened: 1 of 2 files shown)


tboerger:
I have dropped all `return 1` statements.

tboerger:
After removing `|| return 1` the Travis build fails :(

community/hugo/APKBUILD
```diff
@@ -1,40 +1,40 @@
...
+	dep ensure -v
+	go build -v -o bin/$pkgname
 }
 
 check() {
 	rm -f $builddir/releaser/git_test.go
 	cd "$builddir"
 	export GOPATH="$srcdir"
-	govendor test +local || return 1
+	go test
```
clandmeter: >>> hugo*: check
?       github.com/gohugoio/hugo        [no test files]

Could this be the issue? can you check what going on here?

tboerger: Ah yeah, it doesn't find tests in the root package. But this exits with 0.

tboerger: I have enabled the tests recursively now.

tboerger:
Tests are patched to work now, but Travis still fails for a different reason.

andypost:
@tboerger please rebase commit on top of current master to fix travis build

PR #4051, opened 2018-04-16
testing/wireguard: upgrade to 0.0.20180413

zx2c4:
```
  * wg-quick.8: fix typo
  * wg-quick: hide errors on save
  
  This fixes a small regression in the resolvconf save handling on Debian.
  
  * compat: stable kernels are now receiving b87b619
  * compat: silence warning on frankenkernels
  * compat: support OpenSUSE 15
  
  Usual set of fixes for weird kernels.
  
  * curve25519: use precomp implementation instead of sandy2x
  * curve25519: use cmov instead of xor for cswap
  * curve25519: memzero in batches
  * curve25519: precomp const correctness
  
  Rather than using sandy2x, which requires use of the vector registers and simd
  instructions (and therefore thermal throttling and register save/restores), we
  instead use BMI2 and ADX instructions to achieve better performance, using:
    - https://eprint.iacr.org/2017/264
    - https://github.com/armfazh/rfc7748_precomputed
  
  * curve25519: add self tests from wycheproof
  * chacha20poly1305: add self tests from wycheproof
  
  Wycheproof now provides sneaky test vectors, so we've imported them into our
  self-tests to mitigate regressions. More info can be found at:
    - https://github.com/google/wycheproof
```

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -8,8 +8,8 @@ _kpkgrel=0
 
 # when changing _ver we *must* bump _mypkgrel
 # we must also match up _toolsrel with wireguard-tools
-_ver=0.0.20180304
-_mypkgrel=2
+_ver=0.0.20180413
+_mypkgrel=1
 _toolsrel=0
 _name=wireguard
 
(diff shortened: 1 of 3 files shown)


ncopa:
fixed with 7bb43d1ccbd55c0dc19f323256d8cc9c088f9b8b

PR #4055, opened 2018-04-17
testing/meek: fix split package

kpcyrd:
This resolves a dumb mistake from my initial PR, `meek` was supposed to be a meta package, but it still contained both the meek-server and meek-client binary. I've split the package differently, `meek` contains `meek-client`; which is the binary most users need, while `meek-server` is only needed when a user wants to setup their own bridge (and is therefore split into it's own package).

Sorry for the inconvenience.


clandmeter:
Ah i didn't catch that. Did you think about adding a init.d script for the server?

clandmeter:
Ah nice. Thanks!

PR #4061, opened 2018-04-18
main/alpine-conf: add installation support on DASD on s390x

tmh1999:


tmh1999:
This patch is not ready. I'll update soon.

tmh1999:
I will update this commit after https://github.com/alpinelinux/alpine-conf/pull/14 is merged.

PR #4064, opened 2018-04-18
main/debootstrap: upgrade to 1.0.97, clarify license

prspkt:


main/debootstrap/APKBUILD
```diff
@@ -1,24 +1,24 @@
...
 arch="noarch"
-license="GPL-2.0"
+license="MIT"
 options="!check" # upstream doesn't have a test suite
 depends="debian-archive-keyring dpkg tar"
 makedepends=""
 subpackages="$pkgname-doc"
 source="http://ftp.debian.org/debian/pool/main/d/${pkgname}/${pkgname}_${pkgver}.tar.gz"
-builddir="$srcdir"/$pkgname-$pkgver
+builddir="$srcdir/$pkgname-$pkgver"
```
clandmeter: Can you explain why you are making these changes?
If it doesn't make much sense please don't clobber patches with it.

main/debootstrap/APKBUILD
```diff
@@ -1,24 +1,24 @@
...
 
 package() {
 	cd "$builddir"
 	make DESTDIR="$pkgdir" install
-	install -Dm644 $pkgname.8 \
+	install -D -m644 $pkgname.8 \
```
clandmeter: Can you explain why you are making these changes?
If it doesn't make much sense please don't clobber patches with it.

main/debootstrap/APKBUILD
prspkt: @clandmeter Sure.
I extended the double quote to include all variables in case of empty spaces.
Reverted

main/debootstrap/APKBUILD
prspkt: @clandmeter Yes. It looks a bit cleaner but probably redundant.
Reverted.

clandmeter: The thing is, if we don't set a specific format for these kind of things everyone will apply their own preference making diffs larger than needed.

I know its nitpicking so I'm sorry for that :)

main/debootstrap/APKBUILD
clandmeter: While you are right regarding spaces both pkgname and pkgver can never include spaces so this should never happen.

main/debootstrap/APKBUILD
prspkt: Nothing to worry about :) Since there's no specific format your feedback is very much appreciated.

PR #4068, opened 2018-04-18
testing/botan: enable armhf (unblocks build-edge-armhf!)

ollieparanoid:
`testing/kea` depends on `botan-dev`, which is not available for armhf.
This is currently blocking the `build-edge-armhf` bot.

http://build.alpinelinux.org/buildlogs/build-edge-armhf/testing/kea/kea-1.3.0-r0.log says:
```
>>> kea: abuild 3.2.0_rc1-r0
>>> kea: Checking sanity of /home/buildozer/aports/testing/kea/APKBUILD...
>>> kea: Analyzing dependencies...
  .makedepends-kea-0:
    masked in: cache
    satisfies: world[.makedepends-kea]
  botan-dev (missing):
    required by:
>>> ERROR: kea: all failed
>>> kea: Uninstalling dependencies...
```

I have built the `botan` package for `armhf` with qemu and it runs through.

Note: `build-edge-armhf` has been stuck on various packages since the libressl soname bump last Friday, `uboot-tools` for example is not built against the new libressl yet.


tcely:
See https://github.com/alpinelinux/aports/pull/4049

ollieparanoid:
Thanks for fixing it!

Different topic, but it's odd that the build [failed](https://travis-ci.org/alpinelinux/aports/builds/368389714#L1367) with the following, although it compiled successfully on Travis:
> fatal: No names found, cannot describe anything.

I'm not sure why this happens in the first place (Travis is building inside the checked out aports git repo, right?), and why it does [not always](https://travis-ci.org/alpinelinux/aports/builds/368005081#L721) cause the build to fail.

But I think if we modified [this](https://github.com/alpinelinux/abuild/blob/be30cda326f7a4cd6da2077e52297e898de826e3/abuild.in#L1551) back to...
```shell
DESCRIPTION="$repo $(cd $startdir && $git describe || true)"
```
...so the `|| true` is in place like before my commit https://github.com/alpinelinux/abuild/commit/4501a6cc5eb5542c5172ae391ae54377b440e1bf, then at least the builds would not fail there. This is probably not the right fix, but this information could help with creating a proper fix. Another way to avoid this is setting a description with `abuild -D` in the Travis script.

CC: @ncopa, @kaniini, @jirutka

(reporting this here, because somehow I'm not in #alpine-devel anymore in matrix right now, need to investigate that.)

PR #4069, opened 2018-04-19
testing/fbreader: new aport; testing/libunibreak: new aport

duncanguthrie:
testing/fbreader: new aport
https://www.fbreader.org
E-book reader that supports many e-book formats

Fbreader depends on libunibreak:
testing/libunibreak: new aport
https://fbreader.org/content/fbreader-beta-linux-desktop
Implementation of the unicode line breaking algorithm


andypost:
please rebase it to make as 2 commits - each one for separate package

PR #4072, opened 2018-04-19
dnscrypt-proxy - 2.0.10

ianbashford:
move setcap to APKBUILD
minor mods to init script


tanertas:
If you plan to take-over package, could you consider the other changes that I suggested here?

http://lists.alpinelinux.org/alpine-aports/5347.html

I was used setcap at post-install script since I wasn't aware of that apk is capable of keep file system extended attributes.

ianbashford:
Thanks for the comments @tanertas -  I had actually used post-install for the set-cap previously; it was a comment from @jirutka on my original PR who pointed out to me that I should move it to the apk.  I'm happy either way -- I'll take a look at your other comments too.  Thanks for taking the time to review...

tanertas:
I tried to notice that putting setcap in APKBUILD was right thing to do. I think I wasn't clear enough.

andypost:
Please change commit message to- community/dnscrypt-proxy: upgrade to 2.0.10

andypost:
There some guide about commit messages http://wiki.alpinelinux.org/wiki/Creating_patches

ianbashford:
@tanertas  Thanks for taking the time to make all of your comments and provide the feedback so clearly.  I've been through them all and I've had this running for a couple of days now on two clean test installs.  Same @andypost - hopefully the messages are more standard now.  
@fcolista happy to help out supporting this as I'm a user of it - let me know anything else you see as improvements.
All - happy to look at any other guidance you have...

PR #4077, opened 2018-04-20
testing/pcl: disable failing segmentation test on ppc64le and aarch64

mksully22:
Skip failing segementation test until it can be investigated. Opened issue on pcl project:
https://github.com/PointCloudLibrary/pcl/issues/2288

### DIFF
--- a/testing/pcl/APKBUILD
+++ b/testing/pcl/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
 pkgname=pcl
 pkgver=1.8.1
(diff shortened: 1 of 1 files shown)


mksully22:
The pcl build fails on ppc64le and aarch64 because of one of the tests. I submitted PR https://github.com/alpinelinux/aports/pull/4077 to disable the test on these architectures for now. Because of the length of time it takes to build, Travis CI is timing out and err'ing out. Could you please try the patch manually on your setup and then accept the PR if you agree with it?
chambbj, can you evaluate this PR manually on your setup and then accept it if you agree with it?

PR #4079, opened 2018-04-21
community/ffmpeg: upgrade to 4.1

andypost:
Ref http://ffmpeg.org/index.html#pr4.0

### DIFF
--- a/community/ffmpeg/APKBUILD
+++ b/community/ffmpeg/APKBUILD
@@ -3,7 +3,7 @@
 # Contributor: Jakub Skrzypnik <j.skrzypnik@openmailbox.org>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=ffmpeg
-pkgver=4.0.2
+pkgver=4.1
 pkgrel=0
 pkgdesc="Complete and free Internet live audio and video broadcasting solution for Linux/Unix"
 url="http://ffmpeg.org/"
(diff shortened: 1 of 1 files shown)


andypost:
Not sure about usefulness of second commit

community/ffmpeg/APKBUILD
```diff
@@ -50,7 +50,8 @@ build() {
 		--prefix=/usr \
 		--enable-avresample \
 		--enable-avfilter \
-		--enable-gnutls \
+		--enable-libtls \
+		--enable-nonfree \
```
ncopa: why is `--enable-nonfree` needed? can you mention this in ocmmit message?

ncopa:
i had to upgrade to 3.4.4 first to fix security vulnerabilities. Can you please rebase? And remove the `--enable-nonfree` or explain in commit message why its needed.

Also, lets not build it against libtls for now.

ncopa:
I cannot merge this with out alos rebuilding all the packages needing a rebuild. This is a major project, and I cannot merge it like this.

andypost:
@ncopa I think better to remove `--enable-nonfree` https://www.ffmpeg.org/doxygen/4.0/md_LICENSE.html tells about AAC & NVENC 
https://www.ffmpeg.org/legal.html also suggests to exclude this option

community/ffmpeg/APKBUILD
andypost: removed and fixed

andypost:
Added sefixes (not sure makes sense to split into versions) from https://www.ffmpeg.org/security.html

andypost:
Rebased but now in logs
```
Hunk #1 succeeded at 187 (offset 10 lines).
libtls is incompatible with the gpl and --enable-nonfree is not specified.
```

community/ffmpeg/APKBUILD
andypost: Added back because otherwise configure fails with `libtls is incompatible with the gpl and --enable-nonfree is not specified.`

9001:
fdk_aac is the big advantage of adding `--enable-nonfree`, since the standard aac encoder is bad. However since we do not link fdk_aac currently I agree on removing `--enable-nonfree`

andypost:
so second commit is not needed - probably better to file issue about libressl support and codes relation

community/ffmpeg/APKBUILD
andypost: @ncopa I guess we could remove second commit now but libtls is in edge as well?

andypost:
@ncopa removed second commit and rebased on top of current master

PR #4080, opened 2018-04-21
testing/py-flask-peewee: upgrade to 3.0.0

prspkt:


ncopa:
@fabaff is this ok with you?

jirutka:
testing/py-flas-peewee is broken, missing many python dependencies (see https://github.com/coleifer/flask-peewee/blob/3.0.3/setup.py#L4).

PR #4085, opened 2018-04-22
community/caddy: upgrade to 0.10.14

TBK:


andypost:
According changelog https://github.com/mholt/caddy/releases 0.10.13 should be skipped but it fix security flaws

andypost:
@clandmeter probably it needs backport

PR #4088, opened 2018-04-23
testing/mame: new aport

tanertas:
http://mamedev.org
Multi Arcade Machine Emulator with GroovyMAME/Switchres/No-nag patchset.


tanertas:
About Travis CI build issue; it is not clear about error because of log size limitation. This is rather a big package and probably build environment resources exhausted.

tanertas:
Raw log of Travis build says:

```
The job has been terminated
```
It may require to increase log size limitation.

tanertas:
Now Travis build fails with:

`The job exceeded the maximum time limit for jobs, and has been terminated.`

PR #4089, opened 2018-04-23
3.7 stable: Can you please provide packet "lnav"

74cmonty:
Hi,
Can you please provide packet "lnav" (http://lnav.org/).
This program is already listed [here](https://wiki.alpinelinux.org/wiki/Alpine_security), but it's not available.

THX


jirutka:
This PR is non-sense. `lnav` has been moved to unmaintained many releases ago and we definitely cannot introduce it to already released stable branch.

PR #4098, opened 2018-04-23
community/girara: upgrade to 0.2.9

ganwell:
* Upstream switched to meson build


ganwell:
This is merged, but was not closed

PR #4100, opened 2018-04-23
scripts/mkimg*: allow building s390x standard image

tmh1999:
s390x standard media consist of a kernel, an initramfs, a parm file.
Users can tweak parameters for their needs in the example parm file.

### DIFF
--- a/scripts/mkimg.base.sh
+++ b/scripts/mkimg.base.sh
@@ -256,6 +256,10 @@ create_image_iso() {
 		grub-mkrescue --output ${ISO} ${DESTDIR} -follow-links \
 			-sysid LINUX \
 			-volid "alpine-$PROFILE $RELEASE $ARCH"
+	elif [ "$ARCH" = "s390x" ]; then
+		for _image in vmlinuz initramfs modloop; do
+			cp ${DESTDIR}/boot/$_image-$kernel_flavors ${OUTDIR}
+		done
 	else
 		xorrisofs \
 			-quiet \


scripts/mkimg.base.sh
```diff
@@ -245,6 +245,13 @@ create_image_iso() {
 		grub-mkrescue --output ${ISO} ${DESTDIR} -follow-links \
 			-sysid LINUX \
 			-volid "alpine-$PROFILE $RELEASE $ARCH"
+	elif [ "$ARCH" = "s390x" ]; then
+		cp ${DESTDIR}/boot/vmlinuz-$kernel_flavors ${OUTDIR}
+		cp ${DESTDIR}/boot/initramfs-$kernel_flavors ${OUTDIR}
+		# we ship an example of the parm file
+		echo -n "dasd=0.0.01b0,0.0.01b1 net_type=qeth_l2,0.0.0340,0.0.0341,0.0.0342 " > ${OUTDIR}/parmfile
+		echo -n "ip=192.168.0.2::192.168.0.1:255.255.255.0:none::none nameserver=8.8.8.8 " >> ${OUTDIR}/parmfile
+		echo -n "s390x_installer=yes password=alpine " >> ${OUTDIR}/parmfile
```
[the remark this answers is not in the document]
tmh1999: Updated. Thanks.

tmh1999:
@ncopa it would be nice to have this in release candidates.

scripts/mkimg.base.sh
```diff
@@ -256,6 +256,16 @@ create_image_iso() {
...
+	elif [ "$ARCH" = "s390x" ]; then
+		for _image in vmlinuz initramfs modloop; do
+			cp ${DESTDIR}/boot/$_image-$kernel_flavors ${OUTDIR}
+		done
+		# we ship an example of the parm file
+		printf %s "dasd=0.0.00a0,0.0.00a1 s390x_net=qeth_l2,0.0.00b0,0.0.00b1,0.0.00b2 " > ${OUTDIR}/parmfile
+		printf %s "ip=192.168.0.2::192.168.0.1:255.255.255.0:none:eth0:none:8.8.8.8 " >> ${OUTDIR}/parmfile
+		printf %s "alpine_repo=http://dl-cdn.alpinelinux.org/alpine/latest-stable/main " >> ${OUTDIR}/parmfile
+		printf %s "modloop=http://dl-cdn.alpinelinux.org/alpine/latest-stable/releases/s390x/modloop-$kernel_flavors " >> ${OUTDIR}/parmfile
+		printf %s "ssh_pass=secret " >> ${OUTDIR}/parmfile
```
ncopa: what and where is the parmfile used?

Do we need to install boot loader on the .iso image?

tmh1999: The parmfile is used as the configuration file to load into the z/VM hypervisor, along with the kernel and initramfs.

The .iso image is not used for the z/VM hypervisor.

The .iso image is used for another booting scenario (on LPAR) but it requires physical access to the server, which we can't do it right now as no one is sponsoring that.

tmh1999: If you don't want to include the parmfile example in the code, I would create and explain them in the wiki page then.

tmh1999:
@ncopa I updated the wiki page https://wiki.alpinelinux.org/wiki/S390x, so we might not need to put the parmfile content here (kind of ugly imho).

Addition to my previous comment, .iso image both doesn't work with KVM nor z/VM.

scripts/mkimg.base.sh
```diff
@@ -256,6 +256,10 @@ create_image_iso() {
 		grub-mkrescue --output ${ISO} ${DESTDIR} -follow-links \
 			-sysid LINUX \
 			-volid "alpine-$PROFILE $RELEASE $ARCH"
+	elif [ "$ARCH" = "s390x" ]; then
+		for _image in vmlinuz initramfs modloop; do
+			cp ${DESTDIR}/boot/$_image-$kernel_flavors ${OUTDIR}
+		done
```
ncopa: Oh, I see what you do here. You don't create iso image, only kernel. The idea is good but I think we should solve it differently, because:
- the kernel, initramfs and modloop are without version number, which means they will be overwritten next version.
- this is something we may want for netboot of the other arches as well

I did the 3.8.0 rc1 without this and it generated iso image for s390x. you can use `uniso` (from `alpine-conf` package) or `7z` to extract the files from the iso.

I also found out that suse has a perl script to make iso image bootable with zipl: https://github.com/openSUSE/mksusecd/blob/master/isozipl

tmh1999: Agreed. As we talked in IRC, I comment here again for completeness. The ISO might be useful for LPAR boot but we don't have the hardware atm, and we are targeting KVM and z/VM so kernel,initramfs, modloop are enough.
We can still do ISO for KVM thought if I'm not wrong but I don't have time to do it for now, might be 3.8.1.
I'm going to close this pr. Many thanks.

PR #4101, opened 2018-04-23
main/alpine-baselayout: bump pkgrel from the last commit

tmh1999:


jirutka:
It has been already bumped as part of another commit.

PR #4040, opened 2018-04-15
main/wayland: upgrade to 1.15.0

vakartel:
- new library: wayland-egl
- enable tests
- fix libs subpackages depends

PR #4107, opened 2018-04-24
main/mesa: Upgrade to 18.0.2

tanertas:
* Clarify license
* Fix typo related with enabling vulkan drivers
* Add subpackages for ati and intel vulkan drivers
* Enable vc4 on aarch64
* Enable vaapi drivers, include them into corresponding mesa-dri packages
* Re-arrange configure options for easy reading
* --disable-asm and --enable_glx_rts options used on x86 only.
  * x86 build fails during packaging if asm enabled (TEXTREL)
  * since mesa has its own method to detect cross-compile environment, asm should not be disabled globally.
* Remove -j1 from package()
* Modernize APKBUILD


ncopa:
the reference to #4040 in commit message needs to be done differently (we refer to issues https://bugs.alpinelinux.org with `#N`), but i can fix that.

ncopa:
hum
libva here introduced a circular dependency: mesa -> libva -> mesa

tanertas:
@ncopa Can you check out #4106 please? I think there's no need to create an additional package to solve circular dependency.

PR #4121, opened 2018-04-26
main/linux-vanilla: enable VESA framebuffer driver

strfry:
I have a Thinkpad x220 with Coreboot + SeaBIOS instead of the original vendor BIOS.
In this specific configuration, VESA support is necessary to have graphics (or text output) in early boot (before the KMS driver is initialized), since there is no complete VGA BIOS anymore handling the other legacy display methods.

CONFIG_FB_VESA is recommended as a default, and i couldn't any find commits that tell why it should be deactivated, so i hope it's safe to enable this for everyone.

### DIFF
--- a/main/linux-vanilla/APKBUILD
+++ b/main/linux-vanilla/APKBUILD
@@ -7,7 +7,7 @@ case $pkgver in
 	*.*.*)	_kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=0
+pkgrel=1
 pkgdesc="Linux vanilla kernel"
 url="http://kernel.org"
 depends="mkinitfs"


TBK:
Please squash and rename commit + PR: "main/linux-vanilla: enable VESA framebuffer driver"

strfry:
@TBK Ok, changed it.

ncopa:
Should this only be enabled for x86_64 and not 32 bit x86?

strfry:
I added the same options for x86 as well.

ncopa:
I'm not sure what happened but it seems like the patch only bump pkgrel now?

ncopa:
looks like it was fixed with 166813a2251238709cb2fab13e5da856df10a1e4

PR #4124, opened 2018-04-26
main/smartmontools: metadata fixes

awilfox:
* The package can do SATA and NVMe, not just ATA and SCSI drives.

* Specify GPL version.

* Split OpenRC files.

* Add HTTPS site.


jirutka:
Merged as f9c196f688f6320d4b2387b2a4d2a256a9e761fe, thanks!

PR #4125, opened 2018-04-26
community/quassel: upgrade to 0.12.5

starwarsfan:


starwarsfan:
Ah, missed the checksum update... :-/
I'll fix it asap

community/quassel/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Maintainer: Yves Schumann <yves@eisfair.org>
 
 pkgname=quassel
-pkgver=0.12.4
-pkgrel=3
+pkgver=0.12.5
+pkgrel=1
```
TBK: pkgrel should be set to 0 when version is bumped.

TBK: Please remove all the `|| return 1` from the code. It is legacy stuff.

TBK: Please update the license so it is in line with the SPDX format:
https://spdx.org/licenses/
https://github.com/quassel/quassel/blob/master/COPYING

TBK: Squash your commits and rename it and the PR to " community/quassel: upgrade to 0.12.5"

community/quassel/APKBUILD
```diff
@@ -1,8 +1,8 @@
...
 url="http://www.quassel-irc.org/"
```
TBK: https://quassel-irc.org/

community/quassel/APKBUILD
starwarsfan: Thx for the review and your comments. Everything is clear except the notes regarding the license. I have no idea what you mean with this or what I should change. Could you explain please?

TBK: Alpine Linux project has decided to follow the SPDX standard naming convention. The current license mentioned in this APKBUILD file does not follow the convention it just says GPL. SPDX requires the specific variant, so if I need the COPYING file correct it should be `license="GPL-2.0-only GPL-3.0-only"`

community/quassel/APKBUILD
TBK: What I meant here was to update the url and source url to the TLS variant (https://)

starwarsfan:
@TBK: Updated/changed everything as requested.

fcolista:
Superseeded by ciommit 1e108e867674ffe06cf723713a8e2f091fbe2948
Thanks for your contribution anyway!

.: Francesco

starwarsfan:
What!? Can't believe this. :-(

Sorry but to wipe out all other changes from an open PR with a simple oneliner as comment is a showstopper for me! It's of course not the biggest PR but if there was a discussion before including updates to fix all PR requirements, I expect that the PR at least is compared with another one on the same topic and not just thrown away with something like "Thx anyway".

fcolista:
@starwarsfan sorry, I went through the PR from the newer, I didn't see this older PR.
It's my mistake, apologize for that.

PR #4142, opened 2018-04-28
main/ansible: upgrade to 2.5.2

fabaff:


andypost:
Maybe better to keep packages for both py versions?

ncopa:
I agree that we only use python3 and only add python2 support if we get complains. It is a long-term goal to get rid of python2

PR #4145, opened 2018-04-28
testing/py-telegram-bot: upgrade to 10.0.2

fabaff:


droberin:
Hello, @fabaff .
Is there a chance that the dependencies are not really properly set for this package?
While installing py-telegram-bot no problem is found with py-futures but when trying to install py3-telegram-bot it doesn't find py3-futures (py3-**future** IS available but no py3-**futures**).

andypost:
It looks really confusing https://pkgs.alpinelinux.org/packages?name=*-future*&branch=edge&arch=x86_64

andypost:
btw https://code.google.com/archive/p/pythonfutures/ only py2 as it is so this dependency is fine

I filed upgrade to 11.1.0 https://github.com/alpinelinux/aports/pull/5185 (latest for today)

PR #4127, opened 2018-04-26
testing/octave: add support for image and audio

1138-4EB:

PR #4153, opened 2018-04-28
testing/octave: upgrade to 4.2.2

TBK:


1138-4EB:
@TBK, shall we merge the changes proposed in #4127 with these?

Related:

- https://bugs.alpinelinux.org/issues/7753
- https://bugs.alpinelinux.org/issues/5009

TBK:
Sure, I will add the changes.

Have you concidered becomming the maintainer for the package?

TBK:
@1138-4EB I have added the new makedependencies and tried to clean it up a bit (remove redundant). Could you take a look?

TBK:
https://github.com/alpinelinux/aports/pull/4133 needs to be merged before this will succeed.

testing/octave/APKBUILD
```diff
@@ -1,31 +1,30 @@
...
+makedepends="autoconf automake libtool bison curl-dev fftw-dev flex fltk-dev fontconfig-dev
+	freetype-dev gfortran ghostscript-dev glu-dev gnuplot graphicsmagick-dev hdf5-dev
+	icoutils lapack-dev lcms2-dev libltdl librsvg-dev libsm-dev libsndfile-dev pcre-dev
+	perl pkgconf portaudio-dev qt-dev readline-dev texinfo texlive-full zlib-dev"
+source="https://ftp.gnu.org/gnu/octave/$pkgname-$pkgver.tar.gz
 	less-docs.patch
 	libinterp-deps.patch
 	abs.patch
 	gnulib.patch"
-builddir="$srcdir/$pkgname-$pkgver"
```
1138-4EB: `$buildir` is used below (lines 26|24, -|46 and -|51). Shouldn't this be kept?

TBK: Since it is the default path then abuilder does not need it to be declared.

1138-4EB:
@TBK, I had a look at it, and added a comment. However, I did not test it (yet). I need to check how to install apks that I built myself (not form remote repos).

About becoming a maintainer, I'm lacking knowledge. I need to learn about how packaging in alpine works as a whole. The requirement to sign builds is strange to me.

TBK:
@1138-4EB If you have setup a build machine/vm/... then `cd ~/packages/<branch>/<arch>/ && apk add octave-4.2.2-r0.apk octave-doc-4.2.2-r0.apk `

To remove them again `apk del octave octave-doc`

These articles should help you get started:
https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package
https://wiki.alpinelinux.org/wiki/APKBUILD_Reference

You can always get help on IRC/Matrix and via the mailing lists, as well as here and on https://bugs.alpinelinux.org

TBK:
Hmm I think a script for installing the correct mesa-dri-* would be better.

As far as I know the mesa-dri-swrast is a software renderer used as a fallback when the correct driver for the dedicated card is not installed. So using the correct dri will give a lot better performance.

1138-4EB:
That makes sense. I am using it in a docker image, so swrast is the most straightforward to keep it portable. However, users installing alpine natively might use others.

With "a script" you mean something as `$pkgname.post-install` or something external to the package?

TBK:
Could be either. I have not found a good way to determin which dri should be installed so it might just be better with an wiki article or a message through $pkgname.post-install.

PR #4154, opened 2018-04-29
testing/lightdm: Upgrade to 1.26.0

tanertas:
* Clarify license
* Allow null passwords
* Improve APKBUILD, update patches

### DIFF
--- a/testing/lightdm/APKBUILD
+++ b/testing/lightdm/APKBUILD
@@ -15,7 +15,7 @@ makedepends="$depends_dev linux-pam-dev gtk+3.0-dev libxext-dev
 install="$pkgname.pre-install"
 pkgusers="lightdm"
 pkggroups="lightdm"
-subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang $pkgname-openrc"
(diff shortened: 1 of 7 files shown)


testing/lightdm/APKBUILD
```diff
@@ -16,69 +16,49 @@ install="$pkgname.pre-install"
 pkgusers="lightdm"
 pkggroups="lightdm"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
```
TBK: Please add `$pkgname-openrc` to the subpackages.

tanertas: Done.

testing/lightdm/APKBUILD
```diff
@@ -1,84 +1,63 @@
...
-	local i
-	cd "$_builddir"
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
+	cd "$builddir"
+	default_prepare
 }
```
TBK: Missed this before, since no special task is done in prepare() the whole function can be deleted and default_prepare will be executed by default :)

tanertas: Ok.

PR #4157, opened 2018-04-29
community/fuse-exfat: upgrade to 1.2.8

TBK:
These two packages are duplicates:
https://pkgs.alpinelinux.org/contents?branch=edge&name=fuse-exfat-utils&arch=aarch64&repo=community
https://pkgs.alpinelinux.org/contents?branch=edge&name=exfat-utils&arch=aarch64&repo=community

### DIFF
--- a/community/fuse-exfat/APKBUILD
+++ b/community/fuse-exfat/APKBUILD
@@ -2,12 +2,13 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=fuse-exfat
 _pkgreal=exfat
-pkgver=1.2.7
+pkgver=1.2.8
 pkgrel=0
 pkgdesc="Free exFAT file system implementation"
 url="https://github.com/relan/exfat"
 arch="all"
 license="GPL-2.0-or-later"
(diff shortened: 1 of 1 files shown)


TBK:
@awilfox Please review this.

awilfox:
They are the same codebase, but they are not duplicates; `mkfs.exfat` _et al_ do not require FUSE and as such they can be run on a system without FUSE.

Additionally, if someone were brave enough to discount software patents (or live in an area where they do not exist), [there's a kernel module](https://github.com/bkuhn/exfat) for exFAT which would mean FUSE would not be required to mount said, either.

All in all, I'd say the utils should probably be split off of the FUSE package, so that people who only need the `mkfs` or `fsck` do not need to pull in FUSE to use them.  But as my usage of exFAT has gone to virtually nil, I no longer really care either way.

TBK:
@awilfox Thanks for the explanation. I have updated the PR to leave it be.

TBK:
@awilfox I stumpled upon this https://github.com/cryptomilk/kernel-sdfat thought it might be of interest.

PR #4161, opened 2018-04-30
community/openblas: upgrade to 0.2.20, add s390x support

tmh1999:
2 patches based on commits:
7543e578a4c0fc1e23193ea10571d307f0076c28
63cfa32691680505e6b9daf0997755178ddd3144

### DIFF
--- a/community/openblas/APKBUILD
+++ b/community/openblas/APKBUILD
@@ -3,18 +3,20 @@
 # Maintainer: Isaac Dunham <ibid.ag@gmail.com>
 pkgname=openblas
 _pkgname=OpenBLAS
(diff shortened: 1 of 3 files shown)


TBK:
CI fails because:
```

The job has been terminated
```

tmh1999:
After building fine on s390x I assumed it should work on x86_64. Turned out it doesn't. This needs some more work.

tmh1999:
This commit works on x86_64 and s390x. The CI might/might not be correct.

tmh1999:
I'm going to submit PRs for openblas-0.3.0 and its reverse dependencies soon and close this PR.

PR #4176, opened 2018-05-01
main/transmission: upgrade to 2.94

prspkt:


prspkt:
@n3mur1t0r yes, it gets pulled in by `curl-dev`.

n3mur1t0r:
sure, that's fine, but what if tomorrow curl doesn't need it anymore? (extremely unlikely, but it's a good practice to specify all dependencies for each package, just in case something like this happens)

PR #4188, opened 2018-05-03
main/wpa_supplicant: Add a comment to the service configuration file

tanertas:
* NetworkManager can't communicate with wpa_supplicant if dbus option is missing


jirutka:
> NetworkManager can't communicate with wpa_supplicant if dbus option is missing

That’s not a reason for enabling dbus as default; many people use wpa_supplicant without NetworkManager and don’t use DBus at all.

The current default is fine, maybe just add a comment to the config that `-u` is required for running with NetworkManager.

tanertas:
By looking service script and behavior with NetworkManager, I thought that we missed something. Probably real cause is that do not disturb dbus with wireless related info, if dbus not needed.
```
{       
        need localmount
        use logger dbus
        after bootmisc modules
        before dns dhcpcd net
        keyword -shutdown
}
```

PR #4191, opened 2018-05-03
testing/bird: don't use git repo as version number

Fusl:
Makefile.in (used by `make`) contains a shell execution line to read the current git tags and uses that value in the resulting binary. However, since we're not building from a git repository but from a tar file, this results in `bird --version` reporting a version of `v3.7.0-3239-gc13b5c993b` instead of `2.0.2`. The added patch makes sure that bird uses the `BIRD_VERSION` from `sysdep/config.h`


Fusl:
@jirutka Resubmitted the change, please recheck.

PR #4196, opened 2018-05-05
community/pulseaudio: Upgrade to 11.1

strfry:
This PR also adds features, besides upgrading the pulseaudio version:
- pulseaudio-system runs an install script, to adds and configures a user that PA needs for --system mode
- Echo Cancellation support through webrtc_audio_processing library. This requires introducing a new package for said library.


strfry:
Obsoleted by recent upgrade to 12.2 landing in edge

PR #4199, opened 2018-05-05
gcc: fix download URL

ollieparanoid:
This commit switches `ftp://` to `http://`, as the former link does not work anymore. Looks like GNU finally disabled their FTP service, as [announced here](https://ftp.gnu.org/):
> If you maintain scripts used to access ftp.gnu.org over FTP,
we strongly encourage you to change them to use HTTPS instead.

I've changed it to `http://` to match the binutils package.

EDIT: This is the second version, the first version still had the FTP link for the `ecj-4.9.jar` file hosted on sourceware.org. But they seem to [have disabled their FTP port](https://travis-ci.org/alpinelinux/aports/builds/375361469#L637) as well, the HTTP link still works.

### DIFF
--- a/main/gcc/APKBUILD
+++ b/main/gcc/APKBUILD
@@ -143,8 +143,8 @@ if $LANG_ADA; then
 fi
 makedepends="$makedepends_build $makedepends_host"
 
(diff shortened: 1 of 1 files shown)


ollieparanoid:
Now it [fails on Travis](https://travis-ci.org/alpinelinux/aports/builds/375363607#L1276) to something unrelated to this PR:
> configure: error: GNAT is required to build ada

PR #4200, opened 2018-05-06
main/openssh: fix flavour split function, add kerberos

saschazelzer:
The support for different flavours of openssh-server packages
was broken, effectively only supporting a single additional
flavour. The changes now support multiple flavours as
subpackages and also use the '+' sign in the flavour name
as a separator for indicating multiple packages which need
to be enabled in the openssh build. E.g. the "kerberos5+pam"
flavour enables both kerberos and pam support in a single
subpackage named openssh-server-kerberos5+pam.

### DIFF
--- a/main/openssh/APKBUILD
+++ b/main/openssh/APKBUILD
@@ -4,19 +4,19 @@
 pkgname=openssh
 pkgver=7.7_p1
 _myver=${pkgver%_*}${pkgver#*_}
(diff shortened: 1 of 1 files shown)


saschazelzer:
Just found an issue with dependencies being added to flavours where they shouldn't. I will open a new PR with this fixed.

PR #4218, opened 2018-05-07
main/wget: security upgrade to 1.19.5

andypost:
http://lists.gnu.org/archive/html/info-gnu/2018-05/msg00007.html


main/wget/APKBUILD
```diff
@@ -17,6 +17,8 @@ source="ftp://ftp.gnu.org/gnu/$pkgname/$pkgname-$pkgver.tar.gz "
 builddir="$srcdir/$pkgname-$pkgver"
```
TBK: Change the source from ftp:// to https:// and CI should not fail.

PR #4220, opened 2018-05-08
main/ttf-ubuntu-font-family: Fix broken Ubuntu font URL

refacto:
Also, while here, modernize checksums


refacto:
Regarding the somewhat ugly URL: Canonical doesn't serve "pretty" URLs anymore. The URL that's used here is also used in [Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/media-fonts/ubuntu-font-family/ubuntu-font-family-0.83-r1.ebuild#n10) and [Arch.](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/ttf-ubuntu-font-family#n14)

PR #4223, opened 2018-05-08
testing/mbuffer: upgrade to 20180625

prspkt:


prspkt:
Testing succeeds locally but seems to fail on Travis.

ncopa:
hangs on my machine too, til i manually press enter. I think the test suite waits on stdin or something. It needs to be fixed before we can merge.

PR #4228, opened 2018-05-08
main/perl-mail-clamav: fix version comparison that causes build break…

mksully22:
… in Makefile.PL

### DIFF
--- a/main/perl-mail-clamav/APKBUILD
+++ b/main/perl-mail-clamav/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=perl-mail-clamav
 _realname=Mail-ClamAV
 pkgver=0.29
-pkgrel=13
+pkgrel=14
 pkgdesc="Perl extension for the clamav virus scanner"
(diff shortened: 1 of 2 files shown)


main/perl-mail-clamav/APKBUILD
```diff
@@ -14,13 +14,19 @@ makedepends="bzip2-dev clamav-dev zlib-dev
...
-source="http://search.cpan.org/CPAN/authors/id/C/CO/CONVERTER/$_realname-$pkgver.tar.gz"
+source="http://search.cpan.org/CPAN/authors/id/C/CO/CONVERTER/$_realname-$pkgver.tar.gz
+	0001.patch"
 
 _builddir="$srcdir/$_realname-$pkgver"
 
 prepare() {
 	cd "$_builddir"
 	sed -i -e 's%\(clamav-config --version\)%sh /usr/bin/\1%' Makefile.PL || return 1
+	for i in $source; do
```
rgdoliveira: You can call "default_prepare" at the begin of prepare() function so it will apply your patch. Don't need this for.

PR #4232, opened 2018-05-09
community/mdds: Fix a typo ("dimentional")

neuschaefer:


neuschaefer:
(cc @fabled)

nmeum:
Technically you need to bump the pkgrel for this since the description included in the apk package metadata is otherwise not updated. Besides it would be nice if you could split this into separate commis for mdds and mdds1.0

neuschaefer:
Ok, I'll rework it.

PR #4233, opened 2018-05-09
testing/linux-tools: Add gpio-tools subpackage

neuschaefer:
This new package contains the tools lsgpio, gpio-event-mon, and gpio-hammer
from tools/gpio/ in the kernel source code. They are mainly useful for the
interactive testing of GPIOs.

(cc @ncopa)


iggy:
The existing APKBUILD has been updated to version `4.18.13` which doesn't require the patch anymore (it was integrated upstream as of `4.16.0`). Can you fix this up and hopefully we can get it looked at.

PR #4234, opened 2018-05-09
main/yaboot: Move to unmaintained

neuschaefer:
yaboot does not build on Alpine anymore. Some of the problems are:

- arch="powerpc" is wrong; the correct value would be "ppc"
- The bare-metal yaboot binary is linked against libext2fs, which pulls in open
  from the libc

Move yaboot to the unmaintained directory to make it clear that this package
hasn't worked for a while.

(cc @kaniini)


kaniini:
we (neither the adelie ppc team or the IBM ppc team) don't intend to use yaboot on ppc/ppc64/ppc64le anyway, so it can just be deleted.  i will delete it now.

PR #4235, opened 2018-05-09
testing/wireguard*: _toolsrel fixes

neuschaefer:
- The first patch documents the requirement to keep `_toolsrel` in sync with wireguard-tools' `pkgrel`
- The second patch fixes a typo in wireguard-vanilla, where `toolsrel` was used instead of `_toolsrel`

(cc @itoffshore @zx2c4)

### DIFF
--- a/testing/wireguard-hardened/APKBUILD
+++ b/testing/wireguard-hardened/APKBUILD
@@ -10,7 +10,7 @@ _kpkgrel=0
 # we must also match up _toolsrel with wireguard-tools
 _ver=0.0.20180420
 _mypkgrel=2
-_toolsrel=0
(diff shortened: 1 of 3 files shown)


testing/wireguard-hardened/APKBUILD
```diff
@@ -10,7 +10,7 @@ _kpkgrel=0
 # we must also match up _toolsrel with wireguard-tools
 _ver=0.0.20180420
 _mypkgrel=2
```
zx2c4: This is called `_rel` in the vanilla one. Maybe you want to rename this too, to keep it consistent, while you're at it?

PR #4237, opened 2018-05-10
community/mpd: upgrade to 0.20.19

refacto:
also, as protocol wants it, set pkgrel to 0


refacto:
Messed this branch up, sorry.

PR #4255, opened 2018-05-11
community/ghc: upgrade to 8.4.2

mitchty:
There is only one package that depends on ghc right now, and that is
community/cabal.

Removed armhf support for now as there appear to be overall issues with
that architecture across libc's. Additionally, for 8.6 to build we will
need to update to at least 8.2. As such, dropping support for armhf
makes the most sense for right now. I'll continue working with upstream
to fix it.

Removed the haskell-* provides as that was originally a plan to follow
what arch was doing for their haskell setup. In lieu of that not
providing value removing them entirely.

The ghc test suite was updated to use python 3 from python 2.

Finally really late in the release candidate cycle a change to ld
detection necessitated the need to pass --disable-ld-override to the
configure of ghc.

Without that in place there appears to be some shenanigans going on
in exported symbols.

Ghc upstream ticket:
https://ghc.haskell.org/trac/ghc/ticket/13958

Offending commmit:
https://git.haskell.org/ghc.git/commitdiff/2785ef0e31a123400da950ffafebe6cb1ce3f4eb


mitchty:
Not sure how to get travis to get community apk's in place, the failure there is that ghc isn't available. Normally it would timeout trying to build in the past.

mitchty:
Working on updating cabal to 2.0.0 as well, idris to latest after that with 8.4.2 this morning.

jirutka:
> Not sure how to get travis to get community apk's in place, the failure there is that ghc isn't available. 

Take a look how I handled it for rust aport:

https://github.com/alpinelinux/aports/blob/88cb5720c4103520ebe57a9f58d4513314a79c00/community/rust/APKBUILD#L25-L26

https://github.com/alpinelinux/aports/blob/88cb5720c4103520ebe57a9f58d4513314a79c00/community/rust/APKBUILD#L41-L44

mitchty:
Gotcha, guessing this changed recently then. So would that mean I kick off a r7 release of ghc 8.0.1 first and then let that get built? Or does it mean extra work to get an updated ghc apk in place such that travis will at least attempt to build this new apk?

jirutka:
I’ve already fixed it, please rebase this PR against the current master.

mitchty:
@jirutka Rebased, thanks! Interesting dance you did to get that accomplished, well beyond my APK knowledge.

logankaser:
Maybe try adding `--silent` to all the make invocations. I did some research and it seems the only way to solve this is to reduce the size of the travis log either by redirecting as much as possible to a log file or by some other method such as `--silent`. For now I would try adding `--silent` to all the make calls you can and hope that is enough. Down the road some method of redirecting to a file would probably be more useful for debugging.

mitchty:
@logankaser that might kill that but if 8.2.1 and 8.0.1 are goo canary’s it will just get timed out at an hour. Ghc takes a pretty penny to compile, and then the test suite hits. With a skylake i7 and striped nvme drives at home it takes about 50 minutes to build.

jirutka:
> Try adding `--silent` to all the make invocations. 

That’s not needed, don’t worry about Travis.

community/ghc/APKBUILD
```diff
@@ -1,74 +1,54 @@
 # Maintainer: Mitch Tishmack <mitch.tishmack@gmail.com>
 pkgname=ghc
-pkgver=8.0.2
-pkgrel=7
+pkgver=8.4.2
+pkgrel=0
+# Normal non rc candidate
+urlprefix="$pkgver"
+pkgprefix="$pkgname-$pkgver"
```
jirutka: Please prefix non-standard global variables with underscore (e.g. `_urlprefix`).

jirutka:
> PS: Any chance this could still make it in time for 3.8?

@mitchty, should we merge it before v3.8 is released?

mitchty:
@jirutka I don't know if I'm the best person to answer that. The primary reason to get 8.4.2 up is partially that 8.0.2 is a bit outdated by now, and to get 8.6 compiling I'll need a recent ghc.

Ghc 8.6 also just froze recently for feature changes so I need to get back to figuring out the fun of getting the new build system working rather soon. Note if this does get merged, this and cabal both should get updated in tandem.

About the only problem I can see with updating ghc is idris probably won't compile. I've been working on getting that updated too but only so much time in the summer days. The problems there have just been library bounds annoyances however so probably not a hard fix.

ncopa:
I think we should try merge this before v3.8.

mitchty:
@ncopa @jirutka ok sounds good, I think i've got idris fixed/updated as well to work in ghc 8.4.2 and with cabal 2.2.0, just tying to figure out how to get cabal to freeze its constraints by overridding things like I can with cabal install ... --constraint. I might just give up on that and patch the .cabal file instead though.

I still need to actually try to use the built idris but compiling is the first 90% of effort.

mitchty:
I may have spoke too soon >.<

```
[ 13 of 114] Compiling Idris.Core.TT    ( src/Idris/Core/TT.hs, dist/build/Idris/Core/TT.o )

src/Idris/Core/TT.hs:120:10: error:
    • No instance for (GHC.Base.Semigroup FC)
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Monoid FC’
    |
120 | instance Monoid FC where
    |          ^^^^^^^^^
```

Time to look at what version of base I really need.

mitchty:
Ok got it all working, so should get the trifecta of ghc/cabal/idris in all the pull requests I have out. Just ping me if anything goes amiss, it all seems to build on 3.7.0 locally.

neongreen:
@mitchty Do you think it could be bumped to [GHC 8.4.3](https://ghc.haskell.org/trac/ghc/blog/ghc-8.4.3-released) without much pain? The reason I'm worried is that Stackage will have 8.4.3, not 8.4.2, and this incompatibility might lead to difficulties when dockerizing Stack-based projects.

mitchty:
@neongreen not hard at all, I'll do a quick run against that today and make sure cabal+idris are still kosher. I didn't think to look if another point release came out in the past month. So shenanigans on me.

mitchty:
Ok ghc 8.4.3 and cabal 2.2.0 are no problem, idris has some older bounds for base, i'll poke at that later tonight.

ncopa:
I think i just push this, upgrade to ghc-8.4.3 and cabal 2.2.0 so we get this included in v3.8 while you work on idris.

PR #3246, opened 2018-02-09
testing/openresty: new aport

mengskysama:
https://openresty.org/
A Fast and Scalable Web Platform by Extending NGINX with LuaJIT

PR #4264, opened 2018-05-12
testing/openresty: new aport

tcely:
https://openresty.org/
A Fast and Scalable Web Platform by Extending NGINX with LuaJIT

### DIFF
--- a/testing/openresty/APKBUILD
+++ b/testing/openresty/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: tcely <tcely@users.noreply.github.com>
 # Maintainer:
 pkgname=openresty
-pkgver=1.13.6.1
+pkgver=1.13.6.2
 pkgrel=0
 pkgdesc="Scalable Web Platform by Extending NGINX with Lua"
 url="https://openresty.org/"
@@ -63,4 +63,4 @@ package() {
 	make DESTDIR="$pkgdir" install
 }
 
-sha512sums="7f2cb1779bae1967adae9bdcf138e526c20eb2527269bff16d6545c40191e5e4cbd861a41fe8e8b3ddfe442dcf6c2c26992f6de05b591f44406e4e0be028edcb  openresty-1.13.6.1.tar.gz"
+sha512sums="c0584566c06e95b12c5827f7b3185e859cb79a8d4754257246644c19888bfabfcec3577953b1a5a9aa7c0c5627b3e45bd0befacea4e3c64d21f895869ee302d5  openresty-1.13.6.2.tar.gz"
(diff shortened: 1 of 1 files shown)


tcely:
Updated and squashed https://github.com/alpinelinux/aports/pull/3246

testing/openresty/APKBUILD
```diff
@@ -0,0 +1,68 @@
...
+license="BSD"
+options="!check"
+depends="!nginx"
+depends_dev="gd-dev geoip-dev libressl-dev libxslt-dev pcre-dev readline-dev zlib-dev"
+makedepends="$depends_dev linux-headers"
+install=""
+subpackages="$pkgname-dev"
+source="https://openresty.org/download/$pkgname-$pkgver.tar.gz"
+
+builddir="$srcdir/$pkgname-$pkgver"
```
TBK: Can be removed since it is the default path.

testing/openresty/APKBUILD
```diff
@@ -0,0 +1,68 @@
+# Contributor: tcely <tcely@users.noreply.github.com>
+# Maintainer:
+pkgname=openresty
+pkgver=1.13.6.1
+pkgrel=0
+pkgdesc="Scalable Web Platform by Extending NGINX with Lua"
+url="https://openresty.org/"
+arch="all"
+license="BSD"
+options="!check"
```
TBK: Please add a comment why check is disabled.

fabled:
We already ship `nginx-mod-http-lua` and various other openresty modules via the main nginx package's submodules. Does this have something new?

tcely:
@fabled The short answer is yes. This is a bundle that happens to include Nginx.

> OpenResty is a full-fledged web platform that integrates the standard Nginx core, LuaJIT, many carefully written Lua libraries. OpenResty is not an Nginx fork. It is just a software bundle.

You can learn more about what's included at: https://openresty.org/en/components.html

fabled:
We generally prefer not to ship same software multiple times in different bundles. Could the nginx modules be included in the main nginx build, and package the lua modules individually? We in fact already have `lua-resty-http` and `lua-resty-dns` packages in community. We can finally have a meta package for `openresty` that pulls in everything in the bundle. Yes, this is more work, but it'll reduce code duplication and build times.

tcely:
@fabled I broke out sub-packages for the modules the current `nginx` aport is pulling from `openresty` GitHub repositories.

I'm not sure if the current `nginx` build is suitable for use instead of the binary built with this bundle. I don't believe it's worth the time to figure it out as it could very well change in the next release.

My suggestion is to incorporate this aport and stop building `openresty` modules as a third-party addition to the `nginx` aport.

PR #4265, opened 2018-05-12
main/xorgproto: new aport

prspkt:
[Upstream](https://lists.x.org/archives/xorg-announce/2018-February/002835.html) package consolidating the various X11 protocol headers in one.

For `main/xorgproto: new aport` note the `--datadir=/usr/lib` option, which (just like the packages it replaces) installs the `pkgconfig` files under `/usr/lib/pkgconfig` instead of `/usr/share/pkgconfig`. 

Installing the `pkgconfig` files under `/usr/share/pkgconfig`, results in abuild emitting a warning about not being able to find any provider during the `Tracing dependencies` phase.


main/xorgproto/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+	xcmiscproto=$pkgver-r$pkgrel
+	xextproto=$pkgver-r$pkgrel
+	xf86bigfontproto=$pkgver-r$pkgrel
+	xf86dgaproto=$pkgver-r$pkgrel
+	xf86driproto=$pkgver-r$pkgrel
+	xf86miscproto=$pkgver-r$pkgrel
+	xf86vidmodeproto=$pkgver-r$pkgrel
+	xineramaproto=$pkgver-r$pkgrel
+	xproto=$pkgver-r$pkgrel
+	"
```
ncopa: This will cause problems because the build order calculation does not consider provides. We should clean up all the packages that has `*proto` in makedepends.

main/libx11/APKBUILD
```diff
@@ -1,14 +1,14 @@
...
-pkgrel=1
+pkgrel=2
 pkgdesc="X11 client-side library"
 url="http://xorg.freedesktop.org/"
 arch="all"
 license="custom:XFREE86"
 depends=
 subpackages="$pkgname-dev $pkgname-doc"
-depends_dev="libxcb-dev xextproto xf86bigfontproto-dev xtrans
+depends_dev="libxcb-dev xextproto xf86bigfontproto xtrans
```
ncopa: why not use the new `xorgproto`?

ncopa:
i think I'll just merge this and we can do the cleanup afterwards.

main/xorgproto/APKBUILD
prspkt: Yes. I will start the cleanup and submit a PR.

jirutka: @ncopa If you know that it will cause problems, then why did you merge it (before the fix is provided)?!

ncopa: Because it only causes problems build time, to detect the build order, so it will likely not be a problem unless you do a full rebuild, which we do for v3.9. Even if someone get bitten by it before that, it is trivial to fix when you are affected.

The fix affects many packages (~70) so I thought it will be a challenge to do everything in one go without introducing conflicts with other pull requests. So i though we can clean up as we go.

PR #4272, opened 2018-05-14
testing/gist: new aport

refacto:
https://github.com/jdowner/gist
gist command line interface

### DIFF
--- a/testing/gist/APKBUILD
+++ b/testing/gist/APKBUILD
@@ -1,0 +1,35 @@
+# Contributor: Sascha Paunovic <azarus@posteo.net>
+# Maintainer: Sascha Paunovic <azarus@posteo.net>
+pkgname="gist"
+_pkgname="python-$pkgname"
+pkgver=0.6.1
+pkgrel=0
+pkgdesc="gist command line interface"
+url="https://github.com/jdowner/gist"
+arch="noarch"
+license="MIT"
+depends="python3"
+makedepends="python3-dev"
+source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir/$_pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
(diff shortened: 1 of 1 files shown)


testing/gist/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+	mkdir tests # Make checks pass
+	python3 setup.py test
+}
+
+package() {
+	cd "$builddir"
+	python3 setup.py install --prefix=/usr --root="$pkgdir"
+
+	# remove unneeded files
+	rm $pkgdir/usr/share/$pkgname/LICENSE
```
tcely: You should quote `$pkgdir` otherwise this `rm` breaks when the absolute path contains a space.

testing/gist/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Sascha Paunovic <azarus@posteo.net>
+# Maintainer: Sascha Paunovic <azarus@posteo.net>
+pkgname="gist"
+_pkgname="python-gist"
```
tcely: Consider: `_pkgname="python-$pkgname"`

testing/gist/APKBUILD
refacto: Agreed.

testing/gist/APKBUILD
refacto: Thanks!

testing/gist/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+pkgname="gist"
+_pkgname="python-$pkgname"
+pkgver=0.6.1
+pkgrel=0
+pkgdesc="gist command line interface"
+url="https://github.com/jdowner/gist"
+arch="noarch"
+license="MIT"
+depends="python3"
+makedepends="python3-dev py-setuptools"
```
jirutka: `py-setuptools` is for python2, it should not be needed for python3.

maxice8:
Can't be merged, depends on unpackaged python-gnupg

PR #4274, opened 2018-05-14
non-free/adobe-flashplayer: upgrade to 29.0.0.171

refacto:


refacto:
This works with current testing/firefox and community/firefox-esr.

PR #4278, opened 2018-05-14
testing/wireguard-tools: Document pkgrel/_toolsrel requirement

neuschaefer:
This is the remaining part of #4235, just a small comment to help contributors who change the pkgrel variable in wireguard-tools.

cc @zx2c4


andypost:
@itoffshore is this comment enough? Maybe mention that packages should be upgraded same time?

PR #4263, opened 2018-05-11
main/feh: upgrade to 2.26.4

prspkt:

PR #4287, opened 2018-05-15
main/feh: upgrade to 2.26.2

refacto:


refacto:
whoops, #4263 already did this, but the 1.26.2 caught me off-guard

PR #4302, opened 2018-05-17
main/nfs-utils: Fix and enable rpc.svcgssd

jantatje:
Enabled svcgssd build flag, applied patch from gentoo to make it build and wrote another patch to fix segfaults, because musl returns -1 for sysconf(_SC_GETGR_R_SIZE_MAX), which was not handled at all, thus allocating a smaller buffer than needed.

This allows NFSv4 to use kerberos authentication and encryption.

Edit: Apparently the nfs-utils-svcgssd_undefined_reference.patch is from redhat, gentoo just uses it, too.

### DIFF
--- a/main/nfs-utils/APKBUILD
+++ b/main/nfs-utils/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=nfs-utils
 pkgver=2.3.1
-pkgrel=3
+pkgrel=4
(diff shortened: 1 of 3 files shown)


jantatje:
This fixes [bug 8470](https://bugs.alpinelinux.org/issues/8470)

HRio:
@clandmeter can we commit this?

HRio:
@ncopa seems you are assigned to 8470, have you seen this PR?

PR #4315, opened 2018-05-19
testing/wireguard: upgrade to 0.0.20180519

zx2c4:
```
  * chacha20poly1305: add mips32 implementation
  
  "The OpenWRT Commit" - this significantly speeds up performance on cheap
  plastic MIPS routers, and presumably the remaining MIPS32r2 super computers
  out there.
  
  * timers: reinitialize state on init
  * timers: round up instead of down in slack_time
  * timers: remove slack_time
  * timers: clear send_keepalive timer on sending handshake response
  * timers: no need to clear keepalive in persistent keepalive
  
  Andrew He and I have helped simplify the timers and remove some old warts,
  making the whole system a bit easier to analyze.
  
  * tools: fix errno propagation and messages
  
  Error messages are now more coherent.
  
  * wg-quick: use invoking shell in auto rooting
  
  Rather than letting sudo use bash from PATH, we now have it use whatever bash
  is currently executing the script.
  
  * device: remove allowedips before individual peers
  
  This avoids an O(n^2) traversal in favor of an O(n) one. Before systems with
  many peers would grind when deleting the interface.
  
  * dns-hatchet: update paths
  
  Our reorganizing of the wg-quick bash paths was not sync'd with this patch,
  resulting in some trivial problems for Fedora and OpenSUSE.
  
  * compat: backport for OpenSUSE 15
  
  Usual compat fixes.
  
  * wg-quick: add darwin implementation
  
  We released a Darwin implementation of wg-quick(8), to be used with the new
  wireguard-go snapshot.
  
  * wg-quick: darwin: ensure socket directory exists
  * wg-quick: darwin: remove v6 routes after shutdown
  * wg-quick: darwin: bash correctness
  * wg-quick: darwin: restore DNS on down
  * wg-quick: darwin: use bash from environment and require bash 4+
  * wg-quick: darwin: sometimes there are no network services
  * wg-quick: darwin: avoid routing loop if no default
  * wg-quick: darwin: networksetup does not like missing stdio
  * wg-quick: darwin: reorder functions
  * wg-quick: darwin: simpler inclusion check
  
  After a pretty intense first few days of the new macOS port, we've fixed a few
  bugs and improved functionality of wg-quick(8).
  
  * ncat-client-server: add wg-quick variant
  
  We now have client-quick.sh that does the same as client.sh except it builds a
  file for wg-quick(8), which can then be used in `wg-quick up demo`.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 
 pkgname=wireguard-tools
-pkgver=0.0.20180513
+pkgver=0.0.20180519
 pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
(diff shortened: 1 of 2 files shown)


andypost:
Commited http://dup.pw/aports/57554cf6

PR #4320, opened 2018-05-19
main/xfsprogs: upgrade to 4.16.1

refacto:
To fix build, remove assumption that "$pkgdir"{/lib,/usr/lib} are empty.
Add python3 dependency to xfsprogs-extra, as xfs_scrub_all needs it.

### DIFF
--- a/main/xfsprogs/APKBUILD
+++ b/main/xfsprogs/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=xfsprogs
-pkgver=4.14.0
+pkgver=4.16.1
 pkgrel=0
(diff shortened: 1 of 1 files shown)


refacto:
Please don't merge yet, have yet to test it extensively.

refacto:
Tested (with linux-virt kernel):
- [x] Creating and mounting a filesystem
- [x] Reading the installed xfsprogs-doc man pages
- [x] new xfs_scrub program tested (with own kernel, as linux-virt doesn't support it)

refacto:
Hrm. `xfs_scrub_all` needs python3, how should I handle that dependency?

refacto:
Note: xfs_scrub_all is a completely optional program, so the dependency on python3 might not be completely necessary?

ncopa:
python3 needs to be in makedepends as well, so we make sure that python3 is built before xfsprogs. i'm fixing that.

PR #4332, opened 2018-05-20
main/mariadb-connector-c: upgrade to 3.0.4

TBK:


jirutka:
I’m not sure if we need to rebuild the depending packages… [Binary compatibility report for the libmariadbclient.so object between 3.0.3 and 3.0.4](https://abi-laboratory.pro/tracker/compat_report/mariadb-connector-c/3.0.3/3.0.4/8480c/abi_compat_report.html). @ncopa?

ncopa:
I believe it is not necessary. its a header for plugins.

PR #4337, opened 2018-05-21
main/linux-rpi: enable vc4 for rpi

yangxuan8282:


ollieparanoid:
To provide some context: a few people are working on installing postmarketOS on the various raspberry pi versions. We're directly usign the `linux-rpi` package from Alpine for that. Right now, only framebuffer can be used and this PR makes it possible to use DRM as well.

We would also like to update the kernel to 4.14 (that's what `linux-vanilla` has right now) and possibly do other enhanchements on top of that (such as enabling aarch64 for rpi3), but this is blocked by this PR.

@yangxuan8282: could you please increase the pkgrel?

ncopa:
Aw! I missed this! Sorry.

it should be fixed with ff478c74b2ec904371253c4977a16fe0643ec4d8 which uses the `bcmrpi_defconfig`, `bcmrpi3_defconfig` and `bcm2709_defconfig` as base.

PR #4341, opened 2018-05-21
community/xbps: fix ppc64le build break modifying _unused macro to _unused_attr

mksully22:


jirutka:
Please add some comment to the patch file why this change is needed.

mksully22:
This is a rework of Breno Leitao's patch that applied on the prior xbps (0.51) version. 
I pulled his description of the patch directly: 

Currently xbps package fails to build using musl libc because the signal.h
header includes a struct field called _unused, that becomes defined as an
function attribute due to the a CFLAG -D macro definition. This cause the
build to fail.

This patch just rename _unused macro to _unused_attr in order to avoid the
names being clashed

PR #4342, opened 2018-05-21
main/openssh: add start_permit_root function in initd

tmh1999:


tmh1999:
This is a WIP, take it easy.

PR #4350, opened 2018-05-22
main/irqbalance: make numa aware

HRio:


HRio:
@rnalrd is it not better to build irqbalance with numa support? **Note** I have no access to a system with several numa nodes running Alpine.

HRio:
@jirutka what do you think about moving numactl?

main/irqbalance/APKBUILD
```diff
@@ -33,6 +36,12 @@ build() {
 	make
 }
 
+check()
```
TBK: { should be on the same line as check():

```
check() {
.....
```

HRio:
@rnalrd should we not add $pkgname-openrc before 3.8?

numa can wait as its anyhow blocked by its numactl dependency

PR #4355, opened 2018-05-24
main/busybox: don’t silently ignore certificate validation

jirutka:
Ref https://bugs.alpinelinux.org/issues/8917

As I’m thinking about it, it may be better to just print bold warning message about insecure connection for now and switch to error in v3.9. WDYT?


clandmeter:
@jirutka nice! I don't know enough of bb internals and C in general to really review this.
Did you submit your patch to upstream? I wonder what Denys will have to say about it.

jirutka:
http://lists.busybox.net/pipermail/busybox/2018-May/086444.html

jirutka:
Superseded by a93da0e814c542ff3a76cba0b557ee51c8124d1e.

PR #4360, opened 2018-05-24
main/mutt: upgrade to 1.10.0

refacto:
Also change URL to ftp://ftp.mutt.org/, as the previous bitbucket seems
no longer maintained.


refacto:
Travis doesn't seem to like the new url :/

PR #4377, opened 2018-05-28
community/openblas: upgrade to 0.3.0, enable s390x

tmh1999:


tmh1999:
The builder will report failure because the log is too long.

PR #4384, opened 2018-05-28
community/emacs bump to 26.1

stef:
bumped emacs to 26.1 which builds without patches for musl or ASLR.


nmeum:
Does the upgrade also fix [the segfaults reported on the issue tracker](https://bugs.alpinelinux.org/issues/8849)?

stef:
this is the first time i hear about this bug, sorry. but maybe this upgrade fixes it.

stef:
was able to reproduce the segfault, and can confirm the upgrade to 26.1 solves it.

PR #4393, opened 2018-05-29
testing/bazel: new aport

strfry:
This APKBUILD introduces the Bazel build system.

Bazel creates a self-contained binary, but requires a Java Runtime Environment, and a environment variable JAVA_HOME that points to it.
I'd appreciate some feedback if this is correct to ship it that way.


refacto:
Please change PR and commit format to:
```
testing/bazel: new aport
```

PR #4403, opened 2018-05-30
community/rust: Update to 1.26.1

leo-unglaub:


jirutka:
Upstream had already released 1.26.2, so upgraded it right to 1.26.2.

PR #4409, opened 2018-05-30
Update mbedtls

leo-unglaub:


leo-unglaub:
There are ABI changes. Some functions got added and extended. But the API itself should be stable so a simple rebuild should fix all depending programs.

clandmeter:
I read this:

Mbed TLS 2.7.0 maintains source code compatibility with previous versions of Mbed TLS but there are some changes which make the ABI incompatible with the previous version, Mbed TLS 2.6.0.

ncopa:
We cannot merge this unless the packages linked to mbedtls are also rebuilt. Can you please bump the pkgrel of the following packages in this PR aswell:
````
dislocker
hiawatha
monkey
neko
````

PR #4416, opened 2018-06-01
main/pllua: use lua instead of luajit on s390x

tmh1999:


ncopa:
I wonder if it wouldn't be be better to use same lua implementation on all architectures. luajit was introduced with ef609351a6b52c7edab28bd0b549e1170fbf1723

jirutka:
https://github.com/alpinelinux/aports/blob/54f6029e04df3079fbd27f3631e793162618f326/main/pllua/APKBUILD#L20-L21

These variables has to be changed too.

PR #4417, opened 2018-06-01
main/transmission: enable build on new s390x builder

tmh1999:


main/transmission/APKBUILD
```diff
@@ -1,11 +1,11 @@
 # Maintainer: Carlo Landmeter <clandmeter@gmail.com>
 pkgname=transmission
 pkgver=2.94
-pkgrel=0
+pkgrel=1
```
clandmeter: You dont need to bump pkgrel if a package doesn't exist on a specific arch.
The commit msg is also misleading as its only for a specific arch and on master branch which is still edge (and soon to be 3.8).

tmh1999: I would like this to be applied to both v3.8 and edge builder, which are fetching master branch atm. Updated. Thanks.

PR #4421, opened 2018-06-01
testing/openmpi: upgrade to 3.1.0, fix s390x

tmh1999:
openmpi commit bc54c99e12cf33c5a6c70bf1c3952075b66a76ca allows check for
using s390x's builtin asm, previous versions need
--enable-builtin-atomics


tmh1999:
We can just stay with 3.0.0 and add --enable-builtin-atomics for s390x without upgrading.

PR #4427, opened 2018-06-03
community/gimp: upgrade to 2.10.2

refacto:
This PR depends on many others:

PR #4430, opened 2018-06-03
community/gegl: upgrade to 0.4.2

refacto:
Is required for GIMP 2.10.2.


nmeum:
```
-usr/lib/libgegl-0.2.so
+usr/lib/libgegl-0.4.so
```

This includes an soname bump. Please bump the pkgrel of all packages linked against the old version.

nmeum:
We discussed this a bit on IRC the goal is to merge this PR with #4427 thus no rebuild is necessary since gimp is the only package depend on `community/gegl`.

PR #4440, opened 2018-06-06
testing/mopidy: cleanup, add init file, logging, and manpages, and fix permissions

PureTryOut:
- Added init files for OpenRC, the service can now be started with `rc-service mopidy start`
- The service now logs to `/var/log/mopidy`
- System user is created and used for folder permissions

I'm not entirely sure if it's correct to create the `/var/lib/mopidy` folders in the post-install hook. And should I remove the mopidy user (and thus also the `/var/lib/mopidy` folder due to permissions) again in a post-deinstall hook?


testing/mopidy/mopidy.post-install
```diff
@@ -0,0 +1,14 @@
...
+mkdir /etc/mopidy/extensions.d
+mkdir /var/cache/mopidy
+mkdir /var/log/mopidy
+mkdir /var/lib/mopidy/local
+mkdir /var/lib/mopidy/media
+mkdir /var/lib/mopidy/playlists
+
+chown -R mopidy:audio /etc/mopidy/extensions.d
+chown -R mopidy:audio /var/cache/mopidy
+chown -R mopidy:audio /var/lib/mopidy
```
ncopa: I think those should be created in APKBUILD in `package()`.

PureTryOut: But they have to be owned by the correct user and group. Since the user is created in the pre-install hook, I don't think I can add the correct owner from `package()` right? Would the correct fix be creating the directories in `package()`, and then chowning it in the post-install hook?

testing/mopidy/mopidy.post-install
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+
+adduser -S -D -h /var/lib/mopidy -s /sbin/nologin -G audio mopidy 2>/dev/null
+
+chown -R mopidy:audio /etc/mopidy/extensions.d 2>/dev/null
+chown -R mopidy:audio /var/cache/mopidy 2>/dev/null
+chown -R mopidy:audio /var/lib/mopidy 2>/dev/null
```
ncopa: the permissions should also be set in APKBUILD. sorry for not being clear about that last time.

````sh
pkgusers="mopidy"
pkggroups="audio"
...
package() {
    install -dm755 -o modpidy -g audio "${pkgdir}/var/cache/mopidy"
    ....
}
````

File: testing/mopidy/mopidy.post-install [hunk 3]
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+
+adduser -S -D -h /var/lib/mopidy -s /sbin/nologin -G audio mopidy 2>/dev/null
```
ncopa: adduser needs to be in `pre-install`.

testing/mopidy/mopidy.post-install
PureTryOut: This combined with adduser in `pre-install` just causes "unknown user 'mopidy'"...

PureTryOut: Ah I missed the `pkgusers` and `pkggroups` part. Thanks, that did it, will update the PR.

PureTryOut: Hmm, now the directories are owned by `messagebus:audio`? I never used that user anywhere, not sure why this is happening.

testing/mopidy/APKBUILD
```diff
@@ -14,33 +14,60 @@ depends="python py-pykka py-tornado gst-plugins-base gst-plugins-good
...
+		"$pkgdir/etc/init.d/$pkgname"
+
+	install -Dm644 LICENSE "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
+
+	install -dm755 -o mopidy -g audio "${pkgdir}/etc/mopidy/extensions.d"
+	install -dm755 -o mopidy -g audio "${pkgdir}/var/cache/mopidy"
+	install -dm755 -o mopidy -g audio "${pkgdir}/var/log/mopidy"
+	install -dm755 -o mopidy -g audio "${pkgdir}/var/lib/mopidy/local"
+	install -dm755 -o mopidy -g audio "${pkgdir}/var/lib/mopidy/media"
+	install -dm755 -o mopidy -g audio "${pkgdir}/var/lib/mopidy/playlists"
```
clandmeter: Install accepts multiple arguments (dirs/files) so no need to repeat it.

testing/mopidy/APKBUILD
```diff
@@ -14,33 +14,60 @@ depends="python py-pykka py-tornado gst-plugins-base gst-plugins-good
...
+
+	ls -l "${pkgdir}/var/lib/mopidy"
```
clandmeter: This is debug info?

testing/mopidy/mopidy.initd
```diff
@@ -0,0 +1,18 @@
+#!/sbin/openrc-run
+
+name=mopidy
+command="/usr/bin/mopidy"
+command_args="--config /usr/share/mopidy/conf.d:/etc/mopidy/mopidy.conf"
```
clandmeter: Is this correct?

testing/mopidy/APKBUILD
PureTryOut: Yeah I pushed what I had as you asked for the scripts. Will be removed before merging

PureTryOut:
So when installing this package on an already installed system, it works fine; all folders are created with the right owner and permissions. Only when this package is installed as part of the base installation (at least for postmarketOS) the folders are owned by `messagebus:audio` instead.

Seeing this package will probably never be part of any base installation, this is fine for me, but I understand if this is not the case for you guys.

PureTryOut:
I had some mail contact with Marvin Steadfast (the previous maintainer) and he agreed to me taking over the maintenance of this package. I'm hereby also moving it to community, so hopefully it can be included in Alpine Linux 3.9.

PureTryOut:
It seems one of mopidy's dependencies got updated to a version that's not (yet) supported by Mopidy. I guess this PR will have to wait till Mopidy releases an update that's compatible with version 2.7 of py-idna.

PureTryOut:
The package builds again, could somebody do another review?

PureTryOut:
Still waiting for a review!

PureTryOut:
Huh, why did this happen? I commented which should have prevented this...

clandmeter:
Looks like the bot doesn't consider a comment a change.
We will have to ask contributors to rebase instead of comment.

Ikke:
@PureTryOut The first commit lists lots of changes, except the most significant one, moving it to community. Maybe that should even be it's own commit (separating cleanup/changes from moving).

community/mopidy/APKBUILD
```diff
@@ -0,0 +1,79 @@
...
+pkgrel=0
+pkgdesc="Mopidy is an extensible music server"
+url="http://www.mopidy.com/"
+arch="noarch"
+license="Apache-2.0"
+depends="python py-pykka py-tornado gst-plugins-base gst-plugins-good
+	gst-plugins-ugly py-gst py-setuptools py-libxml2 py-requests
+	py-six
+	"
+depends_dev=""
```
Ikke: Empty variables can be removed.

community/mopidy/APKBUILD
```diff
@@ -0,0 +1,79 @@
...
+		"${pkgdir}/etc/mopidy/extensions.d" \
+		"${pkgdir}/var/log/mopidy" \
+		"${pkgdir}/var/cache/mopidy" \
+		"${pkgdir}/var/lib/mopidy/media" \
+		"${pkgdir}/var/lib/mopidy/playlists"
+}
+
+check() {
+	cd "$_builddir"
+	python -m mopidy --help > /dev/null
```
Ikke: Just calling the help function does not suffice as a check function. If there is no test suite, the preference is to mark this package as `!check`.

PureTryOut:
I've rebased the branch, made moving the packages a separate commit, updated `py-pykka` to 2.0.0 and `mopidy` to 2.0.0.

PureTryOut:
@maxice8 I've applied your suggested changes.

Ikke:
Merged as 3284255, 403823e, 10a4515, 7218cf7. Closing PR.

PureTryOut:
Thanks! You seem to have forgotten to close the PR, so I did it myself :wink:

PR #4446, opened 2018-06-07
main/git: add less as dependency

am11:
When we install `git` in a new environment, `git diff` command does not
function well on some shells without `less`.
Granted this is an implicit dependency, but it affects user experience
as there is no error message that indicates what is missing.

### DIFF
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -14,7 +14,7 @@ pkgdesc="Distributed version control system"
 url="https://www.git-scm.com"
 arch="all"
 license="GPL-2.0-or-later"
-depends=""
+depends="less"
 # we need tcl and tk to be built before git due to git-gui and gitk
 makedepends="zlib-dev libressl-dev curl-dev expat-dev perl-dev python2-dev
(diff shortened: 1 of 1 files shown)


refacto:
This would increase the size of git installations that will never see any direct user interaction tough.

am11:
This is what I see as `git diff` output in a new box when `less` is not installed:

image

This patch is being proposed for the future users. I personally don't have strong opinion and have found the reason and fix. Perhaps, if possible, it should show ASCII characters when less is not available?

ncopa:
nack. this would just increase size for no good reason for all the scripts using git. I do agree that this is annoying though.

Possible other solutions:
- make git use `--no-color` by default
- patch busybox less to support colors (see http://lists.busybox.net/pipermail/busybox/2014-January/080353.html)

am11:
I'd vote/hope for busybox option. It seems like -R is already checked into master https://github.com/mirror/busybox/blob/43b17b1/miscutils/less.c#L103-L110, but the latest release is missing past 9 months of commits https://github.com/mirror/busybox/commits/1_28_4/miscutils/less.c (although released 16 days ago) :(

PR #4447, opened 2018-06-07
main/alpine-baselayout: revert 'sysctl security changes'

tstearns:
Lack of agreement on proper values for these settings indicates
that we should be following the Linux kernel's defaults
rather than enforcing our own.

Multiple mistakes in the original implementation reinforce the
justification for this principle.

We can selectively add back any sysctl settings that have a clear
and compelling need to be overriden.


kaniini:
I would prefer to see `fs.protected_symlinks` and `fs.protected_hardlinks` remain.  the network settings are mostly harmless, but i could see how they may be problematic.

PR #4470, opened 2018-06-11
main/gnupg: security fix for CVE-2018-12020

prspkt:
Backport for gnupg-2.1.10 in 3.3-stable.


jirutka:
v3.3 is not supported anymore, see https://wiki.alpinelinux.org/wiki/Releases.

PR #4471, opened 2018-06-11
main/gnupg: security fix for CVE-2018-12020

prspkt:
Backport for gnupg-2.1.12 in 3.4-stable.


jirutka:
I've just renamed the patch file and applied as d159a02c3e2473f94a17f38922f25943b00469af. Thanks!

PR #4484, opened 2018-06-12
main/harfbuzz: build static lib

malbarbo:
This makes easy to use alpine to build a static version of https://github.com/tectonic-typesetting/tectonic (a latex engine).


malbarbo:
Done.

PR #4485, opened 2018-06-12
main/freetype: build static lib

malbarbo:
This makes easy to use alpine to build a static version of https://github.com/tectonic-typesetting/tectonic (a latex engine).

### DIFF
--- a/main/freetype/APKBUILD
+++ b/main/freetype/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Carlo Landmeter <clandmeter@gmail.com>
 pkgname=freetype
 pkgver=2.9.1
-pkgrel=0
+pkgrel=1
 pkgdesc="TrueType font rendering library"
 url="https://www.freetype.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


malbarbo:
Done.

tanertas:
May I ask that why we need an extra package for static libraries? Last time I checked that static files were moving to -dev package already.

ncopa:
We prefer dynamically linked packages and in most cases the dynamic libs are sufficient. By keeping the static libs in separate package we save space and bandwith in most cases. It also makes the intention clearer.

tanertas:
I think that the situation here is different. The corresponding package's default use is to provide a shared object for other dynamically linked binaries. But static libraries has no use except building a statically linked binary which makes them more suitable to keep in development (-dev) package.

In my opinion, default behavior is correct and this is the mainstream approach with all *.a files. A -static package makes only sense for an alternate version of same executable binary like regular apk-tools and static version of it: apk-tools-static. Otherwise we may put ourselves in confusion.

I see that there's also a mariadb-static package. Like regular (or tariditional) users myself would think that this is a statically linked version of regular mariadb package. But it is not. It is just a development package that stores regular *.a files.

To provide an extra package for static objects simply can't be a coherent approach of correctness.

* https://pkgs.alpinelinux.org/contents?branch=edge&name=apk-tools-static&arch=x86_64&repo=main
* https://pkgs.alpinelinux.org/contents?branch=edge&name=mariadb-static&arch=x86_64&repo=main

PR #4507, opened 2018-06-14
testing/wireguard: upgrade to 0.0.20180613

zx2c4:
```
  * wg-quick: android: change name of intent
  * wg-quick: android: delay setting users until end
  
  `ndc users add` eventually invokes SOCK_DESTROY on user sockets, causing
  them to reconnect. By delaying this until after routes are set, we
  ensure that the sockets reconnect using the tunnel, rather than the old
  route. This fixes push notifications on Android.
  
  * chacha20: add missing include to header
  
  Fixes a compile error on some kernels.
  
  * tools: encoding: add missing static array constraints
  
  Makes static analyzers happier.
  
  * tools: support getentropy(3)
  
  This lets us take advantage of both recent glibc calls as well as the long
  standing getentropy functions on the BSDs.
  
  * chacha20poly1305: use slow crypto on -rt kernels
  
  In rt kernels, spinlocks call schedule(), which means preemption can't
  be disabled. The FPU disables preemption. Hence, we can either
  restructure things to move the calls to kernel_fpu_begin/end to be
  really close to the actual crypto routines, or we can do the slower
  lazier solution of just not using the FPU at all on -rt kernels. This
  patch goes with the latter lazy solution. The reason why we don't
  place the calls to kernel_fpu_begin/end close to the crypto routines
  in the first place is that they're very expensive, as it usually
  involves a call to XSAVE. So on sane kernels, we benefit from only
  having to call it once.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 
 pkgname=wireguard-tools
-pkgver=0.0.20180531
+pkgver=0.0.20180613
 pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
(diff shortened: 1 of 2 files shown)


ncopa:
fixed with 37250a4f09d679dc3cdd243dafc16910ac03c3a2

PR #4521, opened 2018-06-15
main/libnotify: enable gobject-introspection

Decatf:
Python depends on this in order to use libnotify.

### DIFF
--- a/main/libnotify/APKBUILD
+++ b/main/libnotify/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=libnotify
 pkgver=0.7.7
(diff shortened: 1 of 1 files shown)


Decatf:
I don't see this flag in the libnotify makefiles. Is it a requirement of gobject-introspection or something?

It appears to me that including the `gobject-introspection-dev` dependency is enough. I looked at a few other aports which provide gobject introspection. Some use this flag and some don't.

Decatf:
[This](https://wiki.gnome.org/Projects/GObjectIntrospection/AutotoolsIntegration) says it's supposed to be added in the makefile of library that needs it but libnotify doesn't.

I've tested and it works either way. I've added the `--enable-introspection` flag so it's clear that we do have it enabled.

PR #4522, opened 2018-06-15
main/nettle: disable arm neon

Decatf:
nettle checks the build machine for cpu features which ends up enabling
neon but armv6 only supports up to vfpv2.

### DIFF
--- a/main/nettle/APKBUILD
+++ b/main/nettle/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=nettle
 pkgver=3.4
-pkgrel=0
+pkgrel=1
 pkgdesc="A low-level cryptographic library"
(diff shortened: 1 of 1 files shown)


Decatf:
The build log of the current package shows arm neon is enabled.
http://build.alpinelinux.org/buildlogs/build-edge-armhf/main/nettle/nettle-3.4-r0.log
```
checking if /proc/cpuinfo claims neon support... yes
...
configure: summary of build options:

  Version:           nettle 3.4
  Host type:         armv6-alpine-linux-musleabihf
  ABI:               standard
  Assembly files:    arm/neon arm/v6 arm
```

After this patch I compiled nettle in [pmbootstrap](https://github.com/postmarketOS/pmbootstrap). It will skip the check of `/proc/cpuinfo` since arm neon is disabled by the configure flag.
```
configure: summary of build options:

  Version:           nettle 3.4
  Host type:         armv6-alpine-linux-musleabihf
  ABI:               standard
  Assembly files:    arm/v6 arm
```

PR #4523, opened 2018-06-15
testing/autofs: new aport

alex-laskin:
A kernel-based automounter for Linux

### DIFF
--- a/testing/autofs/APKBUILD
+++ b/testing/autofs/APKBUILD
@@ -1,0 +1,52 @@
+# Contributor: Alex Laskin <alex@lask.in>
+# Maintainer: Alex Laskin <alex@lask.in>
+pkgname=autofs
+pkgver=5.1.4
+pkgrel=0
+pkgdesc="A kernel-based automounter for Linux"
+url="https://www.kernel.org/pub/linux/daemons/autofs/"
+arch="all"
+license="GPL-2.0-only"
+depends=""
+makedepends="linux-headers automake flex bison rpcgen libtirpc-dev libnfs-dev"
+subpackages="$pkgname-doc $pkgname-openrc"
+source="
+  https://www.kernel.org/pub/linux/daemons/${pkgname}/v5/${pkgname}-${pkgver}.tar.xz
+  autofs.initd
+  autofs.confd
+  10-glibc.patch
+  "
+
+build() {
+  cd "$builddir"
+  ./configure \
+    --prefix=/usr \
+    --sysconfdir=/etc/autofs \
+    --with-mapdir=/etc/autofs \
+    --with-confdir=/etc/conf.d \
+    --without-hesiod \
+    --with-libtirpc \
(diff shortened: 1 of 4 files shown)


testing/autofs/APKBUILD
```diff
@@ -0,0 +1,52 @@
+# Contributor: Alex Laskin <alex@lask.in>
+# Maintainer: Alex Laskin <alex@lask.in>
+pkgname=autofs
+pkgver=5.1.4
+pkgrel=0
+pkgdesc="A kernel-based automounter for Linux"
+url="https://www.kernel.org/pub/linux/daemons/autofs/"
+arch="all"
+license="GPL"
```
Ikke: Use one of the [SPDX](https://spdx.org/licenses/) license descriptions, for example `GPL-2.0-only`

testing/autofs/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+    --with-confdir=/etc/conf.d \
+    --without-hesiod \
+		--with-libtirpc \
+    --enable-ignore-busy
+  make -j8
+}
+
+check() {
+  cd "$builddir"
+  daemon/automount -V
```
Ikke: Just showing the version is not enough for a `check` function. It can give a false sense of security because usually programs exit early when showing the version. Not sure if there is something else to put here though.

ncopa:
looks like a new aport. can you please change commit message to: `testing/autofs: new aport` and squash the 3 commits into a singe?

testing/autofs/APKBUILD
alex-laskin: Ok, sounds reasonable.

testing/autofs/APKBUILD
alex-laskin: This tool has no its own test suite, so I think package maintainer has no more options to test tool than that. I can try to write extensive tests, but should I?

alex-laskin: From https://wiki.alpinelinux.org/wiki/APKBUILD_Reference : «Creating a very simple check function, that calls program --version is better than disabling tests completely.» (https://github.com/alpinelinux/aports/pull/2322#discussion_r142543002)

testing/autofs/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+  cd "$builddir"
+  ./configure \
+    --prefix=/usr \
+    --sysconfdir=/etc/autofs \
+    --with-mapdir=/etc/autofs \
+    --with-confdir=/etc/conf.d \
+    --without-hesiod \
+    --with-libtirpc \
+    --enable-ignore-busy
+  make -j8
```
tanertas: job numbers must define in /etc/abuild.conf.

HRio:
commited as https://github.com/alpinelinux/aports/commit/1b5e1a72021e783153d1a6a6d132a901881e6ae9

PR #4524, opened 2018-06-15
main/dropbear: disable wtmp and lastlog support

nervo:
As spotted on this issue (https://bugs.alpinelinux.org/issues/5075), musl does not support wtmp, nor alpine support lastlog.


nervo:
@ncopa sorry, seems legit :)

PR #4528, opened 2018-06-16
testing/ddcutil: new aport

alex-laskin:
Query and change Linux monitor settings using DDC/CI and USB


maxice8:
Merged, sorry for taking so long.

PR #4531, opened 2018-06-16
certbot-nginx: Nginx plugin for Certbot

74cmonty:
Please provide the following package: certbot-nginx
Source code availabe [here](https://pypi.org/project/certbot-nginx/#files).


andypost:
I see no reason in this PR
Better to implement https://github.com/nginx-modules/ngx_http_acme_module

PR #4533, opened 2018-06-16
main/py-requests: upgrade to 2.19.1

fabaff:


nmeum:
> I have tested, that this PR fixes the compilation error in extra-cmake-modules (meaning the incompatibility is gone).

Good to know, since it is (as you pointed out already) a simple version upgrade otherwise I am merging this right away.

PR #4540, opened 2018-06-16
darktable: upgrade to 2.4.3 and move to community

Ikke:
Darktable has been in testing for a while and has been stable. I've verified that it runs and that the main functionality works.

I also get notifications about the package being out of date, so there are more people interested in darktable.


ncopa:
fixed with 0522443c578b5e623b0f10cfde943209c6c598a5

PR #4541, opened 2018-06-16
testing/snapcast: cleaned, move to community and upgrade to 0.15.0

PureTryOut:
I've basically redone the whole aport.

- Renamed snapcast-client and snapcast-server subpackages to their upstream names
- Added avahi runtime dependency
- Replaced config files for their upstream variants in the upstream defined location
- Removed logging location from `start_pre()` path check, as it's not actually used


PureTryOut:
I've talked with Marvin Steadfast (the previous maintainer of this package) over mail and we agreed to let me take over the maintenance. So hereby I'm taking it over and also moving it to community.

nmeum:
Could you split the commits moving the various packages to `community` in separated commits?

PureTryOut:
Is it better like this, or do you want the separate packages in separate commits?

PureTryOut:
Ok everything should be ok now.

community/snapcast/APKBUILD
```diff
@@ -0,0 +1,69 @@
...
+pkgname=snapcast
+pkgver=0.14.0
+pkgrel=0
+pkgdesc="Synchronous multi-room audio player"
+url="https://github.com/badaix/snapcast"
+arch="all"
+license="GPL-3.0"
+depends="snapserver snapclient"
+makedepends="bash avahi-dev aixlog popl libvorbis-dev flac-dev asio-dev alsa-lib-dev"
+subpackages="$pkgname-doc snapserver snapclient"
```
nmeum: Is there any particular reason why you renamed the subpackages? I would personally prefer `$pkgname-server` and `$pkgname-client` since this is consistent with other packages, e.g. openssh or mosh.

nmeum:
> Added avahi runtime dependency

Is that an optional dependency? Because IIRC I used snapcast without avahi in the past.

testing/snapcast/APKBUILD
```diff
@@ -1,70 +1,69 @@
...
+	install -dm755 "$subpkgdir"/etc/conf.d
+	install -dm755 "$subpkgdir"/etc/init.d
+
 	mv "$pkgdir"/usr/bin/snapserver \
 		"$subpkgdir"/usr/bin/
-	install -Dm755 "$srcdir"/"$subpkgname.initd" \
-		"$subpkgdir"/etc/init.d/"$subpkgname"
-	install -Dm644 "$srcdir"/"$subpkgname.confd" \
-		"$subpkgdir"/etc/conf.d/"$subpkgname"
+	install -Dm644 server/debian/snapserver.default "$subpkgdir"/etc/default/snapserver
```
nmeum: I am not aware of any other packages using `/etc/default` on Alpine is that location even checked by OpenRC?

testing/snapcast/APKBUILD
```diff
@@ -1,70 +1,69 @@
...
-fa353d7974d3a8b3b22ff009b762c7c4ee318f9f83e22762a9ebf550484b9fc3cc4dffd389b6d756630dc035cb86d5c5643541957785f8bbf71fe4915691c337  snapcast-client.confd
-0f6c43e267646ff657bfe184481c7148d3395a6ed2b9c485856b56370dcb5a1d2364f856af683d9af13fc12b8687ba890345673311caabefa09c9471e7b5cf48  snapcast-server.initd
-48d494143d044a413b414b32cd8be51d4242234b6d7cb03102914a42b1ea6b1913c2c31b223c23c6be316918e731875149a2505daa03165659d0a1be3ca7e75c  snapcast-server.confd"
+	install -dm755 "$subpkgdir"/usr/bin
+	install -dm755 "$subpkgdir"/etc/conf.d
+	install -dm755 "$subpkgdir"/etc/init.d
+
+	mv "$pkgdir"/usr/bin/snapclient \
+		"$subpkgdir"/usr/bin
+	install -Dm644 client/debian/snapclient.default "$subpkgdir"/etc/default/snapclient
```
nmeum: Same thing here.

community/snapcast/APKBUILD
PureTryOut: I renamed it because... Well... they're the upstream names. I personally prefer to go with upstream wherever possible, but I guess I can change it back...

testing/snapcast/APKBUILD
PureTryOut: You're right, it isn't checked by OpenRC, I'll change that in a bit.

I'm not sure if any other packages do it, but like with my other comment, this is where upstream places it so I went with that. I mean, I could probably place it somewhere else, but is there any particular reason against having it there?

PureTryOut:
>> Added avahi runtime dependency

> Is that an optional dependency? Because IIRC I used snapcast without avahi in the past.

Yes it is. You can point clients directly at the server by specifying an url or IP, but with avahi clients can autodetect the server without further configuration. The fact that you ask this question makes me think optional dependencies should not be there by default?

nmeum:
> The fact that you ask this question makes me think optional dependencies should not be there by default?

The thing with enabling optional dependencies by default is that you can't uninstall them without uninstalling the package itself. Therefore we actually try to avoid depending on optional dependencies, see e.g. 49794390171fe37829d0f4f162823b43917116dc as an example.

PureTryOut:
I applied all your requested changes, and updated Snapcast to it's latest version as well.

nmeum:
> I applied all your requested changes.

Ok thanks, the things the git commit history is a bit messed up now. Which of your initial changes actually remain now? As far as I am concerned the following changes where made:

* The order of the function declarations was changed
* The package was upgrade from `0.13.0` to `0.14.0`
* The `install_if` rule of the subpackages was removed (why?)
* The `pkgusers` and `pkggroups` variables where removed (is that intentional?)
* You added `.post-deinstall` scripts which remove the user and the group (which is imho a bad idea since there might still be files owned by that user/group (is there any other aport which does this?))
* The OpenRC services files changed, e.g. snapcast client now runs as `snapclient:audio` instead of `snapcast:audio`

The git commit history really doesn't reflect that those changes where made and the diff is pretty hard to read currently because you reverted some changes you made in the "cleanup"-commit later on the "apply requested changes"-commit. Some of those changes, e.g. the package upgrade or the group name change should probably go in their own commit. And some of the changes e.g. the addition of the `.post-deinstall` scripts should probably be reverted.

In any case it would be cool if you could cleanup the commit history somehow because for me this is pretty hard to review as is. I do understand though that this is a lot of work…

PureTryOut:
Ok, I just decided to start over. Everything is now in seperate commits, and the "cleanup" is way smaller than I initially made it, so it should be how you requested it now. *Hopefully* this is finally good to merge.

PureTryOut:
@nmeum could you do one last review?

PureTryOut:
Is anybody willing to do a review still?

Ikke:
Does the project have any tests built-in? If so, can run the tests in the check() function. Otherwise specify 

```sh
options="!checks" # No test suite present
```

PureTryOut:
It does not. There is an option in CMakeLists.txt to build tests, but using that it still doesn't run any tests (`ctest` actually tells us in the log there are no tests). I've added `options="!checks"` for now

PureTryOut:
Is this ready for merging now?

nmeum:
Oopps, I am sorry. I somehow missed my GitHub notifications for this PR. This looks much better now. A few comments:

1. This needs to be rebased as there is a merge conflict currently.
2.  Please add a comment to the `options="!check"` variable explaining that upstream doesn't have a test suite

nmeum:
To make up for my failure to respond in time I made the necessary changes myself (let me know if I fucked something up) and merged your PR. Thanks for taking over maintainership for a bunch of packages and sorry again that it took so long to get this merged.

PureTryOut:
No problem, thanks for your help and thorough review!

PR #4542, opened 2018-06-16
main/py-eyed3: upgrade to 0.8.7

prspkt:


clandmeter:
@prspkt please check next time when moving packages downstream (main to community) nothing in main depends on it. Depending downstream is now allowed.

prspkt:
@clandmeter sorry, my mistake and thank you for the clarification on downstream dependencies.

PR #4549, opened 2018-06-18
community/rust: changes to add initial support for ppc64le

mksully22:
Patch set includes changes to libc and liblibc subdirs to support definitions for ppc64le. An initial PR was submitted to https://github.com/rust-lang/libc/pull/1023 to add these changes in the long term to the libc project.

There are currently some limitations when building for ppc64le. When running with parallelism the build will intermittently stall, even with the "-j 1" parameter (looks like rustc internal job server still splits out multiple instances for rustc work). To workaround that for now taskset to cpu 0 is added. This problem will be investigated further.

Also a number of the check() tests fail on ppc64le which will also need further investigation with the rust team. For now check() had been made conditional for ppc64le. 

These patches have been used to create rustc, rustc-std, and cargo binaries. For reference a github project has been setup that contains the methodology used as well as the output  (tar.gz and apk files) https://github.com/mksully22/ppc64le_alpine_rust_1.26.2


ncopa:
looks like it needs to be rebased against rust 1.27.0

mksully22:
Ok, I'll take a look at rebasing the patches to the new 1.27.0 level

rgdoliveira:
@mksully22 Is there any progress on the rebase of this PR?

mksully22:
@rgdoliveira I spent some time on the rebase to 1.27.0 but couldn't get the get the compiler code to re-compile itself sucessfully. I hope to get back to this later for further investigation.

mksully22:
Sorry but I couldn't get rust to rebuild successfully at the latest levels
on ppc64le and haven't had time to get back to it.

PR #4490, opened 2018-06-12
testing/py-atomicwrites: new aport

ajones-rvbd:
pytest seems to depend on it.

Issue: https://bugs.alpinelinux.org/issues/8975

PR #4557, opened 2018-06-19
community/gcovr: upgrade to 4.1

misery:


community/gcovr/APKBUILD
```diff
@@ -1,30 +1,32 @@
...
+makedepends="python3-dev py3-setuptools"
+checkdepends="py3-pytest"
+options="!check" # ModuleNotFoundError: No module named 'atomicwrites'
 source="gcovr-${pkgver}.tar.gz::https://github.com/gcovr/gcovr/archive/${pkgver}.tar.gz"
 builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
 	cd "$builddir"
-	python2 setup.py build
+	python setup.py build
```
nmeum: You have to use `python3` instead of `python` here.

community/gcovr/APKBUILD
```diff
@@ -1,30 +1,32 @@
...
 check() {
 	cd "$builddir"
-	./scripts/gcovr --help > /dev/null
+	#py.test-3 -v gcovr
 }
 
 package() {
 	cd "$builddir"
-	python2 setup.py install --prefix=/usr --root="$pkgdir"
+	python setup.py install --prefix=/usr --root="$pkgdir"
```
nmeum: Same thing as above.

community/gcovr/APKBUILD
misery: Thanks!

misery:
Maybe wait for checkdepends. See https://github.com/alpinelinux/aports/pull/4490

misery:
Sadly... pyutilib is missing now.

```
Hint: make sure your test modules/packages have valid Python names.
Traceback:
gcovr/tests/test_gcovr.py:11: in <module>
    from pyutilib.misc.pyyaml_util import compare_repn as compare_xml
E   ModuleNotFoundError: No module named 'pyutilib'
```

nmeum:
> Sadly... pyutilib is missing now.

@jirutka since you took care of adding `py-atomicwrites` do you mind taking a look at this?

jirutka:
> @jirutka since you took care of adding `py-atomicwrites` do you mind taking a look at this?

I've added py-atomicwrites just to fix pytest which was broken, because someone recklessly updated it to a new minor version without checking dependencies, not because I'd like creating python aports. ;)

`pyutilib` is a testing dependency of `gcovr`, not `pytest` (see [requirements.txt](https://github.com/gcovr/gcovr/blob/master/requirements.txt)). So, @misery , please create aport `community/py3-pyutilib` and add it to this PR (in a separate commit).

PR #4561, opened 2018-06-19
scripts/mkimg.standard.sh. For ppc64le add ibmvscsi module to command line to avoid ra…

mksully22:
…ce issue during install from iso for qemu guest

During install of qemu guest using iso image the boot intermittently fails and drops into the emergency shell with the message "Mounting boot media failed". It looks like the nlplug-findfs -p /sbin/mdev -d -b /tmp/repositories -a /tmp/apkovls is sometimes not discovering the cdrom (dev/sr0). Adding the ibmvscsi module to the default command line should avoid this race.


ncopa:
I think it could also have been solved by adding a longer timeout, but this is ok too.

The commit message should  have a shorter first line, so splitted it in two.

ncopa:
i also think we should use `case ... in` instead of having multiple `if ...; then`

PR #4566, opened 2018-06-21
Enable login console for ppc64le

mtarsel:
Since ppc64le is using Petitboot, remove comment in order to use the
getty login console via serial cable.

### DIFF
--- a/main/alpine-baselayout/inittab
+++ b/main/alpine-baselayout/inittab
@@ -12,8 +12,10 @@ tty4::respawn:/sbin/getty 38400 tty4
 tty5::respawn:/sbin/getty 38400 tty5
 tty6::respawn:/sbin/getty 38400 tty6
 
-# Put a getty on the serial port
-#ttyS0::respawn:/sbin/getty -L ttyS0 115200 vt100
+# if ppc64le, put a getty on the serial port
+if [ $(uname -m) = "ppc64le" ]; then
+    ttyS0::respawn:/sbin/getty -L ttyS0 115200 vt100
+fi
 
 # Stuff to do for the 3-finger salute
 ::ctrlaltdel:/sbin/reboot


ncopa:
The only thing you need to accomplish what you are trying to do here is to add `console=ttyS0` as boot option.

mtarsel:
The default console for both power8 and power9 machines is `hvc0` so using `console=hvc0` did work with rc6 but I was just trying to shorten the long list of kernel commands to netboot Alpine on bare metal ppc64le. Netbooting rc8 did give me a login shell no problem (no extra kernel commands needed), however this was not the case in rc6 and earlier and this patch made it work in earlier rc releases. Found it here: https://wiki.alpinelinux.org/wiki/Enable_Serial_Console_on_Boot#Enabling_a_login_console

mtarsel:
After doing some testing with rc8 I have found using `console=hvc0 `will produce a login console after a bare metal net boot on ppc64le. Using an `apkovl=` without the `console=` will also produce a login prompt which was causing the false positives in my testing. If no apkovl or console is specified in the kernel, we do not get a login prompt - is this the proper behavior for booting on bare metal? Shouldn't I be able to get a login shell without using console= or apkovl=?

jk-ozlabs:
@mtarsel the way busybox's init works, it will only start a getty (creating a login prompt) on consoles that you explicitly list in `/etc/inittab`.

However, the `/dev/console` node should always point to the console device that the kernel used during boot. So, adding an inittab entry for `console:...` should be a fairly portable way of always starting a getty on the default boot console.

Otherwise, you could add an entry for `hvc0`; either in an architecture-specific way (which would be defined a build time, but busybox does have ways to dynamically create that), or include it in all architectures' builds, and just have it fail (hopefully gracefully) on systems without a `hvc0`.

mtarsel:
@ncopa what do you think about @jk-ozlabs recommendations? I think we should add _something_ in order to get a login console for Alpine on a bare metal ppc64le box. Booting Alpine on ppc64le via an .iso and a docker container work just fine (supplying a login console) and I think net booting Alpine on bare metal ppc64le should be equally supported.

tmh1999:
@mtarsel : have you checked this commit ? https://git.alpinelinux.org/cgit/aports/commit/main/alpine-baselayout?id=bda5751e08a24a6a33f0940c1e7be6d105496c5b

Also, you mentioned ```debug_init``` caused kernel panic, have you tried injecting ```set -x``` into ```initramfs-init``` and re-cpio the ```initramfs``` image?

jk-ozlabs:
My only concern with requiring `console=hvc0` on the kernel command line is that it's an extra setting that's user-visible. If `inittab` can provide an entry, we're not going to see issues if that option has (inadvertantly) been dropped.

The commit that @tmh1999 linked looks like an good implementation for the s390 arch, should we do something similar for power?

mtarsel:
Thank you very much @jk-ozlabs and @tmh1999 I will submit a new PR similar to the s390x patch

tmh1999:
I hope this should work:
```
diff --git a/main/alpine-baselayout/APKBUILD b/main/alpine-baselayout/APKBUILD
index 2f8f896a38..b5a499e94f 100644
--- a/main/alpine-baselayout/APKBUILD
+++ b/main/alpine-baselayout/APKBUILD
@@ -195,6 +195,8 @@ package() {
 			sed -i "s/tty$i::/\#tty$i::/g" "$srcdir"/inittab
 		done
 		echo "console::respawn:/sbin/getty 38400 /dev/console" >> "$srcdir"/inittab
+	elif [ "$CARCH" = "ppc64le" ]; then
+		sed -i "s/\#ttyS0::/ttyS0::/g" "$srcdir"/inittab
 	fi
 
 	install -m644 \
```

PR #4576, opened 2018-06-22
main/usbutils: upgrade to 010

prspkt:
Edits to `hwdata.patch` are due to hunk 2 failing to apply.


prspkt:
@nmeum Yes the removal is from upstream (2nd change from the top): [NEWS](https://github.com/gregkh/usbutils/blob/v010/NEWS)
Depending packages (`testing/usbip-utils` and `spice-gtk`) build fine (locally).

prspkt:
@nmeum I know it is unnecessary but there don't seem to be any packages that would potentially require `usbutils.pc` in the ports tree. I am dropping the above since there's no reason to rebuild them.

PR #4592, opened 2018-06-23
main/cairo: upgrade to 1.15.12

TBK:
https://abi-laboratory.pro/index.php?view=timeline&l=cairo


TBK:
@nmeum It was enabled upstream - https://cgit.freedesktop.org/cairo/commit/src/cairo-ft-font.c?id=52b17c724260512873ab0731e2e2314bb0a15e4b

PR #4594, opened 2018-06-23
main/libbluray: upgrade to 1.0.2

TBK:
100% bw compatible - https://abi-laboratory.pro/?view=timeline&l=libbluray


TBK:
CI fails because the makedepends are in community. I suggest that libbluray is moved to community since there is no package in main that depends on it (according to https://pkgs.alpinelinux.org/package/edge/main/x86_64/libbluray).

nmeum:
> I suggest that libbluray is moved to community since there is no package in main that depends on it

I agree, would be nice if you could do in in this PR because the CI build should succeed then.

TBK:
@nmeum I added a commit moving it to community.

nmeum:
@fabled you are the maintainer of libbluray, right? Are you ok with moving it to `community/`?

fabled:
The move is ok. Though, now the package has `usr/share/java/libbluray-j2se-1.0.2.jar`. I wonder if that should be split separately, or removed from .apk.

fabled:
Merged #4774 which did not have the jar issue asked to be fixed 2+ weeks ago.

PR #4599, opened 2018-06-24
testing/bfs: upgrade to 1.2.2

prspkt:
[Release Notes](https://github.com/tavianator/bfs/blob/master/RELEASES.md).

### DIFF
--- a/testing/bfs/APKBUILD
+++ b/testing/bfs/APKBUILD
@@ -1,29 +1,32 @@
 # Contributor: dai9ah <dai9ah@protonmail.com>
 # Maintainer: dai9ah <dai9ah@protonmail.com>
 pkgname=bfs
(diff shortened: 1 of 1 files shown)


andypost:
Build for ppc64 failed http://build.alpinelinux.org/buildlogs/build-edge-ppc64le/testing/bfs/bfs-1.2.2-r0.log

tavianator:
@andypost What is `getconf ARG_MAX` on ppc64?  It seems to be low enough that `-exec ... {} +` thinks it can only pass one argument at a time while the tests assume a few more can fit.  Actually the output of
```
bfs -D exec -exec echo {} + -quit
```
would be helpful.  I tried to run the Alpine ppc64le ISO under QEMU but it hung during boot.

PR #4624, opened 2018-06-25
main/alpine-baselayout: Set hvc0 for ppc64le console access

mtarsel:
On ppc64le when net booting Alpine, we will not get a login prompt from Petitboot. By default ppc64le uses hvc0 and not ttyS0. This patch starts a getty on hvc0 rather than ttyS0 in order to properly get a login prompt from Petitboot. Thanks to @jk-ozlabs and @tmh1999 for help with this patch.


mtarsel:
So it has come to my attention that not all ppc64le hardware will have a hvc0 console. Closing this PR and submitting a new one.

PR #4626, opened 2018-06-26
main/alpine-baselayout: Use the proper getty for login on ppc64le

mtarsel:
On ppc64le when net booting Alpine we will not get a login prompt from Petitboot. Despite hvc0 being the default TTY, we need to make sure getty uses the correct console for a login prompt. This patch will set TTY as `/dev/console `and the TERMTYPE as `vt100` while using a safer BAUD value in order to work on varying ppc64le hardware.

### DIFF
--- a/main/alpine-baselayout/APKBUILD
+++ b/main/alpine-baselayout/APKBUILD
@@ -195,6 +195,8 @@ package() {
 			sed -i "s/tty$i::/\#tty$i::/g" "$srcdir"/inittab
 		done
 		echo "console::respawn:/sbin/getty 38400 /dev/console" >> "$srcdir"/inittab
+	elif [ "$CARCH" = "ppc64le" ]; then
+		echo "console::respawn:/sbin/getty -L /dev/console 0 vt100" >> "$srcdir"/inittab
 	fi
 
 	install -m644 \


main/alpine-baselayout/APKBUILD
```diff
@@ -195,6 +195,8 @@ package() {
 			sed -i "s/tty$i::/\#tty$i::/g" "$srcdir"/inittab
 		done
 		echo "console::respawn:/sbin/getty 38400 /dev/console" >> "$srcdir"/inittab
+	elif [ "$CARCH" = "ppc64le" ]; then
+		echo "console::respawn:/sbin/getty -L /dev/console 0 vt100" >> "$srcdir"/inittab
```
ncopa: I wonder if we should do this in initramfs init instead. There are other systems that use hvc0, xen for example.

See https://github.com/alpinelinux/mkinitfs/blob/master/initramfs-init.in#L108

mtarsel: I have only tested ppc64le but would other architectures be ok with `console::respawn:/sbin/getty -L /dev/console 0 vt100`? Or would there need to be an `if ppc64le || xen`?

PR #4630, opened 2018-06-26
main/mariadb: upgrade to 10.3.7

TBK:
https://mariadb.com/kb/en/library/changes-improvements-in-mariadb-103/


TBK:
I will split it into several commits tomorrow after my exam.

main/mariadb/APKBUILD
```diff
@@ -28,49 +29,13 @@ source="https://downloads.mariadb.org/interstitial/mariadb-$pkgver/source/mariad
...
-#     - CVE-2017-3238
-#     - CVE-2017-3243
-#     - CVE-2017-3244
-#     - CVE-2017-3257
-#     - CVE-2017-3258
-#     - CVE-2017-3265
-#     - CVE-2017-3291
-#     - CVE-2017-3312
-#     - CVE-2017-3317
-#     - CVE-2017-3318
```
jirutka: Never ever remove `secfixes`!

main/mariadb/APKBUILD
```diff
@@ -28,49 +29,13 @@ source="https://downloads.mariadb.org/interstitial/mariadb-$pkgver/source/mariad
 	$pkgname.initd
 	fix-mysql-install-db-path.patch
 	fix-ucontext-check.patch
-	ppc-remove-glibc-dep.patch
 	pcre.cmake.patch
+	my-medium.cnf
 	"
-# dbug test fails under rootbld
-#options="!check"
-builddir="$srcdir/$pkgname-$pkgver"
```
jirutka: Please stop removing `builddir` from aports.

main/mariadb/APKBUILD
```diff
@@ -5,14 +5,15 @@
...
+pkgver=10.3.7
 pkgrel=0
 pkgdesc="A fast SQL database server"
-url="http://www.mariadb.org"
+url="https://www.mariadb.org"
 pkgusers="mysql"
 pkggroups="mysql"
 arch="all"
 license="GPL-2.0-or-later"
+#options="!check" # dbug test fails under rootbld
```
jirutka: Why do you comment out this instead of removing?

main/mariadb/APKBUILD
```diff
@@ -315,6 +320,15 @@ _server_utils() {
...
 sha512sums="7702af8787973b15784dc8a89f9d56765725a420e2e68f1924b67c88c134c7d83d3249dddd16707fa9e5f05ed747f813f16feefbb477e999a6698b1dcfb5a88f  mariadb-10.2.15.tar.gz
+_plugin_rocksdb() {
+	pkgdesc="MariaDB plugin for RocksDB (MyRocks)"
+	url="https://mariadb.com/kb/en/library/myrocks/"
+	depends="$pkgname=$pkgver-r$pkgrel"
+	mkdir -p "$subpkgdir"/usr/lib/mariadb/plugin
+	mv "$pkgdir"/usr/lib/mariadb/plugin/ha_rocksdb.so \
+		 "$subpkgdir"/usr/lib/mariadb/plugin/ha_rocksdb.so
+}
+
```
jirutka: Ehm...?

andypost:
10.3.8 released

main/mariadb/APKBUILD
```diff
@@ -315,6 +320,15 @@ _server_utils() {
 }
 
 sha512sums="7702af8787973b15784dc8a89f9d56765725a420e2e68f1924b67c88c134c7d83d3249dddd16707fa9e5f05ed747f813f16feefbb477e999a6698b1dcfb5a88f  mariadb-10.2.15.tar.gz
+_plugin_rocksdb() {
```
ncopa: the `sha512sum` line needs to be moved below the `_plugin_rocksdb()` function. It looks like it was pasted wrong location.

main/mariadb/APKBUILD
```diff
@@ -104,7 +104,7 @@ build() {
 		-DPLUGIN_MROONGA=NO \
 		-DPLUGIN_OQGRAPH=NO \
 		-DPLUGIN_PARTITION=YES \
-		-DPLUGIN_ROCKSDB=NO \
+		-DPLUGIN_ROCKSDB=YES \
```
ncopa: does rocksdb work on all arches or only on x86_64?

main/mariadb/APKBUILD
```diff
@@ -90,7 +90,7 @@ build() {
 		-DINSTALL_SUPPORTFILESDIR=share/$pkgname \
 		-DINSTALL_MYSQLSHAREDIR=share/$pkgname \
 		-DINSTALL_DOCDIR=share/doc/$pkgname \
-		-DTMPDIR=/var/tmp \
+		-DTMPDIR=/var/tmp/$pkgname \
```
ncopa: Why is `/var/tmp` wrong?

ncopa:
I fixed some of the commits and upgraded to 10.3.9 with 7fd31e497c7d3dc058c3620358f31d24e83d5007

main/mariadb/APKBUILD
TBK: @jirutka I have had maintainer to ask me to do both.

Where can I ready a "style guide" that the AL team has consensus about?

main/mariadb/APKBUILD
TBK: Same as the comment above.

main/mariadb/APKBUILD
TBK: I did not comment it out. I moved it. Not sure why it was left in from this commit https://github.com/alpinelinux/aports/commit/9f6c5c776a20c9ff350aa839795d1c2bf65fb0ac

main/mariadb/APKBUILD
TBK: @jirutka Is that an attempt at constructive feedback? Or how should it be interpeted?

main/mariadb/APKBUILD
```diff
@@ -290,6 +288,13 @@ _embedded_dev() {
 	mv "$pkgdir"/usr/lib/libmysqld.so "$subpkgdir"/usr/lib/
 }
 
+mytop() {
+	pkgdesc="Command line tool used for monitoring MariaDB performance"
+	depends="perl perl-dbi perl-dbd-mysql perl-getopt-long perl-socket perl-term-readkey"
```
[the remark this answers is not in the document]
TBK: @ncopa k, thanks for enlightening me.

main/mariadb/APKBUILD
TBK: Some docs only mentioned x86_64, but my intention was to see which builds that would fail and then make corrections.

main/mariadb/APKBUILD
TBK: I experienced an issue on a clean baremetal install with only mariadb, where mariadb (innodb) would not start. It turned out to be because caused by the mariadb user dit not have permissions to create its temp files.

I thought that I would be best to fix it by giving mariadb its own subdirectory.

TBK:
@ncopa Thanks for finishing it. I have not had the time nor the drive to work on AL stuff.
The lack of coordination,

main/mariadb/APKBUILD
jirutka: Maintainer of the mariadb aport is @ncopa. Do I understand you correctly that @ncopa told you to remove all `secfixes`? These are not just comments, but source for the security DB.

main/mariadb/APKBUILD
jirutka: I was meant as a notice for your to check what you did here, 'cause there's very obvious mistake - this code is _inside_ the value of the `sha512sums` variable, so quite obvious mistake. Unfortunately if you read it directly on the PR page and not the file page, there's not enough context to see it.

main/mariadb/APKBUILD
jirutka: I'm quite sure that maintainer of this aport did not tell you to remove `builddir=`. I have no idea why you and one or two other contributors have started removing `builddir=`.

main/mariadb/APKBUILD
ncopa: You must have had wrong permissions of `/var/tmp`. It is supposed to be writable by everybody. Must be a bug somewhere else.

jirutka: No, `/var/tmp` should not be writable for everybody, this applies only to `/tmp`! `/var/tmp/$pkgname` is correct.

jirutka: This is weird, alpine-baselayout really creates `/var/tmp` with 1777, but when I look at various Alpine installations I have, everywhere is `0755` (and I did not modify it). o.O

jirutka: I searched web and checked out few other distros, it seems that `/var/tmp` should be really world writable. But all of the Alpine systems I have, including these installed using Alpine scripts, have `/var/tmp` 0755. That's also what I remember.

I thought that the common practice for packaging is to create a subdirectory named after the package in `/var/tmp`, owned by the user the installed program run as and typically with permissions 0700. Maybe I learned this from Gentoo, don't remember.

Anyway, I'm convinced that this practice is better than letting programs write directly to `/var/tmp`. It's more secure, prevents conflicts between programs that may create the same named files/directories and avoids mess in `/var/tmp`.

ncopa: > No, /var/tmp is not and should not be writable for everybody

Do you have any documentation or anything to back up that claim? Are you sure that you are not confusing it with `/var/lib`, which should be writable by root only?

[We set it](https://git.alpinelinux.org/cgit/aports/tree/main/alpine-baselayout/APKBUILD#n110) to `1777`, similar to other distros.

According [FHS 3.0](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html) the purpose is to keep temporary files similiar to `/tmp`, but with the difference that it persist across reboots.

jirutka: I already explained it in my [last comment](https://github.com/alpinelinux/aports/pull/4630#discussion_r213964170), you probably wrote your reply before I posted it.

ncopa: > This is weird, alpine-baselayout really creates /var/tmp with 1777, but when I look at various Alpine installations I have, everywhere is 0755 (and I did not modify it). o.O

This is why I say there is a bug somwhere else.

ncopa: > you probably wrote your reply before I posted it.

I did. sorry.

let us follow up this on https://bugs.alpinelinux.org/issues/9364

PR #4631, opened 2018-06-26
testing/nagios4: new aport

vakartel:
https://www.nagios.org/
Nagios Core 4 - The Industry Standard In IT Infrastructure Monitoring


andypost:
4.4.2 released, meanwhile maybe just update package in main?

vakartel:
This one and main/nagios have very different APKBUILD, not just version bump.
So there will be many questions about many changes in one commit if I push it to main.

Any case it have no sense without applying #4633

PR #4634, opened 2018-06-26
main/apache2: move mod_cgi* to subpackage, optimize APKBUILD

vakartel:


andypost:
May have collision with https://github.com/alpinelinux/aports/pull/4798

andypost:
>  main/apache2/apache2.initd 100755 → 100644 

why this file is not executable?

PR #4657, opened 2018-07-01
main/boost: upgrade to 1.69.0

tcely:
Allows for installing `boost-libs` without the binaries `b2` and `bcp`. Installing just `boost` should result in those binaries and `boost-libs` being installed.

This allows for easier use of compiled software using `boost-dev` because a binary compiled and installed into `/usr/local/bin`, for example, can be used by just installing `boost-libs`.
For anyone unaware of the `-libs` or `-dev` subpackages, the naive solution of installing `boost` will still result in the installation of the needed runtime files too.

`1.69.0` largest subpackages:
```
>>> boost-dev*: Package size: 154.4 MB
>>> boost-static*: Package size: 73.4 MB
```

Fixes https://bugs.alpinelinux.org/issues/10162


main/boost/APKBUILD
```diff
@@ -10,35 +10,53 @@ license="custom"
...
+	timer
+	type_erasure
 	unit_test_framework
 	wave
 	wserialization
 	"
 for _lib in $_libs; do
 	subpackages="$subpackages $pkgname-$_lib:_boostlib"
 done
+subpackages="$subpackages $pkgname-dev"
```
ncopa: Why do we move the `*.a` files to the runtime libs? As far I know those are only needed for static linking.

This commit also moves the `*.so` symlinks to the runtime libs, which breaks build of current packages that uses the shared libs, unless the libs are explicitly added to makedepends. I don't think we want to move there.

If this is needed, then the commit message should explain why.

tcely:
@ncopa Moved all the static libs back into `-dev` subpackage. Also, removed the style change commit and squashed the remaining commits.

ncopa:
What is the purpose of the "full bundle" package? In what situations is that useful? abuild is supposed to autimatically pull in the needed dependencies when they are needed.

ncopa:
the commit message needs to be improved too

ncopa:
build fails for me with:

````
>>> boost-exception*: Running split function _boostlib...
mv: can't rename '/home/ncopa/aports/main/boost/pkg/boost/usr/lib/libboost_exception*': No such file or directory
>>> ERROR: boost-exception*: _boostlib failed
````

**Comment by tcely** [comment]
@ncopa The purpose of the "full bundle" package was to simplify my life when building things that aren't in aports yet but use boost. Having to figure out which of the boost subpackages are needed is painful when I just want to compile something from source.

**Comment by tcely** [comment]
The `boost-dev` package is 200+ megabytes.  Pulling out the `/usr/lib/libboost_*.a` files from `boost-dev` still makes a lot of sense to me. If you just need `libboost_log-mt.a` I'd much rather install just the `libboost_log*` files than all the static libs and headers. I know this is a bit abnormal, but just for the space savings it seems worth it to me.

I'm not sure I understand how the `makedepends` and `libboost_*.so` files interact. Is there somewhere I could look to find out more about why moving the links out of `-dev` would be an issue?

**Comment by ncopa** [comment]
The `*.so` files are normally symlinks to `*.so.$version`. Let us use `libboost_date_time.so -> libboost_date_time.so.1.67.0` as example.

`libboost_date_time.so` is only needed compile time (use linker flag _-lboost_date_time_). The `libboost_date_time.so.1.67.0` is only needed runtime.

We should ship the `libbost_date_time.so` in the `boost-dev` package and the `libboost_date_time.so.1.67.0` in the `boost-date_time` package.

When we need boost at build time, we should only need pull in `boost-dev` as makedepends. abuild will detect that `libboost_date_time.so` is a symlink pointing to a file provided by the `boost-date_time` package and will automatically add `depends=boost-date_time` to `boost-dev`.

This is how almost all `*-dev` packages work.

Now, if you move the `libboost_date_time.so` symlink to the `boost-date_time` instead (there is no reason to do so, because it is not needed runtime), then will there be nothing that makes `boost-dev` pull in the `boost-date_time` as dependency. So if anything using it will break unless you explicitly adds it to makedepends. We dont want to do that.

To conclude, here is what we should do:
- move all `*.a` files to `boost-static` subpackage to save space.
- keep the `*.so` files in `boost-dev` so abuild can manage the `boost-dev`'s dependencies
- remove the -bundle subpackage. It should not be needed when `boost-dev` has dependencies properly detected and abuild manages the runtime dependencies. (eg automatically adds `depends=so:libboost_date_time=1.67.0` when needed)

**Comment by ncopa** [comment]
also, this breaks ABI so there is a number of packages that needs to be rebuilt.

File: main/boost/APKBUILD [hunk 2]
```diff
@@ -38,89 +58,94 @@ _libs="date_time
 	wserialization
 	"
 for _lib in $_libs; do
+	echo "$_options_carch" | grep -F -q -e "--without-${_lib}" || \
```
ncopa: the `... | grep ...` here will make it fork. We don't want fork in global scope (if we do that in every APKBUILD we will have a performance issue when calculating build order).

main/boost/APKBUILD
```diff
@@ -38,89 +58,94 @@ _libs="date_time
...
 	subpackages="$subpackages $pkgname-$_lib:_boostlib"
 done
+subpackages="$subpackages $pkgname-bundle:_boost_bundle:noarch"
```
ncopa: We should not need the `-bundle` subpackage. We should only need use `boost-dev` in makedepends and abuild should automatically figure out the rest.

ncopa:
> The purpose of the "full bundle" package was to simplify my life when building things that aren't in aports yet but use boost. Having to figure out which of the boost subpackages are needed is painful when I just want to compile something from source.

You should only need `boost-dev`, which should pull in all the dependencies for you.

ncopa:
what do you think about this?
````diff
diff --git a/main/boost/APKBUILD b/main/boost/APKBUILD
index 913750a4ae..4bef49e4f5 100644
--- a/main/boost/APKBUILD
+++ b/main/boost/APKBUILD
@@ -10,7 +10,7 @@ license="custom"
 depends=""
 depends_dev="linux-headers"
 makedepends="$depends_dev python2-dev>=2.6 python3-dev flex bison bzip2-dev zlib-dev"
+subpackages="$pkgname-static $pkgname-dev $pkgname-doc"
 source="https://dl.bintray.com/boostorg/release/$pkgver/source/${pkgname}_$_pkgver.tar.bz2
        boost-1.57.0-python-abi_letters.patch
        boost-1.57.0-python-libpython_dep.patch
@@ -58,10 +58,11 @@ _libs="
        wserialization
        "
 for _lib in $_libs; do
-       echo "$_options_carch" | grep -F -q -e "--without-${_lib}" || \
+       case "$_options_carch" in
+               *$_lib*) continue;;
+       esac
        subpackages="$subpackages $pkgname-$_lib:_boostlib"
 done
-subpackages="$subpackages $pkgname-bundle:_boost_bundle:noarch"
 
 _set_options() {
        [ -n "$PY2_VERSION" ] || PY2_VERSION="$(_pyversion python2)"
@@ -148,6 +149,12 @@ package() {
                install
 }
 
+static() {
+       pkgdesc="Static libraries for boost"
+       mkdir -p "$subpkgdir"/usr/lib
+       mv "$pkgdir"/usr/lib/*.a "$subpkgdir"/usr/lib/
+}
+
 _boostlib() {
        local name="${subpkgname#$pkgname-}"
        pkgdesc="Boost $name library"
@@ -156,26 +163,10 @@ _boostlib() {
        esac
 
        mkdir -p "$subpkgdir"/usr/lib
-       # -dev might have moved all these files already
-       local _lib; for _lib in "$pkgdir"/usr/lib/libboost_$name* ; do
-               test ! -e "$_lib" || mv "$_lib" "$subpkgdir"/usr/lib/
-       done
-
+       mv "$pkgdir"/usr/lib/libboost_$name* "$subpkgdir"/usr/lib/
        rmdir "$pkgdir"/usr/lib >/dev/null 2>&1 || :
 }
 
-_boost_bundle() {
-       pkgdesc="Boost bundle meta-package"
-       depends="$pkgname"
-
-       local _pkg
-       for _pkg in ${origsubpackages% *}; do
-               depends="$depends ${_pkg%%:*}"
-       done
-
-       mkdir -p "$subpkgdir"
-}
-
 _pyversion() {
        "$(command -v "$1" || echo :)" -c 'import sys; print("%i.%i" % (sys.version_info.major, sys.version_info.minor))'
 }
````

File: main/boost/APKBUILD [hunk 4]
```diff
@@ -38,89 +58,94 @@ _libs="date_time
...
+	default_prepare
 
 	cd "$builddir"
-
-	PY2_VERSION="$(_pyversion python2)"
-	PY3_VERSION="$(_pyversion python3)"
-	abiflags=$(python3-config --abiflags)
-
+	_set_options
+	local abiflags="$(python3-config --abiflags || :)"
```
ncopa: why do you add the trailing `... || :`? I think we want catch error if the command fails instead of just silently ignore it?

main/boost/APKBUILD
```diff
@@ -131,14 +156,30 @@ _boostlib() {
...
+	for _pkg in ${origsubpackages% *}; do
+		depends="$depends ${_pkg%%:*}"
+	done
+
+	mkdir -p "$subpkgdir"
 }
 
 _pyversion() {
-	$1 -c 'import sys; print("%i.%i" % (sys.version_info.major, sys.version_info.minor))'
+	"$(command -v "$1" || echo :)" -c 'import sys; print("%i.%i" % (sys.version_info.major, sys.version_info.minor))'
```
ncopa: i dont think we want the ` ... || echo :` in there, as it will make us silently ignore errors. I think we want to know if we for some reason fail to detect python version.

tcely:
> You should only need boost-dev, which should pull in all the dependencies for you.

The problem I run into is I build the software by adding `boost-dev` but after installing the binary I don't have any notion of which subpackages the built binary actually needs. The idea was to install the `full bundle` for the runtime modules without including all of the headers and things `boost-dev` has.

tcely:
> also, this breaks ABI so there is a number of packages that needs to be rebuilt.

I realize there will be a lot of packages that need to rebuild after this change. I just figured we would want to review the boost package without all the additional APKBUILD files first.

tcely:
@ncopa I rebased the remaining changes onto your recent commit to master.

main/boost/APKBUILD
```diff
@@ -1,14 +1,14 @@
...
-pkgrel=1
+pkgrel=0
 pkgdesc="Free peer-reviewed portable C++ source libraries"
 url="http://www.boost.org/"
 arch="all"
 license="custom"
-depends=""
-depends_dev="linux-headers"
+depends="$pkgname-libs"
+depends_dev="$pkgname $pkgname-static=$pkgver-r$pkgrel linux-headers"
```
ncopa: why the hard dependency on the `boost-static` package? If it really is a hard dependency (eg. not optional) then we should just include the static libs in the -dev package.

If it is optional (and you can build with only dynamic libs) then the `boost-dev` package should not depend on `boost-static`

main/boost/APKBUILD
```diff
@@ -45,27 +45,35 @@ _libs="
...
 	wave
 	wserialization
 	"
 for _lib in $_libs; do
-	case "$_options_carch" in
-		*$_lib*) continue;;
-	esac
+	for _option in $_options_carch; do
+		case "$_option" in --without-${_lib}) continue ;; esac
+	done
```
ncopa: why is this needed? What is broken with how it was?

main/boost/APKBUILD
```diff
@@ -153,28 +161,42 @@ package() {
 }
 
 static() {
-	pkgdesc="Static libraries for boost"
+	pkgdesc="Boost static libraries"
+	depends="$depends_static"
```
ncopa: where is `depends_static` defined? if it not used other place, then we can remove this line.

main/boost/APKBUILD
```diff
@@ -153,28 +161,42 @@ package() {
...
+
 	mkdir -p "$subpkgdir"/usr/lib
-	mv "$pkgdir"/usr/lib/*.a "$subpkgdir"/usr/lib/
+	mv "$pkgdir"/usr/lib/lib*.a "$subpkgdir"/usr/lib/
```
ncopa: are there any `*.a` files that does not start with `lib`? If so, what are those for and why should they be excluded from the `boost-static`?

andypost:
> after installing the binary I don't have any notion of which subpackages the built binary actually needs.

Meanwhile https://bugs.alpinelinux.org/issues/9245 shows that after build all dependencies listed but leads to specific "sub-libs"

main/boost/APKBUILD
tcely: That's fair. I'll remove that dependency.

main/boost/APKBUILD
tcely: It's not currently defined. However, we can't remove this line without inheriting the depends for `boost` when `boost-static` is created.

main/boost/APKBUILD
tcely: There aren't any. I really don't think there will be any installed into `/usr/lib` but I like being more specific here with the files we are trying to move.

tcely:
@andypost I'm unsure of what I'm supposed to understand from the quoted text and linked issue.

main/boost/APKBUILD
tcely: Dropping the `--without-` part is the main problem.

tcely:
@ncopa Ok, so the latest version is building cleanly and the split functions look reasonable to me. Is there anything left before we merge this and rebuild the many things that depend on boost?

tcely:
`boost-1.70.0` is scheduled for release `201900410`.

ncopa:
i think we may need build boost with `--layout=system`

tcely:
> i think we may need build boost with `--layout=system`

I looked into that. It's a bad idea. We are building different versions (multi/single threaded) that should be distinguished.

There appears to be versioning on the `tagged` layout we have been using, if we really wanted to we could drop back to the `1.66` version of this layout that does not include architecture.

ncopa:
> > i think we may need build boost with `--layout=system`
> 
> I looked into that. It's a bad idea. We are building different versions (multi/single threaded) that should be distinguished.

Do we really need single threaded nowdays? I think we can build multithreaded only. It would save space too.

main/boost/APKBUILD
```diff
@@ -46,27 +46,38 @@ _libs="
...
-done
+	_skip_lib=0
+	for _option in $_options_carch; do
+		case "$_option" in --without-${_lib}) _skip_lib=1; break ;; esac
+	done; unset -v _option
+	if [ ${_skip_lib:-0} -eq 0 ]; then
+		subpackages="$subpackages ${pkgname}-${_lib}:_boostlib"
+	fi
+done; unset -v _lib _skip_lib
+subpackages="$subpackages $pkgname-libs"
```
ncopa: I wonder what this change is supposed to solve in the first place? If you are afraid of overlapping matches, then why not do something like:
````sh
case "$_options_carch " in # the trailing space is intentionally there so we can match last word
    *"--without-$_lib "*) continue;;
esac
````

**Comment by ncopa** [comment]
> > You should only need boost-dev, which should pull in all the dependencies for you.
> 
> The problem I run into is I build the software by adding `boost-dev` but after installing the binary I don't have any notion of which subpackages the built binary actually needs. The idea was to install the `full bundle` for the runtime modules without including all of the headers and things `boost-dev` has.

A better solution than just include them all is to do something like:
````sh
my_binaries=/usr/local
apk add --virtual .my-runtime-deps  \
    $(scanelf -R --needed $my_binaries | awk '{ gsub(/\,/, " so:", $2); print "so:"$2}')
````

**Comment by tcely** [comment]
> > > i think we may need build boost with `--layout=system`
> > 
> > 
> > I looked into that. It's a bad idea. We are building different versions (multi/single threaded) that should be distinguished.
> 
> Do we really need single threaded nowdays? I think we can build multithreaded only. It would save space too.

We can try it to find out. I just would rather not do that now.

File: main/boost/APKBUILD (hunk 10 above)
Comment (tcely) [reply main/boost/APKBUILD]: I don't see any problem with using that kind of case instead.

Comment (tcely) [reply main/boost/APKBUILD]: @ncopa Have a look at https://github.com/alpinelinux/aports/pull/4657/commits/658c6b24d67a901147dda84e9b967d494f09b606 and let me know which you prefer to use.

Comment (tcely) [reply main/boost/APKBUILD]: The point of the change was to handle `_options_carch` containing new lines like we see in `_libs`. Preventing `--without-log` from matching for `--without-log_setup` is also important.

**Comment by tcely** [comment]
> A better solution than just include them all is to do something like:
> 
> ```shell
> my_binaries=/usr/local
> apk add --virtual .my-runtime-deps  \
>     $(scanelf -R --needed $my_binaries | awk '{ gsub(/,/, " so:", $2); print "so:"$2}')
> ```

Thanks! I think I'll use this improvement the next time I hit this problem:
```
apk add --virtual .scanned-runtime-deps \
    $(scanelf -R --needed $prefix | \
        awk '$2 ~ /^lib/ { gsub(/,/, "\nso:", $2); printf "so:%s\n", $2;}' | sort -u)
```

File: main/boost/APKBUILD [hunk 11]
```diff
@@ -154,28 +165,42 @@ package() {
...
+
+	pkgdesc="Boost shared libraries"
+	depends="$depends_libs"
+	local _pkg; for _pkg in $origsubpackages; do
+		case "$_pkg" in
+			*:_boostlib) depends="$depends ${_pkg%%:*}" ;;
+		esac
+	done
+
+	mkdir -p "$subpkgdir"
```
ncopa: I still think the `boost-libs` can be removed. Do you really need it or can you use the `apk add --virtual ...` trick?

I want keep things simple if possible.

main/boost/APKBUILD
```diff
@@ -71,14 +71,12 @@ _libs="
...
-	_skip_lib=0
-	for _option in $_options_carch; do
-		case "$_option" in --without-${_lib}) _skip_lib=1; break ;; esac
-	done; unset -v _option
-	if [ ${_skip_lib:-0} -eq 0 ]; then
-		subpackages="$subpackages ${pkgname}-${_lib}:_boostlib"
-	fi
-done; unset -v _lib _skip_lib
+	# the trailing space in the below case is intentional so that we can match the last word
+	case "${_options_carch//$'\n'/ } " in
```
ncopa: I have doubts about `"${_options_carch//$'\n'/ } "`. It looks like it solves a problem in case `_options_carch` has new lines. It currently does not have newlines, and it will hopefully never need it either, since we want keep build options same across architectures.

It also adds bashism. Even if we tolerate this kind of variable substitution, I want to keep bashisms to a minimal, in case we decide to be strictly POSIX shell.

And finally, does it really work in busybox ash? I will have to verify that it actually works before I can merge it.

So my thinking here is: keep it simple. If we ever will need deal with newlines in `_options_carch` then lets deal with it at that time.

tcely: The for loop version is more flexible in handling spaces, tabs and newline characters. You could use a function to print a newline instead and/or call out to see/awk/tr to format the case argument instead.

You are very welcome to double-check that BusyBox sh works with this case statement.

I would very much like to avoid breaking in unexpected ways when the value of `_oprions_carch` is split across lines.

main/boost/APKBUILD
tcely: I don't understand the reason for removing `boost-libs` when it gives good default behavior for the users at very little cost. There may be an argument for removing the dependency on `boost-libs` from `boost`. However, I do rather like the idea that the user doesn't need to know anything about how we chose to split this package to install a working set of runtime files.

The `scanelf` trick is useful, but encoding the libs list in `boost-libs` is the right choice in my opinion.

tcely:
Rebased on top of commits @ncopa merged to `master`.

PR #4667, opened 2018-07-03
testing/haxe: move to community

andyli:
haxe has been in testing for some time and it is quite stable. I will take care of maintaining it in the long run.


andyli:
Is there any problem or question about this?

PR #4671, opened 2018-07-03
testing/frr: new aport for free range routing, a fork of quagga

ajones-rvbd:
FRRouting is free software that implements and manages various
IPv4 and IPv6 routing protocols.

Currently FRRouting supports BGP4, BGP4+, OSPFv2, OSPFv3, RIPv1, RIPv2,
RIPng, IS-IS, PIM-SM/MSDP, LDP and Babel as well as very early support
for EIGRP and NHRP.

Issue: https://bugs.alpinelinux.org/issues/8795

### DIFF
--- a/testing/frr/APKBUILD
+++ b/testing/frr/APKBUILD
@@ -1,0 +1,64 @@
+# Maintainer: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=frr
+pkgver=5.0.1
+pkgrel=0
+pkgdesc="FRRouting is a fork of quagga"
+pkgusers="frr"
+pkggroups="frr"
+url="https://frrouting.org/"
+arch="x86_64"
+license="GPL-2.0-only and LGPL-2.1-only"
+depends="json-c c-ares iproute2 python2 py-ipaddr bash"
+makedepends="ncurses-dev net-snmp-dev gawk texinfo perl
+    acct autoconf automake bash bison bsd-compat-headers
+    c-ares-dev cryptsetup-libs device-mapper-libs flex
+    json-c-dev kmod lddtree libacl libblkid libburn libedit
+    libisoburn libisofs libltdl libtool libuuid linux-headers lzo m4 mkinitfs
+    mtools pcre perl python2-dev readline-dev
+    squashfs-tools texinfo xorriso py-sphinx"
+checkdepends="pytest py-setuptools"
+install="$pkgname.pre-install"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-dbg $pkgname-openrc"
+source="https://github.com/FRRouting/frr/archive/$pkgname-$pkgver.tar.gz
+    make-check-bypass-isisd-tlv.patch
+    frr.initd
+    "
+builddir="$srcdir"/$pkgname-$pkgname-$pkgver
+
+_sysconfdir=/etc/frr
+
+build() {
+	cd "$builddir"
+	./bootstrap.sh
+	./configure \
+		--prefix=/usr \
+		--sbindir=/usr/lib/frr \
+		--sysconfdir=$_sysconfdir \
+		--libdir=/usr/lib \
+		--localstatedir=/run/frr \
+		--enable-systemd=no \
+		--enable-vtysh \
+		--enable-multipath=64 \
+		--enable-vty-group=frrvty \
+		--enable-user=${pkgusers%% *} \
+		--enable-group=${pkggroups%% *}
+	make
+}
+
+check() {
+	cd "$builddir"
+
+	make -j 1 check
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install
+
+	install -Dm644 "$builddir"/tools/etc/frr/daemons "$pkgdir"$_sysconfdir
+	install -Dm644 "$builddir"/tools/etc/frr/daemons.conf "$pkgdir"$_sysconfdir
+	install -Dm755 "$srcdir"/frr.initd "$pkgdir"/etc/init.d/frr.initd
+}
+sha512sums="e2a3a4e21998f8c85b59fc971190d0b5340bbcd88356a637740ad8e564275b0df14da78b72a24a99a58435725cded1b9089814fc926a7a4080bd3a8762973548  frr-5.0.1.tar.gz
+3c679b857018d82443abd34823e712a8657a597b4210730d260e9796516982a137234354422cb439989985367531d600e205d3fb791cdb7b212425bb704c2016  make-check-bypass-isisd-tlv.patch
+9d2cfa95c3a79fa0da321319dcd6f439a267b69071233d4ad4b56f770eb9852c3c8b3be24c1e396701509a09eabb6ba618af6fa4a0a643bd37eeee7b6496a787  frr.initd"
--- a/testing/frr/frr.initd
+++ b/testing/frr/frr.initd
@@ -1,0 +1,19 @@
+#!/sbin/openrc-run
+
+description="FRRouting"
+
+start() {
+        /usr/lib/frr/frr start
+}
+
+stop() {
+        /usr/lib/frr/frr stop
+}
+
+restart() {
+        /usr/lib/frr/frr restart
+}
+
+status() {
+        /usr/lib/frr/frr status
+}
--- a/testing/frr/frr.pre-install
+++ b/testing/frr/frr.pre-install
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+for g in frr frrvty; do
+	addgroup -S $g 2>/dev/null
+done
+
+adduser -S -D -H -h /var/run/frr -s /sbin/nologin -G frr -g frr frr 2>/dev/null
+adduser frr frrvty 2>/dev/null
+
+exit 0
--- a/testing/frr/make-check-bypass-isisd-tlv.patch
+++ b/testing/frr/make-check-bypass-isisd-tlv.patch
@@ -1,0 +1,49 @@
+commit f404be7c728ca3433bb4aeaeac5e3c450941de35
+Author: Arthur Jones <arthur.jones@riverbed.com>
+Date:   Wed Jun 6 07:47:17 2018 -0700
+
+    tests/isisd: bypass test_fuzz_isis_tlv when inet_ntop is broken
+    
+    On Alpine Linux edge, musl does not seem to be RFC 5952 4.2.2
+    compliant (how to print a single :0: in the IPv6 address).  Let's
+    skip that test, as we get false negatives when running against
+    that version of musl.
+    
+    Credit for the idea for the fix and how to fix it is due to
+    chris@opensourcerouting.org.
+    
+    Testing done:
+    
+    make check on alpine linux passes now
+    
+    Issue: https://github.com/FRRouting/frr/issues/2375
+    Signed-off-by: Arthur Jones <arthur.jones@riverbed.com>
+
+diff --git a/tests/isisd/test_fuzz_isis_tlv.py b/tests/isisd/test_fuzz_isis_tlv.py
+index bf700bfee..d96e3c4fe 100644
+--- a/tests/isisd/test_fuzz_isis_tlv.py
++++ b/tests/isisd/test_fuzz_isis_tlv.py
+@@ -2,10 +2,21 @@ import frrtest
+ 
+ import pytest
+ import platform
++import socket
+ 
+-if platform.uname()[0] == 'SunOS':
++
++##
++# on musl, ntop compresses a single :0: -> :: which is against RFC
++##
++def inet_ntop_broken():
++    addr = '1:2:3:4:0:6:7:8'
++    return socket.inet_ntop(socket.AF_INET6,
++                            socket.inet_pton(socket.AF_INET6, addr)) != addr
++
++
++if platform.uname()[0] == 'SunOS' or inet_ntop_broken():
+     class TestFuzzIsisTLV:
+-        @pytest.mark.skipif(True, reason='Test unsupported on SunOS')
++        @pytest.mark.skipif(True, reason='Test unsupported')
+         def test_exit_cleanly(self):
+             pass
+ else:


testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
+# Contributor: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=frr
+pkgver=5.0
+pkgrel=0
+pkgdesc="Free Range Routing is a fork of quagga"
+url="https://frrouting.org/"
+arch="x86_64"
+license="GPL-2.0"
+depends="json-c c-ares ipsec-tools iproute2 python py-ipaddr bash"
```
TBK: Add the following instead of using `_user=frr`
```
pkgsusers="frr"
pkggroups="frr"
```

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
+# Contributor: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=frr
+pkgver=5.0
+pkgrel=0
+pkgdesc="Free Range Routing is a fork of quagga"
+url="https://frrouting.org/"
+arch="x86_64"
+license="GPL-2.0"
```
TBK: License format is incorrect. Check https://spdx.org/licenses/

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
+# Contributor: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=frr
+pkgver=5.0
+pkgrel=0
+pkgdesc="Free Range Routing is a fork of quagga"
+url="https://frrouting.org/"
+arch="x86_64"
```
TBK: Is it really limited to x86_64? I am asking because we are currently building quagga for a lot more archs - https://pkgs.alpinelinux.org/packages?name=quagga&branch=edge

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
+# Contributor: Arthur Jones <arthur.jones@riverbed.com>
```
TBK: Would you be willing to be the maintainer?

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+		--prefix=/usr \
+		--sbindir=$_sbindir \
+		--sysconfdir=$_sysconfdir \
+		--libdir=$_libdir \
+		--localstatedir=$_localstatedir \
+		--enable-systemd=no \
+		--enable-vtysh \
+		--enable-multipath=64 \
+		--enable-vty-group=frrvty \
+		--enable-user=$_user \
```
TBK: Change:
```
--enable-user=$_user \
--enable-group=$_user
```

To e.g. the following (it will only grap the first word of the string in case more are added in the future):
```
--enable-user=${pkgusers%% *} \
--enable-group=${pkggroups%% *}
```

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+		--enable-group=$_user
+	make
+}
+
+check() {
+	cd "$builddir"
+
+	make -j 1 check
```
TBK: Are the tests flaky since it is limited to one job?

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+    libgomp libisoburn libisofs libltdl libressl libssh2
+    libstdc++ libtool libuuid linux-headers lzip lzo m4 make mkinitfs mpc1
+    mpfr3 mtools musl-dev ncurses-libs ncurses-terminfo ncurses-terminfo-base
+    patch pax-utils pcre perl pkgconf python2 python2-dev readline
+    readline-dev sqlite-libs squashfs-tools sudo tar texinfo xorriso xz-libs
+    py-sphinx"
+checkdepends="pytest py-setuptools"
+install="$pkgname.pre-install $pkgname.pre-deinstall $pkgname.post-deinstall"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-dbg $pkgname-openrc"
+source="https://github.com/FRRouting/frr/archive/$pkgname-$pkgver.tar.gz docker-start make-check-bypass-isisd-tlv.patch tests-isisd-built-sources.patch"
```
TBK: It needs to be styled:
```
source="https://github.com/FRRouting/frr/archive/$pkgname-$pkgver.tar.gz
	docker-start
	make-check-bypass-isisd-tlv.patch
	tests-isisd-built-sources.patch
	"
```

testing/frr/APKBUILD
```diff
@@ -0,0 +1,73 @@
+# Contributor: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=frr
+pkgver=5.0
+pkgrel=0
+pkgdesc="Free Range Routing is a fork of quagga"
+url="https://frrouting.org/"
+arch="x86_64"
+license="GPL-2.0"
+depends="json-c c-ares ipsec-tools iproute2 python py-ipaddr bash"
+makedepends="ncurses-dev net-snmp-dev gawk texinfo perl
```
TBK: The makedepends needs a cleanup. Just to mention some of the redundant dependencie (already part of the build enviroment or unneeded): sudo tar perl patch git curl

testing/frr/docker-start
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
TBK: Not sure if this fiel belongs in the aport.

testing/frr/APKBUILD
ajones-rvbd: Thanks, will do...

testing/frr/APKBUILD
ajones-rvbd: Yep, will fix...

testing/frr/APKBUILD
ajones-rvbd: I don't think it's limited to x86_64, it should support all architectures, but I only have x86_64 setup for testing.

testing/frr/APKBUILD
ajones-rvbd: Yes.  Will fix.

testing/frr/APKBUILD
ajones-rvbd: Thank you!  Will fix...

testing/frr/APKBUILD
ajones-rvbd: Yep.  Unit tests fail about once every 10x unless -j 1...

testing/frr/APKBUILD
ajones-rvbd: Thank you, will fix...

testing/frr/APKBUILD
ajones-rvbd: Sure...

testing/frr/docker-start
ajones-rvbd: OK, I'll take it out...  Thanks for all your help here, I'll fix, test and re-push today...

ajones-rvbd:
I think all comments are addressed in this push.  In the meantime, frr-5.0.1 was released, so I updated to that, which got rid of one patch (5.0.2, if it comes, will remove the other one)...

ajones-rvbd:
re-base and incorporated an openrc fix from Hidde Van der Heide (thank you!)...

ajones-rvbd:
Removed ipsec-tools with last push, all comments should be addressed...

ncopa:
This was reverted by @jirutka with dc261cc04036eb515feb5963a3a200e121fc589d so I re-open.

Comments to 6c21984c5b86ca03c72a88927f8c1de0c420e25f explains what needs to be fixed. I apologize those comments was not given to this PR before it was merged.

ajones-rvbd:
No problem, I'll fix up based on comments and re-submit, really appreciate your tact on this one @ncopa

ajones-rvbd:
Looks like CI failed, but I'm not sure if it is related to the patch...

ncopa:
> frr: FAILED
> sha512sum: WARNING: 1 of 1 computed checksums did NOT match

The CI failed because the `frr` file was modified but checksum in APKBUILD was not updated. You need to run `abuild checksum` before you commit.

testing/frr/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+    json-c-dev kmod lddtree libacl libblkid libburn libedit
+    libisoburn libisofs libltdl libtool libuuid linux-headers lzo m4 mkinitfs
+    mtools pcre perl python2-dev readline-dev
+    squashfs-tools texinfo xorriso py-sphinx"
+checkdepends="pytest py-setuptools"
+install="$pkgname.pre-install"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-dbg $pkgname-openrc"
+source="https://github.com/FRRouting/frr/archive/$pkgname-$pkgver.tar.gz
+    make-check-bypass-isisd-tlv.patch
+    frr
```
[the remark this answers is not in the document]
ajones-rvbd: Will do, thanks...

testing/frr/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+
+build() {
+	cd "$builddir"
+	./bootstrap.sh
+	./configure \
+		--prefix=/usr \
+		--sbindir=/usr/lib/frr \
+		--sysconfdir=$_sysconfdir \
+		--libdir=/usr/lib \
+		--localstatedir=/var/run/frr \
```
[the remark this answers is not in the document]
ajones-rvbd: Cool, thanks, it should be /run then, I'll fixup...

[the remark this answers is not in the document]
ajones-rvbd: Latest push uses /run/frr and it seems to work fine, I think everything is addressed now...

testing/frr/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+pkgname=frr
+pkgver=5.0.1
+pkgrel=0
+pkgdesc="FRRouting is a fork of quagga"
+pkgusers="frr"
+pkggroups="frr"
+url="https://frrouting.org/"
+arch="x86_64"
+license="GPL-2.0-only and LGPL-2.1-only"
+depends="json-c c-ares iproute2 python2 py-ipaddr bash"
```
ncopa: `json-c` and `c-ares` should not be needed in `depends` as abuild will automatically add shared libraries to depends.

Does `frr` support python3 instead of python2? I want avoid adding new packages that depends on python2 it its not necessary.

`py-ipaddr` should be versioned. eg. `py2-ipaddr` or `py3-ipaddr`.

testing/frr/APKBUILD
```diff
@@ -0,0 +1,64 @@
...
+makedepends="ncurses-dev net-snmp-dev gawk texinfo perl
+    acct autoconf automake bash bison bsd-compat-headers
+    c-ares-dev cryptsetup-libs device-mapper-libs flex
+    json-c-dev kmod lddtree libacl libblkid libburn libedit
+    libisoburn libisofs libltdl libtool libuuid linux-headers lzo m4 mkinitfs
+    mtools pcre perl python2-dev readline-dev
+    squashfs-tools texinfo xorriso py-sphinx"
```
ncopa: are all those packages really needed to build this? I doubt that those are really needed:
- acct
- cryptsetup-libs
- device-mapper-libs
- kmod
- lddtree
- libacl
- libblkid
- libburn
- libedit
- libisoburn
- libisofs
- libltdl
- libuuid
- lzo
- m4
- mkinitfs
- mtools
- pcre
- squashfs-tools
- texinfo
- xorriso

Many of those are runtime libraries. some are tools needed to build alpine iso release image and simply does not make sense as makedepends for a routing software package.

This makes me question if the other makedependencies are needed too. It seems like makedepends was constructed by whatever happened to be installed on the build machine.

perl is specified twice.

testing/frr/frr.pre-install
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+for g in frr frrvty; do
+	addgroup -S $g 2>/dev/null
+done
+
+adduser -S -D -H -h /var/run/frr -s /sbin/nologin -G frr -g frr frr 2>/dev/null
```
ncopa: I wonder if it would make more sense to use `/var/empty` as home directory. `/var/run` is a symlink to `/run` which is mounted as tmpfs, which means that the subdirectory `/run/frr` will be removed on reboot.

ncopa:
A few more comments.

I haven't verified it but when I looked in configure.ac it looks like it supports python3.

python seems to only be needed by frr-reload:
````sh
#!/bin/sh

if test -e /usr/lib/frr/frr-reload.py; then
   exec /usr/lib/frr/frr-reload.py --reload /etc/frr/frr.conf
fi
>&2 echo "Please install frr-pythontools package. Required for reload"
exit 1
````
It looks like we could get rid of python dependency by moving this to a `frr-pythontools` package.

Bash is only needed by `/usr/lib/frr/frr` which seems to be a runit script. If we can port it to a proper openrc script then we could get rid of the bash depedency.

**Comment by ncopa** [comment]
I'm gonna push it with a few dependency improvements, so we can test how it works.

Thanks!

File: testing/frr/frr.initd [hunk 15]
```diff
@@ -0,0 +1,19 @@
+#!/sbin/openrc-run
+
+description="FRRouting"
+
+start() {
+        /usr/lib/frr/frr start
```
jirutka: The main executable in `/usr/lib`?!

testing/frr/APKBUILD
ajones-rvbd: I'm not sure about python2 vs python3, right now, it seems like other distro packages are using python2.  The python code needs to be cleaned up upstream (using console scripts, modules, etc) I was hoping to do the python2 and python3 testing once I get those changes upstream.  I do want to go w/ python3.  frr-reload is also due to be replaced upstream, so hoping not to put too much work into that...

Will test/fix json-c c-ares out of depends and py-ipaddr...

Dependencies were cut and paste from other distro packages and docs, @tbk also pointed that out and I trimmed it down, but it looks like I should come at it from the other side and just add in dependencies one by one...

testing/frr/APKBUILD
ajones-rvbd: Will fix perl, makedepends are cut&paste, I'll start fresh and only add in what's necessary, that should clean it up...

testing/frr/frr.pre-install
ajones-rvbd: Right, these are socket and vty files, they should be on tmpfs, nothing is preserved across reboots there...

testing/frr/frr.initd
ajones-rvbd: The user-facing executable is in /usr/bin/vtysh, these are system daemons, docs and other distros have them in /usr/lib/frr.  In frr, the user does not directly interact w/ the routing daemons...

/usr/lib/frr/frr is just a startup shell script.  Upstream didn't want yet another startup script, so we could re-write it and put it in aports, but that seems like a lot of work, esp since I'm not sure of what benefits we'd get back for that...

ajones-rvbd:
I did plan add a python sub-package pretty much as you describe @ncopa, but was hoping I could do that after the python cleanup or, even better, after frr-reload functionality was completely replaced with watchfrr -- this is planned work upstream, but no one is working on it yet...

Removing python drops the base image size down dramatically, it is definitely a big space win...

ajones-rvbd:
Created https://bugs.alpinelinux.org/issues/9172 to track these issues, let me know if I missed anything there.  Thanks all for walking me through this, I'm learning a lot about alpine packaging...

PR #4673, opened 2018-07-04
Add support for singed modloop [WIP]

clandmeter:
For netboot we are unable to verify the modloop (ipxe support signed kernel and initramfs only).

ref: https://github.com/alpinelinux/mkinitfs/pull/31
ref: https://github.com/alpinelinux/alpine-conf/pull/18


clandmeter:
Im not sure about the correct location of the signature. I choose /var/cache/misc as its an existing dir and /tmp sounded too temporary.
Its currently enabled for base profile and should probably be moved to netboot only except if we want to support signed modloop by default.

ncopa:
the modloop itself is in `/boot/`. I think it makes sense to have the signature file there too.

clandmeter:
The location of the modloop differs on boot method. Netboot it wil be in /lib, but on other media it could be a diff location and on iso/cdrom it would be readonly.
So I would opt for a default location like /var/cache/misc.

ncopa:
Ok. lets do that.

PR #4676, opened 2018-07-05
testing/boringssl: new aport

bratkartoffel:
This aport adds boringssl, the SSL implementation by google, to the repository. This adds the possibility to build and statically link other packages against their implementation.

This aport includes two packages:
- boringssl: Contains only the "bssl" executable, similar to the "openssl" command
- boringssl-dev: This contains the headers and the static libs

The headers are installed unter "/usr/include/boringssl" so it doesn't collide with openssl / libressl. The libs will be installed under "/usr/libs/boringssl" for the same reason.
This means that any depending package has to add these two paths to their CFLAGS in order to compile.

### DIFF
--- a/testing/boringssl/APKBUILD
+++ b/testing/boringssl/APKBUILD
@@ -1,0 +1,57 @@
+# Maintainer: Simon Frankenberger <simon-alpine@fraho.eu>
+#
+_gitver=d2f87a777955dd53d7768b236ec3804fe50462a7
+pkgname=boringssl
+pkgver=2018.07.04
+_namever=${pkgname}${pkgver%.*}
+pkgrel=0
+pkgdesc="BoringSSL is a fork of OpenSSL that is designed to meet Google's needs."
+url="https://boringssl.googlesource.com/boringssl/"
+arch="all"
+license="custom"
+depends=""
+makedepends="cmake perl go ninja"
+replaces=
+subpackages="$pkgname-dev"
+source="$pkgname-$_gitver.tar.gz::https://github.com/google/boringssl/archive/$_gitver.tar.gz"
+builddir="$srcdir/$pkgname-$_gitver"
+ldpath="/usr/lib/boringssl"
+
+prepare() {
+	default_prepare
+	cd "$builddir"
+}
+
+build() {
+	cd "$builddir"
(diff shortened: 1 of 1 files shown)


nmeum:
From the boringssl README:

> We don't recommend that third parties depend upon it. Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.

If there is no ABI stability is there any use in packaging it?

The README also states that:

> Programs ship their own copies of BoringSSL when they use it and we update everything as needed when deciding to make API changes.

If programs depending on BoringSSL usually vendor it I don't see any need to package it.

bratkartoffel:
> If programs depending on BoringSSL usually vendor it I don't see any need to package it.
I packaged this library because nghttp2 is compatible with boringssl, but they don't ship it. That's why I created another PR with the nghttp2 using boringssl.

> If there is no ABI stability is there any use in packaging it?
This package only contains the static libs. This way all packages depending on this one will be statically linked. I made this package in mind that it must not collide with any libressl / openssl installed and to ease the adoption of other packages with boringssl.
Anyways, any update to this library needs a recompile of every depending package. I don't know if that can be done in an easy way.

I'm unsure now if it wouldn't be "better" to update the nghttp2-boringssl PR to download and compile boringssl on the fly?

nmeum:
> I packaged this library because nghttp2 is compatible with boringssl, but they don't ship it.

Well…according to the boringssl readme this seems to be a bad idea.

> This way all packages depending on this one will be statically linked

I personally have two issues with this approach:

1. We usually don't link statically against libraries
2. If we rebuild packages against a new boringssl version as part of a release we will run into problems if the boringssl api has changed since the last time the packages have been linked against it.

> I'm unsure now if it wouldn't be "better" to update the nghttp2-boringssl PR to download and compile boringssl on the fly?

Maybe. It would at the very least be good to know why upstream decided not to vendor boringssl even though the readme explicitly recommends doing that.

jirutka:
As @nmeum said, upstream explicitly discourage from third parties to depend on BoringSSL, hence there's no reason to package it.

It's extremely bad idea to provide packages statically linked against notoriously vulnerable library (BoringSSL is just Google's own for of OpenSSL).

PR #4677, opened 2018-07-05
testing/nghttp2-boringssl: new aport

bratkartoffel:
This aport adds the nghttp2 packages compiled and linked against boringssl.
In order for this to work, the testing/boringssl PR  #4676 has to be merged.


jirutka:
Why do you need this? We already provide nghttp2 linked against system-provided SSL/TLS library.

bratkartoffel:
I agree that we don't need this in alpine after thinking about it the last days.
Thanks for your time to look into this.

PR #4682, opened 2018-07-05
testing/bfs: fix broken tests on ppc64le because of 64k pagesize

mksully22:
test_exec_plus test fails because sysconf(_SC_ARG_MAX) is not large enough to contain two 64K pages (ppc64le default page size). Since maximum argument size allowed by BFS is only 16K anyway (BFS_EXEC_ARG_MAX) then update arg size check to make sure it is at least 32k.

### DIFF
--- a/testing/bfs/APKBUILD
+++ b/testing/bfs/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: dai9ah <dai9ah@protonmail.com>
 pkgname=bfs
 pkgver=1.2.2
-pkgrel=0
+pkgrel=1
 pkgdesc="Breadth-first variant of the UNIX find command"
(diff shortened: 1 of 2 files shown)


rgdoliveira:
@mksully22 question: was this patch merged (or sent) upstream?

mksully22:
I issued https://github.com/tavianator/bfs/pull/34 with this patch to the
bfs project as well

tavianator:
Ultimately I went with this fix: https://github.com/tavianator/bfs/commit/30b48746565be758f7014dce9448d8e91febab60.  You can cherry pick it if you want, or just wait for the next release.

mksully22:
Thanks, I will cherry pick your fix and resubmit to alpine.

PR #4683, opened 2018-07-05
main/openvpn: fix init script

wgorski1:
Single quotes were disabling parameter expansion in the regex


jirutka:
Thanks for noticing it, this was very stupid mistake from me. :crying_cat_face:

PR #4693, opened 2018-07-08
main/duplicity: fix dependencies and add check (including missing deps)

ganwell:
I added the dependencies directly to main, because duplicity is currently broken. It would also be nice to backport this to 3.8-stable. (Do I have to do an extra PR?)

### DIFF
--- a/main/py2-fasteners/APKBUILD
+++ b/main/py2-fasteners/APKBUILD
@@ -1,0 +1,30 @@
+# Contributor: Jean-Louis Fuchs <ganwell@fangorn.ch>
+# Maintainer: Jean-Louis Fuchs <ganwell@fangorn.ch>
+pkgname=py2-fasteners
+pkgver=0.14.1
+pkgrel=0
+pkgdesc="A python package that provides useful locks"
+url="https://github.com/harlowja/fasteners/"
+arch="noarch"
+license="Apache-2.0"
+depends="python2 py2-monotonic py2-six"
+makedepends="python2-dev py-setuptools"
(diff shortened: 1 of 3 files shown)


ncopa:
> (Do I have to do an extra PR?)

No, I can handle it, thanks.

The commit message for `py2-fasteners` was wrong and It would have been nice to have it mentioned in commit message why the packages was added directly to main instead of testing.

main/py2-fasteners/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+pkgver=0.14.1
+pkgrel=0
+pkgdesc="A python package that provides useful locks"
+url="https://github.com/harlowja/fasteners/"
+arch="noarch"
+license="Apache-2.0"
+depends="python2 py2-monotonic py2-six"
+makedepends="python2-dev py-setuptools"
+source="fasteners-$pkgver.tar.gz::https://github.com/harlowja/fasteners/archive/$pkgver.tar.gz"
+sha512sums="954d692b7f43563ba1413854c8dec2f5ff98757346c55aaa5f39e9aa777fb746bf4f9506beab6cfa149d5646db4b3de58d35bcef95e3128096c3346c9fd8015c  fasteners-0.14.1.tar.gz"
```
jirutka: Checksums must be always at the end of the APKBUILD. This is a bug in `newapkbuild` script that it adds it in the middle.

ganwell: @jirutka I indeed created the APKBUILD with newapkbuild and I didn't pay attention. Next time I will call abuild checksum afterwards to move them to the end.

PR #4695, opened 2018-07-09
testing/wireguard: upgrade to 0.0.20180708

zx2c4:
```
  * device: print daddr not saddr in missing peer error
  * receive: style
  
  Debug messages now make sense again.
  
  * selftest: ratelimiter: improve chance of success via retry
  * qemu: bump default kernel version
  * qemu: decide debug kernel based on KERNEL_VERSION
  
  Some improvements to our testing infrastructure.
  
  * receive: use NAPI on the receive path
  
  This is a big change that should both improve preemption latency (by not
  disabling it unconditionally) and vastly improve rx performance on most
  systems by using NAPI. The main purpose of this snapshot is to test out this
  technique.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 
 pkgname=wireguard-tools
-pkgver=0.0.20180625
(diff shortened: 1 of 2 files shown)


ncopa:
fixed with 2999aad559a9807f193180c2de7dd52ce6654463

PR #4712, opened 2018-07-11
testing/sane and testing/hplip: Support all architectures, package python part

michitux:
This adds support for all architectures to sane by removing parallel port support for all non-x86 architectures. As the original source location has been shut down (but sane still links to it) I have changed the source url to point to fossies.org.

Further, these changes enable support for all architectures in hplip (my original motivation, thus both in one PR). This also removes the "lite" build option from hplip and adds `python-dev` as build dependency to build the python library that is required by the hpps filter. The hpps filter has already been packaged and installed before, however it was completely broken as it is basically only a wrapper around the Python library that was missing. Without the "lite" build option, it is installed automatically now. With these changes, I can print on my HP OfficeJet Pro 8210 (network) printer from an armhf tablet running Alpine Linux/postmarketOS. Note that curerently, python is not a dependency of hplip (as none of the libraries is linked against python), however, without python being installed, the hpps filter (and possibly also other parts of hplip) won't work. 

[Edit] In the meantime I've been able to test sane on my armhf tablet. After adding the necessary firmware, scanning works with an Epson Perfection 3490 Photo USB scanner using the already packaged simple-scan GUI and sane's snapscan backend.

### DIFF
--- a/testing/hplip/APKBUILD
+++ b/testing/hplip/APKBUILD
@@ -4,13 +4,13 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 pkgname=hplip
 pkgver=3.18.6
-pkgrel=1
+pkgrel=2
 pkgdesc="Drivers for HP printers and scanners"
-arch="x86 x86_64" # missing sane on several arches
+arch="all"
 url="http://hplipopensource.com"
 license="GPL"
 depends=""
-makedepends="libjpeg-turbo-dev net-snmp-dev cups-dev libusb-dev sane-dev"
+makedepends="libjpeg-turbo-dev net-snmp-dev cups-dev libusb-dev sane-dev python-dev"
 subpackages="$pkgname-doc $pkgname-libs sane-backend-hpaio:sane"
 source="http://downloads.sourceforge.net/$pkgname/$pkgname-$pkgver.tar.gz
 	fix-includes.patch"
@@ -25,7 +25,6 @@ build() {
 	cd "$builddir"
 	./configure --prefix=/usr \
 		--with-docdir=/usr/share/doc/$pkgname \
-		--enable-lite-build \
 		--disable-doc-build \
 		--disable-gui-build \
 		--disable-fax-build \
@@ -49,7 +48,6 @@ package() {
 		"$pkgdir"/usr/share/hal \
 		"$pkgdir"/usr/lib/systemd \
 		"$pkgdir"/usr/lib/cups/filter/pstotiff
-	install -m755 -D prnt/filters/hpps "$pkgdir"/usr/lib/cups/filter/hpps
 }
 
 sane() {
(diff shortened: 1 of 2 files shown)


micw:
@nmeum could you please merge this? Looks good and should not break anything on the existing architectures. I'd appreciate if I could scan on my epson printer which is connected to a raspberry.
Thank you a lot!

testing/hplip/APKBUILD
```diff
@@ -4,13 +4,13 @@
...
+pkgrel=2
 pkgdesc="Drivers for HP printers and scanners"
-arch="x86 x86_64" # missing sane on several arches
+arch="all"
 url="http://hplipopensource.com"
 license="GPL"
 depends=""
-makedepends="libjpeg-turbo-dev net-snmp-dev cups-dev libusb-dev sane-dev"
+makedepends="libjpeg-turbo-dev net-snmp-dev cups-dev libusb-dev sane-dev python-dev"
 subpackages="$pkgname-doc $pkgname-libs sane-backend-hpaio:sane"
```
andypost: does it makes sense to create separate `py-sane` kinda package & how to deal with py3?

andypost:
@micw Please point release and when it got broken.

@nmeum I set backport cos looks py part needs review

micw:
@andypost 

> @micw Please point release and when it got broken.

What does this mean?

> I set backport cos looks py part needs review

Should this be split up into 2 PRs? The change to saned modifies it in a way so that x86/x64 builds are unchanged but other archs can be built (mainly by removing the parallel printer dependency sice it's a PC only thing). It also updates the mirror because debian rempoved the ols one months ago. That should be easy to approve.
The other change to hp driver adds some python stuff - probably the original author needs to be involved for questions regarding this.

testing/hplip/APKBUILD
michitux: Back when I tested this, the code worked with python3, i.e., instead of `python-dev`, this could also depend on `python3-dev`. However, at least the `hpps` executable has python (without version) in its shebang and thus breaks when you use python3. Iirc, changing the shebang fixed the issue and I could print with python 3. I found it cleaner to not to change the shebang but instead use python 2 for packaging. If you prefer python 3 I can also try to come up with a patch or sed expression that fixes the files for python 3. Note also that this does not affect sane. This is hplip only, i.e., if we split anything, this should be `py-hplip` (or similar). I have not tested scanning with hplip as I do not have a HP scanner but I assume this python library is only needed for printing. Note that for many printer drivers, `hpps` and thus this python library is not optional but they just do not work without it. See also this [related hplip bug](https://bugs.launchpad.net/hplip/+bug/1453104).

michitux:
The commit that disabled sane for non-x86 architectures was 4a9319a4ebc746898da65138230194ad5c3ff871. Note that both sane and hplip are in testing, so as far as I understood, this shouldn't be included in any Alpine release.

I wouldn't mind splitting this into two PRs, or alternatively you could also just cherry-pick the first commit of this PR. As @micw says, the first commit that affects only `testing/sane` does not change anything for the previously existing architectures and thus should be easier to review than the `hplip` changes. I've only put them together in one commit as I'm mainly interested in `hplip` on armhf (which depends on `sane`).

PR #4725, opened 2018-07-12
main/libgpg-error: upgrade to 1.32

prspkt:


main/libgpg-error/APKBUILD
```diff
@@ -9,6 +9,7 @@ arch="all"
 license="GPL-2.0-or-later LGPL-2.1-or-later"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lisp:lisp:noarch"
 source="https://gnupg.org/ftp/gcrypt/$pkgname/$pkgname-$pkgver.tar.bz2"
+builddir="$srcdir"/$pkgname-$pkgver
```
ncopa: so, this was removed in May 2018 in 440a3d6db2d7b75eff610b73005e6072ea12235e. If I merge this, what are the changes that I get a new PR in 2 months which removes it again and have commit message "improve...".

Don't you think we devs have other things to do than review and comment on completely meaningless changes? Not to mention spend extra time on backporting security fixes to stable branches when that is needed?

This needs to stop.

prspkt: I agree this happens way too often and adds noise. APKBUILD [template](https://raw.githubusercontent.com/alpinelinux/abuild/master/sample.APKBUILD) is explicit and contains _$builddir_ and so does newapkbuild. I don't think we should be removing it.

jirutka: Yes. What we IMHO should remove is repeating `cd "$builddir"` in every single function of every APKBUILD. I wanted to do it (change `abuild` script) after v3.8, but didn't have time yet. Then omitting `builddir=` would result in too magic behaviour and confusions when the default `builddir` value is not right.

PR #4736, opened 2018-07-14
testing/cloud-utils: new aport

mouzfun:
https://launchpad.net/cloud-utils/
This package provides a useful set of utilities for interacting with a cloud

### DIFF
--- a/testing/cloud-utils/APKBUILD
+++ b/testing/cloud-utils/APKBUILD
@@ -1,0 +1,32 @@
+# Contributor: "Vladimir Avinkin <mouzfun@gmail.com>"
+# Maintainer: "Vladimir Avinkin <mouzfun@gmail.com>"
+pkgname=cloud-utils
+pkgver=0.30
+pkgrel=0
+pkgdesc="This package provides a useful set of utilities for interacting with a cloud"
+url="https://launchpad.net/cloud-utils/"
+arch="noarch"
+license="GPL-3.0"
+depends="bash python3 sfdisk sgdisk util-linux qemu-img mtools cdrkit coreutils"
+install=""
+subpackages="$pkgname-doc"
+source="https://launchpad.net/$pkgname/trunk/0.3/+download/$pkgname-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	make
+}
+
+check() {
+	cd "$builddir"
+        # the other tests require root permissions
+        PATH=$PATH:$builddir/bin TMPDIR="$srcdir" ./test/test-growpart-fsimage
+}
+
+package() {
+	cd "$builddir"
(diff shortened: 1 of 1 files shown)


mouzfun:
I'm not sure about sudo in checks, they require root so i used sudo when testing locally.
Also, it somehow failed in travis with 
```
sudo: unable to execute test/test-growpart: No such file or directory
>>> ERROR: cloud-utils*: check failed
>>> ERROR: cloud-utils: all failed
```
But works fine on my box

testing/cloud-utils/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	make
+}
+
+check() {
+	cd "$builddir"
+	PATH=$PATH:$builddir/bin; for i in test/*; do sudo $i; done
```
ncopa: we should not run tests with sudo. we need figure out some other way to run tests.

ncopa:
Also lookslike those tools depends on bash, sfdisk and possibly other tools

ncopa:
this change makes it install, but does not solve the above mentioned issues.
````diff
diff --git a/testing/cloud-utils/APKBUILD b/testing/cloud-utils/APKBUILD
index 829223fbbb..578e616c85 100644
--- a/testing/cloud-utils/APKBUILD
+++ b/testing/cloud-utils/APKBUILD
@@ -8,7 +8,7 @@ url="https://launchpad.net/cloud-utils/"
 arch="noarch"
 license="GPL-3.0"
 depends="sfdisk qemu-img"
-makedepends=""
+checkdepends="coreutils bash"
 install=""
 subpackages="$pkgname-doc"
 source="https://launchpad.net/$pkgname/trunk/0.3/+download/$pkgname-$pkgver.tar.gz"
@@ -21,7 +21,8 @@ build() {
 
 check() {
        cd "$builddir"
-       PATH=$PATH:$builddir/bin; for i in test/*; do sudo $i; done
+       # the other tests require root permissions
+       PATH=$PATH:$builddir/bin TMPDIR="$srcdir" ./test/test-growpart-fsimage
 }
 
 package() {
````

testing/cloud-utils/APKBUILD
mouzfun: Well, it requires root in the check itself, i can skip checks alltogether. Not sure what else can be done here, those checks setup block device file, mount them etc.

mouzfun:
Yeah, sorry. Forgot about python.

Problem is, native ubuntu package lacks those dependencies also and i'm only ever used one script in this package, will investigate deps further

mouzfun:
@ncopa  I've updates dependencies as best i could.

Package builds, check is passing

mouzfun:
@ncopa  ?

PR #4750, opened 2018-07-14
community/deadbeef: enable last.fm plugin

9001:
adding cURL as dependency so the last.fm plugin is built


ncopa:
you need to bump pkgrel too, otherwise will build server not pick it up. (and if it did you wouldnt get it with `apk upgrade` because the version didnt change so apk would think there was nothing to upgrade)

9001:
oh right, thanks!

PR #4751, opened 2018-07-14
community/sox: enable spectrograms

9001:
adding libpng as dependency to enable spectrogram creation


ncopa:
you need to bump pkgrel.

9001:
there, sorry about that :>

PR #4756, opened 2018-07-15
add non-free/b43-firmware-classic

rohanlean:
I am not sure if this is actually helpful – neither this nor the regular b43-firmware (in the updated form that I also submitted) ultimately worked for me.  I also have left the "Maintainer" field blank.

As for b43-firmware I have copied the version and the download-url from Arch Linux. HTTPS unfortunately does not seem to work for the download, but I have verified the checksum against the one from the Arch Linux package. Of course, please double check before merging.


ncopa:
how is this different from #4755?

rohanlean:
This one is an older version. Since both variants are offered in Arch Linux I thought that this one may work better for some chipsets. Neither firmware made my chipset fully functional with the b43 driver.

PR #4757, opened 2018-07-15
add non-free/broadcom-wl-vanilla

rohanlean:
The proprietary wifi kernel module provided by Broadcom.  Package based mostly on the one from Arch Linux.


rohanlean:
I think the ci-build failure is due to linux-firmware being required by linux-vanilla, but somehow not getting installed by abuild/apk.

PR #4773, opened 2018-07-16
 testing/botan: security upgrade to 2.7.0

tcely:


testing/botan/APKBUILD
```diff
@@ -1,7 +1,8 @@
 # Contributor: tcely <tcely@users.noreply.github.com>
 # Maintainer:
 pkgname=botan
-pkgver=2.6.0
+_pkgname=Botan
```
ncopa: I don't think `_pkgname=Botan` bring any value so I'd like that reverted.

tcely: This is just a case of following the "don't repeat yourself" practice. I don't have strong feelings about this change either way.

jirutka: It brings a value - DRY (don't repeat yourself) and it's a common practice.

PR #4774, opened 2018-07-16
main/libbluray: upgrade to 1.0.2

prspkt:


fabled:
Same as #4594 but seems to keep java out. Will look at it.

PR #4777, opened 2018-07-16
testing/pike: new aport

grodzik:
https://pike.lysator.liu.se
Pike Programing language


community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
+# Maintainer: Paweł Tomak <pawel@tomak.eu>
+pkgname=pike
+pkgver=8.0.610
+pkgrel=0
+pkgdesc="Pike Programing language"
+url="https://pike.lysator.liu.se"
+arch="x86_64"
+license="MIT"
```
tcely: The https://pike.lysator.liu.se/download/ site doesn't mention MIT license.

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+			$pkgname-dev \
+			$pkgname-doc \
+			$pkgname-core"
+source="http://pike.lysator.liu.se/pub/pike/all/${pkgver}/Pike-v${pkgver}.tar.gz"
+builddir="$srcdir/"
+
+unpack() {
+	cd "$builddir"
+	tar xzf Pike-v${pkgver}.tar.gz --strip-components=1
+}
```
tcely: You'll want to use a `prepare` function instead.

```
prepare() {
  default_prepare

  cd "$builddir"
  : # your other custom commands
}
```

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+			$pkgname-image \
+			$pkgname-pcre \
+			$pkgname-sdl \
+			$pkgname-svg \
+			$pkgname-sqlite \
+			$pkgname-dev \
+			$pkgname-doc \
+			$pkgname-core"
+source="http://pike.lysator.liu.se/pub/pike/all/${pkgver}/Pike-v${pkgver}.tar.gz"
+builddir="$srcdir/"
```
tcely: `builddir="$srcdir/Pike-v${pkgver}"`

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+		--with-tifflib \
+		--without-freetype \
+		--without-vcdiff \
+		--with-_Image_WebP \
+		--with-zlib"
+}
+
+package() {
+	cd "$builddir"
+	depends=""
```
tcely: Do you have a reason to reset `depends` from the top of the file here? I'd be confused if I added an item to `depends` as I normally would at the top and it didn't end up in the built package.

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+	local dn=$(dirname "$1")
+	local dst="$2/$dn"
+	mkdir -p "$dst"
+	mv "$1" "$dst/$bn"
+}
+
+_install() {
+	local bn=$(basename "$1")
+	local dn=$(dirname "$1")
+	for item in `find $dn -maxdepth 1 -name "$bn"`; do
```
tcely: This is not a particularly safe construction. Consider using something like this:

```
find "$dn" -maxdepth 1 -name "$bn" | while IFS= read item; do _move_item "$item" "$2"; done
```

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+	depends="pike-image=$pkgver-r$pkgrel libglade gtk+2.0"
+	_install "usr/lib/pike/modules/*GTK2*" "$subpkgdir"
+	_install "usr/lib/pike/modules/Tools.pmod/PV.pike" "$subpkgdir"
+}
+
+image() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname image handling modules"
+	depends="pike-core=$pkgver-r$pkgrel giflib-utils libjpeg tiff libwebp"
+	_install "usr/lib/pike/?.?/modules/Image.pmod" "$subpkgdir"
```
tcely: With the quotes, the first argument to `_install` is likely not doing what you expect.

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+}
+
+dev() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname dev files"
+	depends="pike-core=$pkgver-r$pkgrel"
+	_install "usr/lib/pike/include/*.h" "$subpkgdir"
+	_install "usr/include/pike/*.h" "$subpkgdir"
+	mv "$subpkgdir/usr/lib/pike/include/regexp.h" "usr/lib/pike/include/"
+}
```
tcely: You'd probably be better off letting the `default_dev` function do its thing.

community/pike/APKBUILD
```diff
@@ -0,0 +1,235 @@
...
+
+doc() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname documentation"
+	depends="pike-core=$pkgver-r$pkgrel"
+	mkdir -p "$subpkgdir/usr/share/"
+	gzip usr/share/man/man1/pike.1
+	mv usr/doc/ "$subpkgdir/usr/share/"
+	mv usr/share/man/ "$subpkgdir/usr/share/"
+}
```
tcely: Letting the `default_doc` function handle this would be better.

community/pike/APKBUILD
grodzik: I actually need the `regexp.h` file inside `pike-core` package. Or will
```
dev() {
    default_dev
    mv .....
}
```
do the trick?

community/pike/APKBUILD
grodzik: Probably some kind of leftover from development

community/pike/APKBUILD
grodzik: I'll fix that, but can you share why this isn't safe?

grodzik:
As for number of packages, Pike has a lot of dependencies, taking a lot of data, not necessarily needed to have all the time. I don't even made all the packages available, I'll try to add them later. In general Pike is quite a modular package, thus it would be good to use that well. 

My intention with `pike` package was to act only as a meta-package, instead of creating `pike-full` package, and `pike-core` for the very basic stuff.

community/pike/APKBUILD
grodzik: I can change that, no problem. However I would be grateful for explanation why `for` approach is unsafe?

tcely: The path in `$dn` is not under your control and neither is the `find` output.

The easiest problem to understand with that `for` construction is word splitting. You're depending on splitting on `\n` for each line `find` prints, but word splitting also happens on any embedded spaces too.

community/pike/APKBUILD
tcely: Yep. That'll get it done.

https://github.com/alpinelinux/abuild/blob/c0a862930c1bd033be30f2c48e91a6f7e76a5a58/abuild.in#L1706

community/pike/APKBUILD
tcely: The only thing I see that might not be handled by `default_doc` is the `usr/doc` directory. You could change the location of `usr/doc` in the `package` function to `usr/share/doc` so that it is split without any need for your own `doc` implementation.

community/pike/APKBUILD
tcely: You shouldn't need this `unpack` function at all.

community/pike/APKBUILD
tcely: If you want `pike-core` to split before `pike-dev` just change the order in the `subpackages` list.

tcely:
I still think you're making too many complications for yourself by splitting out so much right away. Having `pike` with most of the files, `pike-dev` and `pike-doc` as `subpackages` would be a good starting point.

grodzik:
I've fixed issues you've raised. Thanks for reviewing. I removed `-dev` package, since some header files where still needed and that was creating a bit of mess. I removed subpackages that were not depending on any additional libraries, like `-fuse`.  And I've fixed creating documentation (thanks to Gentoo ebuild)

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	done
+	make INSTALLARGS="--traditional" buildroot="$pkgdir" install
+}
+
+_move_item() {
+	local bn=$(basename "$1")
+	local dn=$(dirname "$1")
+	local dst="$2/$dn"
+	mkdir -p "$dst"
+	mv "$1" "$dst/$bn"
```
tcely: `mv "$1" "$dst/$3"` would let you optionally rename the destination file.

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	depends="pike-image=$pkgver-r$pkgrel libglade gtk+2.0"
+	_install "usr/lib/pike/modules/*GTK2*" "$subpkgdir"
+	_install "usr/lib/pike/modules/Tools.pmod/PV.pike" "$subpkgdir"
+}
+
+image() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname image handling modules"
+	depends="pike-core=$pkgver-r$pkgrel giflib-utils libjpeg tiff libwebp"
+	_install "usr/lib/pike/?.?/modules/Image.pmod" "$subpkgdir"
```
tcely: Does this work the way you are expecting? I don't think the quoted `?.?` will match.

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	local bn=$(basename "$1")
+	local dn=$(dirname "$1")
+	find "$dn" -maxdepth 1 -name "$bn" | \
+		while IFS= read item; do _move_item "$item" "$2"; done
+}
+
+bzip2() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname bzip2 support"
+	depends="pike-core=$pkgver-r$pkgrel bzip2"
```
tcely: Unless you need the bzip2 binaries, you should remove `bzip2` from depends. `libbz2` is being depended on automatically.

> \>>> pike-bzip2*: Tracing dependencies...
	pike-core=8.0.610-r0
	bzip2
	so:libbz2.so.1

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	cd "$pkgdir"
+	pkgdesc="$pkgname SVG support"
+	depends="pike-core=$pkgver-r$pkgrel librsvg"
+	_install "usr/lib/pike/modules/_Image_SVG.so" "$subpkgdir"
+}
+
+sqlite() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname SQLite support"
+	depends="pike-core=$pkgver-r$pkgrel sqlite-libs"
```
tcely: `sqlite-libs` is pulled in automatically. There is no need to list it in depends.

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	cd "$pkgdir"
+	pkgdesc="$pkgname SDL support"
+	depends="pike-image=$pkgver-r$pkgrel sdl sdl_mixer"
+	_install "usr/lib/pike/modules/*SDL.*" "$subpkgdir"
+}
+
+svg() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname SVG support"
+	depends="pike-core=$pkgver-r$pkgrel librsvg"
```
tcely: `librsvg` is not needed here.

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	rm -rf usr/doc/
+	default_doc
+}
+
+core() {
+	cd "$pkgdir"
+	pkgdesc="Pike's core files and executables"
+	depends="gmp nettle nettle-utils zlib readline libgcrypt"
+	mkdir -p "$subpkgdir"
+	mv usr "$subpkgdir/"
```
tcely: I'm going to suggest leaving `usr/bin` in `pike` package.

```
mkdir -p "$subpkgdir/usr"
mv usr/include usr/lib "$subpkgdir/usr/"
```

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+	mkdir -p usr/share/doc/
+	mv "$builddir/refdoc/traditional_manual/" usr/share/doc/pike/
+	rm -rf usr/doc/
+	default_doc
+}
+
+core() {
+	cd "$pkgdir"
+	pkgdesc="Pike's core files and executables"
+	depends="gmp nettle nettle-utils zlib readline libgcrypt"
```
tcely: `gmp`, `zlib` are already included automatically.

I don't see an immediate use for `libgcrypt`. Would you explain why it's here?

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+			nettle-dev \
+			nettle-utils \
+			pcre-dev \
+			sdl-dev \
+			sdl_mixer-dev \
+			sqlite-dev \
+			sqlite-libs \
+			readline \
+			zlib-dev \
+			librsvg-dev"
```
jirutka: It'd not needed to escape newlines inside a string. Please remove these backslahes and use only one tab for indentation.

community/pike/APKBUILD
```diff
@@ -0,0 +1,200 @@
...
+		--without-freetype \
+		--without-vcdiff \
+		--with-_Image_WebP \
+		--with-zlib"
+	PATH="${builddir}/bin:${PATH}" make doc || die "doc failed"
+}
+
+package() {
+	cd "$builddir"
+	for pkg in $(printf "%s" "$subpackages" | tr ' ' '\n'); do
```
jirutka: `pkg` should be declared as local (`local pkg; for pkg in ...`).

community/pike/APKBUILD
grodzik: Ok, so maybe get rid of `pike-core` and leave core stuff in `pike`, and add something like `pike-full` meta package that will pull in all additional `pike-` packages?

tcely: I'd suggest `pike-modules` so that you could choose to install `pike` and specific individual modules or `pike` and `pike-modules` which would depend on all the split up modules.

community/pike/APKBUILD
```diff
@@ -89,96 +85,89 @@ build() {
 
 package() {
 	cd "$builddir"
-	for pkg in $(printf "%s" "$subpackages" | tr ' ' '\n'); do
-		depends="$depends $pkg=$pkgver-r$pkgrel"
-	done
+	depends="gmp nettle nettle-utils zlib readline"
```
tcely: Please put the `depends` for the package in the `depends` variable at the top of the file. This information is used to control the order in which builds happen from what I understand.

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
+# Maintainer: Paweł Tomak <pawel@tomak.eu>
+pkgname=pike
+pkgver=8.0.610
+pkgrel=0
+pkgdesc="Pike Programing language"
+url="https://pike.lysator.liu.se"
+arch="x86_64"
+license="GPL-2.0-or-later"
+options="!check" # there is no tests for those packages
```
tcely: grammar: there are no tests for this package

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
+# Maintainer: Paweł Tomak <pawel@tomak.eu>
+pkgname=pike
```
tcely: `_pkgname=Pike`

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+	sdl_mixer-dev
+	sqlite-dev
+	sqlite-libs
+	readline
+	zlib-dev
+	librsvg-dev"
+install=
+subpackages="$pkgname-doc $pkgname-full"
+source="http://pike.lysator.liu.se/pub/pike/all/${pkgver}/Pike-v${pkgver}.tar.gz"
+builddir="$srcdir/Pike-v$pkgver"
```
tcely:
```
source="http://pike.lysator.liu.se/pub/${pkgname}/all/${pkgver}/${_pkgname}-v${pkgver}.tar.gz"
builddir="${srcdir}/${_pkgname}-v${pkgver}"
```

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
+# Maintainer: Paweł Tomak <pawel@tomak.eu>
+pkgname=pike
+pkgver=8.0.610
+pkgrel=0
+pkgdesc="Pike Programing language"
+url="https://pike.lysator.liu.se"
+arch="x86_64"
+license="GPL-2.0-or-later"
+options="!check" # there is no tests for those packages
+depends=
```
tcely: depends from `package` function should go here

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+subpackages="$pkgname-doc $pkgname-full"
+source="http://pike.lysator.liu.se/pub/pike/all/${pkgver}/Pike-v${pkgver}.tar.gz"
+builddir="$srcdir/Pike-v$pkgver"
+
+_full_deps="$pkgname=${pkgver}-r${pkgrel} ${pkgname}-doc=${pkgver}-r${pkgrel}"
+for _name in bzip2 gdbm gl gtk image pcre sdl svg sqlite
+do
+	subpackages="$subpackages $pkgname-${_name}"
+	_full_deps="${_full_deps} $pkgname-${_name}=$pkgver-r$pkgrel"
+done
```
tcely: Consider adding `$pkgname-dev` to the end of `subpackages` to split out the include files

community/pike/APKBUILD
jirutka: Please don't promote unnecessary curly braces around variables, respect our code-style.

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+		--without-sass \
+		--with-sane \
+		--with-SDL \
+		--with-SDL_Mixer \
+		--with-svg \
+		--with-tifflib \
+		--without-freetype \
+		--without-vcdiff \
+		--with-_Image_WebP \
+		--with-zlib"
```
tcely: Do you need to escape the new lines in this quoted string?

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+	PATH="${builddir}/bin:${PATH}" make doc || die "doc failed"
```
tcely: When make doc returns an error that error having come from make doc should be fairly self evident.

What value does `|| die "doc failed"` add here?

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+		mkdir -p "$subpkgdir/$dn"
+		mv "$1" "$subpkgdir/$dn/$bn"
+		shift
+	done
+}
+
+bzip2() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname support"
+	depends="pike=$pkgver-r$pkgrel bzip2"
```
tcely: You don't need this `depends` line here at all. `bzip2` is the CLI tool for the `libbz2` library that is being found and depended on automatically.

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+	_install usr/lib/pike/modules/*Bz2.*
+}
+
+gdbm() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname GDBM support"
+	depends="pike=$pkgver-r$pkgrel gdbm"
```
tcely: Compare how `depends="pike"` works here in your next build.

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+	cd "$pkgdir"
+	pkgdesc="$pkgname PCRE support"
+	depends="pike=$pkgver-r$pkgrel pcre"
+	_install usr/lib/pike/modules/*Regexp_PCRE.*
+}
+
+sdl() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname SDL support"
+	depends="pike-image=$pkgver-r$pkgrel sdl sdl_mixer"
```
tcely: `depends="pike"` again

community/pike/APKBUILD
```diff
@@ -0,0 +1,187 @@
...
+	_install usr/lib/pike/modules/*Image*.pmod
+	_install usr/lib/pike/modules/Graphics.pmod
+	_install usr/lib/pike/modules/Colors.pmod
+	_install usr/lib/pike/modules/Protocols.pmod/X.pmod/XImage.pmod
+}
+
+pcre() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname PCRE support"
+	depends="pike=$pkgver-r$pkgrel pcre"
```
tcely: This is pulled in by the `so:libpcre.so.1` that is added automatically, so listing `pcre` is redundant.

`depends="pike"`

community/pike/APKBUILD
tcely: I still don't get why you insist on everyone keeping these variable parsing rules in our heads, but I'll edit it to respect the style as you requested.

community/pike/APKBUILD
grodzik: There are at least two headers, that would have to be moved back manually to main package. Pike's preprocesor needs them while parsing some core files.

grodzik:
Hi, is there anything else to fix here?

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+}
+
+package() {
+	cd "$builddir"
+	make INSTALLARGS="--traditional" buildroot="$pkgdir" install
+}
+
+_install() {
+	local _arg
+	for _arg in "$@"; do
```
tcely: The `in "$@"` isn't needed, but won't hurt anything.

Try this and see for yourself.
`$ /bin/sh -c 'foo() { local arg; for arg; do echo $arg; done; }; foo a b c;'`

tcely:
@grodzik Is it possible to clean these up?

```
Could not resolve Getopt. (Perhaps the installed pike tree has been moved.)
lib/master.pike:3163: master()->main_resolv("Getopt",UNDEFINED)
Dumping of some modules failed (not fatal) (0x0000000a):
  "/usr/lib/pike/modules/Crypto.pmod/Random.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA3_512.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA256.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/DH.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/PGP.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/Substitution.pike"
  "/usr/lib/pike/modules/Crypto.pmod/ChaCha20.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA3_384.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/AES.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA3_224.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA512.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SALSA20R12.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/ECC.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SALSA20.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/module.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/Koremutake.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/Camellia.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/DSA.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/Arcfour.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/DES3.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/MD2.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/Twofish.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/SHA224.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/CAST.pmod"
  "/usr/lib/pike/modules/Crypto.pmod/GOST94.pmod"
```

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+	pcre-dev
+	sdl-dev
+	sdl_mixer-dev
+	sqlite-dev
+	sqlite-libs
+	readline
+	zlib-dev
+	librsvg-dev"
+install=
+subpackages="$pkgname-doc $pkgname-full"
```
tcely: The build log is suggesting `$pkgname-full::noarch`

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+	cd "$pkgdir"
+	pkgdesc="$pkgname support"
+	depends="pike"
+	_install usr/lib/pike/modules/*Bz2.*
+}
+
+gdbm() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname GDBM support"
+	depends="pike gdbm"
```
tcely: `gdbm` is not needed in depends here. See https://pkgs.alpinelinux.org/contents?file=libgdbm.so.4&path=&name=&branch=edge&arch=x86_64

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+	_install usr/lib/pike/modules/*Gdbm.*
+}
+
+gl() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname OpenGL/GLUT support"
+	depends="pike freeglut mesa-gl libsm"
```
tcely: `mesa-gl` is not needed in this depends

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+	cd "$pkgdir"
+	pkgdesc="$pkgname PCRE support"
+	depends="pike"
+	_install usr/lib/pike/modules/*Regexp_PCRE.*
+}
+
+sdl() {
+	cd "$pkgdir"
+	pkgdesc="$pkgname SDL support"
+	depends="pike-image sdl sdl_mixer"
```
tcely: `sdl` and `sdl_mixer` aren't needed in this depends line

community/pike/APKBUILD
```diff
@@ -0,0 +1,188 @@
...
+pkgname=pike
+_pkgname=Pike
+pkgver=8.0.610
+pkgrel=0
+pkgdesc="Pike Programing language"
+url="https://pike.lysator.liu.se"
+arch="x86_64"
+license="GPL-2.0-or-later"
+options="!check" # there are no tests for those packages
+depends="gmp nettle nettle-utils readline"
```
tcely: `gmp` (`so:libgmp.so.10`) and `nettle` (`so:libnettle.so.6`) can be removed from this `depends` list

community/pike/APKBUILD
jirutka: I know that this shortened syntax is defined in POSIX, I just don't find it intuitive and reasonable. That's just my opinion.

grodzik:
Thanks for the review @tcely and @jirutka. Anything else needs to be done from my side to push this further?

ncopa:
It would be nice to merge all the commits. We should also move it to `testing/pike` instead of `community/pike`, since it is a new package. I suppose I can do that for you.

We can move it to community once it has been verified that it works, even when built with the official builders.

Thanks!

ncopa:
fixed with eec7db60bd8274db29ce5731c26019778a22f28e

PR #4786, opened 2018-07-18
community/rust: upgrade to 1.27.1

jeizsm:
https://github.com/rust-lang/rust/issues/51650
new patch for this, may be removed in 1.28


jirutka:
I've upgraded it right to 1.27.2 in f64f617640a377343a8836f60e08f627e48b4a78. Thanks for pointing to the issue!

PR #4795, opened 2018-07-19
testing/log4cpp: new aport

hvanderheide:
http://log4cpp.sourceforge.net
Log for C++ is a library of C++ classes for flexible logging to files, syslog and other destinations. It is modeled after the Log for Java library (http://jakarta.apache.org/log4j/), staying as close to their API as is reasonable.

I have no affiliation with this project, nor am I a C++ developer, just need it as a dependency for another package I am preparing. I am happy to maintain it but might not be the best fit.


tcely:
I'm curious what is using this. It's been a year since the last release so this doesn't seem very active.

PR #4799, opened 2018-07-19
testing/monero: upgrade to 0.12.3.0

andypost:
https://github.com/monero-project/monero/releases


prspkt:
@andypost You can use [this patch](https://github.com/void-linux/void-packages/blob/203176dd096fcae7b55268f3465c22fcabd0c355/srcpkgs/monero/patches/system-miniupnpc.patch) to resolve the issue. Also, don't forget to add `rapidjson-dev` in makedepends.

andypost:
Fixed patch a bit - added a/b to file path to 2 & 3 hunk

PR #4804, opened 2018-07-19
community/php7: security upgrade to 7.2.8

andypost:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-9253

When other (older) releases will out I will file new PRs


andypost:
@jirutka It needs to be ported to 3.8

PR #4812, opened 2018-07-20
main/py-dbus: upgrade to 1.2.8, modernize and fix license

rgdoliveira:


ncopa:
this conflicts with #2452

ncopa:
It seems that the CI fails to figure out the proper build order.

ncopa:
I fixed it in py-sphinx: a93d2b7eacde17d8e354f23524aa4b6ab43f3e7b

michitux:
Concerning the license change: are you sure the `py-dbus` package can be licensed under the MIT and X11 license? As far as I can see, the `py-dbus` package links against `dbus-glib` and the the `dbus-glib` package is licensed under the GPL. Therefore, from my understanding as a non-lawyer, `py-dbus` must be licensed under a license that is compatible with the conditions of the GPL which is not the case with the MIT or X11 licenses.

PR #4814, opened 2018-07-20
testing/openmpi: fix build hang by disabling ppc64le

mksully22:
On ppc64le during make check the opal_fifo test is hanging. This effectively blocks the ppc64le build server. This PR disables ppc64le until I can debug and fix the test


mksully22:
I used a script that runs opal_fifo in a loop and found that this hang is reproducible on even the master branch of the ompi project on both Alpine and Ubuntu 17.10 for powerpc64. I've opened an issue on the github ompi project with gdb info on the hang https://github.com/open-mpi/ompi/issues/5470. Once I receive some feedback from the project I will reflect any changes in Alpoine.

PR #4824, opened 2018-07-21
main/xorg-server: upgrade to 1.20.0

prspkt:
The following ~~(3)~~ (1) package fails to rebuild against the new video ABI:

- `main/xf86-video-s3virge-1.10.7`
~~- `main/xf86-video-savage-2.3.9`~~
~~- `main/xf86-video-sis-0.10.9`~~

Those packages are at their latest version as seen [here](https://www.x.org/releases/individual/driver) and I'm not sure if they will continue to be supported going forward (xorg-video-abi-24 included).

### DIFF
--- a/main/xf86-input-keyboard/APKBUILD
+++ b/main/xf86-input-keyboard/APKBUILD
@@ -1,15 +1,15 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=xf86-input-keyboard
 pkgver=1.9.0
-pkgrel=0
+pkgrel=1
 pkgdesc="X.org keyboard input driver"
-url="http://xorg.freedesktop.org/"
+url="https://xorg.freedesktop.org"
 arch="all"
 license="MIT"
 subpackages="$pkgname-doc"
 depends=
 makedepends="libxkbfile-dev xorg-server-dev libxi-dev libxrandr-dev"
-source="http://www.x.org/releases/individual/driver/$pkgname-$pkgver.tar.bz2"
+source="https://www.x.org/releases/individual/driver/$pkgname-$pkgver.tar.bz2"
 
 builddir="$srcdir/$pkgname-$pkgver"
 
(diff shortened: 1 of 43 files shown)


ncopa:
> Failed to build packages: main/xf86-video-s3virge main/xf86-video-savage main/xf86-video-sis

ncopa:
looks like there are a couple of patches we could grab: https://cgit.freedesktop.org/xorg/driver/xf86-video-savage/

ncopa:
there are some patches to gra for xf86-video sis too: https://cgit.freedesktop.org/xorg/driver/xf86-video-sis/log/

main/xf86-video-s3virge/APKBUILD
```diff
@@ -1,17 +1,17 @@
...
-url="http://xorg.freedesktop.org/"
+url="https://xorg.freedesktop.org/"
 arch="all"
 license="MIT"
 subpackages="$pkgname-doc"
 depends=
 makedepends="xorg-server-dev libxi-dev util-macros xorgproto"
-source="http://www.x.org/releases/individual/driver/$pkgname-$pkgver.tar.bz2
+source="https://www.x.org/releases/individual/driver/$pkgname-$pkgver.tar.bz2
+	check-max[HV]Value.patch
```
ncopa: can you please rename the patch file to not include `[` or `]` chars? I believe its generally a bad idea to include those chars in a file name, just like it is a bad idea to include chars like `?` and `*` in filenames.

prspkt: Noted and fixed.

main/xf86-input-keyboard/APKBUILD
```diff
@@ -20,13 +20,13 @@ build() {
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
```
ncopa: trailing `\` needs to be removed.

ncopa:
i have fix for it here. will push it.

Good job!

PR #4825, opened 2018-07-22
community/dnscrypt-proxy upgrade to 2.0.16

ianbashford:
support for supervisor process/unprivileged user and bug fixes
changes made to init.d to support this feature
additional features added to config file

### DIFF
--- a/community/dnscrypt-proxy/config-full-paths.patch
+++ b/community/dnscrypt-proxy/config-full-paths.patch
@@ -1,9 +1,9 @@
 diff --git a/./dnscrypt-proxy.toml b/dnscrypt-proxy/dnscrypt-proxy.toml
 new file mode 100644
-index 0000000..347ada5
+index 0000000..8c4ce26
 --- /dev/null
 +++ b/dnscrypt-proxy/dnscrypt-proxy.toml
-@@ -0,0 +1,465 @@
+@@ -0,0 +1,510 @@
 +
 +##############################################
 +#                                            #
@@ -47,6 +47,18 @@ index 0000000..347ada5
 +max_clients = 250
 +
 +
++## Switch to a non-privileged system user after listening sockets have been created.
++## Two processes will be running.
++## The first one will keep root privileges, but is only a supervisor, that does nothing
++## except create the sockets, manage the service, and restart it if it crashes.
++## The second process is the service itself, and that one will always run as a different
++## user.
++## Note (1): this feature is currently unsupported on Windows.
++## Note (2): this feature is not compatible with systemd socket activation.
++
++user_name = 'dnscrypt'
++
++
 +## Require servers (from static + remote sources) to satisfy specific properties
 +
 +# Use servers reachable over IPv4
@@ -74,16 +86,22 @@ index 0000000..347ada5
 +require_nofilter = true
 +
 +
-+
-+## Always use TCP to connect to upstream servers
++## Always use TCP to connect to upstream servers.
++## This can be can be useful if you need to route everything through Tor.
++## Otherwise, leave this to `false`, as it doesn't improve security
++## (dnscrypt-proxy will always encrypt everything even using UDP), and can
++## only increase latency.
 +
 +force_tcp = false
++
 +
 +## HTTP / SOCKS proxy
 +## Uncomment the following line to route all TCP connections to a local Tor node
 +## Tor doesn't support UDP, so set `force_tcp` to `true` as well.
 +
 +# proxy = "socks5://127.0.0.1:9050"
++
++
 +
 +## How long a DNS query will wait for a response, in milliseconds
 +
@@ -107,7 +125,7 @@ index 0000000..347ada5
 +
 +## log file for the application
 +
-+# log_file = '/var/log/dnscrypt-proxy.log'
++# log_file = '/var/log/dnscrypt-proxy/dnscrypt-proxy.log'
 +
 +
 +## Use the system logger (syslog on Unix, Event Log on Windows)
@@ -165,10 +183,26 @@ index 0000000..347ada5
 +fallback_resolver = '9.9.9.9:53'
 +
 +
-+## Never try to use the system DNS settings; unconditionally use the
-+## fallback resolver.
++## Never let dnscrypt-proxy try to use the system DNS settings;
++## unconditionally use the fallback resolver.
 +
 +ignore_system_dns = false
++
++
++## Maximum time (in seconds) to wait for network connectivity before
++## initializing the proxy.
++## Useful if the proxy is automatically started at boot, and network
++## connectivity is not guaranteed to be immediately available.
++## Use 0 to disable.
++
++netprobe_timeout = 30
++
++
++## Offline mode - Do not use any remote encrypted servers.
++## The proxy will remain fully functional to respond to queries that
++## plugins can handle directly (forwarding, cloaking, ...)
++
++# offline_mode = false
 +
 +
 +## Automatic log files rotation
@@ -190,8 +224,9 @@ index 0000000..347ada5
 +
 +## Immediately respond to IPv6-related queries with an empty response
 +## This makes things faster when there is no IPv6 connectivity, but can
-+## also cause reliability issues with some stub resolvers. In
-+## particular, enabling this on macOS is not recommended.
++## also cause reliability issues with some stub resolvers.
++## Do not enable if you added a validating resolver such as dnsmasq in front
++## of the proxy.
 +
 +block_ipv6 = false
 +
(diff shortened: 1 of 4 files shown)


community/dnscrypt-proxy/dnscrypt-proxy.initd
```diff
@@ -4,17 +4,18 @@
 
 command="/usr/bin/dnscrypt-proxy"
 command_args="${DNSCRYPT_OPTS:--config /etc/dnscrypt-proxy/dnscrypt-proxy.toml}"
-command_user="${DNSCRYPT_USER:-dnscrypt}:${DNSCRYPT_GROUP:-dnscrypt}"
+dir_owner="${DNSCRYPT_USER:-dnscrypt}:${DNSCRYPT_GROUP:-dnscrypt}"
```
fcolista: What is the reason why you changed command_user to dir_owner? 
According with OpenRC manual, this is an option of OpenRC, not a simple variable: 
https://github.com/OpenRC/openrc/blob/master/man/openrc-run.8

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -462,7 +502,7 @@ index 0000000..347ada5
 +
 +  #  [sources.'parental-control']
 +  #  urls = ['https://raw.githubusercontent.com/DNSCrypt/dnscrypt-resolvers/master/v2/parental-control.md', 'https://download.dnscrypt.info/resolvers-list/v2/parental-control.md']
-+  #  cache_file = '/var/cache/dnscrypt-proxy/parental-control.md'
++  #  cache_file = 'parental-control.md'
```
fcolista: See comment on line 495

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -452,7 +492,7 @@ index 0000000..347ada5
 +
 +  [sources.'public-resolvers']
 +  urls = ['https://raw.githubusercontent.com/DNSCrypt/dnscrypt-resolvers/master/v2/public-resolvers.md', 'https://download.dnscrypt.info/resolvers-list/v2/public-resolvers.md']
-+  cache_file = '/var/cache/dnscrypt-proxy/public-resolvers.md'
++  cache_file = 'public-resolvers.md'
```
fcolista: Where is this file downloaded? There's a reason why the full path has been removed?

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -382,7 +422,7 @@ index 0000000..347ada5
 +
 +  ## Optional path to a file logging whitelisted queries
 +
-+  # log_file = '/var/log/whitelisted.log'
++  # log_file = '/var/log/dnscrypt-proxy/whitelisted.log'
```
fcolista: See comment on line 308

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -354,7 +394,7 @@ index 0000000..347ada5
 +
 +  ## Optional path to a file logging blocked queries
 +
-+  # log_file = '/var/log/ip-blocked.log'
++  # log_file = '/var/log/dnscrypt-proxy/ip-blocked.log'
```
fcolista: See comment on line 308

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -326,7 +366,7 @@ index 0000000..347ada5
 +
 +  ## Optional path to a file logging blocked queries
 +
-+  # log_file = '/var/log/blocked.log'
++  # log_file = '/var/log/dnscrypt-proxy/blocked.log'
```
fcolista: See comment on line 308

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -291,7 +331,7 @@ index 0000000..347ada5
 +
 +  ## Path to the query log file (absolute, or relative to the same directory as the executable file)
 +
-+  # file = '/var/log/nx.log'
++  # file = '/var/log/dnscrypt-proxy/nx.log'
```
fcolista: See comment on line 308

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -265,7 +305,7 @@ index 0000000..347ada5
 +
 +  ## Path to the query log file (absolute, or relative to the same directory as the executable file)
 +
-+  # file = '/var/log/query.log'
++  # file = '/var/log/dnscrypt-proxy/query.log'
```
fcolista: If you use a directory here, be sure that is created in  the APKBUILD during the installation, and with the correct permissions.

community/dnscrypt-proxy/config-full-paths.patch
```diff
@@ -107,7 +125,7 @@ index 0000000..347ada5
 +
 +## log file for the application
 +
-+# log_file = '/var/log/dnscrypt-proxy.log'
++# log_file = '/var/log/dnscrypt-proxy/dnscrypt-proxy.log'
```
fcolista: See comment on line 308

community/dnscrypt-proxy/dnscrypt-proxy.initd
ianbashford: True -- this is to accommodate the supervisor.  In this new way of running, the supervisor program runs as root, and starts a child process as the user specified in the config file.   I used dir_owner as I still want to run the directory permission checks later in the file.   We can't run the supervisor as an unprivileged user.    Happy to try for a different method if you have suggestions...

community/dnscrypt-proxy/config-full-paths.patch
ianbashford: My error - will fix

community/dnscrypt-proxy/config-full-paths.patch
ianbashford: I'd left that to the init.d script which has the following to create if doesn't exist :
checkpath -q -d -m 0775 -o "${command_user}" \
		/var/cache/"${RC_SVCNAME}" \
		/var/log/"${RC_SVCNAME}"

Is that OK, or would you prefer a mkdir in the pre-install as well?

ianbashford:
Thanks for the review Francesco -- I've put a couple of questions in there if you could help.  I'll then get an update done.  Would you prefer a second commit, or for me to just submit one cleaned up commit?

community/dnscrypt-proxy/config-full-paths.patch
fcolista: It's better to create it in the APKBUILD itself, since you might run the dnscrypt without using the init.d

community/dnscrypt-proxy/dnscrypt-proxy.initd
fcolista: Ok, yes you mentioned that before. I don't know what is the "standard" way to run processes in this way, actually. Perhaps @ncopa can comment on this.

fcolista:
You can amend the previous commit and send a clean one. Thanks a lot Ian, nice job!

community/dnscrypt-proxy/dnscrypt-proxy.pre-install
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 
-addgroup -S dnscrypt 2>/dev/null
-adduser -S -D -H -h /var/empty -s /sbin/nologin -G dnscrypt -g dnscrypt dnscrypt 2>/dev/null
+addgroup -S dnscrypt 2>/dev/null || true
+adduser -S -D -H -h /var/empty -s /sbin/nologin -G dnscrypt -g dnscrypt dnscrypt 2>/dev/null || true
+
+install -d -o dnscrypt -g dnscrypt -m 755  /var/cache/dnscrypt-proxy 2>/dev/null 
+install -d -o dnscrypt -g dnscrypt -m 755  /var/log/dnscrypt-proxy 2>/dev/null
```
fcolista: This install script should not go here, but inside the APKBUILD.
See, for example, community/pgbouncer
Thanks!

ianbashford: Gotcha - I've tidied up some mkdir's at the same time.
thanks for the pointer

xpecex:
version 2.0.17 released.

ref https://github.com/jedisct1/dnscrypt-proxy/releases/tag/2.0.17

andypost:
https://github.com/jedisct1/dnscrypt-proxy/releases/tag/2.0.18

PR #4829, opened 2018-07-22
main/bind: Upgrade to 9.12.2-P1

tanertas:
* Enable DLZ (Dynamically Loadable Zones) support with  file system, ldap, stub backends
* Enable GSSAPI/Kerberos support
* libressl-2.7 patch removed
* Re-arrange configure options


main/bind/APKBUILD
```diff
@@ -75,19 +75,28 @@ build() {
...
 		--enable-ipv6 \
+		--enable-largefile \
+		--enable-linux-caps \
 		--enable-shared \
 		--enable-static \
-		--with-libtool \
-		--with-randomdev=/dev/random \
-		--mandir=/usr/share/man \
-		--infodir=/usr/share/info
+		--enable-threads \
```
tcely: It'd be good form to remove `\` from the last option.

ncopa:
Do we really need bdb? The recent berkleydb has problematic license so we cannot upgrade it. It has also been problematic to find information regarding sescurity issues since Oracle release very little information and has not answered requests for more info.

I'd like to get rid of `db` if possible.

tanertas:
Of course we don't. Samba4 Active Directory Domain Controller's BIND9_DLZ backend needs dlopen, ldap, kerberos. I just added it because I assumed that it is not a big dependency at all comparing postgres ant mysql. We can discard it of course.

https://wiki.samba.org/index.php/Using_BIND_DLZ_backend_with_secured_/_signed_DNS_updates

I'll edit PR.

tanertas:
Done.

tcely:
Please add `secfixes` comment update.

```
 # secfixes:
+#   9.12.2-r0:
+#     - CVE-2018-5738
 #   9.12.1_p2-r0:
```

tcely:
@andypost Please add `T-security` label to this PR too.

tanertas:
Why?

Security fix already done with upgrade. It is not backported externally. Specifying a secfix only makes sense if a stable-release branch (say Alpine 3.7, 3.8) with a different version of bind which secfix needed to be backported externally via patch. This is master branch.

https://gitweb.gentoo.org/repo/gentoo.git/tree/net-dns/bind/files

tanertas:
@tcely I suggest you for re-open your PR. Because I will not add those secfix lines which was a very bad idea right at the start. They only serve to bloat APKBUILD, if they aren't backported externally for a stable branch where version upgrades done very strictly.

I opt for removing all secfix comments, on every abuild on master branch, if they aren't backported or if they already done with version upgrade. This would be rather an improvement. @ncopa

tanertas:
@ncopa if you prefer a bare bind package without dependencies like libldap and kerberos, I can re-brand this PR as a new package request, like "bind-dlz".

andypost:
According https://www.mail-archive.com/bind-announce@lists.isc.org/msg00516.html
it should be *9.12.2-P1* and sec-fixes with CVE-2018-5740

@ncopa and older releases needs update if `deny-answer-aliases` used

andypost:
Also as @tcely said `CVE-2018-5738` should be added to secfixes

tanertas:
According to ISC's software support policy, **9.11** is the one that has Extended Support which is right through to the **end of 2021**. 9.12 will be EOL if 9.13 announced as new stable.

https://www.isc.org/downloads/software-support-policy/

ncopa:
The format of the commit message should be:
````
One-line description that's less than 72 chars long
<second line empty>
Optional longer description with explanation why changes were made. Links to relevant issues
````

https://wiki.alpinelinux.org/wiki/Development_using_git:Basic_usage#Commit

ncopa:
I will fix the commit message for you this time.

PR #4832, opened 2018-07-23
 main/gcalcli: upgrade to 4.0.0_alpha4, modernize and take maintainership

rgdoliveira:


main/gcalcli/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Roberto Oliveira <robertoguimaraes8@gmail.com>
 pkgname=gcalcli
-pkgver=3.4.0
-pkgrel=1
+pkgver=4.0.0a4
```
ncopa: `4.0.0a4` is not really a valid version. What does the `a4` suffix mean? is `4.0.0` newer/bigger or older/smaller than `4.0.0a4`?

rgdoliveira: @ncopa The last releases versions are using this suffix. The previous versions were v4.0.0a3, v4.0.0a2, v4.0.0a1. In this case what should be the correct version number for apk?

ncopa: That depends on what the `a` in there means. It looks like there are no 4.0.0 [release](https://github.com/insanum/gcalcli/releases) which may indicate that it is some sort of pre release. Maybe the `a` means alpha? I suppose you need to ask upstream what it means.

If it is alfa releases then this may be an option:
````sh
pkgver=4.0.0_alfa4
_ver=${pkgver/_alfa/a}
...
````

ncopa:
if this is alfa release, then we probably dont want it in `main` repo, which we would end up supporting for 2 years. We could add it to testing though.

PR #4833, opened 2018-07-23
main/xen: upgrade to 4.11.0

HRio:
Add flex & bison to build deps (src/xen-4.11.0/README)

Add brige-utils to runtime deps of $pkgname-bridge


HRio:
@rnalrd @clandmeter @ncopa 

Update XEN as 4.10.1 fails to build from source in master.

Build tested and the built-in test suite passes. I have no access to a XEN dom0 running edge (only have test XEN dom0 boxes on v3.5, v3.6 and v3.7).

PR #4840, opened 2018-07-24
main/python3: upgrade to 3.7.1

viktorkholod:


andypost:
3.7.1 is out https://www.python.org/downloads/release/python-371/

Also commits needs rebase

andypost:
Just few tests needs fix
```
4 tests failed:
    test_c_locale_coercion test_embed test_ssl test_urllib2net
```

viktorkholod:
@andypost, there is no way to fix `test_ssl` with current Travis-ci config, details are here https://bugs.python.org/issue32549

andypost:
@viktorkholod the issue is not in travis but in Alpine's openssl library which in progress of transition to openssl 1.1 from libressl

main/python3/APKBUILD
```diff
@@ -27,7 +27,6 @@ prepare() {
 	cd "$builddir"
 	# force system libs
 	rm -r Modules/expat \
-		Modules/zlib \
```
andypost: why that now needed? I mean it was removed but now we gonna compile it?

andypost:
Also another question is how to test all py* packages that they works fine with 3.7

andypost:
Probably it needs `options="net"`

main/python3/APKBUILD
viktorkholod: zlib was removed from the Python source https://bugs.python.org/issue31358

viktorkholod: zlib was removed from the Python source https://github.com/python/cpython/pull/3375

andypost:
@viktorkholod  please rebase commit on current master

haizaar:
Can someone please update what's the status of Python3.7 in alpine? I see that this PR was closed without merging.

maxwell-k:
I think https://bugs.alpinelinux.org/issues/9633 is the relevant feature request in the bug tracker

haizaar:
Yes, I opened that one, and was referred here :)
Still wonder if anyone can provide more details / ETA.

PR #4841, opened 2018-07-24
community/librdkafka: upgrade to 0.11.5

secretmike:
* Upgrade to 0.11.5 https://github.com/edenhill/librdkafka/releases/tag/v0.11.5
* LibreSSL Fix (upstream: https://github.com/edenhill/librdkafka/pull/1901)
* Enable SASL support & use system LZ4 instead of bundled

### DIFF
--- a/community/librdkafka/APKBUILD
+++ b/community/librdkafka/APKBUILD
@@ -1,15 +1,17 @@
 # Contributor: Bennett Goble <nivardus@gmail.com>
 # Maintainer: Bennett Goble <nivardus@gmail.com>
 pkgname=librdkafka
-pkgver=0.11.4
-pkgrel=1
+pkgver=0.11.5
+pkgrel=0
 pkgdesc="The Apache Kafka C/C++ library"
 url="https://github.com/edenhill/librdkafka"
 arch="all"
 license="BSD"
-makedepends="bash python2 libressl-dev zlib-dev"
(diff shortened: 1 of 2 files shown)


secretmike:
Ah, looks like the latest librdkafka has an issue with libressl. librdkafka is checking `OPENSSL_VERSION_NUMBER` to enable a feature or not, and libressl doesn't support it. I'll open an issue with upstream librdkafka.

secretmike:
Possible solution proposed upstream https://github.com/edenhill/librdkafka/pull/1901

If accepted, should be possible to backport here.

secretmike:
I added the fix here as a patch file, so this should be good to go!

secretmike:
I updated the `libressl.patch` file to correspond to the simplified upstream patch here https://github.com/edenhill/librdkafka/pull/1901 .

secretmike:
This patch file has now been accepted upstream, so we can remove this patch file for `0.11.6`

andypost:
I guess patch made from this commit https://github.com/edenhill/librdkafka/commit/b1db8ccb292e12607cb9f60fe18e8a92a8e14546

secretmike:
>I guess patch made from this commit edenhill/librdkafka@b1db8cc

Yes, that's my upstream patch that's now been accepted. It will be part of the `0.11.6` release.

This PR backports the same fix to `0.11.5` which is currently the latest released version of `librdkafka`, but not yet in aports. Thanks!

secretmike:
Hey, any other changes required here?

PR #4846, opened 2018-07-24
main/py-dbus: Add python3 support

michitux:
Note that similar to py-cairo there is only one -doc and one -dev package for both python versions as these are identical for both python versions.

@ncopa Do you know why there is `make test` in `build()`? It doesn't seem to do anything, it just says `make: Nothing to be done for 'test'.`, before and after these changes.

### DIFF
--- a/main/py-dbus/APKBUILD
+++ b/main/py-dbus/APKBUILD
@@ -2,37 +2,78 @@
 pkgname=py-dbus
 _pkgname=dbus-python
 pkgver=1.2.8
-pkgrel=1
+pkgrel=2
 pkgdesc="Python bindings for DBUS"
 url="http://www.freedesktop.org/wiki/Software/DBusBindings"
 arch="all"
 license="MIT X11"
-depends="python2"
+depends=""
 depends_dev="py-dbus"
-makedepends="dbus-glib-dev python2-dev py2-sphinx"
-subpackages="$pkgname-dev $pkgname-doc"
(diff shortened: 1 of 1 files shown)


michitux:
I've just seen (should have looked before...) that there are two other PRs for the same package:

- #2452 also adds python 3 support, but imho the method here is cleaner as it does not use copies of the source, builds the package in build() and uses default doc and dev packaging methods.
- #4812 upgrades and adds a documentation build that requires further changes that let ci fail (though it seems to me that ci should pass with the changes proposed).

If the upgrade or the documentation build are wanted I can integrate them here as these changes are trivial compared to the python 3 support. Alternatively merge #4812 first, then I'll rebase my changes.

ncopa:
> @ncopa Do you know why there is make test in build()? It doesn't seem to do anything, it just says make: Nothing to be done for 'test'., before and after these changes.

We run upstream tests if there are any. It might be they add tests in future. We should do this from `check()` function, not from build.

rgdoliveira:
@ncopa @michitux regarding my commit (#4812) I think something is strange in the Travis CI. I moved a package that was missing in main (py-typing) that is required by py2-sphinx, but the CI is still failing missing the package. Is the CI building all packages at same time and not in properly order?

michitux:
There seem to be upstream tests, there is `make check` but at least in the old version it fails because it tries to run a test script that does not exist. I have not tried if it works in the current version.

z3ntu:
Any update on this PR?

PR #4849, opened 2018-07-25
main/alpine-baselayout: also create /opt

nmeum:
From `hier(7)`:

	/opt   This directory should contain add-on packages that contain
	       static files.

Briefly skipping over `hier(7)` this seems to be the only non-optional directory which alpine-baselayout doesn't create.

---

@ncopa you are the maintainer of this aport, is there any specific reason why you didn't create that directory in the first place?


ncopa:
it was probably just forgotten and nobody asked for it.

PR #4856, opened 2018-07-26
testing/openmpi: fix build hang on ppc64le by disabling builtin atomics

mksully22:
ppc64le hangs intermittently on make check when running the opal-fifo test. An issue was opened at the openmpi project, https://github.com/open-mpi/ompi/issues/5470 

The recommended workaround is to build with atomic builtins disable for powerpc64le

### DIFF
--- a/testing/openmpi/APKBUILD
+++ b/testing/openmpi/APKBUILD
@@ -1,11 +1,11 @@
 # Maintainer: Daniel Sabogal <dsabogalcc@gmail.com>
 pkgname=openmpi
 pkgver=3.1.1
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


testing/openmpi/APKBUILD
```diff
@@ -14,13 +14,20 @@ builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
 	cd "$builddir"
+
+	disable_atomics=""
```
ncopa: this should be declared local so it does not conflict with any variable used internally in abuild.
````sh
    local disable_atomics=""
````

ncopa:
I'm fixing those things for you.

mksully22:
Thanks ncopa!

PR #4861, opened 2018-07-26
testing/crosstool-ng: fix build by updating repo URLs and dependencies

meriac:
- both the project website and package source were updated, build was broken as a result.
- updated project description, dependencies and package source url
  - 'xz' is now a runtime dependency as it is commonly used by ct-ng
  - busybox 'wget' is broken when used by ct-ng for downloading packages,
    added full wget as a run- & build time dependency
  - 'make' is required both during build and runtime

### DIFF
--- a/testing/crosstool-ng/APKBUILD
+++ b/testing/crosstool-ng/APKBUILD
@@ -2,14 +2,14 @@
 pkgname=crosstool-ng
 pkgver=1.23.0
 pkgrel=1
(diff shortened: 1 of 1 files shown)


meriac:
@ncopa Would it make sense to move this package from testing to the community branch, so it becomes easier accessible to users?

meriac:
@ncopa Is there anything I can do to progress this pull request?

PR #4862, opened 2018-07-26
github: pull request template

tcely:
Adds a basic checklist for reminders about pull requests in aports

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -4,5 +4,4 @@ For this PR, I have:
 - [ ] Prefixed repo/name: directory in my commit title line
 - [ ] Started `pkgrel` at `0` for a new version or increased `pkgrel`
 - [ ] Used "upgraded to" or "new aport" phrase in title line
-- [ ] Used global `depends_`_split_ variables (e.g. `depends_dev` or `$pkgname-foo:_foo` => `depends__foo`)
 <!-- Before adding anything else, please consider if it applies to all future PRs. -->
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,8 @@
+For this PR, I have:
+<!-- aports checklist -->
+<!-- You can remove any items that don't apply. This is a helpful starting point. -->
+- [ ] Prefixed repo/name: directory in my commit title line
+- [ ] Started `pkgrel` at `0` for a new version or increased `pkgrel`
+- [ ] Used "upgraded to" or "new aport" phrase in title line
+- [ ] Used global `depends_`_split_ variables (e.g. `depends_dev` or `$pkgname-foo:_foo` => `depends__foo`)
+<!-- Before adding anything else, please consider if it applies to all future PRs. -->


.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+For this PR, I have:
+<!-- aports checklist -->
+<!-- You can remove any items that don't apply. This is a helpful starting point. -->
+- [ ] Prefixed repo/name: directory in my commit title line
+- [ ] Started `pkgrel` at `0` for a new version or increased `pkgrel`
+- [ ] Used "upgraded to" or "new aport" phrase in title line
+- [ ] Used global `depends_`_split_ variables (e.g. `depends_dev` or `$pkgname-foo:_foo` => `depends__foo`)
```
jirutka: How did you came to this? abuild supports only `depends_dev`, that's a special case.

ncopa: yes, this line needs to be removed as it is wrong.

ncopa:
@jirutka are there other common mistakes in APKBUILDs you think we should mention in this checklist?

jirutka:
> are there other common mistakes in APKBUILDs you think we should mention in this checklist?

The list would be long as hell if we mention all of the common mistakes. This is not the right place for this.

ncopa:
> This is not the right place for this.

Where do you think is the right place for checklist of common mistakes?

jirutka:
> Where do you think is the right place for checklist of common mistakes?

Documentation, as a part of the packaging guidelines (something as tl;dr section); in our case it's currently wiki.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
tcely: I put that line in based on previous advice from @ncopa  that depends need to be defined globally rather than in split functions.

I understand that `default_dev` has a depends line using `depends_dev` already. Extending this pattern just makes the most sense to me for custom split functions too.

tcely:
> This is not the right place for this.

I strongly disagree with this statement. Putting this information anywhere outside the normal flow of work for creating a pull request is just asking for it to be overlooked.

I frankly don't care if this turns into a book that people routinely delete before submitting a new pull request as long as they are going to routinely at least see it. Making them go hunting through a wiki or another external website is a less than useful option in my opinion.

tcely:
It's true, I haven't seen a list like this for AlpineLinux aports yet. If you have one I haven't found, I'd appreciate a pointer. Otherwise, I think we should all put anything we think should be commonly checked into a list then decide on which of those things should be in a pull template.

I really don't think we will need to let it get massive to reduce the number of times we tell people things in PR reviews.

tcely:
@jirutka The S-changes-requested  label is outdated.

ncopa:
Would it be better to point them to https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md?

Have a few (4-5) simple check items, like:

[ ] Verify that commit message is according [guidelines](https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md)
[ ] pkgver and pkgrel are valid (link to documentation explaing how pkgver and pkgrel should be)
etc.

So that we don't duplicate the policies and need to change multiple places if when they change.

What do you think about that?

tcely:
That looks like a good first step to me!

PR #4872, opened 2018-07-28
main/linux-vanilla: Add missing build dependencies

deedubs:
Attempted to get kernel building on CircleCI has highlighted these missing deps

flex and bison were causing the configure step to fail

openssl-dev is also required during the build

```
/home/alpine/project/testing/linux-vanilla/src/linux-4.17/scripts/extract-cert.c:21:25: fatal error: openssl/bio.h: No such file or directory
 #include <openssl/bio.h>
```


main/linux-vanilla/APKBUILD
```diff
@@ -12,7 +12,7 @@ pkgdesc="Linux vanilla kernel"
 url="http://kernel.org"
 depends="mkinitfs"
 _depends_dev="perl gmp-dev elfutils-dev bash"
-makedepends="$_depends_dev sed installkernel bc linux-headers linux-firmware"
+makedepends="$_depends_dev flex bison sed installkernel bc linux-headers linux-firmware openssl-dev"
```
ncopa: we currently use libressl, so if this is really needed, then it should be `libressl-dev`.

I wonder why we don't hit this issue on our builders?

deedubs:
Yeah I was chatting about this with _ikke_ on IRC. The assumption was that the build machines aren't as minimal as originally thought. 

I'll update to libressl-dev.

deedubs:
@ncopa Updated to `libressl-dev`

ncopa:
i still don't understand why this is needed. I can not reproduce the problem you are trying to solve.

deedubs:
Sorry, I forgot you were missing some context from IRC.  The builds are failing from a 'clean' build env with missing deps https://circleci.com/gh/massivekube/aports/55.

ncopa:
aha, its for 4.17 kernel.

> linux-vanilla: Applying patch-4.17.8.xz

I suppose we don't need those til we upgrade the kernel.

ncopa:
this should be fixed with d8adf28a6ff7a7527a2d078f5800067b2f506634

PR #4875, opened 2018-07-28
testing/dd_rescue: new aport

jdrowell:
http://www.garloff.de/kurt/linux/ddrescue/
Like dd, but allows for copying from devices with bad blocks


prspkt:
Does `testing/dd_rescue` offer anything in particular over [`community/ddrescue`](https://www.gnu.org/software/ddrescue/ddrescue.html)?
If not, this will mostly add confusion and maintenance overhead.

jdrowell:
They accomplish similar things, but in different ways and with different options. We use dd_rescue for forensic imaging and it is superior to ddrescue in that case (in my opinion of course). Also if I remember correctly ddrescue cannot be piped to stdout, which prevents parallel hashing and compressing.

jdrowell:
Bernhard, I just checked this and it seems that only `autoconf` is required. Should I close this pull request and open a new one with that change?

BernhardGruen:
@jdrowell Please don't close/delete this PR. Just push your changes (if possible in a squashed commit).

@prspkt dd_rescue and ddrescue are two different projects. Both of them are similar but I already had the case that one tool was able to recover data and the other was not able to do so - and vice versa.
Even if I personally don't need those packages in Alpine. I think that having both is great from a users perspective.

BernhardGruen:
Any progress on this @jdrowell ?

jdrowell:
@BernhardGruen the commit above your comment. Is there anything else to do?

BernhardGruen:
@jdrowell Thing is, Travis does not build dd_rescue at the moment (please see the log).
I am not sure if this is really a packaging error or was an error with Travis. Maybe someone is able to answer this.

jdrowell:
@BernhardGruen Oh I see, I thought it was some problem with Travis because it was building fine inside a clean alpine-3.8 container. I looked at the logs and apparently more patching is needed to handle error.h. I'll look into it.

PR #4878, opened 2018-07-28
community/php7-apcu: fix license, use https, add missing file

andypost:
- SPDX identifier fixed according https://pecl.php.net/package/APCu
- added apc.php


jirutka:
I've merged it without the license change. As we've discussed, please open a PR with mass fix of the license in all PHP pkgs.

jirutka:
Should this be backported to 3.8-stable?

andypost:
Makes sense to backport because this file useful for monitoring of shared cache

PR #4883, opened 2018-07-28
community/php7: fix licenses

andypost:
Checked all libs, follow-up to https://github.com/alpinelinux/aports/pull/4879

### DIFF
--- a/community/php7/APKBUILD
+++ b/community/php7/APKBUILD
@@ -4,18 +4,19 @@
 
 # Bundled libraries
 #
-# Name      | License | Location               | State
-# ----------+---------+------------------------+---------
-# bcmath    | LGPL2   | ext/bcmath/libbcmath   | used
-# date      | MIT     | ext/date/lib           | used
-# gd        | BSD     | ext/gd/libgd           | used
-# file      | BSD     | ext/fileinfo/libmagic  | used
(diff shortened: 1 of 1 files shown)


andypost:
Let's use it to fix PHP licenses  for all community packages

andypost:
Does it makes sense to increase `pkgrel`? @jirutka

andypost:
Also commit message needs work

ncopa:
> Does it makes sense to increase `pkgrel`?

Yes. It changes the content so `apk info --license` will display different data. `pkgrel` does not need to be increased when a rebuild produces exactly the same binary (in theory - we still have a way to go to be 100% reproducible)

community/php7/APKBUILD
```diff
@@ -34,7 +35,7 @@ _default_php="yes"
 pkgdesc="The PHP$_suffix language runtime engine"
 url="https://secure.php.net"
 arch="all"
-license="PHP-3.0 BSD LGPL-2.0 MIT Zend"
+license="PHP-3.01 BSD LGPL-2.0-or-later MIT Zend-2.0"
```
ncopa: pkgrel needs to be bumped due to this change

andypost:
@ncopa rebased with bump

ncopa:
fixed with 01e1a5c92706c33217712f5f2cd5aede165bb297

PR #4891, opened 2018-07-29
Redmine #9113 Create /run/redis for docker users

skwashd:
Without this path redis fails to start under docker.

### DIFF
--- a/main/redis/redis.pre-install
+++ b/main/redis/redis.pre-install
@@ -2,5 +2,6 @@
 
 addgroup -S redis 2>/dev/null
 adduser -S -D -H -h /var/lib/redis -s /bin/false -G redis -g redis redis 2>/dev/null
+install -o redis -g redis -m 0770 -d /run/redis
 
 exit 0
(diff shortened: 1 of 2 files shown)


main/redis/redis.pre-install
```diff
@@ -2,5 +2,6 @@
 
 addgroup -S redis 2>/dev/null
 adduser -S -D -H -h /var/lib/redis -s /bin/false -G redis -g redis redis 2>/dev/null
+install -o redis -g redis -m 0770 -d /run/redis
```
andypost: when package changes you need to increase pkgrel

skwashd: Good catch. Fixed

skwashd: @clandmeter this is needed for redis to work under docker. Dockerised services should be run in the foreground not as a daemon managed by initd. See [Remind issue 9113](https://bugs.alpinelinux.org/issues/9113) and linked issues for more info.

andypost: See https://git.alpinelinux.org/cgit/aports/commit/?id=c4d4e9aa7abd163db3a9a4c4e6d6ba4f84d3a64f

skwashd:
Yup @jirutka beat me to it. Closing.

PR #4893, opened 2018-07-30
main/py-google-api-python-client: improved dependencies

maxwell-k:
py3-google-api-python-client should not depend upon python2, python3 is
sufficient.

Lines 30 and 40 add the python2 dependency to
py2-google-api-python-client so specifying python2 on line 10 is not
required.

Fixes issue #9156


rgdoliveira:
This is ok to merge. @ncopa Can you please merge it?

PR #4896, opened 2018-07-30
testing/apparmor: move apparmor_parser to /sbin

allgdante:
There are several applications that have /sbin/apparmor_parser
hardcoded, so it's better to move it and not break things


allgdante:
This change will be addressed in a new version

PR #4898, opened 2018-07-30
main/mesa: Upgrade to 18.1.7, cosmetic changes

tanertas:


tanertas:
It seems have issues with latest kernels. Closed until find the cause of segfaults.

tanertas:
I was getting segfaults only with Urbanterror. Xonotic and glxgeras were fine. I don't get any segfaults after a reboot.

PR #4899, opened 2018-07-30
testing/aws-cli: update to 1.15.66

maxwell-k:


maxwell-k:
Thanks @fcolista ! I appreciate your help. 

Could you possibly take a look at https://github.com/alpinelinux/aports/pull/4897 please?  It solves a bug with `aws s3 help` which is why I started this update in the first place. I understand if we need to wait for  @vakartel 

Thanks again

PR #4905, opened 2018-08-01
community/z3: fix ppc64le test failures by setting proper element width in small allocator 

mksully22:
test-z3 doc test fails because pointer alignment is incorrectly set to 4 bytes for ppc64le. This causes the small_object_allocator::allocate() to incorrectly set the size of the global m_free_list elements to to be too small. The net result is the 8 byte pointer used to link chunks on the free list to overflow into the next element effectively corrupting the free chain. This fix sets the PTR_ALIGNMENT value correctly for ppc64le to use 8 byte elements.

### DIFF
--- a/community/z3/APKBUILD
+++ b/community/z3/APKBUILD
@@ -2,16 +2,17 @@
 # Maintainer: Sören Tempel <soeren+alpine@soeren-tempel.net>
 pkgname=z3
 pkgver=4.7.1
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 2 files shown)


mksully22:
I believe Travis failed CI because the log generated by building this package is too long.

ncopa:
works in my dev env. i'm pushing it. thanks!

mksully22:
This fix might work for s390 and/or aarch64 if someone has a system of this type to give it a try?

PR #4906, opened 2018-08-01
testing/frr: break out a pythontools subpackage to clean up dependencies

ajones-rvbd:
For convenience, frr-reload.py is now packaged with the main package,
but that brings in python2.  Here, we break out an frr-pythontools package
so that we can keep python2 out of the base package.

Fixes #9172

### DIFF
--- a/testing/frr/APKBUILD
+++ b/testing/frr/APKBUILD
@@ -1,14 +1,14 @@
 # Maintainer: Arthur Jones <arthur.jones@riverbed.com>
 pkgname=frr
 pkgver=5.0.1
-pkgrel=0
+pkgrel=1
 pkgdesc="FRRouting is a fork of quagga"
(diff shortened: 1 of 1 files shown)


Shizmob:
I think this can just be called `frr-tools`. Also allows omitting `python` in its `pkgdesc=`.

ajones-rvbd:
Hi @Shizmob, thanks for having a look, I totally agree that frr-tools is a more fitting name for this...  There are three references in the code to frr-pythontools and none to frr-tools.  Also, the debian and the redhat packaging uses frr-pythontools.  So, even though frr-tools is more appropriate, it looks like it will be easier for users if we keep consistent naming.  However, I'll update the PR w/ the pkgdesc change, cuz your suggestion seems a lot clearer to me...

PR #4907, opened 2018-08-01
community/zabbix: upgrade to 3.4.12

akamch:


Shizmob:
`WARNING: zabbix*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc`

Could you add a `$pkgname-openrc` subpackage to `subpackages=`? abuild will take care of its implementation by default.

PR #4911, opened 2018-08-01
community/zabbix: Fix zabbix loadable modules work

am813nt:
Zabbix loadable modules functionality requires that all the symbols in a binary (agent, server or proxy) should be added to dynamic symbol table. Zabbix executables use dlopen() to load modules (shared libraries), which in turn refer to symbols defined in the executable. So --export-dynamic is required to be passed in such cases. Without it zabbix binaries can't load any modules (errroring about undefined symbols in modules).

### DIFF
--- a/community/zabbix/APKBUILD
+++ b/community/zabbix/APKBUILD
@@ -4,7 +4,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=zabbix
 pkgver=3.4.8
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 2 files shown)


endersonmaia:
`A-fix` or `A-improve` tags wanted

Shizmob:
Can you bump `pkgrel` for this change, so that builders (and apk) will pick up that there's a new version?

Shizmob:
There were some whitespace errors and I fixed up the commit message, otherwise good. Thanks!

PR #4920, opened 2018-08-02
main/grub: remove syslinux dependency

losynix:
Avoid sourcing another booloader's configuration files (syslinux's /etc/update-extlinux.conf) when using grub which has it's own (/etc/default/grub).

### DIFF
--- a/main/grub/2.02_beta3-mkconfig-alpine.patch
+++ b/main/grub/2.02_beta3-mkconfig-alpine.patch
@@ -1,17 +1,5 @@
 --- a/util/grub.d/10_linux.in
 +++ b/util/grub.d/10_linux.in
-@@ -21,8 +21,11 @@
- exec_prefix="@exec_prefix@"
- datarootdir="@datarootdir@"
- 
-+. /etc/update-extlinux.conf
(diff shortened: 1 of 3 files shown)


losynix:
(checks failed because of the log length has exceeded the limit of 4 MB)

This fixes bug #8601 : https://bugs.alpinelinux.org/issues/8601

ncopa:
I think users of grub may end up with unbootable systems if i push this as is. Specially if they need some non-default boot options.

We probably also need fix setup-disk at the same time.

losynix:
> I think users of grub may end up with unbootable systems if i push this as is. Specially if they need some non-default boot options.

I guess you're right.

> We probably also need fix setup-disk at the same time.

True, I haven't really tested it but something like this should do (in setup-disk):

```
@@ -1215,7 +1215,9 @@
-       apk add --quiet syslinux
+    [ -z "$BOOTLOADER" ] && BOOTLOADER=$(select_bootloader)
+    apk add --quiet "$BOOTLOADER"

```

ncopa:
fixed with cb6c7c4b66dc4640425f875c7d9545dad9e7823c

PR #4922, opened 2018-08-02
new aport: libyang, a YANG parser and toolkit

ajones-rvbd:
libyang provides a C interface into YANG data models, which are
becoming increasingly important for network configuration and
operational state.  There are patches under review in frr to
use libyang, and the yanglint utility (in the utils package)
is very helpful for debugging and creating yang models.

Fixes: #9133

### DIFF
--- a/testing/libyang/APKBUILD
+++ b/testing/libyang/APKBUILD
@@ -1,0 +1,47 @@
+# Maintainer: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=libyang
+_srcver=0.15-r1
+pkgver=${_srcver/-r/.}
+pkgrel=0
+pkgdesc="yang parser and validator"
+url="https://github.com/CESNET/libyang"
+arch="x86_64"
+license="BSD-3-Clause"
+makedepends="cmake pcre-dev cmocka-dev bison flex doxygen"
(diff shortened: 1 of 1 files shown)


testing/libyang/APKBUILD
```diff
@@ -0,0 +1,47 @@
+# Maintainer: Arthur Jones <arthur.jones@riverbed.com>
+pkgname=libyang
+_srcver=0.15-r1
```
rwestphal: @ajones-rvbd could you update this to the latest release (0.15-r2)?

testing/libyang/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+subpackages="$pkgname-dev $pkgname-doc $pkgname-dbg $pkgname-utils"
+source="https://github.com/CESNET/libyang/archive/v${_srcver}.tar.gz"
+builddir="$srcdir"/$pkgname-$_srcver
+
+build() {
+	cd "$builddir"
+	mkdir build
+	cd build
+	cmake \
+		-DENABLE_BUILD_TESTS=ON \
```
rwestphal: Please add the `-DENABLE_LYD_PRIV=ON` option which is not enabled by default but is required by FRR.

PR #4923, opened 2018-08-02
testing/php7-xhprof: new aport

andypost:
Profiler for PHP 7 https://github.com/longxinH/xhprof

### DIFF
--- a/testing/php7-xhprof/APKBUILD
+++ b/testing/php7-xhprof/APKBUILD
@@ -1,0 +1,54 @@
+# Contributor: Andy Postnikov <apostnikov@gmail.com>
+# Maintainer: Andy Postnikov <apostnikov@gmail.com>
+
+pkgname=php7-xhprof
+_pkgreal=xhprof
+pkgver=2.0.2
+pkgrel=0
+pkgdesc="A Hierarchical Profiler for PHP"
+url="https://github.com/longxinH/xhprof"
+arch="all"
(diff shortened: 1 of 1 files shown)


testing/php7-xhprof/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	make NO_INTERACTION=1 REPORT_EXIT_STATUS=1 -f Makefile.local test_with_exit_status
+}
+
+package() {
+	cd "$builddir"
+	make INSTALL_ROOT="$pkgdir"/ install
+	install -d "$pkgdir"/etc/php7/conf.d
+	echo "extension=$_pkgreal.so" > "$pkgdir"/etc/php7/conf.d/$_pkgreal.ini
+	install -d  "$pkgdir"/usr/share/php7/xhprof
+	cp -R ../xhprof_*  "$pkgdir"/usr/share/php7/xhprof/
```
andypost: probably `xhprof_html` needs to be moved to separate package because it depends on `ghostscript-fonts` & `graphvis`, also it probably needs some webserver with PHP (at least build-in into php works great `php -S`)

also it require `xhprof_libs` which needs `php7-ctype`

andypost: as I see it is not common to separate "-ui" packages
```
1.1M	../xhprof_html/
116.0K	../xhprof_lib/
```

PR #4926, opened 2018-08-03
community/cassandra-cpp-driver: upgrade to 2.9.0

tcely:


community/cassandra-cpp-driver/APKBUILD
```diff
@@ -1,9 +1,9 @@
 # Contributor: Gennady Feldman <gena01@gmail.com>
 # Maintainer: Gennady Feldman <gena01@gmail.com>
 pkgname=cassandra-cpp-driver
-_pkgreal=cpp-driver
-pkgver=2.8.1
-pkgrel=3
+_pkgname=cpp-driver
```
andypost: not sure this change reasonable

tcely: @andypost Why not?

Shizmob: LGTM.

PR #4929, opened 2018-08-03
testing/log4cplus: upgrade to 2.0.1

tcely:
Kept log4cplus1 as 1.2.1 for supporting < C++11

### DIFF
--- a/testing/log4cplus1/APKBUILD
+++ b/testing/log4cplus1/APKBUILD
@@ -1,7 +1,8 @@
 # Contributor:
 # Maintainer:
-pkgname=log4cplus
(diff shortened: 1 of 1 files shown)


Shizmob:
This seems like a broken upgrade path for anything that depends on it, though -- it will get the new version upon `apk upgrade` and not the 1.x it depended on.

tcely:
@Shizmob True enough, but since `kea` is the only thing depending on this, I'm willing to live with that problem.

ncopa:
Do we really need `log4cplus1` if its only `kea` that uses this?

PR #4932, opened 2018-08-04
main/dhcp: improve dhclient package

tcely:


main/dhcp/APKBUILD
```diff
@@ -8,6 +8,9 @@ url="https://www.isc.org/"
 arch="all"
 license="ISC"
 depends=
+# busybox provided features are good enough for these
+#depends_dhclient_optional="bash iputils net-tools sed util-linux"
```
ncopa: i would prefer to change comment to something like:
````sh
# busybox utils are not sufficient for dhclient
depends_dhclient="coreutils iproute2 run-parts"
````

File: main/dhcp/APKBUILD [hunk 2]
```diff
@@ -8,6 +8,9 @@ url="https://www.isc.org/"
...
+depends_dhclient="coreutils iproute2 run-parts $depends_dhclient_optional"
 pkgusers="dhcp"
 pkggroups="dhcp"
 makedepends="perl linux-headers file"
```
ncopa: we should also have depends_dhclient in make depends to make sure those are built before `dhcp`.
````sh
makedepends="$depends_dhclient ...."
````

File: main/dhcp/dhclient-script-alpine [hunk 3]
```diff
@@ -0,0 +1,682 @@
+#!/bin/sh
+# dhclient-script for Linux. Dan Halbert, March, 1997.
+# Updated for Linux 2.[12] by Brian J. Murrell, January 1999.
+# No guarantees about this. I'm a novice at the details of Linux
+# networking.
```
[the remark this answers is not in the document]
tcely: While we could maintain a patch, I don't think that would be better. A fork is easier to maintain than a patch.

As for submitting upstream, we can try, but I don't have a lot of hope since ISC has moved their efforts to kea at this point.

ncopa:
The changes compared to upstream script is pretty massive. @tcely are you willing to take maintainership of this package?

I am hesitant to take responsibility for this alpine specific dhclient script.

ncopa:
related bug report  https://bugs.alpinelinux.org/issues/9553

PR #4937, opened 2018-08-04
main/ansible: security upgrade to 2.6.3

andypost:
Changes https://github.com/ansible/ansible/blob/stable-2.6/changelogs/CHANGELOG-v2.6.rst#v263


andypost:
Not sure about need to backport cos low risk but probably makes sense

@fabaff any thoughts?

PR #4938, opened 2018-08-04
main/busybox: upgrade to 1.29

nmeum:
* Remove all patches already applied upstream
    * 0001-ash-add-support-for-command_not_found_handle-hook-fu.patch
    * 0001-cat-fix-cat-e-and-cat-v-erroneously-numbering-1st-li.patch
    * 0001-wget-emit-a-message-that-certificate-verification-is.patch
    * 0015-ash-introduce-a-config-option-to-search-current-dire.patch
    * 0016-top-handle-much-larger-VSZ-values.patch
    * 0017-ifupdown-do-not-fail-if-interface-disappears-during-.patch
* Rename config option for command_not_found hook
    * upstream patch adding this hook slightly differs from our
      downstream patch in this regard
* Rebase some patches manually:
    * external_ssl_client.patch
    * 0006-ping-make-ping-work-without-root-privileges.patch
    * 0007-fbsplash-support-image-and-bar-alignment-and-positio.patch
* Add support for `-e` to our ssl_client
    * See https://git.busybox.net/busybox/commit/?id=403f2999f94937ba3f37db6d093832f636815bb9
* Update the configuration file
* Regenerate all patches using `git format-patch --no-numbered --no-signature`
  to reduce the diff for future upgrades.


main/busybox/busyboxconfig
```diff
@@ -901,6 +910,8 @@ CONFIG_NETSTAT=y
 CONFIG_FEATURE_NETSTAT_WIDE=y
 CONFIG_FEATURE_NETSTAT_PRG=y
 CONFIG_NSLOOKUP=y
+# CONFIG_FEATURE_NSLOOKUP_BIG is not set
```
nmeum: This option has the following effect:

> Use internal resolver code instead of libc

Do we want to enable it? I would personally prefer not to.

main/busybox/ssl_client.c
```diff
@@ -89,16 +89,19 @@ static void copy_loop(struct tls *ctx, int sfd)
 
 	while (1) {
 		int r = do_poll(fds, 2);
-		if (fds[0].revents)
+		if (fds[0].revents) {
 			copy_from_stdin_to_tls(ctx, &fds[0].fd);
+			if (eofexit && fds[0].fd == -1)
```
nmeum: **TODO:** Is this implemented correctly?

main/busybox/0001-properly-fix-wget-https-support.patch
```diff
@@ -30,23 +40,29 @@ index cd92b3a28..a12c921cd 100644
...
  			argv[0] = (char*)"ssl_client";
- 			argv[1] = (char*)"-s3";
- 			//TODO: if (!is_ip_address(servername))...
+@@ -750,7 +749,13 @@ static void spawn_ssl_client(const char *host, int network_fd, int flags)
  			argv[2] = (char*)"-n";
  			argv[3] = servername;
--			argv[4] = NULL;
+ 			argv[4] = (flags & TLSLOOP_EXIT_ON_LOCAL_EOF ? (char*)"-e" : NULL);
+-			argv[5] = NULL;
 +			if (!ENABLE_SSL_CLIENT &&(option_mask32 & WGET_OPT_NO_CHECK_CERT)) {
```
nmeum: This doesn't work because `argv[4]` might already be null if `!(flags & TLSLOOP_EXIT_ON_LOCAL_EOF)`.

fabled:
> 0008-fbsplash-support-image-and-bar-alignment-and-positio.patch was removed since upstream seems to have implemented configurable image positions. @fabled you original committed the patch, are you ok with removing it?

Alpine initramfs init script uses `IMAGE_ALIGN=CM` currently to center the splash image. I kinda like it that way, but alternate ideas would be good too.

nmeum:
I totally forgot that this is used by initramfs. For a splash image centering the image does indeed make sense. We could just add an `IMG_CENTER` option to `fbsplash` based on [the upstream changes](https://git.busybox.net/busybox/commit/?id=a82fe671f5c7ff96ca4d4d813210d18e0b02c193) to support this, rebasing the old patch doesn't make much sense imho. However, since I don't use a splash image I am currently not motivated enough to do that.

nmeum:
> Alpine initramfs init script uses IMAGE_ALIGN=CM currently to center the splash image.

I totally forgot that the initramfs uses this for the splash image. IMHO centering the splash image is a good idea.

> I kinda like it that way, but alternate ideas would be good too.

We could implement image centering through a shell script utilizing the `IMG_{LEFT,TOP}` configuration variables introduced in `1.29` and `fbset` to retrieve the screen size, e.g.:

```shell
#!/bin/sh

[ $# -eq 1 ] || exit 1
image="${1}"

sw=$(fbset | awk '/geometry/ {print $2}')
sh=$(fbset | awk '/geometry/ {print $3}')

# See: http://netpbm.sourceforge.net/doc/ppm.html
id=$(sed -e '/^#.*/d' -e 's/[[:space:]]/ /g' "${image}" | \
	tr '\n' ' ' | cut -d ' ' -f2,3)
iw=$(printf "%s\n" "${id}" | cut -d ' ' -f1)
ih=$(printf "%s\n" "${id}" | cut -d ' ' -f2)

truncate -s0 config.center
printf "IMG_LEFT=%d\nIMG_TOP=%d\n" \
	$(((sw - iw) / 2)) $(((sh - ih) / 2)) > config.center

fbsplash -i ./config.center -s "${image}"
```

nmeum:
> Any thoughts on this?

I just added an additional patch to implement such a configuration generation procedure in `initramfs-init`. I didn't test this yet though.

main/mkinitfs/0001-initramfs-init-center-the-image-using-the-IMG_-TOP-L.patch
```diff
@@ -0,0 +1,92 @@
...
++	# Normalize whitespaces, extract dimensions (format as above).
++	local id=$(sed -e '/^#.*/d' -e 's/[[:space:]]/ /g' "$image" | \
++		tr '\n' ' ' | awk -vOFS=x '{ print($2, $3) }')
++
++	local sw=${sd%%x*}
++	local sh=${sd##*x}
++	local iw=${id%%x*}
++	local ih=${id##*x}
++
++	# TODO: What should we do if the image
```
nmeum: **TODO**

fabled:
I'd really not use shell math like that. I can rebase fbsplash patch in near future and resubmit to bb.

nmeum:
Alright, fine with me as well.

andypost:
https://busybox.net 1.29.2 id out

nmeum:
> I can rebase fbsplash patch in near future and resubmit to bb.

Any updates on this?

nmeum:
Received a rebased version of the fbsplash patch from @fabled included it and squashed and cleanup everything while at it. Currently testing this a bit in my production setup. The first problem I noticed is that the state detection of ifdown is currently broken. Meaning it doesn't correctly recognize that an interfaces has been configured successfully. Currently trying to figure out why this is the case. Further tests of the changes purposed here are highly appreciated.

nmeum:
> The first problem I noticed is that the state detection of ifdown is currently broken.

This was fixed in 1.29.3 I upgraded the aport accordingly and squashed the changes.

I am currently not aware of any problems with the changes purposed here. I am therefore no longer considering this WIP. However, a thorough review of the changes proposed here would be nice. This is especially true for manually rebased patches and the changes to ssl_client.

andypost:
1.30 released https://busybox.net/news.html

PR #4939, opened 2018-08-04
community/bitcoin: fix $pkgrel

itoffshore:
community/bitcoin: fix `$pkgrel`


andypost:
Not sure it makes sense to lower version

PR #4942, opened 2018-08-05
testing/bchunk: new aport

9001:
conversion utility for CD image files

### DIFF
--- a/testing/bchunk/APKBUILD
+++ b/testing/bchunk/APKBUILD
@@ -23,7 +23,6 @@ package() {
 		"$pkgdir/usr/bin" \
 		"$pkgdir/usr/share/man/man1"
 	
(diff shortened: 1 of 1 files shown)


testing/bchunk/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+	make
+}
+
+package() {
+	cd "$builddir"
+	mkdir -p \
+		"$pkgdir/usr/bin" \
+		"$pkgdir/usr/share/man/man1"
+	
+	sed -ri '/^(BIN|MAN)_DIR /d' Makefile
```
Shizmob: This shouldn't be needed, as `make X=Y` on the command line will override variable `X` in the Makefile already.

9001: you're right - must have been a leftover I missed :p

PR #4944, opened 2018-08-05
main/postfix: security upgrade to 3.3.2

tcely:
- ~~added `m4` to `makedepends`~~
- added `-openrc` subpackage
- added optional license choice as a comment
- removed obsolete `libressl.patch`
- sorted `makedepends` list
- switched to a `https` mirror instead of `ftp` site


tcely:
@clandmeter I rebased this onto your recent commit to master.

main/postfix/APKBUILD
```diff
@@ -3,32 +3,48 @@
...
-	lmdb-dev"
+makedepends="
+	coreutils
+	cyrus-sasl-dev
+	db-dev
+	file
+	libressl-dev
+	linux-headers
+	lmdb-dev
+	m4
```
ncopa: what is `m4` needed for?

main/postfix/APKBUILD
```diff
@@ -3,32 +3,48 @@
...
+	perl
+	postgresql-dev
+	sqlite-dev
+	"
+
 install="$pkgname.pre-install"
-subpackages="$pkgname-doc
+subpackages="
+	$pkgname-doc
+	$pkgname-openrc
```
ncopa: :+1:

main/postfix/APKBUILD
```diff
@@ -3,32 +3,48 @@
...
 	$pkgname-sqlite:_mv_dict
-	$pkgname-stone"
+	$pkgname-stone
+	"
 
 options="!check suid"  # No test suite.
 pkgusers="postfix"
 pkggroups="postfix postdrop"
-source="ftp://ftp.porcupine.org/mirrors/${pkgname}-release/official/${pkgname}-${pkgver}.tar.gz
+source="https://archive.mgm51.com/mirrors/$pkgname-source/official/$pkgname-$pkgver.tar.gz
```
ncopa: :+1:

main/postfix/APKBUILD
```diff
@@ -3,32 +3,48 @@
...
 	$pkgname.initd
 	postfix-install.patch
-	libressl.patch
```
ncopa: You need to also `git rm libressl.patch` to the patch is removed from git repo.

ncopa:
you may also need to change commit message since this no longer is an upgrade to 3.3.1.

tcely:
Fixe commit message and rebased onto master

andypost:
Needs rebase and update to  security release 3.3.2 for https://www.mail-archive.com/postfix-announce@postfix.org/msg00050.html

tcely:
@andypost @ncopa Rebased as requested.

main/postfix/APKBUILD
```diff
@@ -2,33 +2,49 @@
...
 pkgname=postfix
-pkgver=3.3.1
-pkgrel=4
+pkgver=3.3.2
+pkgrel=0
 pkgdesc="Secure and fast drop-in replacement for Sendmail (MTA)"
 url="http://www.postfix.org/"
 arch="all"
 license="IPL-1.0"
+#license="EPL-2.0"
```
andypost: that looks like forgotten

tcely: Do you have a suggestion for software with multiple licenses to choose from?

andypost: All of them should be listed, check php7 as example

tcely: I checked php7 and that is not a case of choice of license. That situation is all of those licenses apply because of the modules included.

I'm not sure how best to handle this.
> Dual license: in addition to the historical IBM Public License 1.0, Postfix is now also distributed with the more recent Eclipse Public License 2.0. Recipients can choose to take the software under the license of their choice. Those who are more comfortable with the IPL can continue with that license.

http://www.postfix.org/announcements/postfix-3.3.0.html

tcely: I found `zstd` has the same situation and lists both. I'll do the same.

@ncopa This really should be handled some way that makes it clear the license is a choice. Perhaps `alpine-devel` is a better place to discuss the situation.

J0WI:
Since this is a security upgrade it would be great to have it in 3.9.

PR #4946, opened 2018-08-05
main/cython: add py2/py3 subpackages + tests

9001:
cython was renamed to py-cython and given py2/py3 subpackages.  
py-pgen is impractical to merge into one aport, so a separate py3-pgen was added.


9001:
the py2-pgen build failed due to a connection loss between the travis instance and www.python.org while the python.tgz was downloading... took that as an excuse to swap out the links with https ones :)

might be a good idea to retry downloading a couple times, in case of transient errors like these?

9001:
Travis timed out after 50 minutes before completing the py2 tests, so under halfways considering there are py3 tests as well. The tests are CPU-bound and parallelize to the number of jobs specified in /etc/abuild.conf, so this would be less of a problem on dedicated build boxes.

Would it be preferable to only run tests for py3 to cut the processing time in half?

Shizmob:
LGTM, with two remarks:

- Maybe the `-pgen` packages should be generated as part of the main Python builds (cc @itoffshore ?);
- I think we should keep running the test suite under both Pythons, can the build server load handle that @ncopa?

Other than that, good for merge by me.

itoffshore:
@Shizmob - see https://github.com/alpinelinux/aports/pull/4963 for tidy up of `pgen`

jirutka:
I'd prefer to rebuild cython against Python 3 and drop support for Python 2 (legacy).

PR #4947, opened 2018-08-05
testing/lsmash: new aport

9001:
MP4 and MOV muxer/demuxer  
Suggesting we use non-standard names for the installed binaries, as they are likely to cause collisions with other packages (/usr/bin/muxer for instance)

### DIFF
--- a/testing/lsmash/APKBUILD
+++ b/testing/lsmash/APKBUILD
@@ -17,7 +17,6 @@ build() {
 	./configure \
 		--prefix=/usr \
 		--extra-cflags=-fPIC \
-		--disable-static \
 		--enable-shared
 	make
 }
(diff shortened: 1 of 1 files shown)


testing/lsmash/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+subpackages="$pkgname-dev"
+source="lsmash-$pkgver.tar.gz::https://github.com/l-smash/l-smash/archive/v$pkgver.tar.gz"
+builddir="$srcdir/l-smash-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure \
+		--prefix=/usr \
+		--extra-cflags=-fPIC \
+		--disable-static \
```
Shizmob: Is there a reason for disabling static builds here?

9001: I thought that was an Alpine policy, forcing everything to dynamic link -- the package and everything that depends on it of course builds just fine with both dynamic and static libs available for linking, so I'll remove that flag.

PR #4950, opened 2018-08-05
Fix reloading of existings ipsets.

liske:
This fixes issue [#9149](https://bugs.alpinelinux.org/issues/9149).

### DIFF
--- a/main/ipset/APKBUILD
+++ b/main/ipset/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Kaarle Ritvanen <kaarle.ritvanen@datakunkku.fi>
 pkgname=ipset
 pkgver=6.38
-pkgrel=0
+pkgrel=1
 pkgdesc="Manage Linux IP sets"
 url="http://ipset.netfilter.org/"
 arch="all"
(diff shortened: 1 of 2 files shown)


jirutka:
Please bump `pkgrel` and update checksums (`abuild checksum`).

liske:
I've updated the PR accordingly @jirutka

ncopa:
@kunkku can you please have a look at this?

kunkku:
The sets are created using a single `ipset restore` command due to performance reasons. This is significant when the number of sets is big. I suggest splitting the loop into two: first collect the names of the existing sets, then generate the creation commands in the second loop.

liske:
I was not aware of the performance impact. I've split the loop as prefered by @kunkku.

kunkku:
Thank you. Seems all right to me.

liske:
Is this fixed to be merged into alpine 3.8? For me this issue looks a little bit serious since ipset is mostly used by iptables rules. *Silently* failing to update the firewall policy might be a security related?

kunkku:
Now backported to all supported releases.

PR #4951, opened 2018-08-05
testing/mutter: upgrade Mutter to 3.28

PureTryOut:
GNOME can't be installed right now due to a partly done upgrade to 3.28. More packages will have to be upgraded after this one, but I'll do it in separate PR's due to requiring new dependencies.


PureTryOut:
Aaaanndddd it seems I'm blind and it's already upgraded to 3.28, what the hell.

PR #4952, opened 2018-08-05
testing/gnome-shell: update to 3.28 and add new required dependencies aports

PureTryOut:
Please have a good look at the new aports, especially ibus, as I did this a bit hasty using the PKGBUILDs from Arch Linux and I'm sure there is something to improve.

Also, Mutter needed a `$pkgrel` bump to being able to be installed.

### DIFF
--- a/testing/cldr-emoji-annotation/APKBUILD
+++ b/testing/cldr-emoji-annotation/APKBUILD
@@ -1,0 +1,26 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+pkgname=cldr-emoji-annotation
+pkgver=33.1.0_p0
+_tag=${pkgver/_p/_}
+pkgrel=0
+pkgdesc="Emoji annotation files in CLDR"
+arch="noarch"
+url="https://github.com/fujiwarat/cldr-emoji-annotation"
+license="Unicode-DFS-2016"
+makedepends="automake autoconf"
+source="$pkgname-$pkgver.tar.gz::https://github.com/fujiwarat/$pkgname/archive/$_tag.tar.gz"
+builddir="$srcdir/$pkgname-$_tag"
+
+build() {
+	cd "$builddir"
+	./autogen.sh --prefix=/usr
+	make
+}
+
+package() {
+	cd "$builddir"
+	make DESTDIR="$pkgdir" install
+
+	install -Dm644 unicode-license.txt "$pkgdir"/usr/share/licenses/$pkgname/license
+}
+sha512sums="57dc2b7c8ce4b287eb83113b4d66f5359ce0e03a86477926705864d997bf87a5638dd66414b8ec4549a887cd328c9f996249269be4ca05f0d86d00393e55c513  cldr-emoji-annotation-33.1.0_p0.tar.gz"
--- a/testing/mutter/APKBUILD
+++ b/testing/mutter/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=mutter
 pkgver=3.28.0
-pkgrel=0
+pkgrel=1
 pkgdesc="clutter-based window manager and compositor"
 url="https://github.com/GNOME/mutter"
 arch="all"
--- a/testing/unicode-character-database/APKBUILD
+++ b/testing/unicode-character-database/APKBUILD
@@ -1,0 +1,24 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+pkgname=unicode-character-database
+pkgver=11.0.0
+pkgrel=0
+pkgdesc="Unicode Character Database"
+arch="noarch"
+license="BSD"
+url="http://www.unicode.org/"
+makedepends="libarchive-tools"
+source="UCD-$pkgver.zip::http://www.unicode.org/Public/zipped/$pkgver/UCD.zip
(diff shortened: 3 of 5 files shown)


testing/cldr-emoji-annotation/APKBUILD
```diff
@@ -0,0 +1,26 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+pkgname=cldr-emoji-annotation
+pkgver=33.1.0.0
+_tag=33.1.0_0
```
ncopa: can we generate `_tag` from `pkgver`? Maybe we can do somethin like:

````sh
pkgver=33.1.0_p0
_tag=${pkgver/_p/_}
````

File: testing/cldr-emoji-annotation/APKBUILD [hunk 2]
```diff
@@ -0,0 +1,26 @@
...
+pkgrel=0
+pkgdesc="Emoji annotation files in CLDR"
+arch="noarch"
+url="https://github.com/fujiwarat/cldr-emoji-annotation"
+license="custom"
```
ncopa: I guess license should be one of:
Unicode-DFS-2015
Unicode-DFS-2016
Unicode-TOU

see https://spdx.org/licenses/

Can you figure out which it is?

testing/ibus/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+source="$pkgname-$pkgver.tar.gz::https://github.com/$pkgname/$pkgname/archive/$pkgver.tar.gz
+	ucd-path.patch::https://github.com/ibus/ibus/commit/5788be80685f397c3db3bdf4e672d67cfb9b3433.patch"
+subpackages="lib$pkgname $pkgname-dev $pkgname-doc $pkgname-lang"
+options="!check"
+
+prepare() {
+	default_prepare
+
+	cd "$builddir"
+	sed -i 's|$(libibus) $(libibus_emoji_dialog)|$(libibus_emoji_dialog) $(libibus)|' ui/gtk3/Makefile.am
```
ncopa: Can we patch the Makefile.am instead for sed'ing it? So we notice if this is fixed upstream (whatever it is supposed to fix).

testing/ibus/ibus.install
```diff
@@ -0,0 +1,16 @@
...
+
+post_upgrade() {
+	post_install
+}
+
+post_remove() {
+	post_install
+
+	rm -f etc/dconf/db/ibus
+}
```
ncopa: no, this will not work. we use dedicated scripts instead of a catchall.

I also think we should implemenr the --update-cache thing as a trigger if possible.

testing/mutter/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=mutter
 pkgver=3.28.0
-pkgrel=0
+pkgrel=1
```
ncopa: why?

testing/unicode-character-database/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+pkgver=11.0.0
+pkgrel=0
+pkgdesc="Unicode Character Database"
+arch="noarch"
+license="BSD"
+url="http://www.unicode.org/"
+makedepends="libarchive-tools"
+source="UCD-$pkgver.zip::http://www.unicode.org/Public/zipped/$pkgver/UCD.zip
+	Unihan-$pkgver.zip::http://www.unicode.org/Public/zipped/$pkgver/Unihan.zip
+	LICENSE"
```
ncopa: I don't think we should provide our own license file.

PureTryOut: Any reason why not? It's the proper license according to https://www.unicode.org/copyright.html, which says that license applies to everything under http://www.unicode.org/Public

The package .zip doesn't have a license file itself, so I thought this was the proper way to do it. I could just keep it out and not install it if that is what you mean.

testing/mutter/APKBUILD
PureTryOut: Because otherwise:

```
>>> gnome-shell: abuild 3.2.0-r0
>>> gnome-shell: Checking sanity of /home/pmos/build/APKBUILD...
>>> gnome-shell: Analyzing dependencies...
  so:libgnome-desktop-3.so.12 (missing):
    required by:
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
                 mutter-3.28.0-r0[so:libgnome-desktop-3.so.12]
  .makedepends-gnome-shell-0:
    masked in: cache
    satisfies: world[.makedepends-gnome-shell]
>>> ERROR: gnome-shell: all failed
```

testing/unicode-character-database/APKBUILD
ncopa: IMHO, The proper way to do it would be report it upstream and ask them to ship the license file.

I think its better to keep it out and not install it.

PureTryOut:
@ncopa could you do another review?

PureTryOut:
In the meantime Ibus has been added by someone else, so I removed it from this PR.

PR #4976, opened 2018-08-08
community/py-lz4: upgrade to 2.1.0

itoffshore:
This release changes the handling of errors for block decompression when
`uncompressed_size > 0`. In this case, we no longer check that the uncompressed
data has exactly the size specified by `uncompressed_size`: this argument is
now used as an upper bound.

In addition, we introduce a new exception:` LZ4BlockError` which is raised
whenever the `LZ4`library fails.

These two changes together allow "guessing" of the uncompressed data size:
simply set uncompressed_size to a number and try decompress, and catch
`LZ4BlockError`. If `LZ4BlockError` is raised, increase `uncompressed_size`
and try again.

See the documentation for more details and examples:

http://python-lz4.readthedocs.io/en/stable/lz4.block.html

### DIFF
--- a/community/py-lz4/APKBUILD
+++ b/community/py-lz4/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=py-lz4
-pkgver=2.0.1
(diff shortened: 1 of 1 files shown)


andypost:
does it still fails `check()` on s390x? could be tested in follow-up

itoffshore:
@andypost - I've commented out the `!check` for `s390x` for the next update

PR #4979, opened 2018-08-09
main/postgresql: security upgrade to 10.5

andypost:
CVE-2018-10915 CVE-2018-10925

Ref https://www.postgresql.org/about/news/1878/


andypost:
@jirutka probably dependent packages need to bump their versions & somehow it needs backport

jirutka:
> probably dependent packages need to bump their versions

PostgreSQL does not change ABI between minor versions, so rebuild of dependent packages should not be needed. Otherwise they'd note it in the release notes (or at least I hope so).

PR #4983, opened 2018-08-10
main/xorg-server: upgrade to 1.20.1

Decatf:


Decatf:
xfwm is broken on xorg-server 1.20.0. A patch to fix it is included in 1.20.1.
reference: https://bugs.archlinux.org/task/58662

PR #4985, opened 2018-08-11
testing/ejabberd: upgrade to 18.03

stwa:


stwa:
I took another look at that: As I remember, the `fast_tls` dependency was added explicitly when Alpine moved from OpenSSL to LibreSSL. The included version of `fast_tls` was incompatible to LibreSSL and we needed to upgrade. So actually there would be no need to overwrite dependencies...

But all of the dependencies required by ejabberd are not included in the source package. They all get downloaded during build (cloned via git). To make all dependencies download before building and to have proper checksums I manually added the required dependencies.

Sadly, I can't think of a way to automate this step. Any suggestions?

stwa:
Another way would be to run `rebar get-deps` during the prepare step, if this is acceptable.

andypost:
Looks good to go

testing/ejabberd/APKBUILD
```diff
@@ -1,15 +1,32 @@
 # Contributor: Leonardo Arena <rnalrd@alpinelinux.org>
 # Contributor: Francesco Colista <fcolista@alpinelinux.org>
+# Contributor: Stefan Wagner <stw@bit-strickerei.de>
 # Maintainer: John Regan <john@jrjrtech.com>
 pkgname=ejabberd
-pkgver=17.09
-pkgrel=3
+pkgver=18.03
```
andypost: btw https://github.com/processone/ejabberd/releases/tag/18.12.1 released

stwa: Thanks for the hint, I will upgrade to 18.12.1 in a separate PR after this PR gets merged.

andypost:
please update to latest version https://github.com/processone/ejabberd/releases

stwa:
This MR does various stuff:
* It upgrades the ejabberd version
* It cleans the APKBUILD
* It names *all* dependencies required by the build to have them downloaded before the actual build step (no network needed during the build step, which is favored by alpine as some other code reviews teached me)
* It enables/fixes the initd script to run ejabberd as the (already created) ejabberd user instead of root (which is recommended by ejabberd, the process says it itself)

The MR is *really* old now, but it never got merged. If this changes are not valuable enough by itself to be merged without another version upgrade I'm not willing to put more time and effort into this to fiddle with a newer version, upgraded dependencies and so on.

PR #4986, opened 2018-08-11
testing/py-anyjson: fix python3 build

stwa:


andypost:
Why it needs to copy build dir? Example https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python does not do that

stwa:
Actually I have no idea why this is required (I don't know much about python). But to reproduce the current package is broken you can run python3 and do `import anyjson` - it fails. Splitting the build directories seems to fix the problem.

I know this is a workaround and of course it's bad to apply a fix to a problem which isn't understood. Maybe someone with more knowledge in python can investigate?

Btw, Arch Linux seems to do the same: https://projects.archlinux.de/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/python-anyjson

stwa:
I found a [pull request](https://bitbucket.org/runeh/anyjson/pull-requests/7/add-support-for-python-34-35-and-36/diff) upstream which isn't merged (yet?). I also added missing tests. This also seems to fix the problem.

testing/py-anyjson/APKBUILD
```diff
@@ -1,22 +1,28 @@
...
 _pkgname=anyjson
 pkgver=0.3.3
-pkgrel=1
+pkgrel=2
 pkgdesc="Wraps the best available JSON implementation available in a common interface"
 url="https://bitbucket.org/runeh/anyjson/"
 arch="noarch"
 license="BSD"
-makedepends="python2-dev python3-dev py-setuptools"
+makedepends="python2-dev python3-dev py-setuptools py-future py-nose"
```
andypost: actually this changes are for `checkdepends` so like
```
checkdepends="py-future py-nose"
makedepends="python2-dev python3-dev py-setuptools"
```

stwa: Didn't know about that, thanks!

testing/py-anyjson/APKBUILD
```diff
@@ -1,22 +1,29 @@
...
+checkdepends="py-future py-nose"
 makedepends="python2-dev python3-dev py-setuptools"
 subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
-source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz
+	python3.patch"
 builddir="$srcdir"/$_pkgname-$pkgver
 
 build() {
-	cd "$builddir"
```
andypost: why you removed that? the same should be added to `check()` - all this commands should start in bulddir

stwa: I thought newest abuild will change to $builddir automatically, but `cd $builddir` should not be removed before 3.9-stable. Sorry for that. Readded.

PR #4991, opened 2018-08-12
community/lxcfs: Modprobe modules instead of failing when absent

nero:
Previously, lxcfs failed when the kernel modules were not listed the lsmod
output. Built-in kernel modules never show up in lsmod, thus the service
could not be properly started on such systems.

Fixes [#9214](https://bugs.alpinelinux.org/issues/9214).

### DIFF
--- a/community/lxcfs/APKBUILD
+++ b/community/lxcfs/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=lxcfs
 pkgver=3.0.1
-pkgrel=0
+pkgrel=1
 pkgdesc="FUSE filesystem for LXC"
 url="https://linuxcontainers.org/lxcfs"
 arch="all"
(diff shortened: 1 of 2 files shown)


nero:
Ping @itoffshore

itoffshore:
needs tabs instead of spaces in the patch to fix the [alignment of `return 1`](https://github.com/alpinelinux/aports/pull/4991/commits/336c55e8346cefb4406d0f97779a9f12848cf454)

nero:
Branch updated.

ncopa:
I don' think this will work. modprobe will return an error when the filesystem is compiled in to kernel instead of module: `modprobe: FATAL: Module ... not found in directory /lib/modules/4.14.65-0-vanilla`

We need to check if the filessystem is available in kernel and try modprobe if it isnt. Something like:
````diff
diff --git a/community/lxcfs/lxcfs.initd b/community/lxcfs/lxcfs.initd
index 1a6e9c9759..496d6156fb 100644
--- a/community/lxcfs/lxcfs.initd
+++ b/community/lxcfs/lxcfs.initd
@@ -19,14 +19,11 @@ depend() {
 }
 
 start_pre() {
-       local module=
+       local filesystem=
        checkpath --directory ${VARDIR}
-       for module in fuse autofs4; do
-               if ! $(lsmod | grep -q ^$module); then
-                       eerror "Enable module: $module"
-                       eerror "modprobe $module"
-                       eerror "echo $module >> /etc/modules"
-                       eend 1
+       for filesystem in fuse autofs4; do
+               if ! grep -q -w $filesystem /proc/filesystems; then
+                       modprobe $filesystem || return 1
                fi
        done
 }
````
modprobe will print error message so I don't think we need print any error message with `eerror`.

**Comment by nero** [comment]
@ncopa There is `/lib/modules/$(uname -r)/modules.builtin.bin`, which contains a list of built-in modules. Modprobe used that as ignore list.

See second headline [here](https://www.kernel.org/doc/Documentation/kbuild/kbuild.txt):
>modules.builtin
>--------------------------------------------------
>This file lists all modules that are built into the kernel. This is used
>by modprobe to not fail when trying to load something builtin.

You can confirm this behavior by modprobing `brd` on alpine, which is not present in the modules directory:

```
nyu:~# modprobe brd; echo $?
0
nyu:~# find /lib/modules/$(uname -r) -name '*brd*'
nyu:~#
```

ncopa:
> 2 files changed, 3 insertions(+), 11 deletions(-)

:heart:

yarons:
You guys ares awesome! Thanks @ncopa, I'm a big fan!

PR #4994, opened 2018-08-13
community/mbedtls: upgrade to 2.12.0

itoffshore:
* includes fixes for **CVE-2018-0497** and **CVE-2018-0498**

https://github.com/ARMmbed/mbedtls/blob/development/ChangeLog

### DIFF
--- a/community/mbedtls/APKBUILD
+++ b/community/mbedtls/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=mbedtls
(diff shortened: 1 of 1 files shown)


community/mbedtls/mbedtls-utils.profile.d
```diff
@@ -0,0 +1 @@
+export PATH=$PATH:/usr/libexec/mbedtls
```
ncopa: Why not install mbedtls in `/usr/bin` if it is supposed to be in `PATH`?

Why is it installed in `/usr/libexec` in the first place?

itoffshore: I thought you put them in `libexec` for a reason - have moved the `utils` into `/usr/bin`

PR #4997, opened 2018-08-13
community/libmbim: upgrade to 1.16.2

itoffshore:
community/libmbim: upgrade to `1.16.2`


andypost:
looks abi is the same and only `community/modemmanager` using `libmbim-dev`

PR #5004, opened 2018-08-14
main/boost: fix and reenable failing debugger tests

mksully22:
In debugger::debug_parent_wait() the variable ch was mistakenly defined to be a char and used to obtain a value from 
ch = fgetc( command_child ); 

Later in the code a compare of the truncated ch (0xff) to the constant EOF (0xffffffff) would fail resulting in errors between the expected and actual values returned in the debugger.py and debugger-mi.py tests.

Since fgetc is defined to return an int the declaration of ch was modified and the truncation eliminated.

I will post the patch to the boost/build community as well.


main/boost/APKBUILD
```diff
@@ -15,7 +15,7 @@ source="https://dl.bintray.com/boostorg/release/$pkgver/source/${pkgname}_$_pkgv
 	boost-1.57.0-python-abi_letters.patch
 	boost-1.57.0-python-libpython_dep.patch
 	revert-wifsignaled-static-assert.patch
-	disable-debugger-test.patch
```
andypost: you need to add removal of the file to commit

mksully22: Thanks...I have removed it

mksully22:
PR opened on boost/build project, https://github.com/boostorg/build/pull/332

PR #5014, opened 2018-08-16
main/nodejs: upgrade to 8.11.4

timbru31:
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.11.4

### DIFF
--- a/main/nodejs/APKBUILD
+++ b/main/nodejs/APKBUILD
@@ -6,6 +6,8 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 #
 # secfixes:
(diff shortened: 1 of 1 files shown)


jirutka:
> deps: Upgrade to OpenSSL 1.0.2p, fixing:
>
>    Client DoS due to large DH parameter (CVE-2018-0732)
>    ECDSA key extraction via local side-channel (CVE not assigned)

Actually this CVE does not affect the package, because we don't compile it with bundled OpenSSL lib. Please remove CVE-2018-0732 from secfixes.

timbru31:
Addressed your comments 👍

PR #5015, opened 2018-08-16
main/sqlite: build static libs

j8r:
Enable the build of static libraries for sqlite. First contribution. I see in [CONTRIBUTING.md](https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md)
> Please do not submit PR's via Github's web interface.

Is it still relevant? Sorry if I haven't done the contribution properly.

### DIFF
--- a/main/sqlite/APKBUILD
+++ b/main/sqlite/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 pkgname=sqlite
 pkgver=3.24.0
-pkgrel=0
+pkgrel=1
 pkgdesc="C library that implements an SQL database engine"
 url="http://www.sqlite.org"
 arch="all"
@@ -54,8 +54,8 @@ build() {
 		--host="$CHOST" \
 		--prefix=/usr \
 		--enable-threadsafe \
-		--disable-static \
 		--enable-readline \
+		--enable-static \
 		--enable-dynamic-extensions
 
 	# rpath removal


andypost:
Looks nice! Does it means we should update all other dependant packages?

j8r:
`--enable-static` add an `sqlite.a` archive file used for static linking, there is no breaking changes :).
From the sqlite's `configure` file:
>   --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-static[=PKGS]  build static libraries [default=yes]

ncopa:
I think I'd like to keep the `--enable-static` in there as it makes it clear what our intention is.

I also think we should have a `sqlite-static` subpackage, unless the file is small.

ncopa:
I don't really know why CONTRIBUTING.md ask you to not use github web interface. @clandmeter, do you know why that was written there?

j8r:
The `/usr/lib/libsqlite3.a` file weights 1.7 MB. I don't know if we can considered it small, I guess yes.

main/sqlite/APKBUILD
```diff
@@ -54,8 +54,8 @@ build() {
 		--host="$CHOST" \
 		--prefix=/usr \
 		--enable-threadsafe \
-		--disable-static \
 		--enable-readline \
+                --enable-static \
```
ncopa: Use tabs instead of spaces.

clandmeter:
@ncopa iirc this was to prevent people to make changes to apkbuilds and submit them via the web interface which would mean they don't have a development environment and don't test their contributions first before submitting.
If you think its incorrect you can remove it or update it to make it more clear.

PR #5021, opened 2018-08-17
community/php5: fix dependency (part 2)

vakartel:
some fixes after #5018

removed php5-cli dependency on all extensions
added php5-common (config) dependency to phpdbg

### DIFF
--- a/community/php5/APKBUILD
+++ b/community/php5/APKBUILD
@@ -4,7 +4,7 @@
 # Maintainer: Matt Smith <mcs@darkregion.net>
 pkgname=php5
 pkgver=5.6.37
-pkgrel=1
+pkgrel=2
 pkgdesc="The PHP language runtime engine"
 url="http://www.php.net/"
 arch="all"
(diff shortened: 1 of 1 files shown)


community/php5/APKBUILD
```diff
@@ -438,6 +439,7 @@ _mv_ext() {
 	local ini=$ext.ini
 	pkgdesc="${ext} extension for PHP"
 	provides="php-$extname"
+	depends=
```
andypost: I think better remove it few lines bellow
```diff
-  depends="${pkgname}-common ${depends}"
+  depends="${pkgname}-common"
```

andypost: ah.. I'm wrong... it just needs to be
```
depends="${pkgname}-common"
	# extension dependencies
	if [ -n "${2-}" ]; then
 		depends="${2-} ${depends}"
fi
```

vakartel: No, we must clean 'depends' variable first. Then populate it with extension-specific depends (if so) and finally add php5-common (config) dependency

andypost: Then just make it `${depends}" depends="${2-}` - to append common if any dependencies added

PR #5023, opened 2018-08-17
testing/aws-ena-driver-vanilla: update to 1.5.3, build virt flavor

tomalok:
Multi-flavoring based on community/virtualbox-guest-modules-vanilla.


tomalok:
@ncopa thank you for merging 2c3f6a6 -- it would also be much appreciated if we could get this one merged as well (at least the part that also builds a virt-flavored version of the driver).

ncopa:
Both `aws-ena-driver-vanilla` and `aws-ena-driver-virt` provides the same file `/etc/modules-load.d/ena.conf`, which means they will conflict if you try install them both at the same time.

I think we should have a common package named `aws-ena-driver` which only provides the config file. The -vanilla and -virt pacakges could then have an `install_if` so you get proper driver for the installed kernels. All you need to do is `apk add aws-ena-driver` and you'll get the proper driver package depending on what kernel you have installed.

I'll try fix that.

PR #5024, opened 2018-08-17
main/linux-vanilla: build NVMe modules for x86* virt flavor

tomalok:
The most recent generation of AWS EC2 instances (c5, i3, m5) expose EBS volums as NVMe devices.
This patch simply builds the same NVMe kernel modules for the 'virt' flavor as it does for 'vanilla'.


tomalok:
TravisCI fails because it takes a long time to build kernels...  Worked fine (but yes, took a long time) when I did an `abuild -r` in my dev environment.

ncopa:
This was merged with 2c3f6a6f647e7e6dce08acf871a7438ee46fc05a

PR #5025, opened 2018-08-17
 main/tiny-ec2-bootstrap: add missing e2fsprogs-extra dependency

tomalok:
tiny-ec2-bootstrap uses resize2fs, which is in e2fsprogs-extra.


tomalok:
superceded by #5034

PR #5029, opened 2018-08-18
community/php7-pecl-redis: renamed from php7-redis, modernize

andypost:

Dependency on `igbinary` allows to use faster protocol that could not be added runtime
Size change on x86-64 *437264* to *441360*


andypost:
@fabiorphp maybe add `provides="php7-redis"` to keep BC?

fabiorphp:
I think it's a good idea

PR #5032, opened 2018-08-18
main/python3: Update to 3.6.6

bratkartoffel:
This PR updates python to the latest 3.6.x and conflicts with PR #4840 (which will update python to the 3.7 branch)

### DIFF
--- a/main/python3/APKBUILD
+++ b/main/python3/APKBUILD
@@ -3,9 +3,9 @@
 
 pkgname=python3
 # the python2-tkinter's pkgver needs to be synchronized with this.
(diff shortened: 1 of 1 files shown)


bratkartoffel:
Don't know why travis fails on this, it seems the network tests fail. Locally everything works fine.

andypost:
Tests fail...
```
1 test failed:
    test_urllib2net
```

bratkartoffel:
As i stated already, the test fails on travis, but works on my machine.
I'll take a look into it, but I'm not that expert in python

PR #5035, opened 2018-08-19
main/ldb: Upgrade to 1.3.5

tanertas:
* Test suite fails if compiler flags set
* Use waf directly instead of disguising as autotools
* Re-arrange and remove unused configure options


tanertas:
Needed for #5005

andypost:
the same tests fail as I reported

algitbot: build-edge-armhf: failed to build ldb: http://build.alpinelinux.org/buildlogs/build-edge-armhf/main/ldb/ldb-1.3.5-r0.log
algitbot: build-edge-x86: failed to build ldb: http://build.alpinelinux.org/buildlogs/build-edge-x86/main/ldb/ldb-1.3.5-r0.log

PR #5036, opened 2018-08-20
main/samba: Various improvements

tanertas:
* Improve initscript
* Fix start service as DC
* Add bind-9.12 support for DC DNS backend
* Enable pam_winbind module and add corresponding subpackage
* uclibc-xattr-create.patch removed


tanertas:
CI build fails because requires #5035

This is a security upgrade:
https://www.mail-archive.com/samba-announce@lists.samba.org/msg00439.html

If non-security related other changes holded for later, @andypost has another PR without additional changes: #5005

andypost:
https://github.com/alpinelinux/aports/pull/5005 was accepted, so this one needs to become "modernize" commit on top of upgrade

main/samba/APKBUILD
```diff
@@ -128,6 +130,7 @@ initscript() {
 	depends=""
 	install -Dm755 $srcdir/$pkgname.initd "$subpkgdir"/etc/init.d/$pkgname
 	install -Dm644 $srcdir/$pkgname.confd "$subpkgdir"/etc/conf.d/$pkgname
+	install -dm755 "$subpkgdir"/run/$pkgname
```
ncopa: `/run` is required to be cleared at reboot, so I don't know if it makes sense to create it on package install time, since it needs to be recreated on service start anyway.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html

tanertas: So, it must have done in service script. I will revise this PR along with other changes.

tanertas:
Requires #5045 first.

main/samba/samba.initd
```diff
@@ -1,24 +1,41 @@
 #!/sbin/openrc-run
 
 extra_started_commands="reload"
+piddir=${piddir:-"/run/samba"}
+daemon_list="${daemon_list//samba/}"
```
ncopa: Why are we always removing `samba` daemon?

It would be nice to avoid `${var//search/replace}` expansion if possible, since it is not specified in POSIX shell.

main/samba/samba.initd
```diff
@@ -1,24 +1,41 @@
...
 
 DAEMON=${SVCNAME#samba.}
 if [ "$DAEMON" != "samba" ]; then
 	daemon_list=$DAEMON
+elif [ "${daemon_list// /}" == "" ] || [ -z '$daemon_list' ]; then
```
ncopa: The `== ` is bashism. use single `=`.

The `[ -z '$daeom_list' ]`  will always return `false`. The single quotes will prevent `$daemon_list` to get expanded and return the literal string `$daemon_list`. Its equivalent to `"\$daemon_list"`. This is most likely not what you intend.

main/samba/samba.initd
```diff
@@ -1,24 +1,41 @@
...
+	daemon_list="samba"
```
ncopa: Your 2 daemon_list changes above will make it impossible only start `nmbd` and `smbd` if you set  `daemon_list="smbd nmbd samba"` (eg. it will not start `samba` service).

I would prefer that `/etc/init.d/samba` would start all services specified in `daemon_list` (including `samba` if its specified), while `/etc/init.d/samba.<daemon>` only starts `<daemon>`, even if `<daemon>` is `samba`.

main/samba/samba.initd
```diff
@@ -1,24 +1,41 @@
...
 fi
 
 depend() {
 	need net
 	after firewall
 }
 
+start_pre() {
+	[ -d "$piddir" ] || mkdir -p "$piddir"
```
ncopa: use `checkpath --directory "$piddir"` here.

main/samba/samba.initd
tanertas: When daemon_list="smbd nmbd samba" samba discarded right at start because daemon_list="${daemon_list//samba/}". We don't want mix them.

I will think about a POSIX friendly solution. Thanks for review. You can hold for this PR for later  because I can do a version upgrade as well because samba is now 4.8.5.  You may do a simple rebuild of samba against #5045 because ldb is more demanding right now.

main/samba/samba.initd
tanertas: I want make sure that samba launches only when $daemon_list is empty (with or without spaces) and not set. I'm aware of the situation with single quotes  but it seems that it misled me while the expression also returns true because '$daemon_list' didn't set as well. If $daemon_list set by any string, /bin/sh seems to append $daemon_list's value to "-z" rather than treating it as a variable identifier. I'm open for any simple solution.

main/samba/samba.initd
tanertas: As far as I know, samba(4) is not compatible with smbd, nmbd. It is compatible with winbindd for purpose but samba launches winbindd itself. So, I'm trying to solve the any possible conflict between old (smbd, nmbd, winbindd) and new (samba). For instance, Debian's default install comes with samba-ad-dc service masked. You have to mask nmbd, smbd, winbindd and then unmask samba-ad-dc to use samba's ad-dc functionality in order to solve the conflicts.

main/samba/samba.initd
tanertas: It seems PR already merged. Thanks.

PR #5037, opened 2018-08-20
Update udhcpc default.script

RostyslavFridman:
to enable the ability to assign IP addresses with /32 netmask


ncopa:
need to update checksum and pkgrel in APKBUILD too. Commit message should be something like:
````
main/busybox-initscripts: fix udhcpc script to assign ip addresses with /32 netmask
````

ncopa:
fixed with c38926656e46f2b1fb6ff4a33b6f43441f854a94

PR #5050, opened 2018-08-23
Upgrade to 0.3.4.6-rc

xpecex:
https://blog.torproject.org/tor-0346-rc-released


tanertas:
As the name implies, this is an Alpha version and probably will rejected.

PR #5055, opened 2018-08-25
community/py3-aiohttp: upgrade to 3.4.0

fabaff:


andypost:
There's 2 deprecations, I guess it needs @fcolista approval

fabaff:
Yes, there are. As it happens with every other release this needs to be addressed by the tools which depends on aiohttp.

fabaff:
Another issue which came up with the RPM for Fedora is the bundling of `http-parser`.

PR #5057, opened 2018-08-26
testing/yubikey-personalization: new aport

dae3:
https://github.com/dae3/alpine-yubikey-personalization
YubiKey Personalization cross-platform library and tool


jirutka:
This is already packaged in Alpine as [ykpers](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/yubico-piv-tool) (the same as other distributions name this package).

PR #5060, opened 2018-08-26
testing/snes9x: Upgrade to 1.56.2

tanertas:
* Switch to Gtk-2


ncopa:
why the Gtk-2 switch? I would like gtk3 whenever possible, since gtk2 is no longer developed upstream.

tanertas:
I thought gkt2 is more compact and few deps. Maybe I'm wrong. You can consider revert the commit.

ncopa:
yes, gtk2 is more compact but i think we want use newer if possible.

PR #5061, opened 2018-08-26
testing/php7-diseval: new aport

itoffshore:
* disables `eval` as a temporary solution for `php7` until `suhosin` is stable for `php7`


testing/php7-diseval/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+pkgname=php7-diseval
+_pkgreal=diseval
+pkgver=0.1
+pkgrel=0
+pkgdesc="Disable eval in php7 (suhosin is not stable yet for php7)"
+url="https://github.com/mk-j/PHP_diseval_extension"
+arch="all"
+license="MIT"
+depends="php7-common"
+makedepends="php7-dev autoconf libtool"
```
andypost: probably makes sense to add `re2c`
```
checking for re2c... no
configure: WARNING: You will need re2c 0.13.4 or later if you want to regenerate PHP parsers.
```

andypost: not sure libtool needed here as builder has it

andypost:
Also it is interesting why suhosin is needed when you can use standard way http://php.net/manual/en/ini.core.php#ini.disable-functions

PR #5068, opened 2018-08-27
Cleanup of go toolchain packages

nmeum:
Things that still need to be addressed: 

* The testsuite should be invoked in `check()`, currently a lot of tests fail and fixing this would probably be a lot of work. I would therefore suggest addressing this problem in a separate pull request.
* We could add a separate package for `go-tools`,  [Arch Linux](https://www.archlinux.org/packages/extra/x86_64/go-tools/) also has such a separate package. However, I would personally prefer not to package it, forcing users to manually install it using go's package management instead.
*  The `loadinternal: cannot find runtime/cgo` warning should be fixed. It was introduced in `1.10` (iirc). I am unsure what's causing this warning to be emitted, however, building with `CGO_ENABLED=1` fixes it. The following issues might be related:
    * https://github.com/golang/go/issues/25762
    * https://github.com/golang/go/pull/25291
    * https://bugs.archlinux.org/task/57921

Any thoughts on this? Should we merge this PR as is or should I add a go-tools package in this PR as well?


ncopa:
whoops. i bumped go to 1.11.

nmeum:
> whoops. i bumped go to 1.11.

No problem, rebased it. Do you think we should add a separate `go-tools` package as a part of this PR?

ncopa:
> Do you think we should add a separate go-tools package as a part of this PR?

Doesn't matter I guess. I'm ok to merge this as is to move things forward.

PR #5080, opened 2018-08-31
testing/py-h5py: new aport

Stegoxorus:
https://www.h5py.org
Read and write HDF5 files from Python


Stegoxorus:
A project that I work on is still on python2, so I included python2 in the build so that I could run A/B testing as I migrate to python3.

ncopa:
I squashed the commits. Thanks!

PR #5083, opened 2018-09-02
community/wait4ports: upgrade to v0.2.2

erikogan:
(These changes are necessary to make the package useful in docker-compose files versions after v1.)

Adds:
* IPv6 support
* Hostname resolution
* Verbosity and sleep interval configuration via environment and command-line


andypost:
Please update commit message to `community/wait4ports: upgrade to 0.2.1`

Also makes sense to modernize package and remove no longer needed `return 1`

erikogan:
I also realized there should be a minor update to the package, so I’ll also bump the version to 0.2.2 now that I’m back online.

PR #5085, opened 2018-09-02
testing/pyscard: new aport

dae3:
https://github.com/LudovicRousseau/pyscard/blob/master/LICENSE
pyscard smartcard library for python


ncopa:
Do we really need python2 support?

dae3:
@ncopa re python2 support: I'm not sure. I built this package as a dependency for testing/yubikey-manager, for which python3 is sufficient. Happy to remove python2 if you recommend.

PR #5101, opened 2018-09-04
main/py-py: upgrade to 1.7.0 and delete in testing

andypost:
https://py.readthedocs.io/en/latest/changelog.html

Ref https://pkgs.alpinelinux.org/packages?name=py-py&branch=edge


andypost:
@fabaff are you ok with this change?

andypost:
@fabaff is any dependent packages needs rebuild?

fabaff:
I don't think so.

PR #5111, opened 2018-09-05
testing/py-tox: move to main

itoffshore:
testing/py-tox: move to main


ncopa:
why is it moved to main and not community? Is this something we want maintain for 2 years?

andypost:
I needs to be in main because `sshfs` and other python packages using it to run tests

PR #5112, opened 2018-09-06
main/bind: Use gss-spnego instead isc-spnego

tanertas:
* Bind's own spnego implementation cause issues with samba. [1]
* Bind must compile with "--disable-isc-spnego" in order make secure updates work with BIND9_DLZ dns backend. [2]
    
    [1] https://lists.samba.org/archive/samba/2018-September/218017.html
    [2] https://wiki.samba.org/index.php/Using_BIND_DLZ_backend_with_secured_/_signed_DNS_update


tanertas:
probably waiting for a version bump for Bind to apply this change is more appropriate.

PR #5116, opened 2018-09-06
testing/py-influxdb: upgrade to 5.2.0

fabaff:


testing/py-influxdb/APKBUILD
```diff
@@ -43,4 +43,4 @@ _py() {
 	$python setup.py install --prefix=/usr --root="$subpkgdir" || return 1
```
andypost: makes sense to remove this "return 1" as well - also why there's no `check()`?

ncopa: i removed the ` || return 1` and merge. It is a step forward.

PR #5118, opened 2018-09-06
main/py-sphinx: upgrade to 1.7.9

fabaff:


andypost:
`py-pluggy` already in main, makes sense to try enable `check`

PR #5120, opened 2018-09-06
community/go: Fix the `cannot find runtime/cgo`-warning

nmeum:
This is basically a follow-up for #5068. As described in this PR our go compiler currently emits a `loadinternal: cannot find runtime/cgo`-warning when compiling certain software. For instance, a simple hello world program. This PR attempts to fix this.

---

This warning seems to be emitted since we build with `-shared` for PIE
by default through our `default-buildmode-pie.patch`. Building with
shared linkage requires depending on CGO. The patch, however, doesn't
force a dependency on CGO. As with android we need to force external
linkage in order to always depend on CGO thereby resolving this warning.
    
I am unsure whether this is really the best way to fix the problem since
I don't know enough about the internals of the go compiler. However, I
am under the impression that the person who originally committed the
patch isn't familiar with the compiler internals either. Thus it's
probably ok to push this as is.
    
The other problem with this patch being that the static buildmode is no
longer supported. Running `go build -buildmode static` no longer works.
I would therefore propose creating a PIE and non-PIE go package, the
Arch Linux folks also seem to have chosen this solution.

### DIFF
--- a/community/go/APKBUILD
+++ b/community/go/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=go
 pkgver=1.11
-pkgrel=2
+pkgrel=3
 pkgdesc="Go programming language compiler"
 url="http://www.golang.org/"
 arch="all"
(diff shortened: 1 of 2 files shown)


nmeum:
@ncopa any chance you can take a look at this within the next week? I haven't noticed any issues with this yet and would assume lazy consent otherwise and merge this as is.

nmeum:
Ok, as stated above I am assuming lazy consent and merging this as is.

PR #5121, opened 2018-09-07
community/mediainfo: update (+lib) to 18.08

frebib:


frebib:
I should mention that this patch also contains [a fix](https://github.com/MediaArea/MediaInfoLib/pull/931) which fixed a crash in Alpine for us

PR #5139, opened 2018-09-11
testing/wireguard: upgrade to 0.0.20180910

zx2c4:
```
  * curve25519: arm: do not modify sp directly
  * compat: support neon.h on old kernels
  * compat: arch-namespace certain includes
  * compat: move simd.h from crypto to compat since it's going upstream

  This fixes a decent amount of compat breakage and thumb2-mode breakage
  introduced by our move to Zinc.

  * crypto: use CRYPTOGAMS license

  Rather than using code from OpenSSL, use code directly from AndyP.

  * poly1305: rewrite self tests from scratch
  * poly1305: switch to donna

  This makes our C Poly1305 implementation a bit more intensely tested and also
  faster, especially on 64-bit systems. It also sets the stage for moving to a
  HACL* implementation when that's ready.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 
 pkgname=wireguard-tools
-pkgver=0.0.20180904
+pkgver=0.0.20180910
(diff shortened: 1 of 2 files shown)


andypost:
Commited as http://dup.pw/aports/1a4a0a7c

PR #5143, opened 2018-09-11
testing/openssl1.1: upgrade to 1.1.1

andypost:
It is suppsoed to be LTS, ref https://www.openssl.org/blog/blog/2018/09/11/release111/


frebib:
If it helps, 1.1.1 on Arch Linux made it out of [testing] in only a few days with none/very few packages needing attention.
According to OpenSSL, 1.1.1 is ABI/API compatible with 1.1.0

PR #5144, opened 2018-09-11
community/tor: upgrade to 0.3.4.8

andypost:
ref https://blog.torproject.org/new-release-tor-0348-also-other-stable-updates-02917-03212-and-03310


andypost:
>>> WARNING:tor*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc

makes sense to file followup

PR #5147, opened 2018-09-12
community/zram-init: upgrade to 8.0

xpecex:
ref https://github.com/vaeth/zram-init/blob/master/ChangeLog

### DIFF
--- a/community/zram-init/APKBUILD
+++ b/community/zram-init/APKBUILD
@@ -9,19 +9,16 @@ arch="noarch"
 url="https://github.com/vaeth/zram-init"
 license="GPL-2.0"
 depends="e2fsprogs-extra"
-subpackages="$pkgname-zsh-completion:zshcomp"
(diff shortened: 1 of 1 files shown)


andypost:
Makes sense cos it is default since kernel 4.14

andypost:
> >>> WARNING:zram-init*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc

It needs subpackage

community/zram-init/APKBUILD
```diff
@@ -35,4 +33,13 @@ zshcomp() {
 		"$subpkgdir"/usr/share/zsh/site-functions/_$pkgname
 }
 
+openrc() {
+	depends=""
+        pkgdesc="$pkgname (OpenRC init scripts)"
```
andypost: did you check default `openrc()` option?

andypost: Also indent of strings different

PR #5149, opened 2018-09-12
testing/foolsm: new aport

Ikke:


testing/foolsm/fix-non-standard-macro.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/forkexec.c b/forkexec.c.new
+index 203d1d3202..e3708bbfa5 100644
+--- a/forkexec.c
++++ b/forkexec.c.new
+@@ -108,7 +108,7 @@ static void sigchld_hdl(int sig)
+ 	int script_status;
+ 	pid_t pid;
+ 
+-	while ((pid = waitpid(WAIT_ANY, &script_status, WNOHANG)) != 0) {
++	while ((pid = waitpid(-1, &script_status, WNOHANG)) != 0) {
```
andypost: `WAIT_ANY` defined in https://www.gnu.org/software/libc/manual/html_node/Process-Completion.html

But I bet it is missing in musl so better to send upstream suggestion to use `-1` for compatibulity

andypost: Btw kinda looks better fix
```diff
+#ifndef WAIT_ANY
+# define WAIT_ANY (-1) /* Any process.  */
+#endif
```

Ikke: I did it this way because someone on IRC said:

> or just change the source to say -1 instead of WAIT_ANY
> since it's using a useless and nonstandard macro for -1

For me it doesn't really matter.

PR #5150, opened 2018-09-12
main/bridge: depends bridge-utils

HRio:
The bridge.pre-up script in this package calls brctl


HRio:
ahh sorry, brctl is in busybox as well

PR #5168, opened 2018-09-14
testing/makeself: new aport

xpecex:
ref https://makeself.io/


tanertas:
Combine commits as single one with `git rebase -i HEAD~2` please. Otherwise doing a review is not going to be easy.

maxice8:
Merged, sorry for taking too long.

PR #5171, opened 2018-09-14
main/gdbm: upgrade to 1.18

xpecex:
ref https://puszcza.gnu.org.ua/projects/gdbm

### DIFF
--- a/main/strongswan/APKBUILD
+++ b/main/strongswan/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=strongswan
 pkgver=5.6.3
 _pkgver=${pkgver//_rc/rc}
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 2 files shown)


nmeum:
This includes an soname change:

```
-usr/lib/libgdbm.so.4
-usr/lib/libgdbm.so.4.0.0
+usr/lib/libgdbm.so.6
+usr/lib/libgdbm.so.6.0.0
```

Various packages are still linked against the old soname:

```
$ apk search -r so:libgdbm.so.4
…
freeradius-3.0.17-r2
python3-3.6.6-r0
…
```

Those packages need to be rebuild. The optimal solution would be if you could bump the pkgrel of those packages in this pull request. Take a look at the `git log` to see how this was done for other upgrades.

PR #5172, opened 2018-09-14
testing/telegraf: upgrade to 1.7.4

Fusl:
https://github.com/influxdata/telegraf/releases/tag/1.7.4


andypost:
Does  it makes sense to add `-openrc` subpackage cos in logs

`>>> WARNING:telegraf*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc`

PR #5177, opened 2018-09-15
testing/spark: new aport

tanertas:
https://www.igniterealtime.org/projects/spark/
Cross-platform IM client optimized for businesses and organizations.

### DIFF
--- a/testing/spark/APKBUILD
+++ b/testing/spark/APKBUILD
@@ -1,0 +1,47 @@
+# Contributor: Taner Tas <taner76@gmail.com>
+# Maintainer: Taner Tas <taner76@gmail.com>
+pkgname=spark
+_pkgname=Spark
+pkgver=2.8.3
+pkgrel=0
+pkgdesc="Cross-platform IM client optimized for businesses and organizations."
+url="https://www.igniterealtime.org/projects/spark/"
+arch="all"
+license="Apache-2.0"
+depends="
+	hicolor-icon-theme
+	openjdk8-jre
+	portaudio
+	"
+makedepends="apache-ant openjdk8"
+source="
+	$pkgname-$pkgver.tar.gz::https://github.com/igniterealtime/$_pkgname/archive/v$pkgver.tar.gz
+	spark.sh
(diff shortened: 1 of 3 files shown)


testing/spark/APKBUILD
```diff
@@ -0,0 +1,48 @@
...
+_pkgname=Spark
+pkgver=2.8.3
+pkgrel=0
+pkgdesc="Cross-platform IM client optimized for businesses and organizations."
+url="https://www.igniterealtime.org/projects/spark/"
+arch="all"
+license="Apache-2.0"
+depends="openjdk8-jre-lib portaudio"
+makedepends="apache-ant openjdk8"
+install=""
```
Ikke: This could be left out

testing/spark/APKBUILD
```diff
@@ -0,0 +1,48 @@
...
+	rm -rf target/build/bin
+	rm -rf target/build/logs
+	rm -rf target/build/lib/linux
+	rm -rf target/build/lib/linux64
+	rm -rf target/build/lib/windows
+	rm -rf target/build/lib/windows64
+	cp -R target/build "$pkgdir"/usr/share/$pkgname
+}
+
+doc() {
```
Ikke: Is there a reason you implemented your own doc function? If you need to do something extra, you can call `default_doc` to do the default thing.

tanertas: Thanks for noticing. I removed doc subpackage since help menu uses online documentation.

testing/spark/APKBUILD
tanertas: Yes.

Ikke:
@tanertas Where did you get Spark.png from? We prefer not to have binary files in the Aports repository.

tanertas:
> @tanertas Where did you get Spark.png from? We prefer not to have binary files in the Aports repository.

I found somewhere on internet. Btw, I just noticed that distribution file contains some icon files for different sizes. I will use them.

tanertas:
@ikke I removed png file from PR.

PR #5095, opened 2018-09-04
testing/py3-pathlib2: New aport

bratkartoffel:
This PR adds python pathlib2, a library to work with filesystem paths in an object-oriented way.

PR #5178, opened 2018-09-15
testing/ipython: new aport + dependencies

Ikke:
A nice REPL for python. Only supports python3 since 6.x.

### DIFF
--- a/testing/py-backcall/APKBUILD
+++ b/testing/py-backcall/APKBUILD
@@ -1,0 +1,46 @@
+# Contributor: Kevin Daudt <kdaudt@alpinelinux.org>
+# Maintainer: Kevin Daudt <kdaudt@alpinelinux.org>
+pkgname=py-backcall
+pkgver=0.1.0
+pkgrel=0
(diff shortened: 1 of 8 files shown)


testing/py-backcall/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+# Maintainer: Kevin Daudt <kdaudt@alpinelinux.org>
+pkgname=py-backcall
+pkgver=0.1.0
+pkgrel=0
+pkgdesc="Backwards compatible callback APIs"
+url="https://github.com/takluyver/backcall"
+arch="noarch"
+license="BSD-3-Clause"
+makedepends="python2-dev python3-dev
+options="!check" # pytest broken on alpine
```
andypost: what is broken? does it needs fix?

Ikke: https://bugs.alpinelinux.org/issues/9420

andypost: Then we need it for both versions https://github.com/alpinelinux/aports/pull/5095

PR #5179, opened 2018-09-16
main/gcc: patch to remove call to glibc-specific function in Ada library

ids1024:
Musl does not provide a pthread_rwlockattr_setkind_np() function.
Fixes https://bugs.alpinelinux.org/issues/9389


ids1024:
The Travis failure is not related to this change, and due to the Ada compiler not being installed.
Coincidentally, I filled a bug for the missing dependency earlier: https://bugs.alpinelinux.org/issues/9390

andypost:
> configure: error: GNAT is required to build ada
yes, this dependency needs to be resolved as @ncopa pointed in issue #9390

PR #5182, opened 2018-09-16
community/phonon: upgrade to 4.10.1

xpecex:
ref https://community.kde.org/Phonon/Releases/Core

### DIFF
--- a/community/phonon/APKBUILD
+++ b/community/phonon/APKBUILD
@@ -1,18 +1,20 @@
 # Contributor: William Pitcock <nenolod@dereferenced.org>
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=phonon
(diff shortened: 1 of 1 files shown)


andypost:
@xpecex please fix remains
```
>>> WARNING:phonon*: APKBUILD does not run any tests!
```
and a bit of clean-up
```
>>>phonon-dev*: Tracing dependencies...
>>> WARNING:phonon-dev*: qt-dev should be removed from depends_dev
>>> WARNING:phonon-dev*: qt-dev should be removed from depends_dev
>>> WARNING:phonon-dev*: qt-dev should be removed from depends_dev
	qt-dev
	glib-dev
	pc:QtCore
	pc:QtDBus
	pc:QtGui
	phonon=4.10.1-r0
	pkgconfig
```

PR #5184, opened 2018-09-16
main/xtables-addons: upgrade to 3.2

xpecex:


andypost:
`options="!check"` needed cos package has no test suite

ncopa:
we need update xtables-addons-vanilla too while at it. I can take care of that. Thanks!

PR #5187, opened 2018-09-17
main/nano: Upgrade to 3.1

tanertas:
https://savannah.gnu.org/bugs/?54642
https://lists.gnu.org/archive/html/nano-devel/2018-09/msg00045.html


tanertas:
New version is just released. PR is modified as a bare upgrade accordingly.

PR #5190, opened 2018-09-18
testing/py-backcall: enable test suite

Ikke:
pytest was broken due to missing dependencies, so the test suite did not
work. This has been fixed in the mean time, so the test suite can now be
run.

### DIFF
--- a/testing/py-backcall/APKBUILD
+++ b/testing/py-backcall/APKBUILD
@@ -8,14 +8,22 @@ url="https://github.com/takluyver/backcall"
 arch="noarch"
 license="BSD-3-Clause"
 makedepends="python2-dev python3-dev"
-options="!check" # pytest broken on alpine
+checkdepends="pytest"
 subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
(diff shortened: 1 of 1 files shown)


testing/py-backcall/APKBUILD
```diff
@@ -8,14 +8,24 @@ url="https://github.com/takluyver/backcall"
 arch="noarch"
 license="BSD-3-Clause"
 makedepends="python2-dev python3-dev"
-options="!check" # pytest broken on alpine
+#options="!check" # pytest broken on alpine
```
andypost: please remove this line instead of commenting out

testing/py-backcall/APKBUILD
```diff
@@ -8,14 +8,24 @@ url="https://github.com/takluyver/backcall"
...
+	python2 setup.py build
+	python3 setup.py build
+}
+
+check() {
+	cd "$builddir"
+
+	py.test-2
+	py.test-3
+
```
andypost: nitpick - no reason for empty line

testing/py-backcall/APKBUILD
Ikke: Yes, forgot to remove it.

PR #5196, opened 2018-09-18
testing/wireguard: upgrade to 0.0.20180918

zx2c4:
```
  * blake2s-x86_64: fix whitespace errors
  * crypto: do not use compound literals in selftests
  * crypto: make sure UML is properly disabled
  * kconfig: make NEON depend on CPU_V7
  * poly1305: rename finish to final
  * chacha20: add constant for words in block
  * curve25519-x86_64: remove useless define
  * poly1305: precompute 5*r in init instead of blocks
  * chacha20-arm: swap scalar and neon functions
  * simd: add __must_check annotation
  * poly1305: do not require simd context for arch
  * chacha20-x86_64: cascade down implementations
  * crypto: pass simd by reference
  * chacha20-x86_64: don't activate simd for small blocks
  * poly1305-x86_64: don't activate simd for small blocks
  * crypto: do not use -include trick
  * crypto: turn Zinc into individual modules
  * chacha20poly1305: relax simd between sg chunks
  * chacha20-x86_64: more limited cascade
  * crypto: allow for disabling simd in zinc modules
  * poly1305-x86_64: show full struct for state
  * chacha20-x86_64: use correct cut off for avx512-vl
  * curve25519-arm: only compile if symbols will be used
  * chacha20poly1305: add __init to selftest helper functions
  * chacha20: add independent self test
  
  Tons of improvements all around the board to our cryptography library,
  including some performance boosts with how we handle SIMD for small packets.
  
  * send/receive: reduce number of sg entries
  
  This quells a powerpc stack usage warning.
  
  * global: remove non-essential inline annotations
  
  We now allow the compiler to determine whether or not to inline certain
  functions, while still manually choosing so for a few performance-critical
  sections.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 
 pkgname=wireguard-tools
-pkgver=0.0.20180910
+pkgver=0.0.20180918
 pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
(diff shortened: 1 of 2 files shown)


andypost:
Commited http://dup.pw/aports/f745e49c

PR #5198, opened 2018-09-18
Use generic travis image

micw:
Simple change to travis.yml to use a "generic" image - saves a minute per build, see https://docs.travis-ci.com/user/languages/minimal-and-generic/ for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 dist: trusty
+language: generic
 sudo: required
 install:
   - sudo .travis/install-alpine


andypost:
Interestig idea, can you update description with alink to docs?

micw:
Sure, done.

andypost:
Looking at instal&setup files it could be https://docs.travis-ci.com/user/languages/minimal-and-generic/#minimal

andypost:
there some packages that lack of disk while bulding and that's why there no tests/check

jirutka:
Great catch! I think that `minimal` will be even better for our use case.

andypost:
@micw can you `commit --amend` with `minimal`?

PR #5164, opened 2018-09-13
main/libjpeg-turbo: Backport fix for CVE-2018-1152

euanh:
Cherry-pick commit f1322ac from the 1.5.x branch

Signed-off-by: Euan Harris <euan.harris@docker.com>

PR #5205, opened 2018-09-19
main/libjpeg-turbo: apply all patches applied to the 1.5.x branch

nmeum:
Since the 1.5.3 release (patch was generated using `git show
1.5.3..HEAD`). This includes the CVE-2018-1152 patch added in the
previous version of this aport. This change is basically a follow up to
a GitHub discussion https://github.com/alpinelinux/aports/pull/5164

---

Open question:

1. Why doesn't upstream do a new 1.5.4 release?
2. Does this fix any additional CVEs? See https://github.com/alpinelinux/aports/pull/5164


andypost:
is that really good idea to use "dev version" from some commit and how to track what really got changes

ncopa:
The patch does not apply cleanly. I think we should just cherry-pick specific issues that affects us til 1.5.4 is released.

> 1. Why doesn't upstream do a new 1.5.4 release?

I filed an issue upstream:
https://github.com/libjpeg-turbo/libjpeg-turbo/issues/285

> 2. Does this fix any additional CVEs? See #5164

Yes this fixes at least CVE-2018-11813

nmeum:
> is that really good idea to use "dev version"  …

This is not a "dev version" this is from the 1.5.x stable branch.

> The patch does not apply cleanly.

Oops, I fucked up. I can simply fix this.

> I filed an issue upstream […]

In that case we should just wait for an upstream response. 

> Yes this fixes at least CVE-2018-11813

We should definitely update the secfixes comment then.

nmeum:
After reading the [upstream respones](https://github.com/libjpeg-turbo/libjpeg-turbo/issues/285#issuecomment-423206397) I would suggest simply fetching the tarball for the latest commit from the `1.5.x` branch from GitHub. Ok?

nmeum:
Implemented this accordingly.

PR #5208, opened 2018-09-20
Add armv7 next to armhf in the arch line

ollieparanoid:
And also !armv7 where the arch line said !armhf. Done with the following script.

```sh
#!/bin/sh
cd ~/code/aports
for i in */*/APKBUILD; do
	arch_line=$(grep '^arch=' "$i")
	case $arch_line in
		*armv7*)
			continue
			;;
		*!armhf*)
			sed -i '/^arch=/ s/!armhf/!armhf !armv7/' "$i"
			;;
		*armhf*)
			sed -i '/^arch=/ s/armhf/armhf armv7/' "$i"
			;;
		*)
			continue
			;;
	esac
done
echo "Done!"
```

Related mail thread:
https://lists.alpinelinux.org/alpine-devel/6271.html


ollieparanoid:
Travis failed because the log was too long:
> The job exceeded the maximum log length, and has been terminated.

Ikke:
The commit message tells what is done, not why (and it basically tells what we can read from the diff anyway).

How about something like

> aports: add support for armv7
>
> More details about how and why

ollieparanoid:
@Ikke: right, thanks for taking a look at it. Is this one better?

fabled:
Applied. Thanks.

PR #5216, opened 2018-09-21
testing/py-zeroconf: upgrade to 0.21.2

fabaff:


ncopa:
maybe we can move this to community if its working?

PR #5217, opened 2018-09-22
[various]: enable stripping of go binaries

nmeum:
Stripping go binaries used to be an issue but nowadays it should work as
expected.

> We don't intentionally do anything that would make stripping a binary
> not OK, and strip has worked for the past five years or so.

https://groups.google.com/forum/?_escaped_fragment_=topic/golang-dev/ABppMOjYP6w#!topic/golang-dev/ABppMOjYP6w


nmeum:
gogs and hub fail. A fix for gogs is provided in #5218, hub likely only fails on travis but passes locally should thus probably work fine on the builders :man_shrugging:

PR #5219, opened 2018-09-22
Initial package build script that uses docker

micw:
That scripts allows easy package builds using docker images containing build environments for different archs. Proposal and discussion started in alpine-devel mailing list.


micw:
(currently broken, still working on it)

PR #5221, opened 2018-09-23
community/audacious: upgrade to 3.10

xpecex:
ref https://audacious-media-player.org/news/43-audacious-3-10-released


ncopa:
we also need update audacious-plugins at the same time.

PR #5223, opened 2018-09-23
main/py-requests: fix dependency requirements for idna

Ikke:
py-requests has a dependency requirement for idna <2.7, but alpine ships
idna-2.7. When a project uses setuptools, this version constraint is
checked and will raise an error.

Relax the version constraint by allowing idna-2.7 as well.

---
Not sure if this is the correct way to fix this nor if there are any incompatibilities with idna-2.7. This is meant to fix:

* [#9238](https://bugs.alpinelinux.org/issues/9238) 
* [#9053](https://bugs.alpinelinux.org/issues/9053 )


myhro:
Hi Ikke,

The last `py-requests` release doesn't seems to have this problem, as the [idna version constraint was already raised in there][1].

Wouldn't it be better to update the package instead of applying this patch?

[1]: https://github.com/requests/requests/blob/v2.19.1/setup.py#L53

Ikke:
Depends if we want to upgrade the package in 3.8 or not. If so, that would be indeed better. If we don't want to update, this can act as a fix.

ncopa:
I think we want upgrade py-requests in v3.8, and test that nothing breaks.

ncopa:
> Ok, I'll close this PR then and open a new one for that.

already fixed it add7b6dd3d789534d4cf60dbb7a049da403ee613

Ikke:
> > Ok, I'll close this PR then and open a new one for that.
> 
> already fixed it [add7b6d](https://github.com/alpinelinux/aports/commit/add7b6dd3d789534d4cf60dbb7a049da403ee613)

Thanks

PR #5235, opened 2018-09-24
testing/mobile-broadband-provider-info/APKBUILD -> community/mobile-b…

tanertas:
…roadband-provider-info:

Upgrade to 20180813

PR #5236, opened 2018-09-24
community/network-manager-applet: Upgrade to 1.8.18

tanertas:
Requires #5235


tanertas:
#5235 is move from testing to community.

ncopa:
> Requires #5235

It would be nice if you could have the commit in same PR if they depend on each other

PR #5241, opened 2018-09-25
main/hiredis: Fixed build hiredis-dev

sergey-safarov:


sergey-safarov:
Now include dir is pointed into `/usr/local/include/hiredis`. That is wrong value.
`sed` command is fix this

sergey-safarov:
Hello @nero 
Can you merge PR?

main/hiredis/APKBUILD
```diff
@@ -12,6 +12,7 @@ makedepends="$depends_dev"
 install=""
 subpackages="$pkgname-dev"
 source="$pkgname-$pkgver.tar.gz::$url/archive/v$pkgver.tar.gz"
+options="!check"
```
ncopa: would be nice with a comment explaining why checks are disabled

main/hiredis/APKBUILD
```diff
@@ -34,6 +35,7 @@ package() {
 	cd "$_builddir"
 
 	make PREFIX="$pkgdir"/usr install
+	sed -i -e 's:^libdir.*$:libdir=${exec_prefix}/lib:' -e 's:^includedir.*:includedir=${prefix}/include/hiredis:' $pkgdir/usr/lib/pkgconfig/hiredis.pc
```
ncopa: we normally prefer patches instead of sed, so we can detect if upsteram has fixed things.

There is also no explanation in commit message or in comment that explains why this is needed. Why can we not just leave it as it is?

PR #5243, opened 2018-09-26
main/nfs-utils: drop build dep on libnfsidmap-dev

HRio:
libnfsidmap-dev has been integrated into nfs-utils since 2.2.1

Fixes: #8487


HRio:
@ncopa can you help me out on this change? I am not sure how to best replace a package with subpackages in an other...

PR #5245, opened 2018-09-26
main/gtest: Upgrade to 1.8.1 and add gmock

z3ntu:
I copied most parts from the Arch Linux PKGBUILD. At first I was only missing gtest-config but gmock was missing for the software I was trying to compile too.

Also, I broke something in the subpackages because I cannot install `gtest-dev` due to the following error:
```
  gmock-1.8.1-r0:
    conflicts: gtest-dev-1.8.1-r0[pc:gmock=1.8.1] gtest-dev-1.8.1-r0[pc:gmock_main=1.8.1]
    satisfies: world[gmock><Q1PiuVDvsbsl8HcUkVmhrksOT8N3I=]
  gtest-dev-1.8.1-r0:
    masked in: cache
    conflicts: gmock-1.8.1-r0[pc:gmock=1.8.1] gmock-1.8.1-r0[pc:gmock_main=1.8.1]
    satisfies: world[gtest-dev><Q126Z+vjL6E0VwZHR0jFTXqgz6yBM=]
```


ncopa:
not really sure what the intention here is. `libgmock` seems to be shipped with `gtest`. the dev stuf is shipped with `gtest-dev`. gmock-config is not shipped at all and should not be needed, since gmock.pc is there.

I am ok with upgrading the package and split gmock out as separate package, but the rest should not be needed or should be fixed other places.

PR #5246, opened 2018-09-26
V38 libnfsidmap

HRio:


HRio:
wrong baseline in PR web ui.

HRio:
> @HRio Hint: you can modify an PR by force-pushing to the source branch, so you dont need to close pull requests because you accidentally picked the wrong heads.

ah did not know that, pushed with the right baseline, but the PR wui wanted master, I was too quick to notice. Thanks for the hint.

PR #5247, opened 2018-09-26
main/nfs-utils: drop build dep on libnfsidmap-dev

HRio:
libnfsidmap-dev has been integrated into nfs-utils since 2.2.1

Fixes: #8487


HRio:
@ncopa fix for alpine v3.8

ncopa:
i suppose we need to remove the libnfsidmap package too then, from v3.8?

HRio:
> i suppose we need to remove the libnfsidmap package too then, from v3.8?

yes, that right. Thanks!

PR #5252, opened 2018-09-27
main/xen: upgrade to 4.10.2

HRio:
Remove XSA-270 and XSA-274 from secfix list as they are not in XEN but in the Linux kernel.

Remove XSA-271 for the list due to that tha bug is in XAPI and not in XEN.


HRio:
@rnalrd XEN 4.10.2 for Alpine v3.8.

Note: only build tested, have no v3.8 bare metal test environment yet.

PR #5270, opened 2018-09-29
community/easy-rsa: upgrade to 3.0.5

J0WI:


andypost:
Makes sense to add `check()` same time like travis running the https://github.com/OpenVPN/easy-rsa/blob/v3.0.6/op_test.sh

J0WI:
`op_test.sh` does not pass with `abuild`

community/easy-rsa/APKBUILD
```diff
@@ -6,7 +6,7 @@ pkgdesc="Simple shell based CA utility"
 url="https://github.com/OpenVPN/easy-rsa"
 arch="noarch"
 license="GPL-2.0"
-depends="openssl1.0"
+depends="libressl"
```
andypost: that could wait til transition to `openssl`

J0WI: The transition to `opnessl` seems to block a lot of things here (e.g. #5406). Is there a timeline this?

andypost: transition done!

J0WI:
Rebased on top of new `openssl` package

PR #5273, opened 2018-09-29
community/gcompat: upgrade to 0.3.0

xpecex:
ref https://code.foxkit.us/adelie/gcompat/blob/master/CHANGELOG.rst


andypost:
Does it needs to be linked against `libucontext` now?

kaniini:
yes, needs to be linked against libucontext.  will do the bump myself when we are ready to upstream it.

PR #5280, opened 2018-09-30
community/aws-ena-driver-vanilla: move from testing, update to 1.6.0

tomalok:
Need to move this out of testing so that https://github.com/mcrute/alpine-ec2-ami
can eventually switch to more stable repositories for default AMI image builds.

Also, updating to the latest version of the driver.


tomalok:
going to redo PR from scratch.

PR #5285, opened 2018-10-02
testing/openttd: fix crash on save due to insufficient stacksize

dl00:
Seems like a pretty straightforward fix to me.


dl00:
is there something else I have to do or can this be merged?

PR #5298, opened 2018-10-02
community/kbuild: upgrade to 0.1.5

xpecex:
ref http://trac.netlabs.org/kbuild/wiki/Download


gena01:
I thought this was the latest version, however it seems that this one might be older than what I need for the other PR.

PR #5304, opened 2018-10-03
main/linux-vanilla: enable NFS v4.1 & v4.2

tomalok:
AWS Elastic File System works best with NFS v4.1, Docker's cloudstor:aws plugin requires it.
This is already enabled in `config-vanilla.s390x`; simply enabling it in the other configs.

@ncopa would be nice to get this in with the next round of kernel updates, thanks!


ncopa:
> @ncopa would be nice to get this in with the next round of kernel updates, thanks!

agree. Sorry that i missed it last time.

PR #5306, opened 2018-10-03
main/dovecot: upgrade to 2.3.3

andypost:
https://www.dovecot.org/list/dovecot-news/2018-October/000389.html


andypost:
@ncopa finally passed

maxwell-k:
The older version (2.3.2.1) switched from libressl to openssl and was bumped to -r4 in d7dec352994041227f126e10aa2c0bfcdbbc956f

fabled:
Needs rebase.

andypost:
@fabled rebased

fabled:
Could the `libressl`dependency be switched to `openssl`? It seems the only use case is the post-install script which create the default cert.

andypost:
yes, it works fine

andypost:
it builds locally fine but CI has not enough time limit to finish

PR #5319, opened 2018-10-04
testing/lttng-tools: upgrade to 2.10.6

z3ntu:


testing/lttng-tools/APKBUILD
```diff
@@ -13,34 +13,32 @@ depends_dev="popt-dev libxml2-dev userspace-rcu-dev lttng-ust-dev kmod-dev
 makedepends="$depends_dev bash grep coreutils automake autoconf"
 install=""
 subpackages="$pkgname-dev $pkgname-doc"
-source="https://lttng.org/files/$pkgname/$pkgname-$pkgver.tar.bz2
-	musl-has-no-dlmopen.patch"
-
-builddir="$srcdir/$pkgname-$pkgver"
```
andypost: why builddir removed

testing/lttng-tools/APKBUILD
```diff
@@ -13,34 +13,32 @@ depends_dev="popt-dev libxml2-dev userspace-rcu-dev lttng-ust-dev kmod-dev
...
+source="https://lttng.org/files/$pkgname/$pkgname-$pkgver.tar.bz2"
+options="!check" # Tests fail
```
andypost: then no reason to add `check()`

z3ntu:
```
============================================================================
Testsuite summary for lttng-tools 2.10.5
============================================================================
# TOTAL: 4742
# PASS:  2891
# SKIP:  1782
# XFAIL: 0
# FAIL:  65
# XPASS: 0
# ERROR: 4
```

z3ntu:
@andypost Could you please review again? Also 2.10.6 has been released meanwhile which I'll update to when there's some progress on this PR.

testing/lttng-tools/APKBUILD
andypost: did you check why they fails?

z3ntu: I'm not familiar with lttng-tools but looking at the test output I can't spot anything that would be useful

andypost: @z3ntu btw how many tests fail in this version?

andypost:
I think it makes sense to report upstream about failed tests

testing/lttng-tools/APKBUILD
z3ntu:
```
============================================================================
Testsuite summary for lttng-tools 2.10.6
============================================================================
# TOTAL: 4756
# PASS:  2905
# SKIP:  1782
# XFAIL: 0
# FAIL:  65
# XPASS: 0
# ERROR: 4
```

z3ntu: (4 more pass instead of fail when util-linux is added as checkdepends installed)

z3ntu:
There seems to be going on more or less duplicate work on the mailing list: 

* https://patchwork.alpinelinux.org/patch/4643/
* https://patchwork.alpinelinux.org/patch/4693/

z3ntu:
Tests succeed now when lttng-ust has the following patch applied: https://patchwork.alpinelinux.org/patch/4643/

Ready to merge, I'd say!

ncopa:
> Tests succeed now when lttng-ust has the following patch applied: https://patchwork.alpinelinux.org/patch/4643/
> 
> Ready to merge, I'd say!

How long does it take to run the tests? more than one hour? I don't think running tests that long is worth it. Are there some quicktest we could use instead?

PR #5323, opened 2018-10-04
testing/mxml: upgrade to 2.12, use https

prspkt:
[Release notes](https://github.com/michaelrsweet/mxml/releases/tag/v2.12).


andypost:
While on it makes sense to add `check()` with `make testmxml` or option
```
>>> WARNING: mxml: APKBUILD does not run any tests!
```

prspkt:
@andypost added check()

PR #5341, opened 2018-10-06
community/zabbix: update to 4.0.0

Ikke:


Ikke:
FYI: zabbix.a.o is already running 4.0.0 ;-)

PR #5342, opened 2018-10-06
[various]: force doc mantype in all aports setting it explicitly

nmeum:
See, for instance, https://bugs.alpinelinux.org/issues/7324

### DIFF
--- a/main/libasr/APKBUILD
+++ b/main/libasr/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=libasr
 pkgver=1.0.2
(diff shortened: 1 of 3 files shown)


nmeum:
libasr doesn't seem to build with gcc 8 and fetching the tarball for ruby failed. Will take a look at the former later.

nmeum:
The libasr issue is that GCC8 emits a new warning which GCC7 didn't seem to emit. This causes a build failure since we compile with `-Werror`. Possible solutions:

1. Apply https://github.com/OpenSMTPD/libasr/pull/18 locally. I would personally prefer to wait for an upstream response.

2. Don't compile with `-Werror`

PR #5349, opened 2018-10-07
main/nfs-utils: Fix and enable rpc.svcgssd

HRio:
enabled svcgssd build flag, applied patch from gentoo to make it
build and wrote another patch to fix segfaults, because musl
returns -1 for sysconf(_SC_GETGR_R_SIZE_MAX), which was not handled
at all, thus allocating a smaller buffer than needed.

(cherry picked from commit 16354c44b431a0c9861ad169a78490e6df3f283a)

### DIFF
--- a/main/nfs-utils/APKBUILD
+++ b/main/nfs-utils/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=nfs-utils
 pkgver=2.3.1
-pkgrel=4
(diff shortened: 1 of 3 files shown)


HRio:
@ncopa as requested in https://bugs.alpinelinux.org/issues/8470#note-3 I have backported this to v3.8.

I have tested it on a Kerberized Alpine v3.8 nfs-client mounting the nfs-server with ` mount -t nfs4 -o sec=krb5p,vers=4.0 ...`

PR #5354, opened 2018-10-08
testing/gnome-control-center: fix deprecated set_color call

mksully22:
migrate gnome_bg_set_color() to gnome_bg_set_rgba(). Also fix abuild check failure by adding transmission-lang package so test-endianess.c won't fail.


ncopa:
Wonderful! Thank you for looking into this.

The `transmission-lang` dependency looks weird though. It is the translations for the bittorrent client transmission. I cannot see how adding that could fix anything?

ncopa:
I am pushing this without `transmission-lang` and see what happens.

mksully22:
The endianess-test that was failing depends on the presence of /usr/share/locale which transmission-lang package added

PR #5356, opened 2018-10-08
testing/flatbuffers: fix broken const cast

mksully22:
cherry picked fix https://patch-diff.githubusercontent.com/raw/google/flatbuffers/pull/4698.patch
to eliminate build error "error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]"


andypost:
package already got upgrade to 1.10 so this patch no longer needed

andypost:
Ref https://github.com/alpinelinux/aports/commit/75676e55acefaa5499981ac04bdbca6033ba4dec

PR #5362, opened 2018-10-09
testing/openrct2: new aport

PureTryOut:
The patch has been upstreamed and will be included in the next release.

This application requires an installation of a retail copy of RollerCoaster Tycoon for the assets, but the code ran is completely GPL-3.0. This requirement is printed out by the .post-install script, but I'm not sure if that's the best way to do it.

I already played a few games this way and it works well, so I would love to add this directly to community rather than testing, but I don't believe that's what you guys want.


PureTryOut:
I've updated the PR to fetch the required objects and title sequences manually rather than having `make` download and install them.

PureTryOut:
Still waiting for a review!

PureTryOut:
I've updated the aport to the latest 0.2.2 release. The patch needed previously is not needed anymore.

PR #5363, opened 2018-10-09
testing/bwrap-oci: new aport

shawrkbait:
https://github.com/projectatomic/bwrap-oci
Run an OCI container using Bubblewrap


maxice8:
Merged, sorry for taking too long.

PR #3803, opened 2018-03-28
Add -openrc subpackage to all aports shipping runscripts

nmeum:
Probably want to backport https://github.com/alpinelinux/abuild/commit/54dbf5f383a7a4248517fb7ddcb3bbf7d22f0d8b first

CC: @kaniini, @awilfox

PR #5367, opened 2018-10-10
main/varnish: upgrade to 6.1.0

J0WI:
https://varnish-cache.org/releases/rel6.1.0.html

### DIFF
--- a/main/varnish/APKBUILD
+++ b/main/varnish/APKBUILD
@@ -3,17 +3,17 @@
 # Contributor: V.Krishn <vkrishn4@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=varnish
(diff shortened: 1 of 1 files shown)


andypost:
btw Please add `-openrc`
```
>>> varnish*: Running postcheck for varnish
>>> WARNING: varnish*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc
```

nmeum:
I would rather like to add an `-openrc` subpackages to all aports in one commit, as in #3803. However, I fell that we haven't discussed this extensively enough yet, so I guess not shipping the OpenRC services in a separate package is fine for now as well.

I merging this as is

nmeum:
Merged 31210df3074a646ce59de51d85b381ecad39e76b

PR #5370, opened 2018-10-11
main/opensmtpd: checkconfig should only output messages on error

nmeum:
Other OpenRC services don't output anything on standard error unless
something is not ok as well. The message `configuration OK` is therefore
pretty much useless.

Discussion: This is somewhat ugly. A proper fix would be to patch smtpd
and make sure that the "configuration ok" message is written to standard
output instead of standard error (which is were it belongs anyhow). After
doing so the message can simply be suppressed by running `smtpd -n >/dev/null`.

@jirutka You originally committed the checkconfig function in
47b9ff1206f. Any objections to the changes purposed here?

### DIFF
--- a/main/opensmtpd/APKBUILD
+++ b/main/opensmtpd/APKBUILD
@@ -4,7 +4,7 @@
 # Maintainer: Jonathan Curran <jonathan@curran.in>
 pkgname=opensmtpd
 pkgver=6.0.3p1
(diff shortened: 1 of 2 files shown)


nmeum:
See also: https://github.com/OpenSMTPD/OpenSMTPD/pull/881

PR #5372, opened 2018-10-11
testing/wok: fix checksum

andypost:
For some reason checksum changed http://build.alpinelinux.org/buildlogs/build-edge-aarch64/testing/wok/wok-2.5.0-r1.log


andypost:
@fcolista as tests does not pass I guess better just fix checksum

PR #5373, opened 2018-10-11
testing/py-ev: upgrade to 0.9.5 and modernize

andypost:
The package is moved and that's why build failed http://build.alpinelinux.org/buildlogs/build-edge-ppc64le/testing/py-ev/py-ev-0.9.0-r1.log


andypost:
@fabaff are you fine with this upgrade?

PR #5374, opened 2018-10-12
main/chrony: decrease startup time

nmeum:
While chrony can fork by itself it takes a while to do so. The startup
time of the chrony services can be significantly reduced by forking the
process into background using start-stop-daemon.

This significantly reduced the startup time of my laptop. Not sure if
this has any obvious disadvantages, I haven't discovered any so far.


ncopa:
I think the only disadvantage is that time may be wrong for a short time right after chrony has started. So if any other service require correct time and depends on "ntp", it may fail. I believe previous behavior was to block til time is synchronized.

I think we want have this configurable with a safe (slow) default.

PR #5375, opened 2018-10-12
main/busybox: fix incorrect timestamp generation in syslogd -Z patch

nmeum:
By using the non-standard timegm(3) function.

While it is not standardized by POSIX and a GNU extension musl has
support for it since 0.9.2 so we might as well use it.

See: https://bugs.alpinelinux.org/issues/9279#note-5

Fixes #9279
Fixes #7818

While at it also make sure that tm_year is initialized correctly (the
problem here being that the year is not part of the strptime format
string).

Fixes #8028


main/busybox/0011-sysklogd-add-Z-option-to-adjust-message-timezones.patch
```diff
@@ -79,7 +79,7 @@ index 4265f4f90..eca955891 100644
 -		now = 0;
 -		timestamp = msg;
 +		if (G.adjustTimezone && strptime(msg, "%b %e %T", &nowtm)) {
-+			now = mktime(&nowtm) - timezone;
++			now = timegm(&nowtm) - timezone;
```
nmeum: While at it we might as well check the return value of timegm, any thoughts on this?

nero:
I can confirm this is improving the behavior with the timezones.

8028 is not fixed by this. `mktime` is not failing anymore, but instead of getting 1970 - 1, i get a date on the 2^32 border in 2036.

nmeum:
@nero I investigated #8028 further and noticed that the problem described there is related to the fact that tm_year is not initialized correctly. I fixed that as well now and it seems to work in my x86 alpine chroot. Could you confirm that?

nmeum:
Just for the record: Busybox Git HEAD has a different solution for the same problem: https://git.busybox.net/busybox/commit/sysklogd/syslogd.c?id=9d539f9fbd0dc4ea70ed8ba66e3c78150fa8a8b2

We should probably migrate to that in the long run.

main/busybox/0011-sysklogd-add-Z-option-to-adjust-message-timezones.patch
```diff
@@ -76,19 +78,23 @@ index 4265f4f90..eca955891 100644
...
-+			now = mktime(&nowtm) - timezone;
-+			timestamp = ctime(&now) + 4; /* skip day of week */
-+		} else {
-+			now = 0;
-+			timestamp = msg;
++			/* strptime doesn't set the year */
++			tmp = time(NULL);
++			nowtm.tm_year = gmtime(&tmp)->tm_year;
++
++			if ((tmp = timegm(&nowtm))) {
```
nmeum: That needs to be changed to `if ((tmp = timegm(&nowtm) == (time_t)-1)))`

nero:
I found it very unpleasant to read and understand. If someone can confirm that this works, step forward.

nmeum:
> […] very unpleasant to read and understand […]

Well…it's busybox C code, it's generally not optimized for readability. That being said, I don't think that this patch is less readable than the rest of `syslogd.c`. Apart from the fact that "readability" is of cause a rather subjective attribute.

Keeping in mind that busybox 1.30 will have an [upstream solution](https://git.busybox.net/busybox/commit/sysklogd/syslogd.c?id=9d539f9fbd0dc4ea70ed8ba66e3c78150fa8a8b2) for the issue solved by the `-Z` patch I would not like to discuss cosmetic issues here, unless you could be more specific, but would rather focus on functional issues and from a functional standpoint this should correctly work as is and should fix both #9279 and #8028. If that's not true for your setup let me know.

nmeum:
Superseded by 397f0cd902eaadce3b74e1e3021e941d9bb6ba72

PR #5379, opened 2018-10-14
main/zip: Fix location for manpages

nero:
Previously:

```
nero@rei:/home/nero[0]: ls -la  /usr/share/man/                                                                                                       
total 1052
[snip]
-rw-r--r--    1 root     root         28496 Jul 27 14:26 zip.1.gz
-rw-r--r--    1 root     root           951 Jul 27 14:26 zipcloak.1.gz
-rw-r--r--    1 root     root           793 Jul 27 14:26 zipnote.1.gz
-rw-r--r--    1 root     root           596 Jul 27 14:26 zipsplit.1.gz
```

### DIFF
--- a/main/zip/APKBUILD
+++ b/main/zip/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Carlo Landmeter <clandmeter@gmail.com>
 pkgname=zip
 pkgver=3.0
-pkgrel=6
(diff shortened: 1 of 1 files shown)


nero:
The build failure is due to upstream server unreliability. Locally, retrying fixes it.

PR #5387, opened 2018-10-17
main/gpgme: create a new python bindings package - py3-gpgme

mikedlr:
This a change to the main/gpgme package which creates a package, `py3-gpgme` with the python bindings.  

the source for the bindings is in gpgme and can be found online at 
https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gpgme.git;a=tree;f=lang/python

compared with py3-pygpgme this set of bindings should be more up to date and should be maintained in future by the GNU Privacy Guard project. 

Compared to most other GNUPG related libraries this uses the recommended direct interface and so _should in principle_ reduce the risk of vulnerabilities due to problems with parsing GPG output or due to changes in the format of `gpg` output.

### DIFF
--- a/main/gpgme/APKBUILD
+++ b/main/gpgme/APKBUILD
@@ -25,7 +25,7 @@ build() {
 		--infodir=/usr/share/info \
 		--localstatedir=/var \
 		--enable-languages="python cpp" \
-		PYTHON_VERSIONS="2.7 3.6" \
+		PYTHON_VERSIONS="3.6" \
 	make
 }
 
@@ -42,7 +42,7 @@ check() {
 gpgmepp() {
 	pkgdesc="C++ bindings for GPGME"
 	mkdir -p "$subpkgdir"/usr/lib
-	cp "$pkgdir"/usr/lib/libgpgmepp.so.* "$subpkgdir"/usr/lib/
+	mv "$pkgdir"/usr/lib/libgpgmepp.so.* "$subpkgdir"/usr/lib/
 }
 
 py3() {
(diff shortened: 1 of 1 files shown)


mikedlr:
Marking this as WIP for now since I'm still in process of verifying that this is the right approach for the packaging and also doing tests to see if this actually does what I need.

main/gpgme/APKBUILD
```diff
@@ -40,8 +42,27 @@ check() {
 gpgmepp() {
 	pkgdesc="C++ bindings for GPGME"
 	mkdir -p "$subpkgdir"/usr/lib
-	mv "$pkgdir"/usr/lib/libgpgmepp.so.* "$subpkgdir"/usr/lib/
+	cp "$pkgdir"/usr/lib/libgpgmepp.so.* "$subpkgdir"/usr/lib/
```
ncopa: why is this changed to `cp`? it means that `libgpgmepp.so.*` will be provided by both main package and gpgmepp, which I don't think we want.

main/gpgme/APKBUILD
```diff
@@ -23,7 +23,9 @@ build() {
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
-		--localstatedir=/var
+		--localstatedir=/var \
+		--enable-languages="python cpp" \
+		PYTHON_VERSIONS="2.7 3.6" \
```
ncopa: Do we need support python 2.7?

mikedlr:
I'll revert the cp change. 

As far as python 2.7 goes, if it's not supported in during gpgme packaging then it's a complete pain for the user to add it later - they need to get exact version matching to avoid binary incompatibilities.  However I have no reason, personally to add it and I don't know of any particular software that requires it so I guess I could delete it for now on the grounds that any new software related to python should be using python 3.

PR #5390, opened 2018-10-17
main/python3: skip minor test failure on ppc64le

mksully22:
The test failure in question (test_memoryview_struct_module (test.test_buffer.TestBufferProtocol) ... FAIL) seems to have been introduced by the move to gcc8 (the test runs successfully on the prior gcc6 version). 

FYI this same failure was also encountered in Fedora https://bugzilla5.redhat.com/show_bug.cgi?id=1561011 on the gcc upgrade, deemed minor, and the test was disabled. 

This patch also disables this test for now on ppc64le for Alpine.

### DIFF
--- a/main/python3/APKBUILD
+++ b/main/python3/APKBUILD
@@ -5,7 +5,7 @@ pkgname=python3
 # the python2-tkinter's pkgver needs to be synchronized with this.
 pkgver=3.6.6
 _basever="${pkgver%.*}"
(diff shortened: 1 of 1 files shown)


mksully22:
I'm not sure why Travis is failing the test_urllib2net test. I can't recreate this failure on my x86_64 setup and the change in this PR should only affect ppc64le?

PR #5391, opened 2018-10-17
main/gtk+3.0: Fix touchscreen tapping menu items on X11

Decatf:
On GTK+3.0 and X11 tapping menus are broken since GTK 3.22.30 and higher.
This bug has been reported upstream[1] but there hasn't been a solution to
it.

The cause of the problem is an upstream GTK commit[2] which fixes
a particular touchscreen issues with Inkscape[3].

The scope of issue addressed in the breaking commit is much narrower
than tapping menus which affects any application that uses menus.
This patch reverts the upstream commit so that touchscreen devices using
GTK 3.0 on X11 are more usable again.

[1] https://gitlab.gnome.org/GNOME/gtk/issues/945
[2] https://gitlab.gnome.org/GNOME/gtk/commit/5c700cf575a341482d802edce999789feb17849a
[3] https://bugzilla.gnome.org/show_bug.cgi?id=781757


rnalrd:
@Decatf is there a fix from upstream as of now? Otherwise I'll just merge this.
Thanks

Decatf:
Nothing has changed upstream as far as I can see.

PR #5401, opened 2018-10-19
main/dnsmasq: upgrade to 2.80

J0WI:


main/dnsmasq/uncomment-conf-dir.patch
```diff
@@ -1,8 +1,11 @@
...
 +++ b/dnsmasq.conf.example	2016-08-02 11:30:34.094695721 +0200
-@@ -663,4 +663,4 @@
+@@ -671,7 +671,7 @@
  #conf-dir=/etc/dnsmasq.d,.bak
  
  # Include all files in a directory which end in .conf
 -#conf-dir=/etc/dnsmasq.d/,*.conf
 +conf-dir=/etc/dnsmasq.d/,*.conf
+ 
+ # If a DHCP client claims that its name is "wpad", ignore that.
```
nmeum: I personally think that this comment is confusing and I don't believe that this is the right place to document something like this.

main/dnsmasq/uncomment-conf-dir.patch
```diff
@@ -1,8 +1,11 @@
 --- a/dnsmasq.conf.example	2016-08-02 11:30:24.207034952 +0200
 +++ b/dnsmasq.conf.example	2016-08-02 11:30:34.094695721 +0200
-@@ -663,4 +663,4 @@
+@@ -671,7 +671,7 @@
  #conf-dir=/etc/dnsmasq.d,.bak
  
  # Include all files in a directory which end in .conf
 -#conf-dir=/etc/dnsmasq.d/,*.conf
 +conf-dir=/etc/dnsmasq.d/,*.conf
```
nmeum: Could you elaborate a bit why you enabled this by default?

nmeum:
Oh indeed, sorry. Would still be good to know why that configuration change is necessary ;)

J0WI:
The change is necessary if you want to use additional `/etc/dnsmasq.d/*.conf` files. Otherwise all files in this directory are ignored by default.

nmeum:
> The change is necessary if you want to use additional `/etc/dnsmasq.d/*.conf` files. 

The semantics of this option are clear to me. I am simply wondering why this change is necessary now as this option was disabled by default before. Is this at all related to the upgrade?

J0WI:
I kept this to avoid breaking existing configurations.

nmeum:
So this option was enabled by default in dnsmasq `2.79` and is disabled by default in `2.80`?

J0WI:
If by default means by default in Alpine Linux, then yes.

mikecook:
The conf-dir option was enabled 2 years ago for dnsmasq 2.76-rc1 in alpine v3.5.0_rc1 with this commit: https://github.com/alpinelinux/aports/commit/50d3f9a02bd5f9c4c442d65dd0bf673be249ad26#diff-98b1d8190d65b0fa848f242accdd3512L3

Github is improperly highlighting as new a change that was made in alpine 2 years ago. Changing it at this point would break expectations. If you still want to change it back to upstream, can that be done in a different PR?

This is a clean, meticulous patch (that now needs rebase due to 2.79-rc2) for 2.80.

J0WI:
@nmeum am I requested to do any further changes?

nmeum:
b14fc437abacc82496ff9f347444eb0f5ebd2f3d merged. Thanks!

PR #5402, opened 2018-10-19
main/imagemagick: disable openmp for stability for 3.8

csandanov:
https://bugs.alpinelinux.org/issues/9190

### DIFF
--- a/main/imagemagick/APKBUILD
+++ b/main/imagemagick/APKBUILD
@@ -25,10 +25,6 @@ build() {
 	sed -i -e \
 		's:DOCUMENTATION_PATH="${DATA_DIR}/doc/${DOCUMENTATION_RELATIVE_PATH}":DOCUMENTATION_PATH="/usr/share/doc/imagemagick":g' \
 		configure
-	local _openmp=
-	case "$CARCH" in
-	s390x) _openmp="--disable-openmp"
-	esac
 
 	./configure \
 		--build=$CBUILD \
@@ -38,7 +34,7 @@ build() {
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
 		--disable-static \
-		$_openmp \
+		--disable-openmp \
 		--with-threads \
 		--with-x \
 		--with-tiff \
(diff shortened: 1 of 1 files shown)


andypost:
It was disabled in edge https://github.com/alpinelinux/aports/commit/a23a75a8962b419f92fe63ba60317496bbfbe712

andypost:
This PR needs work to increase `pkgver` 

Meanwhile I suggest to use `graphicsmagick` as faster and more stable alternative

J0WI:
The checksum needs to be updated.

csandanov:
Can we merge this please?

csandanov:
Ok, I see @ncopa included the original fix to [3.8.2 roadmap](https://bugs.alpinelinux.org/versions/128) but it looks far from being ready to release. Perhaps I can downgrade ImageMagic to 7.0.7.x so we keep the version relatively the same (few patches will be included though) and release with disabled openmp in 3.8 directly. 

On the other hand, [here](https://github.com/ImageMagick/ImageMagick/issues/1272) ImageMagick maintainers suggested to try the latest version. So, there's a chance the latest 7.0.8.x woudl work even with openmp enabled. This issue is easy-to-reproduce on vanilla WordPress and I would test it but edge repository already has this package with disabled openmp. 

@ncopa, any suggestions?

andypost:
7.0.8.14 released 
@csandanov please squash commits for upgrade & checksum

meanwhile why trying to fix tests for `php7-gmagick` I came to the same option for `graphicsmagick`

Also voidlinux made the same https://github.com/void-linux/void-packages/issues/3649

andypost:
It was fixed in edge in https://github.com/alpinelinux/aports/commit/a23a75a8962b419f92fe63ba60317496bbfbe712 by @ncopa

andypost:
@csandanov is there a steps to reproduce the issue like it was with graphicsmagick?
Meanwhile there is long tread in void-linux and few other places. As I got it slightly depends on gcc version

csandanov:
I'm not aware of graphicsmagick issues. If you run vanilla WordPress with enabled imagick extension the default image upload just doesn't work. There's a similar issue has been reported with easier way to reproduce at https://github.com/ImageMagick/ImageMagick/issues/1272

ncopa:
this was fixed with edd37627b316f881325df6073fb379a4671d366e.

We cannot upgrade to 7.0.8. See https://bugs.alpinelinux.org/issues/9742

PR #5414, opened 2018-10-20
main/perl-mozilla-ca: modernize and upgrade to 20180117

J0WI:


fabled:
Is this manual work, or did you use apkbuild-cpan to rewrite the aport?

J0WI:
This was manually done. I wasn't aware of `apkbuild-cpan` and it does not work for me.

ncopa:
fixed with 3f2fb910adfa1aedc48160b3810617da62809dcc

PR #5420, opened 2018-10-20
community/php7-pecl-igbinary: upgrade to 2.0.8

andypost:
ref https://pecl.php.net/package-changelog.php?package=igbinary&release=2.0.8


andypost:
Now can move with https://github.com/alpinelinux/aports/pull/5029

PR #5421, opened 2018-10-20
main/perl-net-http: modernize and upgrade to 6.18

J0WI:


ncopa:
fixed with 050b3fb3ba6966538cb063e6f37acd5f07f1f5a1

PR #5422, opened 2018-10-20
main/perl-datetime: modernize and upgrade to 1.50

J0WI:


ncopa:
fixed with 5a827272ce743e0cb3a875c4c70df3144a3ce42b

PR #5424, opened 2018-10-20
main/perl-http-body: modernize and upgrade to 1.22

J0WI:


ncopa:
fixed with 7454a74b607d2ea338d1742b0ae51257b74fafe9

PR #5425, opened 2018-10-20
main/perl-devel-checkbin: modernize and upgrade to 0.4

J0WI:


ncopa:
fixed with 8ce6a205ff92427784b9c822c5229a6dc7c1e093

PR #5427, opened 2018-10-20
main/perl-class-accessor: modernize and upgrade to 0.51

J0WI:


ncopa:
fixed with 60eb89d6a116718372f0c795c86b6abde053db18

PR #5431, opened 2018-10-20
community/php7-pecl-event: renamed from php7-event, modernize

andypost:


andypost:
Can't reproduce failure locally and on arm8

andypost:
The same test failed on aarch64 builder, guess because the test makes external connection https://bitbucket.org/osmanov/pecl-event/src/c0f1f1060100f999652db2fb5376ad070086dfd0/tests/04-bevent-socket.phpt?at=master&fileviewer=file-view-default

andypost:
Follow-up https://github.com/alpinelinux/aports/pull/5439

PR #5435, opened 2018-10-20
community/php7-pecl-mailparse: renamed from php7-mailparse, modernize…

andypost:
… and add check 


andypost:
@fabiorphp are you fine with move?

fabiorphp:
Absolutely!

PR #5439, opened 2018-10-20
community/php7-pecl-event: fix check() which require net

andypost:
WIP trying to fix http://build.alpinelinux.org/buildlogs/build-edge-aarch64/community/php7-pecl-event/php7-pecl-event-2.4.1-r1.log


andypost:
Looks like network issue, now package build fine

PR #5440, opened 2018-10-20
community/php7-pecl-amqp: renamed from php7-amqp, modernize

andypost:


andypost:
@fabiorphp are you fine with changing desription to be more inline with https://pecl.php.net/package/amqp

fabiorphp:
Yes, go ahead!

PR #5450, opened 2018-10-20
community/php7-pecl-imagick: renamed from php7-imagick, modernize, fi…

andypost:
fix license https://pecl.php.net/package/imagick


andypost:
@vakartel are you fine with move?

PR #5452, opened 2018-10-21
main/glib: don't depend on perl

pksadiq:
All glib perl scrips have been ported to python


pksadiq:
This has been fixed in 3d02166a3dfa7b700716202544ed8b5eab146ee4

PR #5456, opened 2018-10-21
community/py-pyldap: upgrade to 3.2.0

J0WI:


andypost:
There's dependent packages
```
community/ssh-ldap-pubkey/APKBUILD:10:depends="python3 py3-pyldap py3-docopt"
testing/wok/APKBUILD:11:        py2-pyparted  py2-pyldap py2-psutil py2-m2crypto py-cheetah
```
so they also needs testing with this major upgrade

andypost:
Checked both packages and they have no restrictions on specific versions for py-ldap

@jirutka It looks good to go

community/py-pyldap/APKBUILD
```diff
@@ -1,17 +1,15 @@
...
+pkgver=3.1.0
 pkgrel=0
 pkgdesc="Python modules for implementing LDAP clients"
-url="https://github.com/pyldap/pyldap"
+url="https://github.com/python-ldap/python-ldap"
 arch="all"
 license="PSF"
-depends="libldap>=$_vermajor.$_verminor"
-makedepends="python2-dev python3-dev openldap-dev>=$_vermajor.$_verminor"
+depends="libldap"
```
jirutka: This version requires new dependencies, see [setup.py](https://github.com/python-ldap/python-ldap/blob/ec41070d3fa1d00cb272f86e0c8ca55db51bdf62/setup.py#L166-L169):

```
  install_requires=[
    'pyasn1 >= 0.3.7',
    'pyasn1_modules >= 0.1.5',
],
```

J0WI: Fixed, thanks!
See also #5641

J0WI:
@jirutka could you please have a second look here?

andypost:
I think it should be one commit, because without fixed deps first one makes no sense

J0WI:
I squased commits.

J0WI:
I got the following error with `noarch`:
```
>>> py2-python-ldap*: Preparing subpackage py2-python-ldap...
>>> ERROR: py2-python-ldap*: Arch specific binaries found so arch must not be set to "noarch"
>>> ERROR: py2-python-ldap*: prepare_package failed
>>> ERROR: py-pyldap*: prepare_subpackages failed
>>> ERROR: py-pyldap: rootpkg failed
```

Ikke:
@J0WI Can you rebase this to fix conflicts?

The arch warning is normal when you have a dummy main package which has no arch, but the subpackages do have arch related files. (You cannot specify the main package to have noarch, but subpackages to have arches).

And only if it's no hassle, would it be a problem to drop python2 support, which is reaching EOL soon?

J0WI:
Done.

PR #5471, opened 2018-10-23
main/busybox: no acpid logrotate mail

HRio:
Before this change a mail with the following is generated when acpid.log is rotated:
```console
 * Stopping busybox acpid ... [ ok ]
 * Starting busybox acpid ... [ ok ]
```


HRio:
other log-rotation scripts does not generate useless cron mails, I would like to only get mails from cron when something actually happened that I need to investigate.

HRio:
> The checksum for acpid.logrotate has not been updated

thanks, fixed.

HRio:
@ncopa it this OK with you?

PR #5475, opened 2018-10-23
main/py-markupsafe: add check() and modernize

andypost:


andypost:
@ncopa I bet it help to move on armv7 build

PR #5476, opened 2018-10-23
main/py-jinja2: add check() and modernize

andypost:
The package itself must define its dependency to point right build order
This is why jinja2 was build without its real dependency, adding `check()` should prevent it in future

Ref http://build.alpinelinux.org/buildlogs/build-edge-armv7/main/py-sphinx/py-sphinx-1.8.1-r0.log

### DIFF
--- a/main/py-jinja2/APKBUILD
+++ b/main/py-jinja2/APKBUILD
@@ -3,14 +3,14 @@
 pkgname=py-jinja2
 _pkgname=Jinja2
 pkgver=2.10
-pkgrel=1
+pkgrel=2
 pkgdesc="A small but fast and easy to use stand-alone python template engine"
 url="http://jinja.pocoo.org/"
 arch="noarch"
 license="BSD"
-_py2_depends="py2-markupsafe"
-_py3_depends="py3-markupsafe"
-makedepends="python2-dev python3-dev py-setuptools $_py2_depends $_py3_depends"
+depends="py-markupsafe"
+checkdepends="pytest"
+makedepends="python2-dev python3-dev py-setuptools"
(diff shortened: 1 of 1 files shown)


andypost:
Tested on arm8 & armhf

ncopa:
i fixed it differently, maybe you can rebase it to include the check()?

andypost:
Fixed deps according suggested template https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python#Multiversion_package_without_executables

main/py-jinja2/APKBUILD
```diff
@@ -3,14 +3,14 @@
...
-pkgrel=1
+pkgrel=2
 pkgdesc="A small but fast and easy to use stand-alone python template engine"
 url="http://jinja.pocoo.org/"
 arch="noarch"
 license="BSD"
-_py2_depends="py2-markupsafe"
-_py3_depends="py3-markupsafe"
-makedepends="python2-dev python3-dev py-setuptools $_py2_depends $_py3_depends"
+depends="py-markupsafe"
```
ncopa: I am not sure i like this change, which requires that py2/py3-markupsafe are built as sub packages from py-markupsafe. If we for some reason in future split py-markupsafe into 2 separate APKBUILDs, py2-markupsafe and py3-markupsafe, then things will break.

I think its better to explicitly depend on py2/py3-markupsafe and let build servers pull in whatever provides those.

ncopa: ok, i looks like this is according the documentation.

andypost:
@ncopa yes, I think better to have all packages more inline withdocs so further conversions will be easy

PR #5478, opened 2018-10-23
community/opam: upgrade to 2.0.1

avsm:


avsm:
@ncopa - would it be possible to get this into alpine 3.9 if the patch is ok? We've deprecated opam 1.2.2, so its package repository no longer regularly updates.

ncopa:
Yes, i think we want this for v3.9, thanks!

avsm:
Thanks @ncopa!

community/opam/APKBUILD
```diff
@@ -1,18 +1,17 @@
 # Contributor: Michael Zuo <muh.muhten@gmail.com>
 # Maintainer: Anil Madhavapeddy <anil@recoil.org>
-# FIXME: lib-ext downloads dependencies; package them and use system-provided!
```
jirutka: Have you fixed this? Or opam 2.0.x does not download deps anymore?

PR #5480, opened 2018-10-23
community/drupal7: security upgrade 7.60 - SA-CORE-2018-006

xpecex:
ref: https://www.drupal.org/SA-CORE-2018-006 and https://www.drupal.org/project/drupal/releases/7.60

### DIFF
--- a/community/drupal7/APKBUILD
+++ b/community/drupal7/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Andy Postnikov <apostnikov@gmail.com>
 _php=php5
 pkgname=drupal7
-pkgver=7.59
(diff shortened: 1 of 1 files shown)


community/drupal7/APKBUILD
```diff
@@ -16,6 +16,7 @@ pkggroups="www-data"
 source="http://ftp.drupal.org/files/projects/drupal-$pkgver.tar.gz"
 
 builddir="$srcdir/drupal-$pkgver"
+options="!check" # This package not have testsuite
```
andypost: it has but it require at least sqlite to run - I will try to fix it

PR #5486, opened 2018-10-25
main/fail2ban: improve sshd filter

HRio:


HRio:
@clandmeter can you check this? Based on what I get in `/var/log/messages` for a default alpine v3.8 installation directly on Internet (a few days worth of logging).

HRio:
@rnalrd can you check these? thanks

PR #5488, opened 2018-10-25
testing/openssl: remove support for TLS compression

J0WI:
https://en.wikipedia.org/wiki/CRIME_(security_exploit)
It's already disabled in other distros, e.g. Debian.

### DIFF
--- a/testing/openssl/APKBUILD
+++ b/testing/openssl/APKBUILD
@@ -2,14 +2,13 @@
 pkgname=openssl
 pkgver=1.1.1
 _abiver=${pkgver%.*}
-pkgrel=2
-pkgdesc="Toolkit for SSL v2/v3 and TLS v1"
+pkgrel=3
+pkgdesc="Toolkit for Transport Layer Security (TLS)"
 url="https://www.openssl.org"
 arch="all"
 license="OpenSSL"
-depends_dev="zlib-dev"
 makedepends_build="perl"
-makedepends_host="zlib-dev linux-headers"
+makedepends_host="linux-headers"
 makedepends="$makedepends_host $makedepends_build"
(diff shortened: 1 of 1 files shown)


ncopa:
I was unsure that we could do this, but it seems we can. I'm all for it.

ncopa:
commit that removed it from debian: https://salsa.debian.org/debian/openssl/commit/37e59f2704565b52c59e7162caf21fb760e763e3

J0WI:
Other flags to consider are `no-des`, `no-idea`, `no-seed` or `no-rc4`: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/INSTALL#L541-L547

I think they are already dropped from LibreSSL (not sure about RC4).

J0WI:
`SEED` ciphers don't occur in the current `openssl ciphers` list anyway.

J0WI:
> I think we disable as much as we can and add things back if/when something shows up that needs it.

@ncopa should I remove everything in this PR?

ncopa:
I wonder if you can merge the commits and have commit message to something like: "update description and remove more weak cryptos" and in the body of commit message you give a more complete list of what was removed. Also mention there that those cryptos are removed in libressl or debian.

J0WI:
I squashed commits and added references, but I just realized that this has already been merged in https://github.com/alpinelinux/aports/commit/abe1dc5988d12f5aca771605b109390f33ce7519.

I'm not sure about DES, RC4 and PSK. But let's see if they are required somewhere.

ncopa:
> I'm not sure about DES, RC4 and PSK. But let's see if they are required somewhere.

DES is needed by NTLM support in cyrus-sasl. RC4 is needed by freeradius. I havent found anything that needs PSK yet. I will fix those things when they show up.

Thanks!

PR #5501, opened 2018-10-25
main/syslinux: Changed update-extlinux to detect intel-ucode

HRio:
If `/boot/intel-ucode.img` is present, update-extlinux will add it as
(first) INITRD. This way the microcode update in package non-free/intel-ucode
will be automatically loaded, if installed. This package fixes, among other
processor bugs, the hyper-threading bug in Intel Skylake and Kaby Lake
processers currently in the news [1]. More details here [2] and here [3].

[1] https://lists.debian.org/debian-devel/2017/06/msg00308.html
[2] https://wiki.archlinux.org/index.php/Microcode#Syslinux
[3] https://wiki.gentoo.org/wiki/Intel_microcode


HRio:
@ncopa as discussed here is the intel-ucode support from https://patchwork.alpinelinux.org/patch/3443/

nero:
Is there a strong reason against including it into mkinitfs itself, via `/etc/mkinitfs/features.d`? Both methods would result in the microcode being present in the initramfs. Not entirely sure of what else `intel-ucode.img` would be doing.

nero:
Nevermind, this mechanism is documented in [Documentation/x86/microcode.txt](https://github.com/torvalds/linux/blob/master/Documentation/x86/microcode.txt) of the linux kernel.

HRio:
@nero yes it implements this method https://wiki.archlinux.org/index.php/Microcode#Syslinux (ref 2). One good thing with this method is that we do not have to trigger rebuild of all initramfs files in /boot on intel-ucode updates.

nero:
Your links do not answer my question: "Is there a strong reason against including it into mkinitfs itself".

I did some research and i'll quote the kernel code to answer it myself:

Quote from kernel `lib/earlycpio.c`
> /*
>  * earlycpio.c
>  *
>  * Find a specific cpio member; **must precede any compressed content**.
>  * This is used to locate data items in the initramfs used by the
>  * kernel itself during early boot (before the main initramfs is
>  * decompressed.)  It is the responsibility of the initramfs creator
>  * to ensure that these items are uncompressed **at the head of the**
>  * **blob**.  Depending on the boot loader or package tool that may be a
>  * separate file or part of the same file.
>  */

This all happens pre-SMP and before parsing of the regular initrd.

This mechanism is used for AMD & Intel ucode and ACPI tables.

main/syslinux/update-extlinux
```diff
@@ -165,21 +165,16 @@ for kernel in $(find /boot -name "vmlinuz*" -type f); do
...
-	else
-		if [ -f "/boot/intel-ucode.img" ]; then
-			everbose "Found microcode for Intel CPUs: /boot/intel-ucode.img"
-			initrd="intel-ucode.img${initrd}"
-			echo "  INITRD intel-ucode.img" >> $conf.new
-		fi
 	fi
+	if [ -f "/boot/intel-ucode.img" ]; then
+		everbose "Found microcode for Intel CPUs: /boot/intel-ucode.img"
+		[ -n "$initrd" ] && initrd="intel-ucode.img,${initrd}" || initrd="intel-ucode.img"
```
ncopa: Alternative:
````sh
initrd="intel-ucode.img${initrd:+,}${initrd}"
````

HRio: thanks, changing.

PR #5502, opened 2018-10-25
community/shellcheck: new aport

myhro:
ShellCheck, a static analysis tool for shell scripts - https://www.shellcheck.net

Requested on issues [#6417][1] and [#7105][2], currently part of the [3.9.0 roadmap][3].

I have a couple questions about this aport:

- As it is distributed as a single binary (like the Go ones), I'm considering that its build dependencies doesn't need to be packaged. I think this is OK, based on what I saw on other packages that fetches dependencies from the internet, but I never asked if there's a intention to aim for completely-offline-reproducible-builds.
- Is it enough to distribute it for `x86_64` only? Or should we work improving the Haskell infrastructure to support more architectures before offering it?

[1]: https://bugs.alpinelinux.org/issues/6417
[2]: https://bugs.alpinelinux.org/issues/7105
[3]: https://bugs.alpinelinux.org/versions/127

### DIFF
--- a/community/shellcheck/APKBUILD
+++ b/community/shellcheck/APKBUILD
@@ -1,0 +1,34 @@
+# Contributor: Tiago Ilieve <tiago.myhro@gmail.com>
+# Maintainer: Tiago Ilieve <tiago.myhro@gmail.com>
+pkgname=shellcheck
+pkgver=0.5.0
+pkgrel=0
+pkgdesc="ShellCheck, a static analysis tool for shell scripts"
+url="https://www.shellcheck.net/"
+arch="x86_64" # Haskell infrastructure is only supported on x86_64
+license="GPL-3.0-only"
+makedepends="cabal ghc"
+source="shellcheck-$pkgver.tar.gz::https://github.com/koalaman/shellcheck/archive/v${pkgver}.tar.gz"
+
+build() {
+	cd "$builddir"
+
+	cabal update
+	# A symlink to the binary has to be discarded, as it can't be disabled.
+	cabal install --bindir "$builddir" --symlink-bindir /tmp
+}
+
+check() {
+	cd "$builddir"
+
+	cabal test
+	./shellcheck --version
+}
+
+package() {
+	cd "$builddir"
+
+	install -Dm755 ./shellcheck "$pkgdir"/usr/bin/shellcheck
+}
+
+sha512sums="87861cddb353262630e4370e12e508224b3c14e128082909b4b35f0526dfe648a744d68cc27f77f2f8bb098af37a2af7bdc805d88018bba5e48b6c1ff1749f10  shellcheck-0.5.0.tar.gz"


nero:
I want to add that the review relates to the successful build and functional result.

There is currently no policy for the issues you mentioned in the contributing guide. Depending on what the devs say, i it should probably be added there.

andypost:
Needs more opinion on haskell dependency and now sure it will be useful because only x86_64

nero:
Im having some afterthoughts on this since it uses ~/.cabal/ outside of the designated build directories. I dont think we can keep it like that.

myhro:
@nero This bothers me as well. I'll try to change it to `$builddir` and let you know.

myhro:
@nero Done. Can you please take another look?

myhro:
Hi @nmeum,

> Especially since most of the build system downloading dependencies from the internet do not properly verify their integrity.

I'll compare the Haskell situation to Go, because the later is the one which I have experience with its dependencies ecosystem.

I looked around for options in this regard and [`cabal` allows dependencies to be frozen][1], similar to what we do [using `glide.lock` files][2] on some Go packages. But lock files aren't meant for integrity checks - that's why [`go.sum` files were introduced on Go 1.11][3].

So the question is: should we block packages from entering Alpine because their build system doesn't offer integrity checks for its dependencies?

I'm asking that because this one doesn't looks to be the first case. Our very own [`community/docker` package calls `go get` directly][4], which not only doesn't offer integrity checks, but also doesn't even support versioning.

[1]: https://www.haskell.org/cabal/users-guide/developing-packages.html#freezing-dependency-versions
[2]: https://glide.readthedocs.io/en/latest/glide.lock/
[3]: https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using
[4]: https://github.com/alpinelinux/aports/blob/9485160/community/docker/APKBUILD#L139-L149

myhro:
> If you ask me: Yes. The best way to handle dependencies is to package them. Though I am aware that this is not feasible with some systems (go is one example but a lot of go software vendors dependencies these days which fixes this). Others might see this differently.

I'm not (nor intend to be) a Haskell expert, but it seems to be the case here as well.

> Yep, that's also what `testing/idris` seems to do.

Thanks for pointing out that example. I've also checked again how `cabal` and the Hackage repository works and (although I haven't found that stated in their docs) it looks like they do properly signing of the packages in there. In `~/.cabal/packages/hackage.haskell.org/root.json` there are lines like:

```
{
  "signatures": [
    {
      "keyid": "fe331502606802feac15e514d9b9ea83fee8b6ffef71335479a2e68d84adc6b0",
      "method": "ed25519",
      "sig": "SmAxDgEMLLKJl0sJnHh/uLwZx8CMP0zdrMqpS/19eDpeH/MHt6a4ydDBeuRtEOn3xH52bfqhRGXVaKi7aWSfBw=="
    },
(...)
  "signed": {
    "_type": "Root",
    "expires": "2019-07-21T18:34:14Z",
    "keys": {
      "0a5c7ea47cd1b15f01f5f51a33adda7e655bc0f0b0615baa8e271f4c3351e21d": {
        "keytype": "ed25519",
        "keyval": {
          "public": "zazm5w480r+zPO6Z0+8fjGuxZtb9pAuoVmQ+VkuCvgU="
        }
(...)
```

Package versions in their repository are also granted to be immutable, [as they can't be deleted from there][1] - hence can't be replaced.

So pinning the versions would have the same effect as `go.mod`/`go.sum` in the Go ecosystem. Relying on the programming language packaging tools for trusted builds -  instead of redoing this work using Alpine tooling - makes total sense to me.

I really don't want to contribute to the madness of flooding Alpine repositories with thousands of Go/Haskell/etc. packages, for the sake of building a single CLI tool, like other distributions are doing. It's not the same thing as interpreted languages, like Python, where the dependencies are needed not just for build but also during runtime.

[1]: https://hackage.haskell.org/

tomalok:
FWIW, I've submitted PR #5973 to provide a common executable for all the packages that seem to want to use it for processing markdown to man pages.

oxr463:
@myhro any updates on this?

tcely:
See also #9936

PR #5514, opened 2018-10-27
community/py-greenlet: fix patch for ppc64le

andypost:
Ref https://github.com/python-greenlet/greenlet/issues/139#issuecomment-433650221

Meanwhile probably now we could remove `-O0` optimization and use standard one `-Os` but that will need pkgrel bump


andypost:
@awilfox I guess Adélie could backport it as well

awilfox:
Thank you for poking me, but we don't ship greenlet right now.  Nice to see ppc64 support improving upstream!

PR #5515, opened 2018-10-28
community/fzf: upgrade to 0.17.5

myhro:
Drop `*-doc` package and move keybindings to the respective shell
completions package, making them consistent with their descriptions.


Ikke:
@myhro Thanks, looks good. Note that `0.17.5` is out, we might bump the version as well?

myhro:
Thanks for the quick response. I've upgraded the package to the newest version.

PR #5518, opened 2018-10-29
community/py-greenlet: fix test fail on ppc64le by adding omit-frame-…

mksully22:
…pointer and removing obsolete no-tree-dominator-opts option


andypost:
See https://github.com/alpinelinux/aports/pull/5514

mksully22:
Thanks Andy, I will close

PR #5522, opened 2018-10-31
community/php7-pecl-igbinary: fix load order

andypost:
it must be loaded before php7-pecl-redis and php7-pecl-couchbase

Ref https://github.com/gliderlabs/docker-alpine/issues/440


andypost:
> andypost: should I try to rm previous ini file or apk will do it for me?
> ncopa: apk will do it for you, unless the config file is modified by user

PR #5533, opened 2018-11-01
testing/py-astral: upgrade to 1.7.1

fabaff:


andypost:
The `check()` in package is wrong)
So better disable it because it requires maps api key https://travis-ci.org/sffjunkie/astral/jobs/446135937

fabaff:
Thanks for the feedback. `check()` disabled.

PR #5536, opened 2018-11-01
testing/rakudo: fix ppc64le test break by increasing performance tolerance

mksully22:
t/08-performance/99-misc.t subtest4 collects the time taken to do a grep vs hyper.grep and compares the elapsed time. The tolerance is subjective and and the results are just out of range on ppc64le. This is a minor performance test issue and can be fixed by increasing the comparison tolerance for the subtest.

### DIFF
--- a/testing/rakudo/APKBUILD
+++ b/testing/rakudo/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Curt Tilmes <Curt.Tilmes@nasa.gov>
 pkgname=rakudo
 pkgver=2018.10
-pkgrel=0
+pkgrel=1
 pkgdesc="A compiler for the Perl 6 programming language"
 url="http://rakudo.org/"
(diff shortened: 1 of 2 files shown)


mksully22:
seems like Travis failure is due to nqp and moarvm packages still being at the 09 level

andypost:
It makes sense to promote upstream

andypost:
please squash commits, no reason to bump pkgrel cos package was not build and test does no affect on binary

mksully22:
K I will close

andypost:
btw it can increase pkgrel for all 3 packages so ci can bebuild them all

andypost:
@mksully22 looking in issue from your patch https://github.com/rakudo/rakudo/commit/057966987937d645b74cab33ae95766896ff07c6 mentioned as solution - we can disable this test while it fails on all builders except `aarch64`

PR #5539, opened 2018-11-02
main/varnish: upgrade to 6.1.1

andypost:
Ref https://varnish-cache.org/releases/rel6.1.1.html#rel6-1-1


andypost:
https://github.com/varnishcache/varnish-cache/issues/2808 The same issue

PR #3468, opened 2018-03-03
main/py-flake8: fixes

orsonteodoro:
Enabled check()
Added check()
Added py-tox to checkdepends
Increment pkgrel
Added py2-configparser ( #3467 ) as python2 dependency

PR #5559, opened 2018-11-04
main/py-flake8: upgrade to 3.6.0 and add check()

andypost:
Ref http://flake8.pycqa.org/en/latest/release-notes/3.6.0.html

The `check()` still downloads missing dependencies because they are in testing

PS: follow-up to https://github.com/alpinelinux/aports/pull/3468


andypost:
The test shows that new version require
- `pyflakes<2.1.0,>=2.0.0` (1.6.0 in main) 
- `pycodestyle<2.5.0,>=2.4.0` (2.3.1 in testing)

Also py2 needs `configparser` & `enum34`

PR #5558, opened 2018-11-04
main/nodejs: upgrade to 10.13.0

andypost:
Ref https://nodejs.org/en/blog/release/v10.13.0/

Depends on openssl 1.1 & https://github.com/alpinelinux/aports/pull/5561

PR #5561, opened 2018-11-05
main/libuv: upgrade to 1.23.2

andypost:
Upgrade of nodejs require 1.23.0 at least https://github.com/alpinelinux/aports/pull/5558 
because `uv_os_setpriority` introduced there http://docs.libuv.org/en/v1.x/misc.html#c.uv_os_setpriority

no abi changes https://abi-laboratory.pro/index.php?view=timeline&l=libuv

https://github.com/libuv/libuv/blob/master/ChangeLog


andypost:
Patch removed because https://github.com/libuv/libuv/pull/1836

PR #5569, opened 2018-11-07
PHP5 EOL

andypost:
Closes #9291

Removes
- php5
- php5-apcu
- php5-suhosin

Summary in https://bugs.alpinelinux.org/issues/9291


andypost:
Now all brokers commited and php5 could be removed

andypost:
@ncopa would be great to remove it before 3.9

PR #5574, opened 2018-11-08
community/phpldapadmin: modernize and bump dependencies to use php7

andypost:
Ref https://github.com/alpinelinux/aports/pull/5569

Patch from https://github.com/breisig/phpLDAPadmin/commit/db8a98c7a5ae6352018e2a2673e3d24df03f3e48

### DIFF
--- a/community/phpldapadmin/APKBUILD
+++ b/community/phpldapadmin/APKBUILD
@@ -1,17 +1,20 @@
 # Maintainer: Leonardo Arena <rnalrd@alpinelinux.org>
-_php=php5
+_php=php7
 pkgname=phpldapadmin
 pkgver=1.2.3
-pkgrel=4
+pkgrel=5
 pkgdesc="Web front-end for managing OpenLDAP"
 url="http://phpldapadmin.sourceforge.net"
 arch="noarch"
 license="GPL-2.0"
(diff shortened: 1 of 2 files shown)


andypost:
All dependencies updated - mbstring & openssl optional but their methods used all over code conditionally.
I used to test it with edge php7 
- copy `/etc/phpldapadmin/config.php` and add server's host
- the run build-in server `cd /usr/share/webapps/phpldapadmin/htdocs && php -S 127.0.0.1:8080`

andypost:
@rnalrd it needs you approval to move ahead

community/phpldapadmin/APKBUILD
```diff
@@ -1,17 +1,12 @@
 # Maintainer: Leonardo Arena <rnalrd@alpinelinux.org>
-_php=php5
```
rnalrd: I'd prefer to do not remove it. Other packages use the same approach. It's handy for the next php version.

andypost: I will revert template but package is broken cos at least php5-session is needed, olso I'd added opcache or apc in phpt

andypost: Also I'm not sure zlib is bundled in php5 build like in php7

andypost: Pushed updates

PR #5576, opened 2018-11-08
Remove LLVM 3

J0WI:
Seems to be unused, LLVM 4/5+ are widely supported.


ncopa:
seems like ponyc still depends on llvm3.9.

@jirutka can we rebuild ponyc to newer llvm so we can remove llvm3.*?

jirutka:
https://github.com/ponylang/ponyc/blob/master/README.md:
> Pony supports LLVM 3.9 and on an experimental basis it supports LLVM 4.0 and 5.0.

So no.

jirutka:
> But ok to remove llvm3.7?

Yes, there's nothing depending on 3.7 left.

martell:
@J0WI can you resolve the conflicts here?
https://github.com/alpinelinux/aports/commit/610072d44b6b988b341e2890b63de436228819a7
Is probably the cause

J0WI:
@martell done

martell:
@andypost @jirutka @ncopa can someone merge when possible. :)

PR #5599, opened 2018-11-09
[3.6] community/pdns-recursor: security upgrade to 4.0.9

J0WI:


ncopa:
we should probably move pdns-recursor to main if we want maintain this for more than latest stable

ncopa:
fixed with a82df6a94c042f4bf27b6d162cda7fb7d7f513b5

J0WI:
> we should probably move pdns-recursor to main if we want maintain this for more than latest stable


I'm following this policy:

> We support each stable release for a certain time, normally for 2 years. We can do security fixes beyond that on request and when patches are available.

https://wiki.alpinelinux.org/wiki/Releases

So 3.6 should receive security updates at least until 2019-05-01 (or 2019-05-24 to be exact).

PR #5622, opened 2018-11-12
community/wireshark: rebuild against openssl

J0WI:


ncopa:
seems like nothing links to LibreSSL/openssl. Maybe the build time dependency can be removed?

J0WI:
Dependency removed.

fabled:
Applied.

PR #5623, opened 2018-11-12
community/mktorrent: rebuild against openssl

J0WI:


andypost:
@fcolista also it makes sense to upgrade

community/mktorrent/APKBUILD
```diff
@@ -2,12 +2,12 @@
 # Maintainer:  Francesco Colista <fcolista@alpinelinux.org>
 pkgname=mktorrent
 pkgver=1.1
-pkgrel=1
+pkgrel=3
```
andypost: I bet it should be 2)

PR #5625, opened 2018-11-12
community/knot: rebuild against openssl

J0WI:


J0WI:
Fixed in https://github.com/alpinelinux/aports/commit/ea05d8c376a161a4813663e1952477468765759a

PR #5449, opened 2018-10-20
community/php7-pecl-gmagick: renamed from php7-gmagick, modernize

andypost:
Renamed according https://bugs.alpinelinux.org/issues/9277

Fix license https://pecl.php.net/package/gmagick

PR #5627, opened 2018-11-12
community/graphicsmagick: disable openmp for stability

andypost:
This is a blocker for #5449 

Performance gain is not huge enough and can't compensate stability http://www.graphicsmagick.org/OpenMP.html

Also the same was done for imagemagick in https://github.com/alpinelinux/aports/commit/a23a75a8962b419f92fe63ba60317496bbfbe712

Related issue for imagemagick is https://github.com/alpinelinux/aports/pull/5402


andypost:
Commited as 5f97babe

PR #5650, opened 2018-11-15
community/chntpw: rebuild against libgcrypt

J0WI:


ncopa:
I wonder why libressl/openssl is a makedepends in the first place. It is not linked against it:
````
chntpw-140201-r1 depends on:
so:libc.musl-x86_64.so.1
````

ncopa:
the build is broken. It does not build with crypto support at all. Seems like the correct define is `-DDOCRYPTO` instead of `-DUSEOPENSSL`. Enabling that make it try build with crypto, but it fails.

It seems that fedora uses some patch from debian to make it build against gcrypt instead. Maybe we should do that too.

ncopa:
It seems that also fedora's package is broken and built without libgcrypt.

J0WI:
I've added Debian's patch to build against libgcrypt

ncopa:
The patch is not used. Travis build log shows that libgcrypto is not pulled in as a runtime dependency (https://travis-ci.org/alpinelinux/aports/builds/455496671#L748). This is because the code is broken.

The `#ifdef USELIBGCRYPT` is within an `#ifdef DOCRYPTO`. If I add `-DDOCRYPTO` to CFLAGS it will enable the libgrypto code, which ends up in another compile error, due to a couple of places DOCRYPTO seem to be misspelled to DOCRYPT. Adding both, ie. `-DDOCRYPT -DDOCRYPTO -DUSELIBGCRYPTO` should make it build... but...

I found a comment in the code:
````c
/* Define DOCRYPTO in makefile to include cryptostuff to be able to change passwords to                          
 * a new one.                                                                                                    
 * Changing passwords is seems not to be working reliably on XP and newer anyway.                                
 * When not defined, only reset (nulling) of passwords available.                                                
 */                                                                                                              
````

So I don't think the code will work, even if we make it build.

I think we should disable the crypto, and re-enable (and potensially fix) it if someone asks for it. According the comment it should still be possible to clear the password with the tool when its built without crypto.

The other patches from Fedora seems to be needed though.

@fabaff you are the maintainer for this package. Is this ok with you?

fabaff:
It's fine with me.

I haven't used the tool for quite some time. So, I don't really know what's going on with it.

J0WI:
@ncopa Can you point me to the required patches and build options?

ncopa:
fixed with 2e4139498d33005f4d9c56db360024b11ae5a42e.

Thanks!

PR #5658, opened 2018-11-15
testing/wireguard: upgrade to 0.0.20181115

zx2c4:
```
  * Zinc no longer ships generated assembly code. Rather, we now
    bundle in the original perlasm generator for it. The primary purpose
    of this snapshot is to get testing of this.
  * Clarify the peer removal logic and make lifetimes more precise.
  * Use READ_ONCE for is_valid and is_dead.
  * No need to use atomic when the recounter is mutex protected.
  * Fix up macros and annotations in allowedips.
  * Increment drop counter when staged packets are dropped.
  * Use static constants instead of enums for 64-bit values in selftest.
  * Mark large constants as ULL in poly1305-donna64.
  * Fix sparse warnings in allowedips debugging code.
  * Do not use wg_peer_get_maybe_zero in timer callbacks, since we now can
    carefully control the lifetime of these functions and ensure they never
    execute after dropping the last reference.
  * Cleanup hashing in ratelimiter.
  * Do not guard timer removals, since del_timer is always okay.
  * We now check for PM_AUTOSLEEP, which makes the clear*on-suspend decision a
    bit more general.
  * Set csum_level to ~0, since the poly1305 authenticator certainly means
    that no data was modified in transit.
  * Use CHECKSUM_PARTIAL check for skb_checksum_help instead of
    skb_checksum_setup check.
  * wg.8: specify that wg(8) shows runtime info too
  * wg.8: AllowedIPs isn't actually required
  * keygen-html: add missing glue macro
  * wg-quick: android: do not choke on empty allowed-ips
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,7 +3,7 @@
 
 # NOTE: pkgrel must match _toolsrel in wireguard-vanilla
 pkgname=wireguard-tools
-pkgver=0.0.20181018
+pkgver=0.0.20181115
 pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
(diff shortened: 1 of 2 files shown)


andypost:
Thanks, commited 11eee56b

PR #5659, opened 2018-11-15
main/valgrind: upgrade to 3.14.0

HRio:
valgrind-3.13.0-ld-separate-code.patch now in upstream


HRio:
@ncopa or @fabled can you have a look? thanks.

PR #5660, opened 2018-11-16
chrony: add pidfile to config

ollieparanoid:
Fix rc-status from reporting chrony as crashed.

Thanks to Chris Clonch for reporting and proposing this fix:
https://bugs.alpinelinux.org/issues/9645

(Tested and working for me.)


tcely:
This is the wrong direction in my opinion.

`s%\@DEFAULT_PID_FILE\@%/var/run/chrony/chronyd.pid%g;`

The pid file used by default is at `/var/run/chrony/chronyd.pid` and `chronyd.initd` should be updated to become aware of that default value.

HRio:
info https://github.com/alpinelinux/aports/commit/a1a3fd0238ae1c2a46ab402e9967162f2d1b1460

tcely:
Please close this.

PR #5661, opened 2018-11-16
Adding /lib64 sysmlink for ppc64le arch in musl

Pensu:
Go binaries use ```/lib64/ls64.so.2```. This PR creates the symlink to /lib64, so that binaries can be run on ppc64le.

Fixes: https://bugs.alpinelinux.org/issues/9648

### DIFF
--- a/main/musl/APKBUILD
+++ b/main/musl/APKBUILD
@@ -132,7 +132,10 @@ compat() {
 		ln -s lib "$subpkgdir"/lib64
 		;;
 	mips* | s390*) _ld="ld.so.1" ;;
-	ppc64le) _ld="ld64.so.2" ;;
+	ppc64le) _ld="ld64.so.2"
+		# go binaries use /lib64/ld64.so.2
+		ln -s lib "$subpkgdir"/lib64
+		;;
 	esac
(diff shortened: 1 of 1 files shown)


Pensu:
@fabled Hi, Can you please take a look at this?

ncopa:
> Go binaries use /lib64/ls64.so.2.

This sounds like something that should be fixed in Go.

Pensu:
@nero I think libc6-compat is part of musl in Alpine. I tried to look the package origin: https://pkgs.alpinelinux.org/package/edge/main/x86/libc6-compat, it says musl there. I couldn't find libc6-compat. Can you please point me to the right package to change in this case?

nero:
You were right there. I confused it with `community/gcompat`. The musl `APKBUILD` is the right place to fix this, ncopa's comment is probably caused by the typo in the interpreter name.

Pensu:
@nero Can we get this merged?

nero:
I don't have push rights.

The commit messages should adhere to the contribution guidelines.

Pensu:
Closing it as this is fixed via: https://github.com/alpinelinux/aports/pull/5926

PR #5681, opened 2018-11-18
testing/android-tools: enable build on armhf and aarch64

drebrez:
This PR will enable compilation of `android-tools` package for `armhf` and `aarch64` architectures.


nmeum:
Yes, I will take a look at this soonish.

nmeum:
Does the build failure also occur with clang? If so: We should also disable that warning when compiling with clang as [nmeum/android-tools](https://github.com/nmeum/android-tools) should optimally work with both clang and gcc.

drebrez:
@nmeum I don't know much about clang, and I don't really know how to check if it fails.

nmeum:
I would usually test this myself but since I currently don't have access to an aarch64 or armhf machine I can't. Could you install clang, invoke cmake as `CC=clang CXX=clang++ cmake ..` and then just built the aport as usual with make?

nmeum:
Superseded by 67283cdbf6ef0ce2912bd28297714d6bb3976a38

PR #5685, opened 2018-11-18
main/py-pillow: rebuild against libjpeg-turbo and add check

J0WI:


main/py-pillow/APKBUILD
```diff
@@ -23,6 +23,12 @@ build() {
 	python3 setup.py build
 }
 
+build() {
+	cd "$builddir"
+	python2 setup.py test
+	python3 setup.py test
```
andypost: why tests running in `build()`?

J0WI: Thanks for the hint, I was already wondering about the strange output.

J0WI:
:ping_pong: @fabaff

J0WI:
Fixed in 659180a3e5f5997a5b55774c1079c52ed097577e

PR #5689, opened 2018-11-19
testing/wireguard: upgrade to 0.0.20181119

zx2c4:
```
  * chacha20,poly1305: fix up for win64
  * poly1305: only export neon symbols when in use
  * poly1305: cleanup leftover debugging changes
  * crypto: resolve target prefix on buggy kernels
  * chacha20,poly1305: don't do compiler testing in generator and remove xor helper
  * crypto: better path resolution and more specific generated .S
  * poly1305: make frame pointers for auxiliary calls
  * chacha20,poly1305: do not use xlate
  
  This should fix up the various build errors, warnings, and insertion errors
  introduced by the previous snapshot, where we added some significant
  refactoring. In short, we're trying to port to using Andy Polyakov's original
  perlasm files, and this means quite a lot of work to re-do that had stableized
  in our old .S.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,7 +3,7 @@
 
 # NOTE: pkgrel must match _toolsrel in wireguard-vanilla
 pkgname=wireguard-tools
-pkgver=0.0.20181115
+pkgver=0.0.20181119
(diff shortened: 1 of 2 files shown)


andypost:
Commited as 2d6d4f4f

PR #5693, opened 2018-11-20
main/chrony: set default PID file path

tcely:
https://bugs.alpinelinux.org/issues/9645
https://bugs.alpinelinux.org/issues/9674


Anthony25:
Hi,
I tested this PR, it works perfectly and fixes the issues I got with the PID.

Thanks!

tcely:
@Anthony25 Thanks for testing. It's good to have confirmed that it works as expected.

I'm just waiting fir a developer to merge the changes at this point.

HRio:
info https://github.com/alpinelinux/aports/commit/a1a3fd0238ae1c2a46ab402e9967162f2d1b1460

tcely:
@HRio I saw the commit by @nmeum and have resolved the resulting conflicts.

Just changing the default value (twice, violating DRY) is insufficient in my opinion. The rest of the changes in this PR should be used instead.

nmeum:
Sorry. Didn't See your PR. Why do you believe that my changes are insufficient? I would like to keep the diff for fixing this as small as possible.

tcely:
@nmeum You did not remove the duplicate tests and default values. Also, it is better to have the `-openrc` subpackage.

We should optimize the resulting file not the diff size, in my opinion. I prefer having a single location for the assignment of the default values and only a single check on the value derived from the configuration file. These changes make future updates easier and safer, since there is no possibility of a typo resulting in disagreement about where the PID file is located.

nmeum:
> @nmeum You did not remove the duplicate tests and default values.

Those are just stylistic changes they should have the same functional outcome.

> Also, it is better to have the -openrc subpackage.

I agree. In fact I would preferably have such a subpackage for all aports shipping runscripts (see #3803). To me it seems ineffective to incrementally add this to every single aport. Besides this is not related to the pidfile issue.

> We should optimize the resulting file not the diff size, in my opinion.

Sure, but we are [close to a new release](http://dev.alpinelinux.org/buildlogs/build-3-9-x86_64.log) so fixing this in the most simple way guarantees that the fix doesn't introduce new bugs.

> These changes make future updates easier and safer, since there is no possibility of a typo resulting in disagreement about where the PID file is located.

TBH: I would prefer hardcoding a path in the APKBUILD instead of relying on the default value. Maybe even substituting a keyword in the service file using sed from the APKBUILD to make sure that the paths always match. Your changes do not address this since the service would still break if the default pid file location changes again.

**tl;dr** I am all for cleaning up the chronyd OpenRC service (we could probably just rewrite it from scratch without manual `start()/stop()` functions) but to me this is independent of the pidfile issue. If there are any functional issues with the fix I pushed let me know. If you want to cleanup the chronyd OpenRC services that's fine by me but I would personal prefer doing so after 3.9 has been released.

tcely:
> Those are just stylistic changes they should have the same functional outcome.
Should being the operative word in that statement. Leaving the duplication in place just invites error for no gain.

> > Also, it is better to have the -openrc subpackage.
> I agree. In fact I would preferably have such a subpackage for all aports shipping runscripts (see #3803). To me it seems ineffective to incrementally add this to every single aport. Besides this is not related to the pidfile issue.
The approach I see as ineffective is to not add it while touching the aport anyway. Waiting for one huge update that may be massively outdated seems like a poor idea to me.

> > We should optimize the resulting file not the diff size, in my opinion.
> Sure, but we are [close to a new release](http://dev.alpinelinux.org/buildlogs/build-3-9-x86_64.log) so fixing this in the most simple way guarantees that the fix doesn't introduce new bugs.
I don't have any input on the timing of this fix or it would not have remained unfixed for so long. I also don't have any input on when the next release drops, so that seems out of scope from my point of view.

I don't agree that this was the most simple fix or even the one least likely to introduce bugs. Your rationale seems to be that the smallest change is best. I've always found that argument is flawed and encourages churn rather than better software. Let's get to the better state with as few changes and testing cycles as we can so that we ship better software this release rather than at some possible future release.

> > These changes make future updates easier and safer, since there is no possibility of a typo resulting in disagreement about where the PID file is located.
> 
> TBH: I would prefer hardcoding a path in the APKBUILD instead of relying on the default value. Maybe even substituting a keyword in the service file using sed from the APKBUILD to make sure that the paths always match. Your changes do not address this since the service would still break if the default pid file location changes again.

True and including the script file in those using the generated / configured value is a good idea for improvement.

> **tl;dr** I am all for cleaning up the chronyd OpenRC service (we could probably just rewrite it from scratch without manual `start()/stop()` functions) and I think your changes are step in the right direction but to me this is independent of the pidfile issue. If there are any functional issues with the fix I pushed let me know. If you want to cleanup the chronyd OpenRC services that's fine by me but I would personal prefer doing so after 3.9 has been released.

I think rewriting it is ultimately the correct approach. But since this was about making the pidfile handling better I limited my changes to that and pulling out a duplicated hardcoded value.

tcely:
@nmeum Please merge these changes.

tcely:
@nmeum Please merge these changes now that we aren't waiting on a release.

nmeum:
Whoops, sorry somehow forgot about this. Will merge is ASAP, thanks!

PR #5700, opened 2018-11-20
main/openssl: security upgrade to 1.0.2q 

J0WI:
https://www.openssl.org/news/cl102.txt


ncopa:
we normally don't fix security issues in unsupported branches.

PR #5706, opened 2018-11-21
testing/traefik: new aport

joeholden:
As per the title, new port for Traefik - build and runtime tested.

Not sure what else you guys need other than a PR but reviews would be appreciated.

### DIFF
--- a/testing/traefik/APKBUILD
+++ b/testing/traefik/APKBUILD
@@ -1,0 +1,62 @@
+# Contributor: Joe Holden <jwh@zorins.us>
+# Maintainer: Joe Holden <jwh@zorins.us>
+pkgname=traefik
+pkgver=1.7.5
+pkgrel=0
+pkgdesc="The Cloud Native Edge Router"
+url="https://traefik.io"
+arch="all"
+license="MIT"
+makedepends="go go-bindata"
+install="$pkgname.pre-install"
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+subpackages="$pkgname-openrc"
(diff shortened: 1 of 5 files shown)


testing/traefik/APKBUILD
```diff
@@ -0,0 +1,60 @@
+pkgname=traefik
+pkgver=1.7.4
+pkgrel=7
```
andypost: why 7 instead of 0?

joeholden:
Yup will do

testing/traefik/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+pkgdesc="The Cloud Native Edge Router"
+url="https://traefik.io"
+arch="all"
+license="MIT"
+makedepends="go go-bindata"
+install="$pkgname.pre-install"
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+subpackages="$pkgname-openrc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/containous/$pkgname/archive/v$pkgver.tar.gz
```
andypost: please change it to `source="$pkgname-$pkgver.tar.gz::https://github.com/containous/traefik/archive/v$pkgver.tar.gz` to make navigation easy

PS: https://github.com/containous/traefik/releases/tag/v1.7.5 is out

joeholden: Ok I'll update and fix that, thanks

joeholden: done, updated to 1.7.5 also

PR #5639, opened 2018-11-13
testing/clang6: reenable tests

J0WI:
Depends on #5638

PR #5638, opened 2018-11-13
testing/llvm6: new aport

J0WI:

PR #5717, opened 2018-11-24
main/llvm: upgrade to 7.0.1

martell:
Taking simple steps and adding a main/llvm package.
I presume the latest should always be this version.
older versions can be main/llvm6 or main/llvm5 etc

### DIFF
--- a/main/llvm/APKBUILD
+++ b/main/llvm/APKBUILD
@@ -4,13 +4,13 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 _pkgname=llvm
-pkgver=7.0.0
+pkgver=7.0.1
 _majorver=${pkgver%%.*}
 pkgname=$_pkgname$_majorver
 pkgrel=0
 pkgdesc="Low Level Virtual Machine compiler system, version $_majorver"
 arch="all"
-url="http://llvm.org/"
+url="https://llvm.org/"
 license="UOI-NCSA"
 depends_dev="$pkgname=$pkgver-r$pkgrel"
 makedepends="binutils-dev chrpath cmake file libffi-dev paxmark python2 py-setuptools zlib-dev"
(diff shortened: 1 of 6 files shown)


misery:
By the way.... is it possible to enable sanitizier support? https://bugs.alpinelinux.org/issues/9643

J0WI:
See also #5638 (some pkgs requires LLVM 6)

martell:
The job exceeded the maximum log length, and has been terminated

> By the way.... is it possible to enable sanitizer support?

Sure but I'd rather do it in a followup PR


Thanks @J0WI for doing other packages

main/llvm/APKBUILD
```diff
@@ -0,0 +1,252 @@
...
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
+_pkgname=llvm
+pkgver=7.0.1
+_majorver=${pkgver%%.*}
+pkgname=$_pkgname$_majorver
+pkgrel=0
+pkgdesc="Low Level Virtual Machine compiler system, version $_majorver"
+arch="all"
+url="http://llvm.org/"
```
J0WI: HTTPS works here too ;)

martell: done, thanks

martell:
rebased. @ncopa :)

J0WI:
@martell if you rebase again you will get a try build from Drone CI

ncopa:
i think the directory needs to be renamed to `llvm7` because that is what the `pkgname` is.

main/llvm/APKBUILD
```diff
@@ -0,0 +1,252 @@
+# Contributor: Martell Malone <martell@marinelayer.io>
+# Contributor: Travis Tilley <ttilley@gmail.com>
+# Contributor: Mitch Tishmack <mitch.tishmack@gmail.com>
+# Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Maintainer: Jakub Jirutka <jakub@jirutka.cz>
```
ncopa: @martell have you asked if @jirutka is ok with being maintainer for llvm7?

jirutka: He did not. I don’t wanna maintain LLVM packages.

jirutka:
> Taking simple steps and adding a main/llvm package.

Please read main/llvm5 abuild properly before doing anything. There’s a good reason why it is as is, we already support multiple LLVM versions in parallel, it’s not needed nor desired to reinvent a wheel again.

ncopa:
> > Taking simple steps and adding a main/llvm package.
> 
> Please read main/llvm5 abuild properly before doing anything. There’s a good reason why it is as is, we already support multiple LLVM versions in parallel, it’s not needed nor desired to reinvent a wheel again.

Are there anything specific in the main/llvm5 APKBUILD you think needs proper reading? The llvm 7 APKBUILD is a copy of llvm5's. I am not sure what I missed this time.

J0WI:
@ncopa `main/clang` is still on version 5 and the build is now failing (same on #5639).

martell:
I feel like a couple of comments here need addressing

@jirutka you are still marked as the maintainer of llvm 3.9 and 4.0 https://github.com/alpinelinux/aports/blob/master/community/llvm3.9/APKBUILD#L4
https://github.com/alpinelinux/aports/blob/master/unmaintained/llvm4/APKBUILD#L4

@ncopa this this PR was made about 2 months before @jirutka gave up maintainership of the llvm5 package
https://github.com/alpinelinux/aports/commit/8bc22b8a46a30aa4e64f128b008dd6e9c01dfea4#diff-b014f8d33984357df79bf97e7ee4bad9

Back in Nov 2018 I made what I believed was a reasonable assumption in that the maintainer of llvm 3.9, 4 and 5 would still be the maintainer of the next aport version.
Assuming I didn't put any thought into that is a little unfair to say the least.
I would have gladly offered to maintain it myself but I didn't want to suggest taking anyones thunder.

I am at fault however for not noting in a comment that this PR was not ready to be merged as it was mostly a rebase of llvm5 with patches removed that were upstreamed and checksum updates etc, we were waiting on longer build bot times which @J0WI alerted me to recently.

In general the feeling (and I'm probably not the only one) I get is that some PRs come to alpine to die a slow death like this one I created almost a year ago for wine
https://github.com/alpinelinux/aports/pull/3576

The lack of action with respect to llvm aport PRs got to the point where I had a colleague add musl host support to rust so that we can avoid dealing with llvm updates in alpines aports all-together and just use rustup directly, which is mostly why there has been very little updates from me here.

I'd like to note that all this is not a direct criticism of people but instead should be thought of as a frustration at a lack of resources. I do appreciate that there are higher priorities and so little time in package/distro management and people always do as much as they can where possible.
My understanding of this is rooted in a past experience of helping to maintain msys2 a few years ago along with a few dedicated developers without any sponsors.

I am however glad to see @ncopa picking up llvm stuff as of today which gives me hope and I'll try my best to do a more complete PR of llvm 8 when it is tagged.

main/llvm/APKBUILD
martell: see https://github.com/alpinelinux/aports/pull/5717#issuecomment-474686094

PR #5721, opened 2018-11-25
testing/perl-db_file-lock: Locking with flock wrapper for DB_File

timlegge:
Needed it for a foswiki docker I was creating the APKBUILD generated with the updated apkbuild-cpan that I submitted that has support for  metacpan


timlegge:
cancelling to resubmit squashed with correct syntax

PR #5723, opened 2018-11-25
testing/wireguard-tools: add missing depends

itoffshore:
`wg-quick` does not work with some of the `busybox` built-ins.

reported by Nathan Caldwell (`saintdev at gmail dot com`)

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -10,9 +10,12 @@ arch='all'
 url='https://www.wireguard.com'
 license="GPL-2.0"
 makedepends="libmnl-dev"
+depends="iproute2 procps coreutils"
 subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch"
 options="!check"
-source="https://git.zx2c4.com/WireGuard/snapshot/WireGuard-$pkgver.tar.xz"
+source="https://git.zx2c4.com/WireGuard/snapshot/WireGuard-$pkgver.tar.xz
+	alpine-compat.patch
+	"
 builddir="$srcdir"/WireGuard-$pkgver
 
 build() {
@@ -43,4 +46,5 @@ bashcomp() {
 	mv "$pkgdir"/usr/share "$subpkgdir"/usr
 }
 
-sha512sums="73c8e9b37d857349b75df776607c15ea2082814952acdba3ad6379c4ce631601db2767603e46ecadf1bce9348a0c26d07f4f6b5857ddd72bb4f4411d1d13d88c  WireGuard-0.0.20181218.tar.xz"
+sha512sums="73c8e9b37d857349b75df776607c15ea2082814952acdba3ad6379c4ce631601db2767603e46ecadf1bce9348a0c26d07f4f6b5857ddd72bb4f4411d1d13d88c  WireGuard-0.0.20181218.tar.xz
+be16a4b547b9ed60d7cd2b7dddb1ec8001f38faf8340bd76013e631c4c376a2fcd45b757296ac7422b53d7f14c759eecc41376bc6e01159c3bc5bdb121413598  alpine-compat.patch"
--- a/testing/wireguard-tools/alpine-compat.patch
+++ b/testing/wireguard-tools/alpine-compat.patch
@@ -1,0 +1,25 @@
+diff --git a/src/tools/wg-quick/linux.bash b/src/tools/wg-quick/linux.bash
+index 48ce163..a2a0738 100755
+--- a/src/tools/wg-quick/linux.bash
++++ b/src/tools/wg-quick/linux.bash
+@@ -1,9 +1,18 @@
+-#!/bin/bash
++#!/bin/sh
+ # SPDX-License-Identifier: GPL-2.0
+ #
+ # Copyright (C) 2015-2018 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
+ #
+ 
++if [ -z "$BASH_VERSION" ]; then
++	if ! type bash >/dev/null 2>&1; then
++		echo "Please run 'apk add bash' and try again." >&2
++		exit 1
++	fi
++	exec bash "$0" "$@"
++fi
++set +o posix
++
+ set -e -o pipefail
+ shopt -s extglob
+ export LC_ALL=C
+


testing/wireguard-tools/APKBUILD
```diff
@@ -10,6 +10,7 @@ arch='all'
 url='https://www.wireguard.com'
 license="GPL-2.0"
 makedepends="libmnl-dev"
+depends="bash iproute2 procps"
```
andypost: does it really require `bash`?

itoffshore: `wg-quick` specifies:

```
#!/bin/bash
# SPDX-License-Identifier: GPL-2.0
```

& makes use of `bash` arrays in various places:
```
LXC-alpine [~]$ grep "]}" $(which wg-quick)
SELF="$(readlink -f "${BASH_SOURCE[0]}")"
	INTERFACE="${BASH_REMATCH[2]}"
	[[ $UID == 0 ]] || exec sudo -p "$PROGRAM must be run as root. Please enter the password for %u to continue: " -- "$BASH" -- "$SELF" "${ARGS[@]}"
		output="$(ip route get "${BASH_REMATCH[1]}" || true)"
		[[ ( $output =~ mtu\ ([0-9]+) || ( $output =~ dev\ ([^ ]+) && $(ip link show dev "${BASH_REMATCH[1]}") =~ mtu\ ([0-9]+) ) ) && ${BASH_REMATCH[1]} -gt $mtu ]] && mtu="${BASH_REMATCH[1]}"
		[[ ( $output =~ mtu\ ([0-9]+) || ( $output =~ dev\ ([^ ]+) && $(ip link show dev "${BASH_REMATCH[1]}") =~ mtu\ ([0-9]+) ) ) && ${BASH_REMATCH[1]} -gt $mtu ]] && mtu="${BASH_REMATCH[1]}"
		echo "${BASH_REMATCH[1]}." && return 0
	[[ ${#DNS[@]} -gt 0 ]] || return 0
	printf 'nameserver %s\n' "${DNS[@]}" | cmd resolvconf -a "$(resolvconf_iface_prefix)$INTERFACE" -m 0 -x
	[[ ${#DNS[@]} -gt 0 ]] || return 0
	for address in ${BASH_REMATCH[1]}; do
		[[ $address =~ ^nameserver\ ([a-zA-Z0-9_=+:%.-]+)$ ]] && new_config+="DNS = ${BASH_REMATCH[1]}"$'\n'
	[[ -n $MTU && $(ip link show dev "$INTERFACE") =~ mtu\ ([0-9]+) ]] && new_config+="MTU = ${BASH_REMATCH[1]}"$'\n'
	for cmd in "${PRE_UP[@]}"; do
	for cmd in "${POST_UP[@]}"; do
	for cmd in "${PRE_DOWN[@]}"; do
	for cmd in "${POST_DOWN[@]}"; do
	execute_hooks "${PRE_UP[@]}"
	for i in "${ADDRESSES[@]}"; do
	execute_hooks "${POST_UP[@]}"
	execute_hooks "${PRE_DOWN[@]}"
	execute_hooks "${POST_DOWN[@]}"
```

ncopa:
This was discussed in https://github.com/alpinelinux/aports/pull/3903

I think the better solution would be to to move `wg-quick` to its own package.

@zx2c4 what do you think?

zx2c4:
I certainly don't like the idea of moving wg-quick to its own package. The shebang in wg-quick clearly needs bash; don't try to hack it to use a busybox shebang. Rather, just leave it as is, and if the user doesn't have bash installed, the error running wg-quick will indicate that clearly they need bash.

itoffshore:
I've removed `bash` as a depends & revised the commit message explaining why

ncopa:
> I certainly don't like the idea of moving wg-quick to its own package.

What are the drawbacks? Or what are the benefits in shipping it non-functional with main package?

> Rather, just leave it as is, and if the user doesn't have bash installed, the error running wg-quick will indicate that clearly they need bash.

The error message does not indicate that bash i missing:

````
/ # apk add wireguard-tools
fetch http://dl-cdn.alpinelinux.org/alpine/edge/testing/x86_64/APKINDEX.tar.gz
(1/2) Installing libmnl (1.0.4-r0)
(2/2) Installing wireguard-tools (0.0.20181119-r0)
Executing busybox-1.29.3-r3.trigger
OK: 5 MiB in 15 packages
/ # wg-quick
/bin/sh: wg-quick: not found
/ # ls -l /usr/bin/wg-quick
-rwxr-xr-x    1 root     root         10459 Nov 20 15:34 /usr/bin/wg-quick
````

So I disagree, and would prefer to move wg-quick to a separate subpackage. It does not increase any maintenance work, and for the end user the difference is `apk add wg-quick` instead of `apk add bash` when hitting the _wg-quick: not found_ error.

Something like this:
````diff
diff --git a/testing/wireguard-tools/APKBUILD b/testing/wireguard-tools/APKBUILD
index dfb5c83750..7239d9dcda 100644
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -9,8 +9,9 @@ pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
 url='https://www.wireguard.com'
 license="GPL-2.0"
-makedepends="libmnl-dev"
-subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch"
+makedepends="libmnl-dev bash"
+subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch
+       wg-quick:wg_quick:noarch"
 options="!check"
 source="https://git.zx2c4.com/WireGuard/snapshot/WireGuard-$pkgver.tar.xz"
 builddir="$srcdir"/WireGuard-$pkgver
@@ -43,4 +44,10 @@ bashcomp() {
        mv "$pkgdir"/usr/share "$subpkgdir"/usr
 }
 
+wg_quick() {
+       depends="bash"
+       mkdir -p "$subpkgdir"/usr/bin
+       mv "$pkgdir"/usr/bin/wg-quick "$subpkgdir"/usr/bin/
+}
+
 sha512sums="dc9286e536bb3f4a6545261b98b5bb056a085f2e95b1b571350168af23cda5231a90b915b03dd29d195412fc61efc781b7d57b2a15a42a34b97e989a5105198e  WireGuard-0.0.20181119.tar.xz"
````

**Comment by zx2c4** [comment]
```diff
diff --git a/src/tools/wg-quick/linux.bash b/src/tools/wg-quick/linux.bash
index 48ce163..a2a0738 100755
--- a/src/tools/wg-quick/linux.bash
+++ b/src/tools/wg-quick/linux.bash
@@ -1,9 +1,18 @@
-#!/bin/bash
+#!/bin/sh
 # SPDX-License-Identifier: GPL-2.0
 #
 # Copyright (C) 2015-2018 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
 #
 
+if [ -z "$BASH_VERSION" ]; then
+	if ! type bash >/dev/null 2>&1; then
+		echo "Please run 'apk add bash' and try again." >&2
+		exit 1
+	fi
+	exec bash "$0" "$@"
+fi
+set +o posix
+
 set -e -o pipefail
 shopt -s extglob
 export LC_ALL=C
```

This will give better error messages.

andypost:
@itoffshore please add patch from previous comment to commit - having this message is needed

druilaap:
wg-quick also depends on a couple more packages. Namely:

- procps for systctl
- openresolv for dns
- iproute2 for /sbin/ip and ipv6 routing

If you do not want to add these dependencies to wireguard-tools, should wg-quick not be patched to to print error messages in wg-quick for those missing dependencies? Currently this leads to hard to parse error messages. e.g. missing command line options for ip, etc. [1]

Cheers,
Druilaap

[1]: https://lists.zx2c4.com/pipermail/wireguard/2018-December/003608.html

itoffshore:
@andypost - Alpine compat patch added

@zx2c4 - added `coreutils` depends as Alpine will use them over the `busybox` builtin tools (for `sysctl -r`)

zx2c4:
That sounds good to me.

ncopa:
you forgot to bump `pkgrel`, but i fixed that for you. Thanks!

PR #5731, opened 2018-11-26
community/z3: fix ppc64le build break to skip use of immintrin.h

mksully22:
Upgrade to z3 version 4.8.3 includes new use of immintrin.h in mpz.cpp resulting in error on ppc64.  This file only exists in the x86_64 gcc-8.2.0-r1 package. This patch adds condition to not include it on ppc64le. 

Build error:
/home/buildozer/aports/community/z3/src/z3-z3-4.8.3/src/util/mpz.cpp:33:10: fatal error: immintrin.h: No such file or directory
 #include <immintrin.h>
          ^~~~~~~~~~~~~
compilation terminated.

### DIFF
--- a/community/z3/APKBUILD
+++ b/community/z3/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Sören Tempel <soeren+alpine@soeren-tempel.net>
 pkgname=z3
 pkgver=4.8.3
-pkgrel=0
+pkgrel=1
 pkgdesc="Theorem prover from Microsoft Research"
 url="https://github.com/Z3Prover/z3"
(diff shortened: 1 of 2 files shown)


mksully22:
Travis CI fails  because the log generated during the z3 build is too long which appears normal for the z3 build. "The job exceeded the maximum log length, and has been terminated."

andypost:
does it makes sense to promote upstream?

ncopa:
> Travis CI fails because the log generated during the z3 build is too long which appears normal for the z3 build. "The job exceeded the maximum log length, and has been terminated."

I'll test build it locally. thanks!

PR #5735, opened 2018-11-28
main/nodejs: upgrade to 10.14.0

timbru31:
This updates Node.js to v10.14.0 which includes the following secfixes:
https://nodejs.org/en/blog/vulnerability/november-2018-security-releases/


andypost:
@jirutka probably it needs backport

andypost:
Current is upgraded in https://github.com/alpinelinux/aports/pull/5736

PR #5736, opened 2018-11-28
community/nodejs-current: upgrade to 11.3.0

timbru31:
This updates Node.js to v11.3.0 which includes the following secfixes:
https://nodejs.org/en/blog/vulnerability/november-2018-security-releases/


andypost:
Checked secfixes - they are valid for 11.x

andypost:
No idea how to backport it to 3.8 which is using 9.x as current

PR #5737, opened 2018-11-28
testing/gcc6: fix ppc64le build break due to vector parsing

mksully22:
Adding fix https://gcc.gnu.org/viewcvs/gcc?view=revision&revision=261621

to overcome build error on ppc64le: 
build/genmatch --gimple /home/buildozer/aports/testing/gcc6/src/gcc-6.4.0/gcc/match.pd \
    > tmp-gimple-match.c
/home/buildozer/aports/testing/gcc6/src/gcc-6.4.0/gcc/match.pd:120:1 error: expected (, got NAME
   negative value by 0 gives -0, not +0.  */

### DIFF
--- a/testing/gcc6/APKBUILD
+++ b/testing/gcc6/APKBUILD
@@ -3,7 +3,7 @@ pkgname=gcc6
 pkgver=6.4.0
 _ver=${pkgver%%.*}
 pkgname="$pkgname"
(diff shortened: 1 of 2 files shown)


mksully22:
Travis CI failed because of source retrieval (although it probably would have failed due to duration or log length checks)

PR #5740, opened 2018-11-29
main/lighttpd: upgrade to 1.4.52

andypost:
Ref http://www.lighttpd.net/2018/11/28/1.4.52/

Closes https://bugs.alpinelinux.org/issues/9725


tcely:
It is probably not important, but this is a security upgrade since 1.4.51 was, correct?

andypost:
As I see by ref link only bugfixes and no CVEs found

tcely:
I agree no CVE was listed in the announcement, but it does say "security fixes" right near the top.

https://www.lighttpd.net/2018/10/14/1.4.51/

andypost:
Thanks, missed this 2
- [core,security] process headers after combining folded headers
- [mod_userdir] security: skip username “.” and “..”

PR #5742, opened 2018-11-30
testing/openocd: enable bcm2835gpio for ARM (RPi) architectures

bfritz:
Follow-up to 958a8ae7e7 because the 0.10.0 [release notes] also request `--enable-bcm2835gpio` for builds that support Raspberry Pi:

> 7.  Please add --enable-sysfsgpio for all targets running Linux, the
> kernel (esp. useful for SBCs), additionally for builds that support
> RaspberryPi, please add --enable-bcm2835gpio

`--enable-bcm2835gpio` is ignored when building on non-ARM architectures.  This configuration is also consistent with the [Debian packaging].

[release notes]: https://sourceforge.net/p/openocd/mailman/message/35570895/
[Debian packaging]: https://salsa.debian.org/electronics-team/openocd/blob/debian/0.10.0-4/debian/rules#L33

### DIFF
--- a/testing/openocd/APKBUILD
+++ b/testing/openocd/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Sören Tempel <soeren+alpine@soeren-tempel.net>
 pkgname=openocd
 pkgver=0.10.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Free and Open On-Chip Debugging, In-System Programming and Boundary-Scan Testing"
 url="http://openocd.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


bfritz:
@nmeum Seem reasonable?  Let me know if there's anything I should change or if I should submit via patchwork instead.  Thank you!

nmeum:
What does that option do? Can we just enable it unconditionally? Does it have any negative side effects on non-rasperrypi hardware? And if so: What about non-raspberry pi armhf hardware?

PR #5745, opened 2018-12-02
testing/razercfg: new aport

tanertas:
https://bues.ch/cms/hacking/razercfg.html
Razer mouse configuration tool.

### DIFF
--- a/testing/razercfg/APKBUILD
+++ b/testing/razercfg/APKBUILD
@@ -1,0 +1,42 @@
+# Contributor: Taner Tas <taner76@gmail.com>
+# Maintainer: Taner Tas <taner76@gmail.com>
+pkgname=razercfg
+pkgver=0.40
+pkgrel=0
+pkgdesc="Razer mouse configuration tool."
+url="https://bues.ch/cms/hacking/razercfg.html"
+arch="all"
+license="GPL-2.0-or-later"
+makedepends="
+	cmake
+	libusb-dev
+	python3
+	"
+subpackages="$pkgname-openrc $pkgname-gui::noarch"
+source="
+	https://github.com/mbuesch/razer/archive/$pkgname-$pkgver.tar.gz
+	razerd.initd
+	"
+options="!check" # No test suite
+builddir="$srcdir/razer-$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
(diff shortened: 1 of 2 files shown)


tanertas:
@andypost can you take a look at this?

testing/razercfg/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+source="
+	https://github.com/mbuesch/razer/archive/$pkgname-$pkgver.tar.gz
+	razerd.initd
+	"
+options="!check" # No test suite
+builddir="$srcdir/razer-$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	cmake . -DCMAKE_INSTALL_PREFIX='/usr'
```
clandmeter: You are not running make here?
I think make install is not triggering it which is not the correct location.

clandmeter:
Please also rebase against master so CI picks it up.

testing/razercfg/APKBUILD
tanertas: I'm not sure what is your confusion bu I have to say that his is the cmake equivalent of --prefix=/usr.

tanertas:
> Please also rebase against master so CI picks it up.

Why rebase? There's no conflict with master anyway since this is a new aport PR. CI already triggered when I update PR anytime.
Btw, I updated initscript.

tanertas:
> > Please also rebase against master so CI picks it up.
> 
> Why rebase? There's no conflict with master anyway since this is a new aport PR. CI already triggered when I update PR anytime.
> Btw, I updated initscript.

@clandmeter I realized that you were referring to the "drone" CI specifically. Sorry for misunderstood. I rebased to master.

testing/razercfg/APKBUILD
maxwell-k: Maybe I can explain a little more:

The README says that both cmake and make are used for the build step. cmake creates the Makefile. make uses the Makefile to build the binaries. https://github.com/mbuesch/razer#building

make install is correctly called in the package step. If the binary files are not available, so if make has not been called in the build step, make will notice and compile them. Compilation should not happen in the package step, so we're suggesting you add a make call to build().

Ikke:
Is this significant?

> WARNING: ldconfig failed: 1
Your system will probably be unable to locate librazer.so library

PR #5751, opened 2018-12-03
main/perl: upgrade to 5.28.1

andypost:
https://metacpan.org/pod/release/XSAWYERX/perl-5.28.0/pod/perldelta.pod


ncopa:
this is an upgrade from 5.26 and will require rebuild of alot of packages. I don't think we should wait til after v3.9.

ncopa:
we should upgrade to 5.26.3 for now.

andypost:
Makes a lot of sense - better to delay upgrade til next minor because there's a lot of changes (unicode is main one)

andypost:
@ncopa any idea how to identify packages which needs rebuild?

andypost:
@ncopa guess it should wait for 3.10 now?

andypost:
Fixed with 86b5b86c35 6d70352d3a

PR #5760, opened 2018-12-04
main/xfsprogs: fix ppc64le bld break by checking for MAP_SYNC define

mksully22:
/usr/include/bits/mman.h provided by musl-dev-1.1.20-r2 #undef MAP_SYNC causing a build error for undefined variable in mmap.c. This patch checks that the define is present before using it.


ncopa:
i think Adelie's patch is better: https://code.foxkit.us/adelie/packages/blob/master/system/xfsprogs/fix-mmap.patch

ncopa:
fixed with commit 1da104d7a14ed33450a512bdc2b4d7a43484a48a

PR #5803, opened 2018-12-10
main/alpine-baselayout: preserve value of PATH

am11:
In order to exercise interoperability with wsl, subsystem process
injects PATH variable to the Linux OS, this delta preserves its value.

ref: https://github.com/Microsoft/WSL/issues/3710

cc @agowa338

### DIFF
--- a/main/alpine-baselayout/profile
+++ b/main/alpine-baselayout/profile
@@ -1,5 +1,5 @@
 export CHARSET=UTF-8
-export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
 export PAGER=less
 export PS1='\h:\w\$ '
 umask 022


agowa338:
This may be a better fix than commenting it out, like I currently do to archive interoperability.
https://github.com/agowa338/WSL-DistroLauncher-Alpine/blob/master/DistroLauncher/DistroSpecial.h#L20

agowa338:
> Could this be made conditional based on feature/version detection?

You mean, like it currently is?

am11:
cc @ncopa, @andypost, PTAL

agowa338:
Do you have a /etc/resolv.conf file within the wsl?
And if so, what's it's content.
Please also try `ping google.de` and `ping 8.8.8.8`.

am11:
@agowa338, I have `/etc/resolv.conf` in WSL and content is:

```conf
# This file was automatically generated by WSL. To stop automatic generation of this file, remove this line.
nameserver 192.168.1.1
nameserver fe80::222:7ff:fe60:9003
nameserver fe80::222:7ff:fe60:9003
search wlan
```

Both ping commands are getting 64 bytes response.

agowa338:
That's odd, as from your above output `apk update` did not reach the internet...

agowa338:
Ok, I found the bug, the /etc/resolv.conf file does not reliably get created. Therefore accessing internet fails in some cases. (I just had to reconnect to my wifi in order for it to be created...)
I'll update my windows device and try if it still occurs and open a bug report...

main/alpine-baselayout/profile
```diff
@@ -1,5 +1,24 @@
+# /etc/profile
+
+appendpath () {
```
agowa338: I think this function is not needed.
Doing
`export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH` would be enough.

am11: See the implementaiton on other distros, ArchLinux: https://git.archlinux.org/svntogit/packages.git/tree/trunk/profile?h=packages/filesystem#n6 and CentOS: https://releases.pagure.org/setup/setup-2.12.6.tar.bz2. The reason why they use this approach is to deduplicate the entries in PATH in a safe manner.

agowa338:
Just tested with 1903 (Fast Ring), the file /etc/resolv.conf is correctly created. So it is fixed.

nero:
> This patch is aligning the value of PATH variable with other shells/distros, which in their default configuration preserve (append) the old value. For instance: https://git.archlinux.org/svntogit/packages.git/tree/trunk/profile?h=packages/filesystem#n6.

Devuan, Gentoo or CRUX overwrite the PATH on login. Like Alpine, they use SysV or OpenRC init. Lets align with them instead of Arch Linux, which is using an init system **incompatible to Alpine** and should not used as reference for anything related to init or startup.

The login mechanism is used to reset the environment to a known default. If you want custom PATH settings, please use the `/etc/profile.d/` mechanism like anyone else.

@agowa338 you are already installing non-base packages for WSL per default, so it should be easy to add a `/etc/profile.d/` script or replace `/etc/profile`.

am11:
@nero, I was looking at CentOS and ArchLinux as example. Thanks for pointing out that they are not a good source of reference. Would you instead have this follow Gentoo way (excerpt from Gentoo's /etc/profile):

```sh
# Set up PATH depending on whether we're root or a normal user.
# There's no real reason to exclude sbin paths from the normal user,
# but it can make tab-completion easier when they aren't in the
# user's PATH to pollute the executable namespace.
if [ "${EUID-}" = "0" ] || [ "${USER-}" = "root" ] ; then
        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin${ROOTPATH:+:}${ROOTPATH-}"
else
        PATH="/usr/local/bin:/usr/bin:/bin${PATH:+:}${PATH-}"
fi
export PATH
```

Note that @agowa338's is one implementation of WSL, there are other WSL implementers as well as qemu scenarios, which can profit from such default (i agree this is debatable).

> Edit: There are expectations how a commit message in aports should look like, it might be helpful to read the contribution guidelines. From the title its not evident that its about login or baselayout.

Thank you, I will shortly amend the commit as suggested.

PR #5811, opened 2018-12-10
testing/litespeed: don't remove directories needed later, fix depende…

joeholden:
…ncies


testing/litespeed/APKBUILD
```diff
@@ -56,8 +56,7 @@ package() {
 		"$pkgdir"/var/log || return 1
 
 	# remove trash
-	rm -fr "$pkgdir"/$_pkghome/php* \
```
andypost: any idea what stored here? 
Guess lsapi for php https://openlitespeed.org/kb/installation-from-source-code/#Install_PHP

needs @vakartel to approve

joeholden: There is something wrong with this package, the above only fixed some of it so I need to take another look

PR #5773, opened 2018-12-07
testing/gx: new aport

OTLabs:
https://github.com/whyrusleeping/gx
Packaging tool build around the distributed, content addressed filesystem IPFS

Needed for addition of go-ipfs, Inter Planetary File System (IPFS) package.

PR #5772, opened 2018-12-07
testing/gx-go: new aport

OTLabs:
https://github.com/whyrusleeping/gx-go
A tool to use with the gx package manager for packages written in go

Needed for addition of go-ipfs, Inter Planetary File System (IPFS) package.

PR #5812, opened 2018-12-10
testing/go-ipfs: new aport

OTLabs:
https://ipfs.io/
Inter Platnetary File System (IPFS), a peer-to-peer hypermedia distribution protocol

Requieres  #5773  and  #5772  packages.

### DIFF
--- a/testing/go-ipfs/APKBUILD
+++ b/testing/go-ipfs/APKBUILD
@@ -1,0 +1,80 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=go-ipfs
+pkgver=0.4.19
+pkgrel=0
+pkgdesc="Inter Platnetary File System (IPFS), a peer-to-peer hypermedia distribution protocol"
+url="https://ipfs.io/"
+arch="x86_64 x86"
+license="MIT"
+pkgusers="ipfs"
+pkggroups="ipfs"
+options="!check" # No test suit
+depends="musl"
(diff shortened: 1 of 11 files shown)


clandmeter:
If a PR depends on other PR's you need to merge the commits into a single one.
This will make sure the dependencies will be met.

OTLabs:
Build on arch architectures is failing as gcc unable to find ld.

clandmeter:
I'm sorry but i think I explained it incorrectly, i mean put the commits into a single PR. Sorry about that. 
The build scripts will automatically figure out the build order.

clandmeter:
nice thx, If you cannot fix builds on other archs please disable them.

OTLabs:
Could it be a misconfiguration of drone for arch architectures?

```
# github.com/ipfs/go-ipfs/cmd/ipfs
754 | /usr/lib/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
755 | collect2: fatal error: cannot find 'ld'
756 | compilation terminated.
```

Ikke:
Also make sure all dependencies that go pulls in are fixed. If you just do `go get` it will just get the latest dependencies, which can cause the build to be different each time.

OTLabs:
I have restricted the architectures to x86_64 and x86 only.

OTLabs:
The multihashes presented in https://github.com/ipfs/go-ipfs/blob/master/package.json are the same you see in build log, so the go dependencies are fixed.

Ikke:
gx in the last version looks good, thanks! Now you can apply the same to gx-go.

OTLabs:
Applied glide to gx-go!

PR #5814, opened 2018-12-11
testing/libupnpp: upgrade to 0.17.0

xsteadfastx:


andypost:
I bet `testing/upmpdcli` also needs update

andypost:
Yep, ABI changed
```
-usr/lib/libupnpp.so.7
-usr/lib/libupnpp.so.7.0.0
+usr/lib/libupnpp.so.8
+usr/lib/libupnpp.so.8.0.0
```
so dependent package needs bump

xsteadfastx:
On Di, 11. Dez 12:23, Andy Postnikov wrote:
> so dependent package needs bump

do i need a seperate PR for this?

andypost:
@xsteadfastx I think better make it separate PR and point that depends on this one

andypost:
As I see dependent package needs upgrade, and this upgrade is major - so probably better new PR

xsteadfastx:
i cant get upmpdcli updated. maybe even other dependencies are outdated.

andypost:
Looks yep, it needs to upgrade `libupnp` (used to test upgrade to 1.8.4, except failed test it works) @fcolista needed here 
then `libupnpp` upgrade (works fine) and then rebuild `upmpdcli` or upgrade

but neither current 1.2.16 no latest 1.3.8 can't be build with this new lib
``
src/ohservice.hxx: In constructor 'OHService::OHService(const string&, const string&, UpMpd*)':
src/ohservice.hxx:35:54: error: no matching function for call to 'UPnPProvider::UpnpService::UpnpService(const string&, const string&, UpMpd*&)'
         : UpnpService(servtp, servid, dev), m_dev(dev) {
                                                      ^
```

**Comment by xsteadfastx** [comment]
got it built. i think its ready for merge!

**Comment by xsteadfastx** [comment]
and i have no idea why the travis build fails. i rechecked the checksums... and it should work.

**Comment by xsteadfastx** [comment]
the problem is that upmpdcli needs the newer version of libupnpp. thats why the build fails on travis.

**Comment by clandmeter** [comment]
its normal that would happen as the local package failed to build (checksum error) so it will pull in the old one. If you fix the checksum error the new pkg will be pulled and should resolve your issue.

**Comment by xsteadfastx** [comment]
> its normal that would happen as the local package failed to build (checksum error) so it will pull in the old one. If you fix the checksum error the new pkg will be pulled and should resolve your issue.

a "abuild checksum" doesnt have anything to do. the checksum stays the same. maybe it was a temp failure.

**Comment by clandmeter** [comment]
thats because it creates checksums from the local archive which seems different from the  archive that travis pulled. You should remove the local archive on your system (probably in /var/cache/distfiles or similar) and run checksums again.

**Comment by xsteadfastx** [comment]
> thats because it creates checksums from the local archive which seems different from the archive that travis pulled. You should remove the local archive on your system (probably in /var/cache/distfiles or similar) and run checksums again.

that helped :) thanks

File: testing/upmpdcli/APKBUILD [hunk 1]
```diff
@@ -1,7 +1,7 @@
 # Contributor: Marvin Steadfast <marvin@xsteadfastx.org>
 # Maintainer: Marvin Steadfast <marvin@xsteadfastx.org>
 pkgname=upmpdcli
-pkgver=1.2.16
+pkgver=1.4.1
```
andypost: please bump it to 1.4.2

testing/libupnpp/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: Marvin Steadfast <marvin@xsteadfastx.org>
 # Maintainer: Marvin Steadfast <marvin@xsteadfastx.org>
 pkgname=libupnpp
-pkgver=0.16.1
-pkgrel=1
+pkgver=0.17.0
```
andypost: please bump it to 0.17.1

PR #5817, opened 2018-12-11
community/geth: upgrade to 1.8.20

misery:


community/geth/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: André Klitzing <aklitzing@gmail.com>
 # Maintainer: André Klitzing <aklitzing@gmail.com>
 pkgname=geth
-pkgver=1.8.16
+pkgver=1.8.19
```
andypost: btw why not latest release? https://github.com/ethereum/go-ethereum/releases/tag/v1.8.20

misery: Because it was released "at the same time" ;-)

PR #5822, opened 2018-12-12
testing/flawfinder: upgrade to 2.0.7

HRio:
add check()


HRio:
@fabaff @rgdoliveira please check, should we not move this package to community as well?

fabaff:
I would like to suggest that instead of disabling Python 3 we do it the other way around and do it for Python 2.

Can be moved to community from my point of view.

HRio:
@fabaff OK, I will switch to python3.

HRio:
@fabaff OK if I commit as in d049ee1 and a move to community/ ?

PR #5823, opened 2018-12-12
community/gst-plugins-bad: enable webrtc support

huww98:


huww98:
Not ready to merge now, I can not get WebRTC to work. It's reporting segmentation fault now

huww98:
OK, it's working. Previous segmentation fault is due to stack overflow. I can fix this in my code with a `pthread_setattr_default_np` call to set default stack size to 1M.

However, I cannot figure out where does this troublesome thread is created, thus cannot apply a patch to fix this. Can we just publish this and let the users to solve this themselves, or we need to fix this before merge?

ncopa:
> **update**:
> After further investigation, this thread is from a thread pool in glib. So it seems there are no good way to solve this problem. We can:
> 
> * patch glib to create threads with larger stack
> * patch gstreamer to call pthread_setattr_default_np for us
> * let users to deal with this themselves.

Another option may be to find where the large buffer is allocated on stack, and the make it use malloc instead. I think this should be reported upstream.

ncopa:
do you have a backtrace?

huww98:
> Another option may be to find where the large buffer is allocated on stack, and the make it use malloc instead. I think this should be reported upstream.

@ncopa that's a good option.

And I do have a backtrace.

```
Thread 8 "nicesrc1:src" received signal SIGSEGV, Segmentation fault.
[Switching to LWP 32]
priv_conn_keepalive_tick_unlocked (agent=agent@entry=0x5555558c8060 [NiceAgent]) at conncheck.c:829
829	conncheck.c: No such file or directory.
(gdb) bt
#0  0x00007ffff542f3e6 in priv_conn_keepalive_tick_unlocked (agent=agent@entry=0x5555558c8060 [NiceAgent]) at conncheck.c:829
#1  0x00007ffff542fa0d in priv_update_selected_pair
    (agent=agent@entry=0x5555558c8060 [NiceAgent], component=component@entry=0x555555e82760 [NiceComponent], pair=pair@entry=0x555555f22800) at conncheck.c:1433
#2  0x00007ffff542fd24 in priv_mark_pair_nominated
    (agent=agent@entry=0x5555558c8060 [NiceAgent], stream=stream@entry=0x555555e822e0 [NiceStream], component=component@entry=0x555555e82760 [NiceComponent], localcand=localcand@entry=0x555555ec08d0, remotecand=remotecand@entry=0x555555eec200)
    at conncheck.c:1574
#3  0x00007ffff5432081 in priv_reply_to_conn_check
    (use_candidate=1, msg=0x7ffff3c60a60, rbuf_len=112, sockptr=0x555555ec8c90, toaddr=0x7ffff3c6132c, rcand=0x555555eec200, lcand=0x555555ec08d0, component=0x555555e82760 [NiceComponent], stream=0x555555e822e0 [NiceStream], agent=0x5555558c8060 [NiceAgent]) at conncheck.c:2454
#4  0x00007ffff5432081 in conn_check_handle_inbound_stun
    (agent=agent@entry=0x5555558c8060 [NiceAgent], stream=stream@entry=0x555555e822e0 [NiceStream], component=component@entry=0x555555e82760 [NiceComponent], nicesock=0x555555ec8c90, from=0x7ffff3c6132c, buf=buf@entry=0x555555f1e860 "", len=120)
    at conncheck.c:3654
#5  0x00007ffff542c086 in agent_recv_message_unlocked
    (agent=agent@entry=0x5555558c8060 [NiceAgent], stream=stream@entry=0x555555e822e0 [NiceStream], component=component@entry=0x555555e82760 [NiceComponent], nicesock=<optimized out>, message=message@entry=0x7ffff3c613e0) at agent.c:3706
#6  0x00007ffff542c58e in component_io_cb (gsocket=<optimized out>, condition=<optimized out>, user_data=0x5555559d5c40)
    at agent.c:4982
#7  0x00007ffff645a93f in socket_source_dispatch
    (source=0x555555ef1f20, callback=0x7ffff542c264 <component_io_cb>, user_data=0x5555559d5c40) at gsocket.c:3857
#8  0x00007ffff65a01f6 in g_main_dispatch (context=0x555555be6ca0) at gmain.c:3182
#9  0x00007ffff65a01f6 in g_main_context_dispatch (context=context@entry=0x555555be6ca0) at gmain.c:3847
#10 0x00007ffff65a046f in g_main_context_iterate (context=0x555555be6ca0, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at gmain.c:3920
#11 0x00007ffff65a0728 in g_main_loop_run (loop=0x555555e8a200) at gmain.c:4116
#12 0x00007ffff3f25a1e in gst_nice_src_create (basesrc=<optimized out>, buffer=0x7ffff3c71668) at gstnicesrc.c:291
#13 0x00007ffff5246dc6 in  () at /usr/lib/libgstbase-1.0.so.0
#14 0x00007ffff524869d in  () at /usr/lib/libgstbase-1.0.so.0
#15 0x00007ffff5721549 in  () at /usr/lib/libgstreamer-1.0.so.0
#16 0x00007ffff65bcf45 in g_thread_pool_thread_proxy (data=<optimized out>) at gthreadpool.c:307
#17 0x00007ffff65bc9c0 in g_thread_proxy (data=0x555555ebb540) at gthread.c:784
#18 0x00007ffff7fbe95b in  () at /lib/ld-musl-x86_64.so.1
#19 0x0000000000000000 in  ()
```

huww98:
And https://stackoverflow.com/questions/48576955/program-segfaults-on-alpine-linux-how-do-i-resolve-it contains more details to the same problem.

huww98:
After play with gdb, I figure out that most of stack space is taken by [`local_buf`](https://gitlab.freedesktop.org/libnice/libnice/blob/6bfea0d063fb358a44e8800eae314520a2babd61/agent/agent.c#L4976) local variable which is a 64KB array, in the `#6` stack frame

I'm going to fix this with a patch if I have time. However, is it OK for me to report this to upstream? I think allocating a 64KB array on stack is quite normal and should not cause problem in most system.

PR #5826, opened 2018-12-12
main/qemu: upgrade to 3.1.0

andypost:
Ref http://lists.nongnu.org/archive/html/qemu-devel/2018-12/msg02415.html


mor1:
The Travis failure appears to be due to the `configure` script options having changed. 
I think `--with-gtkabi=3.0` needs to be removed (ca. line 275 in APKBUILD) as it looks like gtk.3.0 is the only one supported now.
(Am waiting for a build to complete to check this, but QEMU build is long!)

andypost:
@mor1 please file pr if gtkapi should be cleaned

mor1:
ncopa beat me to it :)

PR #5827, opened 2018-12-12
main/nodejs: upgrade to 10.14.2

timbru31:
Ref https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#10.14.2


timbru31:
branch rebased.

PR #5829, opened 2018-12-13
main/openssl: enable psk

misery:
There is no need to disable psk. Qt5 supports it. I will bump Qt5 after this.

### DIFF
--- a/main/openssl/APKBUILD
+++ b/main/openssl/APKBUILD
@@ -56,7 +56,7 @@ build() {
 		--openssldir=/etc/ssl \
 		shared no-zlib $_optflags \
 		no-async no-comp no-idea no-mdc2 no-rc5 no-ec2m \
-		no-sm2 no-sm4 no-ssl2 no-ssl3 no-seed no-psk \
+		no-sm2 no-sm4 no-ssl2 no-ssl3 no-seed \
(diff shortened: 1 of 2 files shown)


ncopa:
> Qt5 supports it.

But does Qt5 *need* PSK? libressl removed it, and since we used libressl for a while, I was kinda hoping nobody would miss it.

What is the usecase for PSK?

misery:
No, psk is not required for Qt5. But I like to compile AusweisApp2 (see: https://github.com/Governikus/AusweisApp2 and my Alpine-Docker: https://github.com/misery/DockerAusweisApp2) with Alpine and OpenSSL. The AusweisApp2 requires RSA-PSK support of Qt5.

btw: TLS 1.3 requires psk, too. Or is it separated?

ncopa:
@fabled what do you think about enabling PSK in openssl?

fabled:
I would in favor of enabling PSK as it is TLS feature that people seem to be using. Though, I have no strong opinion on this. I did bring up this also earlier in https://github.com/alpinelinux/aports/commit/abe1dc5988d12f5aca771605b109390f33ce7519#commitcomment-31279291.

misery:
I'm not the only one ;-) https://github.com/libressl-portable/portable/issues/450

PR #5830, opened 2018-12-13
community/qt: upgrade to 5.12.0

misery:
Qt 5.12.0 is an LTS release!

This contains https://github.com/alpinelinux/aports/pull/5829 because of "psk".


misery:
The job exceeded the maximum time limit for jobs, and has been terminated

andypost:
Needs to check `wkhtmltopdf` patch, kinda https://github.com/Opencomp/Dockerfiles/blob/master/wkhtmltopdf/Dockerfile

ollieparanoid:
Thank you, @misery!

@ncopa: can this be merged next week (but not earlier, so we can finish the kde framework upgrade in postmarketOS first)? Writing that since you asked me if that would cause problems in #postmarketOS.

PR #5837, opened 2018-12-15
main/rsyslog: upgrade to 8.40.0

andypost:
Ref https://github.com/rsyslog/rsyslog/blob/master/ChangeLog


andypost:
Fixed 272a1c64

PR #5839, opened 2018-12-15
testing/mitmproxy: security upgrade to 4.0.4 and modernize

andypost:
Ref https://github.com/mitmproxy/mitmproxy/releases

Package very outdated


andypost:
@fabaff looks a lot of py- deps needs to change

misery:
mitmproxy requires Python 3 and a lot of new deps. Remove py2 here.

PR #5848, opened 2018-12-15
main/xen: upgrade to 4.8.5 (Alpine 3.6)

HRio:
Security update


HRio:
@rnalrd tested on an Intel server

PR #5849, opened 2018-12-16
main/xen: add missing CVE refs

HRio:
While at it, drop depend that was not needed after all


HRio:
@rnalrd and @ncopa please check

PR #5850, opened 2018-12-16
WIP upgrade lxc to v3.1

clandmeter:
This should upgrade LXC from 2.x to 3.1

* upgrade lxc and add pam module as subpkg
* cleanup of lxc package and dependencies
* added man pages to lxc doc package
* split lua and py modules to seperate projects
* split templates to legacy templates (should be replaced by distrobuilder)
* introduce distrobuilder to testing repo (there is no stable release yet)
* add umoci and skopeo to community to support the new lxc-oci template


clandmeter:
@ncopa I wonder if we should merge this for v3.9. 
The oci template is depending on go based tools which I needed to keep in community so I didn't add them as deps.

testing/distrobuilder/APKBUILD
```diff
@@ -0,0 +1,73 @@
...
+url="https://linuxcontainers.org/"
+arch="all"
+license="Apache-2.0"
+depends="xz gnupg1"
+makedepends="go linux-headers glide"
+subpackages="$pkgname-lxc::noarch $pkgname-lxd::noarch"
+options="!check" # has no check
+source="$pkgname-$pkgver.tar.gz::https://github.com/lxc/distrobuilder/archive/$_commit.tar.gz
+	glide-$pkgver.yaml
+	glide-$pkgver.lock
```
ncopa: i think go modules is replacing glide, so we probably want to use go modules instead.

ncopa:
yes, i think it would be good to have this with v3.9.

testing/distrobuilder/APKBUILD
clandmeter: We don't have a proper way to handle go modules atm so its better to use glide for now.
We first need to write something to the ml on how to use them properly.

PR #5857, opened 2018-12-17
[3.8] main/libressl: upgrade to 2.7.5

J0WI:
> Added a blinding value when generating DSA and ECDSA signatures, in
> order to reduce the possibility of a side-channel attack leaking the
> private key.

https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.7.5-relnotes.txt


ncopa:
fixed with 4d33ed061d332d4a03432791a0973177a619919f

PR #5858, opened 2018-12-18
testing/wireguard: upgrade to 0.0.20181218

zx2c4:
```
  * jerry-rig: replace S_shipped with pl
  
  This fixes our jerry rig script, so that people can patch WireGuard into
  arbitrary kernel trees, instead of building as a standalone module.
  
  * chacha20,poly1305: simplify perlasm fanciness
  
  Numerous cleanups and correctness fixes in the assembly generators.
  
  * compat: don't undef BUILD_BUG_ON for Clang >=8
  
  The clang bug was finally fixed upstream, so we remove the workaround, while
  retaining the hack in our compat layer.
  
  * embeddable-wg-library: do not warn on unrecognized netlink attributes
  
  This brings behavior into parity with wg(8), and also allows more graceful
  addition of netlink attributes.
  
  * chacha20: do not define unused asm function
  
  This not only decreases code size, but also makes PaX's RAP happier.
  
  * compat: account for Clang CFI
  
  It turns out that RAP is no longer the only game in town, when it comes to CFI
  on kernels before Kees' epic timer_list refactoring. The Pixel 3/3XL kernels
  are based on 4.9 and come with Clang CFI on by default, so we account for this
  in our compat layer, so that we don't get CFI violation's and thus crashes.
  
  * wg-quick: bring interface up while setting MTU
  
  A small optimization to save a fork/exec.
  
  * makefile: use immediate expansion and use correct template patterns
  
  Coming up with the right combination of makefile template params that work on
  many kernels hasn't been easy, but hopefully this should solve building in a
  number of strange circumstances. If you're still having issues and you're
  using ccache, be sure to flush your cache first, as the cache might be serving
  stale copies of gcc's dependency info.
```

### DIFF
--- a/testing/wireguard-tools/APKBUILD
+++ b/testing/wireguard-tools/APKBUILD
@@ -3,7 +3,7 @@
 
 # NOTE: pkgrel must match _toolsrel in wireguard-vanilla
 pkgname=wireguard-tools
-pkgver=0.0.20181119
+pkgver=0.0.20181218
 pkgrel=0
 pkgdesc="Next generation secure network tunnel: userspace tools"
 arch='all'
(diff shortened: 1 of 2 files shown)


ncopa:
fixed with 932513c69d63b19b87b3d516e076d15c0baccc87

PR #5867, opened 2018-12-20
testing/py3-pyte: New aport

bratkartoffel:
This PR adds a new python module for pyte.

From the developers:
It's an in memory VTXXX-compatible terminal emulator. XXX stands for a series of video terminals, developed by DEC between 1970 and 1995. The first, and probably the most famous one, was VT100 terminal, which is now a de-facto standard for all virtual terminal emulators. pyte follows the suit.

This library will be used by another aport I'm about to create right now.


bratkartoffel:
will be part of #5868

PR #5871, opened 2018-12-21
main/sed: upgreade to 4.6

andypost:
Ref https://www.mail-archive.com/info-gnu@gnu.org/msg02532.html


tcely:
There are two options lines in the APKBUILD.

Why are we patching tests when the check is disabled?

andypost:
@tcely nice catch - fixed

andypost:
But not sure about tests - I just bumped version

PR #5878, opened 2018-12-21
communit/qt5-* rebuild package for psk

misery:
Can be ignored if https://github.com/alpinelinux/aports/pull/5830 lands into Alpine v3.9.


ncopa:
#5830 was merged

PR #5880, opened 2018-12-21
community/imagemagick6: fix broken download link

mksully22:
https://www.imagemagick.org/download/releases/ImageMagick-6.9.9-47.tar.xz is no longer available for download. Moving forward to https://www.imagemagick.org/download/releases/ImageMagick-6.9.9-51.tar.xz

### DIFF
--- a/community/imagemagick6/APKBUILD
+++ b/community/imagemagick6/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=imagemagick6
 _pkgname=ImageMagick
-pkgver=6.9.9.47
+pkgver=6.9.9.51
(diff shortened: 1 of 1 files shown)


community/imagemagick6/APKBUILD
```diff
@@ -3,7 +3,7 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=imagemagick6
 _pkgname=ImageMagick
-pkgver=6.9.9.47
+pkgver=6.9.9.51
```
dunielpls: As of 21. Dec. 2018, `6.9.10-19` is the latest version, which successfully builds.

ncopa: 6.9.10-x may require other packages to be rebuilt. I guess 6.9.9.51 is safe.

PR #5909, opened 2018-12-24
main/perl-sys-hostname-long: Use hostname -f instead of -fqdn

bratkartoffel:
The hostname bundled with busybox doesn't know the -fqdn switch. The
same functionality can be reached using the -f switch.


tcely:
I would almost prefer installing the version of `hostname` that is expected instead of patching.

bratkartoffel:
I'm not a fan of an extra dependency if the same functionality can be achieved by an 1-2 line patch.
As the last upgrade to this package was in 2015 I'm pretty sure that there won't be that much work with future upgrades.

bratkartoffel:
Can anyone please reopen, review and merge this?

PR #5913, opened 2018-12-24
community/qt5-qtcharts: New aport

bratkartoffel:
This PR adds the charts module of qt5

Relates to #5830


misery:
Should be switched to 5.12

PR #5917, opened 2018-12-25
community/ltb-project-ssp: upgrade to 1.3 and modernize

andypost:
Ref https://github.com/ltb-project/self-service-password/releases

It also points to outdated `mcrypt` which was fixed in 1.1 https://github.com/ltb-project/self-service-password/pull/92

Tha's why sodium & openssl added


andypost:
@rnalrd I did remove second patch because of crypty got rework and it is no longer needed

andypost:
One more reason http://build.alpinelinux.org/buildlogs/build-3-9-aarch64/community/ltb-project-ssp/ltb-project-ssp-1.0-r0.log

andypost:
@rnalrd on other hand the symlink to `config.inc.local.php` should also work after 1.3 https://github.com/ltb-project/self-service-password/pull/202/files

So we can remove second patch!

PR #5921, opened 2018-12-27
testing/newt: new aport

OTLabs:
https://mynewt.apache.org
Apache Newt is a smart build and package management tool for Apache Mynewt Operating System

### DIFF
--- a/testing/newt/APKBUILD
+++ b/testing/newt/APKBUILD
@@ -1,0 +1,31 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=newt
+pkgver=1.5.0
+_ver=${pkgver//./_}_tag
+pkgrel=0
+pkgdesc="Apache Newt is a smart build and package management tool for Apache Mynewt Operating System"
+url="https://mynewt.apache.org"
+arch="x86_64 x86"
+license="Apache-2.0"
+options="!check" # No test suit
+depends="git"
+makedepends="go"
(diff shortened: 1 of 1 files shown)


testing/newt/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=newt
+pkgver=1.5.0
+_ver=1_5_0_tag
```
Ikke: Note entirely sure if it's acceptable, because it's not POSIX complient, but ash does support it, but if you use this, then we don't have to update both `pkgver` and `_ver`.
```suggestion
_ver=${pkgver//./_}_tag
```
There are existing packages that use this syntax.

testing/newt/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+builddir="$srcdir/mynewt-newt-mynewt_${_ver}"
+
+build() {
+	cd "$builddir"
+	./build.sh
+}
+
+check() {
+	cd "$builddir"/newt
+	./newt version
```
Ikke: Just showing a version is not suitable for a `check` function, because it usually only tests a very small amount of code. We prefer to mark the package with `options="!check"` if there is no test suite available.

testing/newt/APKBUILD
OTLabs: Done! Thank you!

testing/newt/APKBUILD
OTLabs: Done! Thank you!

PR #5920, opened 2018-12-27
community/php7-phalcon: upgrade to 3.4.2

vakartel:

PR #5923, opened 2018-12-27
community/rspamd: upgrade to 1.8.3

andypost:
https://rspamd.com/announce/2018/12/03/rspamd-1.8.3.html and follow-up to https://github.com/alpinelinux/aports/pull/5920


andypost:
@vakartel btw looks this patch no longer needed, I build it fine locally without it

PR #5926, opened 2018-12-28
main/musl: Fix interpreter paths for libc6-compat

nero:
Paths for mips and ppc64el were wrong, several CARCH'es were missing.

On x86_64, this changes /lib64 from being a symlink to a directory containg a symlink.

I added a link to official glibc documentation to avoid guesswork in the future.

This supersedes #5661.

### DIFF
--- a/main/musl/APKBUILD
+++ b/main/musl/APKBUILD
@@ -121,20 +121,25 @@ compat() {
 	pkgdesc="compatibility libraries for glibc"
 
 	mkdir -p "$subpkgdir"/lib
+	# definitive source is https://sourceware.org/glibc/wiki/ABIList
 	case "$CARCH" in
-	armel) _ld="ld-linux.so.3" ;;
-	armhf | armv7) _ld="ld-linux-armhf.so.3" ;;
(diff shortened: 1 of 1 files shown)


nero:
Update: I renamed `$links` to `$_links` to avoid name clashes in the future.

ncopa:
Using symlinks to dirs is generally a bad idea so I am in favor of this.

main/musl/APKBUILD
```diff
@@ -119,25 +119,29 @@ utils() {
 
 compat() {
 	pkgdesc="compatibility libraries for glibc"
+	_links="lib/libc.so.6 lib/libcrypt.so.1 lib/libm.so.6 lib/libpthread.so.0 lib/librt.so.1 lib/libutil.so.1"
```
fabled: Should be "local _links=...".
But in fact I would have preferred to not touch the `for` loop. Just keep `_ld` and make it have the path prefix. This would make the patch less intrusive.
But other than this minor issue looks good to me.

nero:
Should be better now.

ncopa:
I think you forgot to bump pkgrel. I fixed that for you. Thanks!

PR #5930, opened 2018-12-28
main/openldap: Update to 2.4.47

bratkartoffel:
https://www.openldap.org/software/release/changes.html

### DIFF
--- a/main/openldap/APKBUILD
+++ b/main/openldap/APKBUILD
@@ -9,7 +9,7 @@
 #   - CVE-2017-9287
 #
 pkgname=openldap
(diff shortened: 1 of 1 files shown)


main/openldap/APKBUILD
```diff
@@ -9,7 +9,7 @@
 #   - CVE-2017-9287
 #
 pkgname=openldap
-pkgver=2.4.46
+pkgver=2.4.47
 pkgrel=2
```
tcely: @bratkartoffel It's not a huge thing, but `pkgrel` should be set to `0` when upgrading.

PR #5933, opened 2018-12-28
main/perl-http-body: Update to 1.22

bratkartoffel:
https://metacpan.org/source/GETTY/HTTP-Body-1.22/Changes


main/perl-http-body/APKBUILD
```diff
@@ -12,35 +12,28 @@ makedepends="perl-dev perl-test-deep"
...
 source="http://search.cpan.org/CPAN/authors/id/G/GE/GETTY/HTTP-Body-$pkgver.tar.gz
-	CVE-2013-4407.patch
 	"
 
-_builddir="$srcdir"/HTTP-Body-$pkgver
+builddir="$srcdir"/HTTP-Body-$pkgver
+
+# secfixes:
+#   1.22-r0:
+#     - CVE-2013-4407
```
ncopa: This secfixes comment is not correct. It indicates that the versions prior to 1.22-r0 is vulnerable to CVE-2013-4407, which is not true. We had a patch for it, which is removed with this commit.

The secfix comment needs to say which .apk version we fixed this CVE in alpine.

ncopa: use git log to find out exactly when we fixed this issue.

ncopa:
CVE-2013-4407 was fixed with 213ebd008f5f44ea962bebeb139ec959c8c5ce4d, so the `1.17-r0` had the fix.

main/perl-http-body/APKBUILD
bratkartoffel: Thank you, I just corrected that line

bratkartoffel:
This has already been applied with 7454a74b607d2ea338d1742b0ae51257b74fafe9

PR #5940, opened 2018-12-28
main/vanilla-linux: build amazon ena module

tomalok:
Build amazon's ena network driver module on x86, x86_64, and aarch64 (for EC2 a1 instances).
At some point we will probably want to add config-virt.aarch64, too.

The ena driver in recent kernels (4.19.x) is in sync with the one in https://github.com/amzn/amzn-drivers, eliminating the necessity for the 'community/aws-ena-driver' aport.

@ncopa


ncopa:
oh! very nice!

PR #5948, opened 2018-12-30
main/bind: update root hint file contents and naming

tanant:
I noticed when spinning up a simple bind container (https://hub.docker.com/r/resystit/bind9/) that the defaults for the root hints are a smidge out of date. It *works*, but it's possibly somewhat sub-optimal, so thought I'd suggest updating the file accordingly.

(also, if this isn't the right way to submit a PR, let me know - the contrib guidelines start by saying to *not* do a PR via the web interface.. but then in the detailed steps talk about going to github.com and creating the pull?)

 - update to the current (2018111402) root hints from
   ftp://rs.internic.net/domain/named.cache

 - rename file to named.cache from named.ca to match source
   and update recursive named.conf example to match

### DIFF
--- a/main/bind/named.cache
+++ b/main/bind/named.cache
@@ -1,85 +1,92 @@
-;       This file holds the information on root name servers needed to
+;       This file holds the information on root name servers needed to 
 ;       initialize cache of Internet domain name servers
 ;       (e.g. reference this file in the "cache  .  <file>"
-;       configuration file of BIND domain name servers).
-;
+;       configuration file of BIND domain name servers). 
+; 
 ;       This file is made available by InterNIC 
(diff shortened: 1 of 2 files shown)


tcely:
It'd be great if you could make a bind-root-hints subpackage out of this.

tanant:
@tcely sure, I'll have a stab at it and propose a couple edits - want to mark this one as changes request til i get that done, or want me to close and resubmit the new one?

tanant:
@tcely ah, gotcha. Would probably have never figured that out quite so quickly - neato

tanant:
( obsoleted/incorporated in #5950 )

PR #5106, opened 2018-09-05
testing/knot-resolver: upgrade to 3.2.1

tcely:
See https://github.com/alpinelinux/aports/pull/5950 for ideas about better future solutions for DNS root hints file.

PR #5950, opened 2018-12-31
main/dnshints-root: new aport

tcely:
Add `bind-dnssec-root` subpackage to handle `bind.keys` more like `dnssec-root`

Use `dnssec-root` with:
* knot-resolver
* unbound

Add and use `dnshints-root` package with:
* bind
* knot-resolver
* unbound

Fixes https://bugs.alpinelinux.org/issues/10160


main/bind/APKBUILD
```diff
@@ -68,6 +74,13 @@ prepare() {
 	default_prepare
 	cd "$builddir"
 
+	# Fetch and verify named.root
+	gpg --import < "$srcdir/verisign-grs-nstld-key.asc"
+	rm -rf named.root named.root.sig
+	curl -RLO 'https://www.internic.net/domain/named.root{.sig,}'
```
ncopa: I don't think the `{.sig,}` part works in busybox ash. Its a bash extension.

I don't think we should fetch this from `prepare()`, as it does not make the build reproducible. We may not end up with same .apk if we rebuild this in the future.

ncopa:
Is there a specific reason that we ship the root hints with bind instead of unbound? we ship an update scripts with unbound?

Or even better, why don't we have dns-root-hints as a separate package, similar to what we do with dnssec-root?

main/bind/APKBUILD
tcely: @ncopa curl is handling the brace expansion in the URL itself, not the shell.

The only way I can see to make this reproducible is to archive every version of `named.root` that is published on our own. It doesn't seem highly worthwhile to do that though.

tcely:
@ncopa We were shipping the same file (different revisions) from each of `bind` and `unbound` and they were both out of date. I don't know why this came about for sure.

The `unbound` update script was pretty much just a blind download in a cron task. I would much prefer to automate the updating of this file as part of the build.

We could have a `dns-root-hints` package, but what would prompt us to rebuild that? At least as part of bind, when a new version comes out we would get the latest file.

main/bind/APKBUILD
tanant: Slight pedant note, but the file references itself as named.cache. All three (named.cache, named.root, db.cache) all have the same signature etc and are identical, should we use that internal naming? There's no *material* difference, just curious mainly.

main/bind/named.ca
```diff
@@ -1,85 +0,0 @@
...
-;       configuration file of BIND domain name servers).
-;
-;       This file is made available by InterNIC 
-;       under anonymous FTP as
-;           file                /domain/named.root
-;           on server           FTP.INTERNIC.NET
-;       -OR-                    RS.INTERNIC.NET
-;
-;       last update:    Feb 04, 2008
-;       related version of root zone:   2008020400
```
tcely: So this very old version calls itself `named.root` and was just rotting away in our git repository.

main/unbound/root.hints
```diff
@@ -1,91 +0,0 @@
...
-;       configuration file of BIND domain name servers).
-;
-;       This file is made available by InterNIC 
-;       under anonymous FTP as
-;           file                /domain/named.cache
-;           on server           FTP.INTERNIC.NET
-;       -OR-                    RS.INTERNIC.NET
-;
-;       last update:    August 25, 2016
-;       related version of root zone:   2016082500
```
tcely: This file from a few years ago calls itself `named.cache` and presumably was being updated for some people by the cron job, but not being updated with `unbound` releases.

main/bind/APKBUILD
tanant: fair enough, I myself also have no idea behind the naming - I've always called it root.hints or something meaningful like that myself.

ncopa:
so, both `main/bind/named.ca` and `main/unbound/root.hints` were not updated. I am in doubt if it is a good idea to ship the root.hints with any apk at all.

Some alternatives:
- We ship a separate `update-root-hints` package that fetches the root.hints and store it in `/var` seomwhere. Both unbound and bind (and potentially other DNS recursors) depends on this package. It installs the update script as a cron job so systems automatically gets updated. It is also called from pre-install (or post-install) so it gets ifetched during first install. The benefit here is that we don't keep any copy of `root.hints` so we don't need to update any package, so no maintenance work for us. The drawback is that you may not have internet during install, for example for diskless installs. So in this case we'd need to make sure that the root.hints get saved with the config. (we could store it under `/etc` or add an lbu include file)

- We make a separate `dns-root-hints` package, where the version number is the date (eg `dns-root-hints-2018111402`. Since the version number is the timestamp, it will be more visible when root.hints is outdated. We could also as a part of the `check()` function fetch it form internet and compare with current version and give us an error if its outdated. An error would be catched every 6months when we build world for new releases. Or we could set up a service that monitors this package and gives us notification when its outdated. Benefit is that you get updates with `apk upgrade`. Drawback is that it require more maintenance work for us to keep this package updated.

- A combination of the above.  A `dns-root-hints` package that comes with a bundled `root.hints`, (which we store in our git repo together with the APKBUILD) but also  an `update-root-hints` to keep it up-to-date. We'd need some maintenance routine to update it regularly, maybe a monitoring service that send the maintainer a notification or similar.

The goals are:
- Reproducible builds (so a package `bind-x.y.z-r0` always has same content, regardless of when it was built)
- Low maintenance work.

tcely:
@ncopa I reworked this with your `dns-root-hints` alternative. The error from check should spit out the new version number for the root zone. Updating it should be as simple as copying that number into `pkgver` then trying again.

It's called `dnshints-root` to match the `dnssec-root` package. I'd like to put it in `main` so that it is available. For now, I've also copied it to `testing` so that TravisCI is happy to build my `testing/knot-resolver` changes.

tcely:
Now that 3.9 is released could someone commit 2e5a60b89c670bbb7b53acdc4c1fb7df42cac007 please. This will allow me to remove the work around from this branch.

@ncopa @andypost @nmeum

tcely:
Merging https://github.com/alpinelinux/aports/pull/5106 before this PR would be best.

tcely:
@clandmeter Can this and #6563 get merged soon?

rnalrd:
@ncopa any objection in merging this patch?

nmeum:
> I am in doubt if it is a good idea to ship the root.hints with any apk at all.

I agree and I didn't read the entire thread, but regarding unbound there is also an alternative regarding root-hints handling: using the builtin hints for the IN class which is apparently the default:

```
root-hints: <filename>
      Read the root hints from this file. Default is nothing, using
      builtin hints for the IN class. The file has the format of zone
      files, with root nameserver names and addresses only. The
      default may become outdated, when servers change, therefore it
      is good practice to use a root-hints file.
```

That's what I've been using [for the past years]( https://github.com/nmeum/aports/tree/master/8pit/unbound) since I didn't want to rely on the `root.hints` file shipped by the aports. IIRC this also what OpenBSD (which ships unbound in the base system) does.

tcely:
@nmeum Please read the thread. Your linked configuration is using the file included in the `unbound` source release. The last sentence of your quoted block recommends using a file that is kept up to date. That is what this new aport provides 

It is much better to share a common file that was known to be fresh than to ship out of date copies with our various resolvers.

tcely:
Please merge this before #6717

tcely:
Incorporated `bind` change by @HRio from #6833

rnalrd:
@tcely please look at https://github.com/alpinelinux/aports/commit/7ff07a2d4ff2b2784e9edf2fdb919f4207a49015
It's largely based on your work and implementing reproducibility and snapshot to self-update the package.
Thanks!

tcely:
@rnalrd you may have missed

tcely:
@rnalrd How does this detect when the hints are out of date?

rnalrd:
IMO it's not an abuild job. We already have a release monitoring service that notify the maintainers when new releases are out. Probably it could be used as well for this or something similar be setup.


Il giorno 4 apr 2019, 18:35, alle ore 18:35, tcely <notifications@github.com> ha scritto:

main/dns-root-hints/APKBUILD
```diff
@@ -13,27 +13,34 @@ options="net"
...
 
-build() {
-	mkdir -p "$builddir"
+_check_sig() {
+	local GNUPGHOME="$builddir/.gpg"
+	install -d -m 0700 "$GNUPGHOME"
+	gpg --import < "$srcdir"/verisign-grs-nstld-key.asc
+	gpg --verify "${_tmp}"/named.root.sig "${_tmp}"/named.root
+	rm -rf "$GNUPGHOME"
+	mv "${_tmp}"/named.root.sig "${_tmp}"/named.root "$builddir"/
```
rnalrd: I'd move the files in build() because it's not strictly part of the _check_sig() process

main/dns-root-hints/APKBUILD
```diff
@@ -13,27 +13,34 @@ options="net"
...
-	install -D -m 644 -o root -g root "$srcdir"/verisign-grs-nstld-key.asc \
-		"$pkgdir"/usr/share/$pkgname/verisign-grs-nstld-key.asc
 
-	# compatibility links
-	cd "$pkgdir/usr/share/$pkgname"
-	ln -s named.root named.cache
-	ln -s named.root db.cache
+	_fetch_new_version
+
+	sed -e 's/[[:space:]]\+$//' named.root > root-ns.hints
```
rnalrd: We prefer to use the original zone file, without touching it. It this way the user can verify anytime the signature of the zone file used in production.

main/dns-root-hints/APKBUILD
```diff
@@ -45,40 +52,42 @@ check() {
...
+
+package() {
+	install -D -m 644 -o root -g root named.root \
+		"$pkgdir"/usr/share/$pkgname/named.root
+	install -D -m 644 -o root -g root named.root.sig \
+		"$pkgdir"/usr/share/$pkgname/named.root.sig
+	install -D -m 644 -o root -g root "$srcdir"/verisign-grs-nstld-key.asc \
+		"$pkgdir"/usr/share/$pkgname/verisign-grs-nstld-key.asc
+
+	# install our cleaned up hints file
```
rnalrd: See previous comment

main/dns-root-hints/APKBUILD
tcely: It absolutely is. Unverified files should not be moved.

main/dns-root-hints/APKBUILD
tcely: That preference does not make a lot of sense. We don't store other sources in the repository or install signature / hash files into the running system.

I see the verification of new versions of this file as the job of this build process, not something for the users to do.

main/dns-root-hints/APKBUILD
rnalrd: If _check_sig() fails, it won't move on into build().

tcely:
> IMO it's not an abuild job. We already have a release monitoring service that notify the maintainers when new releases are out. Probably it could be used as well for this or something similar be setup.

An important property of this package is that it will not build / check an old version of the `named.root` file. Your changes violated this property. I'm also not convinced the snapshot function even works as intended.

There is no checkout, the directories don't seem to line up correctly and nothing pushes the commit.

main/dns-root-hints/APKBUILD
tcely: Indeed. You are arguing for moving the operation of moving verified files out of the `_check_sig` function to every place that function is called. Why does that seem better to you?

rnalrd: By saying "I'd" (I would) usually I mean a preference, not a strong opinion and that's the general meaning in english I believe but I could be mistaken since I'm not a native english speaker.

For the sake of the argument in this case "every place" equals to 1.

If you have a strong opinion on this you may leave it where it is. :smile:

PR #5953, opened 2018-12-31
testing/lzbench: new aport

OTLabs:
https://github.com/inikep/lzbench
lzbench is an in-memory benchmark of open-source LZ77/LZSS/LZMA compressors

### DIFF
--- a/testing/lzbench/APKBUILD
+++ b/testing/lzbench/APKBUILD
@@ -1,0 +1,31 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=lzbench
+pkgver=1.7.3
+pkgrel=0
+pkgdesc="lzbench is an in-memory benchmark of open-source LZ77/LZSS/LZMA compressors"
+url="https://github.com/inikep/lzbench"
+arch="x86_64 x86"
+license="GPL zlib MIT Unlicense BSD Apache-2.0 CDDL CC0 custom"
+subpackages="$pkgname-doc"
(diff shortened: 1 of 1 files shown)


andypost:
Interesting about how that works for arm according to https://github.com/inikep/lzbench/pull/26

OTLabs:
Thank you for pointing this out. I have removed ARM architectures.

testing/lzbench/APKBUILD
```diff
@@ -0,0 +1,31 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=lzbench
+pkgver=1.7.3
+pkgrel=0
+pkgdesc="lzbench is an in-memory benchmark of open-source LZ77/LZSS/LZMA compressors"
+url="https://github.com/inikep/lzbench"
+arch="x86_64 x86"
+license="unknown"
```
andypost: It looks like here should be list of licenses because a lot of different sublibs commited to repo like https://github.com/inikep/lzbench/blob/master/lzf/lzf.h#L2

OTLabs: I have updated APKBUILD with a list of licenses in use.

andypost:
it fails on 32 bit https://build.alpinelinux.org/buildlogs/build-edge-x86/testing/lzbench/lzbench-1.7.3-r0.log

OTLabs:
I will remove x86, thank you for checking this out.

andypost:
it builds on arm8 so I commited d12f36a1

andypost:
On s390x it fails as well https://build.alpinelinux.org/buildlogs/build-edge-s390x/testing/lzbench/lzbench-1.7.3-r0.log
So pushed 513342ec

PR #5954, opened 2018-12-31
community/geth: fix ppc64le build break by re-enabling isatty termina…

mksully22:
…l functions

https://github.com/ethereum/go-ethereum/pull/17740 disabled building isatty_linux.go for ppc64,ppc64le which makes these functions unavailable when building flags.go and color.go modules for geth. This patch allows the isatty_linux.go module to be built for ppc64le so geth build requirements are fixed. 

A post was added to https://github.com/ethereum/go-ethereum/pull/17740 to try and understand better why this was disabled in the first place for ppc64le.

### DIFF
--- a/community/geth/APKBUILD
+++ b/community/geth/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: André Klitzing <aklitzing@gmail.com>
 pkgname=geth
 pkgver=1.8.20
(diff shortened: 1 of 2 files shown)


hackmod:
it is just not updated (outdated isatty) not disabled.

PR #5956, opened 2019-01-01
Expose Travis Environment Variables

tcely:
None of the variables set by TravisCI were seen by scripts running under `alpine_run`.

This attempts to fix that situation so that `TRAVIS_COMMIT_RANGE` and the like are useful again.


tcely:
@clandmeter ping

PR #5939, opened 2018-12-28
main/http-parser: upgrade to 2.9.0

misery:

PR #5957, opened 2019-01-02
main/http-parser: upgrade to 2.9.0

timbru31:
This is required by recent Node.js versions, builds will fail otherwise (see e.g. https://api.travis-ci.org/v3/job/474337566/log.txt)


tcely:
See #5939

timbru31:
Oh missed that. Thanks for the hint.

PR #5958, opened 2019-01-02
testing/gdal: upgrade to 2.4.0

hjaekel:


hjaekel:
Thank you for your review, @andypost . I hope is what is needed to deal with the ABI chages.

hjaekel:
There seems to be a time limit of 50 mins for Travis, so the build was canceled. I tested the build locally, everything builds fine if there is enough time. Is there a way to increase the Travis time limit?

clandmeter:
This seems to break things depending on gdal. Builders are stuck now.
Can you make sure you can build things against new gdal?

andypost:
Something went wrong with builders gdal builds fine (https://build.alpinelinux.org/buildlogs/build-edge-x86_64/testing/gdal/gdal-2.4.0-r0.log ) but still not yet uploaded

hexer which fails right now builds fine on travis https://travis-ci.org/alpinelinux/aports/builds/475076389#L3892

andypost:
Builders except x86-64 unblocked, meanwhile it's strange that this builder have no sqlite-libs installed (at least gdal depends on it)

hjaekel:
Whats with armhf, armv7, and s390x? For [armhf](https://build.alpinelinux.org/buildlogs/build-edge-armhf/testing/gdal/gdal-2.4.0-r0.log) and [s390x](https://build.alpinelinux.org/buildlogs/build-edge-s390x/testing/gdal/gdal-2.4.0-r0.log) we have (successful) build log files, for armv7 I cannot find a build log.

hjaekel:
I tested the build of postgis against the new gdal version before creating this PR. The relevant part from my build logs is:

```
checking for gdal-config... /usr/bin/gdal-config
checking GDAL version... 2.4.0
checking for OGR enabled... yes
checking gdal.h usability... yes
checking gdal.h presence... yes
checking for gdal.h... yes
checking ogr_api.h usability... yes
checking ogr_api.h presence... yes
checking for ogr_api.h... yes
checking cpl_conv.h usability... yes
checking cpl_conv.h presence... yes
checking for cpl_conv.h... yes
checking for library containing GDALAllRegister... none required
checking for library containing OGRRegisterAll... none required
checking for library containing GDALFPolygonize... none required
enabling PostgreSQL extension support...
```

whereas the log on the alpine build server (see [here](https://build.alpinelinux.org/buildlogs/build-edge-x86/testing/postgis/postgis-2.5.0-r2.log)) fails with `GDALAllRegister`:

```
checking for gdal-config... /usr/bin/gdal-config
checking GDAL version... 2.4.0
checking for OGR enabled... yes
checking gdal.h usability... yes
checking gdal.h presence... yes
checking for gdal.h... yes
checking ogr_api.h usability... yes
checking ogr_api.h presence... yes
checking for ogr_api.h... yes
checking cpl_conv.h usability... yes
checking cpl_conv.h presence... yes
checking for cpl_conv.h... yes
checking for library containing GDALAllRegister... no
configure: error: could not find GDAL
>>> ERROR: postgis: build failed
```

I have no idea what causes the different behaviour. Maybe it is enough to trigger the build again. But because of the build failure, the build for postgis was disabled.

andypost:
In follow-up I enable build but to fix it for other arches `testing/geos` needs fixes https://github.com/alpinelinux/aports/pull/6024

hjaekel:
I investigated the build error of postgis and found the following reason:

The gdal package defines `makedepends=sqlite-dev` and in the resulting package description we find the line `depend = so:libsqlite3.so.0`. This shared object dependency is auto-detected.

When building postgis, Travis (and my local build environement) correctly installs `sqlite-libs` from the main repository to the build environment. Unfortunately the alpinelinux build environment chooses `lxd` from the testing repository, which also provides `libsqlite3.so.0`, but `libgdal.so.20` is not build against this library from `lxd` so it crashes when the function `GDALAllRegister` is called.

What can we do? Should we use `!tracedeps` and specify the dependencies of the gdal package manually? Should we fix the alpinelinux build environment so that it prefers packages from main over packages from testing?

ncopa:
> Unfortunately the alpinelinux build environment chooses lxd from the testing repository, which also provides libsqlite3.so.0

This sounds like a bug. Why does lxd provide libsqlite3.so.0? I think it should use system sqlite instead.

andypost:
Does it needs to bump gdal to unblock postgis?

sphrak:
@andypost thank you and everybody else for your efforts on getting this to work again :)

PR #5963, opened 2019-01-03
testing/postfwd2: New aport

bratkartoffel:
This PR adds postfwd2, a tool to write postfix restrictions in a ruleset file similar to those of the most firewalls.

### DIFF
--- a/testing/postfwd2/APKBUILD
+++ b/testing/postfwd2/APKBUILD
@@ -1,0 +1,45 @@
+# Contributor: Simon Frankenberger <simon-alpine@fraho.eu>
+# Maintainer: Simon Frankenberger <simon-alpine@fraho.eu>
+pkgname=postfwd2
+_pkgname=${pkgname:0:-1}
+pkgver=1.39
+pkgrel=0
+pkgdesc="Combines complex postfix restrictions in a ruleset similar to those of the most firewalls"
+url="https://postfwd.org/"
+arch="noarch"
+license="BSD"
+depends="perl-net-server perl-net-dns perl-io-multiplex"
+subpackages="$pkgname-doc $pkgname-openrc"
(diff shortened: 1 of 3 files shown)


tanertas:
Shouldn't this run as service/daemon? If so, you may enhance the package with required openrc files (postfwd.confd, postfwd.initd) and dedicated user and group (postfwd) to run service.

bratkartoffel:
Yes, you're right. I'll add the required files and the -openrc subpackage

bratkartoffel:
Done, this package now adds the "postfwd2" user and group and installs an openrc-configuration

testing/postfwd2/postfwd2.initd
```diff
@@ -0,0 +1,31 @@
...
+# Copyright 1999-2011 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+PIDFILE=/run/postfwd2.pid
+
+depend() {
+	need net
+}
+
+start() {
```
Ikke: If possible, could you use the declarative syntax rather than [defining your own start/stop functions][0]?

[0]:https://github.com/OpenRC/openrc/blob/master/service-script-guide.md#dont-write-your-own-startstop-functions

testing/postfwd2/APKBUILD
```diff
@@ -0,0 +1,48 @@
+# Contributor: Simon Frankenberger <simon-alpine@fraho.eu>
+# Maintainer: Simon Frankenberger <simon-alpine@fraho.eu>
+pkgname=postfwd2
+_pkgname=${pkgname:0:-1}
+pkgver=1.39
+pkgrel=0
+pkgdesc="postfwd is written in perl to combine complex postfix restrictions in a ruleset similar to those of the most firewalls."
```
Ikke: The description should not repeat the package/project name. Maybe you can think of a more succinct description?

bratkartoffel:
Thanks @Ikke, I just applied your suggestions

PR #5973, opened 2019-01-06
community/go-md2man: new aport

tomalok:
Adding an explicit `go-md2man` package, as a number of other packages (`runc`, `containerd`, `docker`, etc.) are building it for the sole purpose of converting markdown documentation to man pages... (often with 'go get', which has raised some unanswered questions.)

@ncopa @nmeum

PR #5713, opened 2018-11-23
community/docker: upgrade to 18.09.0 (containerd and runc splitted into own packages)

BernhardGruen:
community/docker: upgrade to 18.09.0 (containerd and runc splitted into own packages)
community/containerd: upgrade to 1.2.0 (moved from testing)
community/runc: new aport - based on runc 1.0.0rc5 + fixes

PR #5974, opened 2019-01-06
community/runc: new aport

tomalok:
Splitting `runc` out of community/docker, should be a little more straightforward and self-contained than PR #5713; also makes use of the new `go-md2man` package (PR #5973).

@ncopa

### DIFF
--- a/community/runc/APKBUILD
+++ b/community/runc/APKBUILD
@@ -26,7 +26,7 @@ build() {
 	mkdir -p $(dirname "$builddir")
 	ln -s "$PWD/$pkgname-$_commit" "$builddir"
 	cd "$builddir"
-	make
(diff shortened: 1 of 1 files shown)


tomalok:
build will fail until `go-md2man` is committed to edge.

Taking a look at `containerd` build docs, and I'm seeing that they have a dependency on an explicit `runc` commit, a bit after 1.0.0-rc6, for "dealing with advanced features"...  Currently considering how to deal with that, I may change `runc`'s `pkgver` to include the trailing commit hash.

tomalok:
`apk version --check` didn't exactly like the trailing hash, unless there's a better way, this will have to do.

tomalok:
Now bundled in #5986.

PR #5975, opened 2019-01-06
community/containerd: update to 1.2.1, remove from testing

tomalok:
Splitting `containerd` out of community/docker, should be a little more straightforward and self-contained than PR #5713.

Also makes use of the new `go-md2man` package (PR #5973) and the new `runc` package (PR #5974) -- build will not succeed until these two are committed.

@ncopa


tcely:
https://probot.github.io/apps/dep/

This is another PR where dep might be useful. @ncopa

tomalok:
Now bundled in #5986.

PR #5318, opened 2018-10-04
main/llvm5: upgrade to 5.0.2

J0WI:
See also #5317

PR #5977, opened 2019-01-06
Mesa/Clang/LLVM Upgrade

emolitor:
Upgrade LLVM to 7.0.1, Clang to 7.0.1 and update Mesa to 18.3.1

This includes patches to disable just the DynamicLibrary unload test in clang which fails due to musl not supporting library unloads. https://wiki.musl-libc.org/functional-differences-from-glibc.html

The clang build uses the llvm packaged lit to run clang tests. This works around a known issue in Clang 7.0.1. 

Upgrade Mesa to 18.3.1 - This was tested against Intel and AMD gpus. Limited NVidia testing.

Patches were pulled from the existing packages in Alpine, from void-linux or cleaned up from embedded projects I'm working on. x86_64 has been tested the most, x86 and aarch64 have had only limited testing. 

All versions are intentionally packaged in testing to leave main untainted. (This may not be correct)


J0WI:
See also #5717 for LLVM 7

martell:
Awesome work @emolitor
It seems as though time and resources are an issue to get large items like llvm and co reviewed.
So we might be waiting awhile to see all this in tree.

As this is the most recent PR maybe we can suggest a plan of attack.

- [ ] merge #5576 -> Remove llvm 3.7
Already approved, just need someone to merge it

- [ ] merge #5318 and #5318 -> LLVM / CLANG 5.0.1 -> 5.0.2
should be easy to approve
https://github.com/llvm-mirror/llvm/commits/release_50
https://github.com/llvm-mirror/clang/commits/release_50
Mostly contains spectre meltdown fixes and llvm ir fixes

The LLVM 7 changes don't need LLVM 6 so we can come back to 6 later.

After the above is merged. I will rebase #5717

- [ ] Merge #5717 -> LLVM 7.0.1

Then finally revisit the awesome mesa and clang work here.

craftyguy:
There's talk on [the Mesa mailing list about deprecating/removing autotools support](https://lists.freedesktop.org/archives/mesa-dev/2018-December/212166.html). I don't know if it makes sense to also include a move to meson in this PR, but it may want to be considered soon :smile:

emolitor:
Added in the 7.0.1 upgrades for both LLD and LLDB. Flirted with updating to python3 but realized that will complicate the rust-lldb package so reverted back to python2. Along the way found a few issues with the standalone build of lldb which I will try to get merged upstream.

community/rust/APKBUILD
```diff
@@ -16,7 +16,7 @@ license="Apache-2.0 BSD ISC MIT"
 # gcc is needed at runtime just for linking. Someday rustc might invoke
 # the linker directly, and then we'll only need binutils.
 # See: https://github.com/rust-lang/rust/issues/11937
-depends="$pkgname-stdlib=$pkgver-r$pkgrel gcc llvm-libunwind-dev musl-dev"
```
cjdesno: How much testing did you do after making this change? I was under the impression that on musl, rustc requires llvm-libunwind-dev to be present during compilation (but maybe I'm wrong, of course, or maybe it's a makedepends and not a normal dependency)

cjdesno:
How much testing have you done since removing rust's dependency on llvm-libunwind-dev? I was under the impression that on musl, rustc requires llvm-libunwind-dev to be present during compilation (but maybe I'm wrong, of course, or maybe it's a makedepends and not a normal dependency)

ncopa:
#5717 was merged. Can you please rebase? And also clean up the git commit messages. See https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request-pr

ncopa:
Clang test suite fails. I think th
````
Command Output (stderr):                                                                                                                                                                         --                                                                                                                                                                                               /home/ncopa/aports/main/clang/src/cfe-7.0.1.src/test/Driver/fsanitize.c:599:8: error: SP: expected string not found in input                                                                     // SP: -stack-protector                                                                                                                                                                          
       ^           
````

ncopa:
I'm gonna take the clang, lldb and llvm-unwind, squash the into fewer commits, and drop mesa and the rust thing for now. Mesa and firefox will have to go into separate PRs after the oter have been merged.

ncopa:
fixed with 09ed5f798886f450d1c66f0af50111e54e3c8870, a84ab695096337553ae79182143730853331e0e5 and 3c1b4c4949c0c0f431d2a0330aa69b37f668d9f5

PR #5979, opened 2019-01-06
 main/bind: upgrade to 9.12.3-P1

tcely:
Also, bring `named.ca` up to the current revision.

See https://github.com/alpinelinux/aports/pull/5950 for ideas about better future solutions for DNS root hints file.


tcely:
@rnalrd please commit this

PR #5981, opened 2019-01-06
testing/hexer: fix build error by adding sqlite-libs makedepends

mksully22:
Build failing with error: /usr/lib/gcc/i586-alpine-linux-musl/8.2.0/../../../../i586-alpine-linux-musl/bin/ld: warning: libsqlite3.so.0, needed by /usr/lib/gcc/i586-alpine-linux-musl/8.2.0/../../../libgdal.so, not found (try using -rpath or -rpath-link)

fix by explicitly adding sqlite-libs to make dependancies

### DIFF
--- a/testing/hexer/APKBUILD
+++ b/testing/hexer/APKBUILD
@@ -2,14 +2,14 @@
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
 pkgname=hexer
 pkgver=1.4.0
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


andypost:
I did this fix in https://github.com/alpinelinux/aports/pull/5983 and cd7e796072b2defa799dc93bcdeee8b5fc25de9e
But surely that is not needed - `gdal` package reports that it has `sqlite-libs` dependency https://github.com/alpinelinux/aports/pull/5958#issuecomment-451786067

PR #5983, opened 2019-01-06
testing/hexer: rebuild with sqlite

andypost:
Follow-up https://github.com/alpinelinux/aports/pull/5958


andypost:
For some reason buiilders does not set dependency for new gdal on `sqlite-libs`

PR #5986, opened 2019-01-07
community/docker: upgrade to 18.09.0, etc.

tomalok:
* update docker to 18.09.0
* `containerd` and `rund` are now built separately
* add dependency on `tini-static`, no need to build it here
* use new `go-md2man` package to process manpages

Alternative to  PR #5713, accomplishes the same thing, and some more.

@ncopa @andypost


community/runc/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+# and https://github.com/containerd/containerd/blob/v1.2.1/vendor.conf
+_commit=96ec2177ae841256168fcf76954f7177af9446eb
+
+pkgver=1.0.0_rc6
+pkgrel=0
+pkgdesc="CLI tool for spawning and running containers according to the OCI specification"
+url="https://www.opencontainers.org"
+arch="all"
+license="Apache-2.0"
+depends="libseccomp"
```
ncopa: I don't think the explicit depend of libseccomp is needed. It is automatically detected:
````
>>> runc*: Tracing dependencies...
	libseccomp
	so:libc.musl-x86_64.so.1
	so:libseccomp.so.2
````

tomalok: i'll keep that in mind for the next update, thanks!

PR #6002, opened 2019-01-09
main/bash: upgrade to 5.0

andypost:
Ref http://lists.gnu.org/archive/html/bug-bash/2019-01/msg00063.html


andypost:
Related upgrade of readline https://github.com/alpinelinux/aports/pull/6004

andypost:
Fixed with 7b8fc884

PR #3387, opened 2018-02-23
testing/py-boto3: depend on s3transfer

z38:
boto3 uses a separate library for managing Amazon S3 transfers.

Exception:
```text
Traceback (most recent call last):
  File "/app/app.py", line 52, in <module>
    main()
  File "/app/app.py", line 36, in main
    bucket = boto3.resource('s3').Bucket(bucket_name)
  File "/usr/lib/python3.6/site-packages/boto3/__init__.py", line 92, in resource
    return _get_default_session().resource(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/boto3/session.py", line 389, in resource
    aws_session_token=aws_session_token, config=config)
  File "/usr/lib/python3.6/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/usr/lib/python3.6/site-packages/botocore/session.py", line 861, in create_client
    client_config=config, api_version=api_version)
  File "/usr/lib/python3.6/site-packages/botocore/client.py", line 70, in create_client
    cls = self._create_client_class(service_name, service_model)
  File "/usr/lib/python3.6/site-packages/botocore/client.py", line 95, in _create_client_class
    base_classes=bases)
  File "/usr/lib/python3.6/site-packages/botocore/hooks.py", line 227, in emit
    return self._emit(event_name, kwargs)
  File "/usr/lib/python3.6/site-packages/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/usr/lib/python3.6/site-packages/boto3/utils.py", line 61, in _handler
    module = import_module(module)
  File "/usr/lib/python3.6/site-packages/boto3/utils.py", line 52, in import_module
    __import__(name)
  File "/usr/lib/python3.6/site-packages/boto3/s3/inject.py", line 15, in <module>
    from boto3.s3.transfer import create_transfer_manager
  File "/usr/lib/python3.6/site-packages/boto3/s3/transfer.py", line 127, in <module>
    from s3transfer.exceptions import RetriesExceededError as \
ModuleNotFoundError: No module named 's3transfer'
```

PR #6012, opened 2019-01-09
testing/py-boto3: added missing py-s3transfer dependency

dimon222:
I was going across requirements.txt in official boto3 repo and found out that we missed one of reqs.
This PR is just a fresh version of #3387 for latest boto3 lib.

S3 manager references - https://github.com/boto/boto3/blob/develop/boto3/s3/transfer.py
reqs file reference - https://github.com/boto/boto3/blob/develop/requirements.txt#L3


andypost:
Commited 64f42e16

PR #6016, opened 2019-01-09
main/py-six: upgrade to 1.12.0, added tests

dimon222:
https://github.com/benjaminp/six/blob/master/CHANGES#L6

### DIFF
--- a/main/py-six/APKBUILD
+++ b/main/py-six/APKBUILD
@@ -2,22 +2,27 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=py-six
 _pkgname=six
-pkgver=1.11.0
+pkgver=1.12.0
 pkgrel=0
 pkgdesc="Python 2 and 3 compatibility library"
 url="https://pypi.python.org/pypi/six"
 arch="noarch"
 license="MIT"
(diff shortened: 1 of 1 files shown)


dimon222:
Added tests.

main/py-six/APKBUILD
```diff
@@ -20,6 +21,12 @@ build() {
 	python3 setup.py build
 }
 
+build() {
```
rnalrd: I believe you meant to write check() ... :smile:

main/py-six/APKBUILD
```diff
@@ -20,6 +21,12 @@ build() {
...
+	cd "$builddir"
```
rnalrd: This can be omitted nowdays..

main/py-six/APKBUILD
dimon222: Eh, true.

dimon222: Done, forcepushed new version.

main/py-six/APKBUILD
dimon222: Removed in new forcepush.

rnalrd:
Merged

PR #6022, opened 2019-01-10
testing/gdal: Added missing dependancy for `ogr2ogr`

ellery:
`ogr2ogr` throws a missing shared library when run on alpine 3.8. Added
sqlite-lib to address issue.

Error Message
```
Error loading shared library libsqlite3.so.0: No such file or directory
(needed by /usr/lib/libgdal.so.20)
Error relocating /usr/lib/libgdal.so.20: sqlite3_bind_int: symbol not
found
Error relocating /usr/lib/libgdal.so.20: sqlite3_column_blob: symbol not
found
Error relocating /usr/lib/libgdal.so.20: sqlite3_column_int: symbol not
found
Error relocating /usr/lib/libgdal.so.20: sqlite3_column_int64: symbol
not found
Error relocating /usr/lib/libgdal.so.20: sqlite3_create_module_v2:
symbol not found
Error relocating /usr/lib/libgdal.so.20: sqlite3_bind_double: symbol not
found
Error relocating /usr/lib/libgdal.so.20: sqlite3_user_data: symbol not
found
...
```

### DIFF
--- a/testing/gdal/APKBUILD
+++ b/testing/gdal/APKBUILD
@@ -19,6 +19,7 @@ makedepends="
 	postgresql-dev
 	python2-dev
 	sqlite-dev
+	sqlite-libs
 	swig
 	tiff-dev
 	zlib-dev


ncopa:
I think this is wrong fix. Problem here is that testing/gdal was built on edge, which has newer version of sqlite. You need install sqlite from edge and it should be ok.

ncopa:
in general, mixing stable repos (v3.8) with edge is not supported. It may work, but there are no guarantee that it does, so when using testing repos, you should use everything from edge. Once it is confirmed to work the package can be moved to community (or main) for inclusion to next stable.

hjaekel:
@ncopa mixing stable and edge is not the root of the problem in this case, I tried to explain it in #5958 

Adding `sqlite-lib` does solve the symptoms, but not the cause.

andypost:
I bet this fix no longer needed

hjaekel:
I agree, the fix is no longer needed. The build of postgis can be enabled after the fix in f89e1def8cc32f8e631bd4b7c9adc9918ed9ef4c.

PR #6023, opened 2019-01-10
main/py-requests: upgrade to 2.21.0

dimon222:
Ref http://www.python-requests.org/en/master/community/updates/#release-history

### DIFF
--- a/main/py-requests/APKBUILD
+++ b/main/py-requests/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: William Pitcock <nenolod@dereferenced.org>
 pkgname=py-requests
 _pkgname=requests
-pkgver=2.19.1
+pkgver=2.21.0
 pkgrel=0
 pkgdesc="A HTTP request library for Python"
 url="http://www.python-requests.org/"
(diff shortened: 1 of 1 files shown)


andypost:
There's **CVE-2018-18074** mentioned in 2.20.0 (2018-10-18) so update could to add sec-fix

- now supports urllib3 v1.24 but https://pkgs.alpinelinux.org/packages?name=py-urllib3&branch=edge is 1.22

dimon222:
I'm looking in consideration for adding tests but sadly:
1. Way too many dependencies for checkdepends. Some are not available in alpine.
2. One of dependencies is statically pointing to very old version, so not sure if there's any point to push that package of this much old version (pytest-httpbin)

I'll see if i can publish missing ones from #1, but pytest-httpbin is sadly probably a bad idea to publish old version of package (especially considering alpine is keeping stuff mainly recent)

dimon222:
Any update on this? What is outstanding?
@andypost

andypost:
@dimon222 I mean there secfix comment should be added (use `git grep secfixes`) to find examples

dimon222:
@andypost added the comment and rebased

PR #6031, opened 2019-01-11
main/py-setuptools: upgrade to 40.6.3

dimon222:


dimon222:
On the side note, it actually does have tests (!check is not true), but they will fail because Alpine is missing nested dependencies for them (particularly pytest-virtualenv)

PR #6035, opened 2019-01-12
testing/py-mongo: python3 version added, upgrade to 3.7.2

dimon222:


andypost:
Makes sense to add `check()` cos package has tests https://github.com/mongodb/mongo-python-driver/blob/master/setup.py#L389

But travis reports
```
>>> WARNING: py-mongo: APKBUILD does not run any tests!
```

andypost:
I guess @fabaff agree on upgrade

dimon222:
I'll add tests and see how it goes.

andypost:
Commited 4cae36ba 6f9ec4d2

PR #6038, opened 2019-01-12
main/newsboat: upgrade to 2.14.1 (blocked on cargo)

mcrute:
`newsboat` introduced a hard dependency on `cargo` which is still in community and not main. Looks like this will need to wait till that moves to main. Verified that this builds if community is in the repos file.


mcrute:
`newsboat` introduced a hard dependency on `cargo` which is still in community and not main. Looks like this will need to wait till that moves to main. Verified that this builds if community is in the repos file.

Ikke:
@mcrute It might make more sense to move newsboat to community though.

PR #6052, opened 2019-01-15
testing/py-setuptools_scm: new aport

dimon222:
Ref https://github.com/alpinelinux/aports/pull/6030#issuecomment-454111875

PR #6053, opened 2019-01-15
testing/pytest-forked: new aport

dimon222:
Depends on #6052


andypost:
Commited 123fc84e but builders broken except armv7

dimon222:
Is it the code issue that caused builders to break? Anything needs to be done additionally from this package side?

PR #6057, opened 2019-01-16
Enabled CONFIG_PPS and CONFIG_PTP_1588_CLOCK

dheimerl:
Hyper-V time sync uses ptp, this enables that in the -virt kernel.


dheimerl:
Enabled on 32-bit as well and squashed into a single commit.

PR #6059, opened 2019-01-16
testing/py-apipkg: new aport

dimon222:
New aport + tests

### DIFF
--- a/testing/py-apipkg/APKBUILD
+++ b/testing/py-apipkg/APKBUILD
@@ -1,0 +1,53 @@
+# Contributor: Dmitry Romanenko <dmitry@romanenko.in>
+# Maintainer: Dmitry Romanenko <dmitry@romanenko.in>
+pkgname=py-apipkg
+_pkgname=apipkg
+pkgver=1.5
+pkgrel=0
+pkgdesc="apipkg: namespace control and lazy-import mechanism"
+url="https://github.com/pytest-dev/apipkg"
+arch="noarch"
+license="MIT"
+options="check"
+checkdepends="pytest"
+makedepends="python2-dev python3-dev py-setuptools py-setuptools_scm"
+subpackages="py3-$_pkgname:_py3 py2-$_pkgname:_py2"
+source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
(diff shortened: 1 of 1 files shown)


testing/py-apipkg/APKBUILD
```diff
@@ -0,0 +1,53 @@
...
+
+build() {
+	cd "$builddir"
+	python2 setup.py build
+	python3 setup.py build
+}
+
+check() {
+	cd "$builddir"
+	tox -e py27,py36
```
andypost: why 36 used when 3.7 in repo?

dimon222: Oops, going to update it shortly

dimon222: I actually took a look and didn't find py37.
tox.ini contains py36 only, as well as Alpine repo simply has only Python 3.6.7
https://github.com/pytest-dev/apipkg/blob/master/tox.ini

andypost: Btw ehst about using `pytest` here instead of tox,flakes, like https://github.com/pytest-dev/apipkg/blob/master/.travis.yml#L19

andypost: I checked and found no way to run tests... please remove the `check()` at all and `checkdepends` line

testing/py-apipkg/APKBUILD
```diff
@@ -0,0 +1,53 @@
...
+pkgname=py-apipkg
+_pkgname=apipkg
+pkgver=1.5
+pkgrel=0
+pkgdesc="apipkg: namespace control and lazy-import mechanism"
+url="https://github.com/pytest-dev/apipkg"
+arch="noarch"
+license="MIT"
+options="!check" # Tests are outdated
+depends=""
```
andypost: no reason to keep empty defaults

testing/py-apipkg/APKBUILD
dimon222: I was thinking about keeping tests inplace but disabling them (until it actually gets fixed in their repo... ever?)
But i'm okay to remove them to avoid noise.

testing/py-apipkg/APKBUILD
dimon222: Ok, I'll clean it from empty defaults

dimon222:
I found the way to get tests to work with native pytest + applied suggested corrections on removing empty defaults. I think this should be ok now.

andypost:
Commited a76860cf

PR #6073, opened 2019-01-18
Added support for copy_modules kernel command line argument.

herrfrei:
If used, the /.modloop/modules directory is copied to /lib instead of linked,
afterwards depmod is executed.
This allows to add other modules via apk to a USB-based system (like my acpi_call port).


nero:
This functionality is supported via OverlayFS, which is a preferable default over copying it to memory. Whats the reasoning behind this?

Please use the `KOPT_` variables instead of grepping through `/proc/cmdline`, like its done for `modloop=...`.

herrfrei:
The overlay_size argument didn't work for me. Only after issuing a /etc/init.d/modloop restart the overlay was created. The kernel module  was then loaded from the apk when doing apk add, but on the next reboot it was lost again.
I'd like to have an option to load kernel modules on boot, like Tiny Core Linux. They install them to /usr/local/lib/modules/... and link this to /lib/modules/.../kernel.tclocal. A depmod and the modules are available.

main/openrc/modloop.initd
```diff
@@ -99,7 +99,35 @@ start() {
 		mount -t overlay -o upperdir=/.modoverlayfs/modules,lowerdir=/.modloop/modules,workdir=/.modoverlayfs/work overlay /lib/modules
 		eend $? || return 1
 	else
-		rm -rf /lib/modules && ln -sf /.modloop/modules /lib/
+		# if we have a directory /lib/modules/$(uname -r)/local with modules, preserve these
+		LOCAL_MODULE_DIR=/lib/modules/$(uname -r)/local
```
nero: How do modules get into the local/ directory? Does this need to be manually?

main/openrc/modloop.initd
```diff
@@ -99,7 +99,35 @@ start() {
...
+		if [ -z $KOPT_local_modules ]; then
```
nero: This is missing from `init_KOPT()` above, otherwise it wont be set

main/openrc/modloop.initd
```diff
@@ -99,7 +99,35 @@ start() {
...
+			if [ ! -d $LOCAL_MODULE_DIR ]; then
+				mkdir $LOCAL_MODULE_DIR
+			fi
+			ebegin "Combining local kernel modules with /.modloop/modules"
+			# move local modules dir to tmp
+			mv $LOCAL_MODULE_DIR /tmp
+			# remove all modules and link all directories from /.modloop/modules/$(uname -r)/
+			rm -rf /lib/modules && mkdir -p /lib/modules/$(uname -r)
```
nero: Please don't delete `/lib/modules`, since it might contain package-supplied files.

main/openrc/modloop.initd
```diff
@@ -99,7 +99,35 @@ start() {
...
+			for d in $(find /.modloop/modules/$(uname -r)/ -mindepth 1 -maxdepth 1 -type d)
+			do
+				ln -sf $d /lib/modules/$(uname -r)
+			done
+			# copy original depmod files (to supress depmod warning)
+			cp /.modloop/modules/$(uname -r)/* /lib/modules/$(uname -r)/ 2>/dev/null
+			# link firmware 
+			ln -sf /.modloop/modules/firmware /lib/modules
```
nero: Symlink name is wrong here...

main/openrc/modloop.initd
```diff
@@ -99,7 +99,35 @@ start() {
...
+			# move local modules back
+			mv /tmp/local /lib/modules/$(uname -r)
```
nero: ... so this would fail because `/lib/modules` is now a symlink into the read-only `/.modloop`.

herrfrei: No, the subdirs of /.moduloop/modules/$(uname -r)/ are linked, not the dir itself. That's why I'm able to create the local dir at the same level. And, it works (I use it).

PR #6081, opened 2019-01-18
main/librsvg: upgrade to 2.45.4

J0WI:
https://gitlab.gnome.org/GNOME/librsvg/blob/master/NEWS


ncopa:
we cannot upgrade librsvg til we have rust available for all architectures. librsvg newer than 2.40 depends on rust.

J0WI:
Any update from `community/rust`?
https://forge.rust-lang.org/platform-support.html

PR #6085, opened 2019-01-18
main/postfix: Modify postfix-files so set-permissions works now

bratkartoffel:
This PR fixes /etc/postfix/postfix-files, a file containing a list of all files and permissions of the postfix package. This list is used to create/fix permission issues in the spool-directory or to setup a chrooted postfix installation.
Sadly this file contained non existant files (e.g. the postfix-ldap.so) when not all postfix packages were installed. This file also contained some entries about files which where moved to /usr/share/doc (LICENSE).

The "postfix set-permissions" aborts on the first missing file and refuses to continue. After removing the moved / not needed / missing files from the "postfix-files", the command now finishes without error.

### DIFF
--- a/main/postfix/APKBUILD
+++ b/main/postfix/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=postfix
 pkgver=3.3.2
-pkgrel=0
+pkgrel=1
 pkgdesc="Secure and fast drop-in replacement for Sendmail (MTA)"
 url="http://www.postfix.org/"
 arch="all"
@@ -164,6 +164,16 @@ package() {
 		_mvline "^\s*$map" "$map" dynamicmaps.cf
 	done
 	rm makedefs.out
+
+	# fix /etc/postfix/postfix-files so that "postfix set-permissions" can run
+	sed -i \
+		-e '/shlib_directory\/postfix-/d' \
+		-e '/meta_directory\/makedefs.out/d' \
+		-e '/manpage_directory/d' \
+		-e '/config_directory\/LICENSE/d' \
+		-e '/config_directory\/TLS_LICENSE/d' \
+		-e '/config_directory\/[^/]\+\.cf\.default/d' \
+		"$pkgdir"/etc/postfix/postfix-files
 }
 
 _mv_dict() {


main/postfix/APKBUILD
```diff
@@ -164,6 +164,18 @@ package() {
...
+	# fix /etc/postfix/postfix-files so that "postfix set-permissions" can run
+	sed -i \
+		-e "/shlib_directory\/postfix-/d" \
+		-e "/meta_directory\/makedefs.out/d" \
+		-e "/manpage_directory/d" \
+		-e "/config_directory\/LICENSE/d" \
+		-e "/config_directory\/TLS_LICENSE/d" \
+		-e "/config_directory\/.\+\.cf\.default/d" \
+		-e "/config_directory\/main.cf/d" \
+		-e "/config_directory\/master.cf/d" \
```
tcely: Why are main and master configuration files removed?

bratkartoffel:
You're right, I'll change this

Am 19. Januar 2019 18:28:45 MEZ schrieb tcely <notifications@github.com>:
>tcely commented on this pull request.
>
>
>
>> @@ -164,6 +164,18 @@ package() {
> 		_mvline "^\s*$map" "$map" dynamicmaps.cf
> 	done
> 	rm makedefs.out
>+
>+	# fix /etc/postfix/postfix-files so that "postfix set-permissions"
>can run
>+	sed -i \
>+		-e "/shlib_directory\/postfix-/d" \
>+		-e "/meta_directory\/makedefs.out/d" \
>+		-e "/manpage_directory/d" \
>+		-e "/config_directory\/LICENSE/d" \
>+		-e "/config_directory\/TLS_LICENSE/d" \
>+		-e "/config_directory\/.\+\.cf\.default/d" \
>
>Using `[^/]` is a better wildcard pattern than `.` for file paths.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/alpinelinux/aports/pull/6085#pullrequestreview-194363680

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

main/postfix/APKBUILD
tcely: I don't find a read-only mount used with some Docker containers to be a compelling reason for removing these two files. Particularly when as you said, postfix is going to fail when they are missing.

tcely:
@rnalrd please commit this

PR #6060, opened 2019-01-16
testing/py-execnet: new aport

dimon222:
Depends on #6059

PR #6090, opened 2019-01-19
testing/pytest-xdist: new aport

dimon222:
Depends on #6059 #6060


testing/pytest-xdist/APKBUILD
```diff
@@ -0,0 +1,54 @@
...
+
+build() {
+	cd "$builddir"
+	python2 setup.py build
+	python3 setup.py build
+}
+
+check() {
+	cd "$builddir"
+	tox -e py27,py36
```
andypost: Probably tox could be replaced with pytest

testing/pytest-xdist/APKBUILD
```diff
@@ -0,0 +1,54 @@
...
+# Maintainer: Dmitry Romanenko <dmitry@romanenko.in>
+pkgname=pytest-xdist
+_pkgname=pytest-xdist
+pkgver=1.26.0
+pkgrel=0
+pkgdesc="pytest xdist plugin for distributed testing and loop-on-failing modes"
+url="https://github.com/pytest-dev/pytest-xdist"
+arch="noarch"
+license="MIT"
+#options="!check" # Tests are outdated
```
andypost: no need in this line

dimon222: Good point, will be removed.

testing/pytest-xdist/APKBUILD
dimon222: Tried that, seems it works using "python2 -m pytest" but couple of tests are looking for filelock python dependency that is not part of alpine packages yet (huh).

dimon222: Got tests running, but also added checkdependency on #6142

andypost:
Restarted travis build after dependency commited
> 15 failed, 174 passed, 7 skipped, 9 xfailed, 1 xpassed in 312.52 seconds

Locally getting the same result

dimon222:
Interesting because all the tests pass 100% successfully for me on my alpine.

dimon222:
Found potential solution by manual install of lib in temporary directory, check latest build

andypost:
restarted travis and now it running tests

PR #6092, opened 2019-01-19
testing/nginx-mod-http-modsecurity: new aport

adamrushad:
New aports for ModSecurity. The first is the ModSecurity nginx connector, testing/nginx-mod-http-modsecurity. The second is a dependency of that nginx module, the ModSecurity 3.x library. Both packages included in this PR to ensure automated builds work. Comments also present in the nginx-mod-http-modsecurity APKBUILD related to why it is separate and how to merge into main/nginx if/when that is appropriate.

### DIFF
--- a/testing/modsecurity/APKBUILD
+++ b/testing/modsecurity/APKBUILD
@@ -1,0 +1,35 @@
+# Contributor: AdamRushad <adamrushad@adamrushad.com>
+# Maintainer: AdamRushad <adamrushad@adamrushad.com>
+pkgname=modsecurity
+pkgver=3.0.3
+pkgrel=0
+pkgdesc="ModSecurity WAF engine"
+url="https://modsecurity.org"
+arch="all"
+license="Apache-2.0"
+depends=""
+makedepends="libxml2-dev pcre-dev linux-headers flex-dev yajl-dev geoip-dev curl-dev valgrind"
+install=""
+# ModSecurity does not build documentation automatically. Seems to be a bug upstream. Both perl and doxygen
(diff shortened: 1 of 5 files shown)


testing/modsecurity/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+install=""
+# ModSecurity does not build documentation automatically. Seems to be a bug upstream. Both perl and doxygen
+#  will be required for generating documentation once it is fixed/implemented.
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::https://github.com/SpiderLabs/ModSecurity/releases/download/v$pkgver/$pkgname-v$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-v$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure --prefix=/usr --enable-valgrind
```
ncopa: Why is valgrind enabled? Valgrind is normally only used for debugging purposes.

adamrushad: No reason in particular. I had missed the email about your comment while I was migrating my domain, but some maxice8's comment below. I'll fix this while I fix their comment below.

testing/nginx-mod-http-modsecurity/APKBUILD
```diff
@@ -0,0 +1,217 @@
+# Maintainer: AdamRushad <adamrushad@adamrushad.com>
```
Ikke: Where does this APKBUILD come from? It seems to have a lot of history already.

PR #6118, opened 2019-01-21
main/py-pylast: upgrade to 3.0.0

fabaff:
pylast > 3.0.0 no longer supports Python 2.

Should we rename the package?


ncopa:
> pylast > 3.0.0 no longer supports Python 2.
> 
> Should we rename the package?

I think that makes sense, yes.

PR #6119, opened 2019-01-21
testing/mame: fix ppc64le build break for float handling

mksully22:
Submitting patches from https://github.com/bkaradzic/bx/commit/684f7ec5d6a8e0d9c396bdd2dfc1f19244800557?diff=unified

to fix ppc64le build issue described in https://github.com/alpinelinux/aports/commit/8b6964b6832e341d8c5514411c5f3e8af47ac6a4

with this integration ppc64le can be enabled again

### DIFF
--- a/testing/mame/APKBUILD
+++ b/testing/mame/APKBUILD
@@ -3,10 +3,10 @@
 pkgname=mame
 pkgver=0.205
 _pkgver=${pkgver/.}
-pkgrel=0
+pkgrel=1
 pkgdesc="Multi Arcade Machine Emulator with GroovyMAME/Switchres/No-nag patchset."
 url="https://mamedev.org"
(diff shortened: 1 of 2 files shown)


tanertas:
Thanks for PR and formatting patch. Mame monthly release routine is close and probably will be better push this with 0.206. I want to prepare a native arm build env otherwise it takes too much time in qemu. It took tens of hours to come where ppc64le build fails with qemu-system-ppc64.

andypost:
I going to commit it to check builders and result, on upgrade this patch will be just removed

andypost:
Now builders reports `clang` issue for `armhf` https://build.alpinelinux.org/buildlogs/build-edge-armhf/testing/mame/mame-0.205-r1.log

PR #6121, opened 2019-01-22
main/p11-kit: upgrade to 0.23.15

fabaff:


tcely:
Can the `etc/init.d` directory be removed? It appears to be empty.

PR #6126, opened 2019-01-22
community/py-mechanize: upgrade to 0.4.0

fabaff:


andypost:
Also you need to move to community `py-webencodings` and `pytest-expect`

PR #6142, opened 2019-01-23
testing/py-filelock: new aport

dimon222:
tl;dr dependency for tests of #6090 
Currently pointing to github releases instead of Pypi releases because tests somehow not included in tarball on pypi (?)
https://github.com/benediktschmitt/py-filelock/issues/48

### DIFF
--- a/testing/py-filelock/APKBUILD
+++ b/testing/py-filelock/APKBUILD
@@ -1,0 +1,52 @@
+# Contributor: Dmitry Romanenko <dmitry@romanenko.in>
+# Maintainer: Dmitry Romanenko <dmitry@romanenko.in>
+pkgname=py-filelock
+_pkgname=filelock
+pkgver=3.0.10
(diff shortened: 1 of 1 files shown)


testing/py-filelock/APKBUILD
```diff
@@ -0,0 +1,53 @@
...
+	python2 -m pytest -xvv test.py
+	python3 -m pytest -xvv test.py
+}
+
+package() {
+	mkdir -p "$pkgdir"
+}
+
+_py2() {
+	replaces="$pkgname"
```
fabaff: Not needed as it's a new aport and there was never a `py-filelock` package.

dimon222: Tho, for longterm simple upgrades would it give some benefit?

andypost:
Commited e03e40f3

PR #6156, opened 2019-01-25
community/qt5-qtwebengine: Fix x86

GrantM11235:
I took the patch from https://codereview.qt-project.org/#/c/247777/
and I changed the file paths from chromium to src/3rdparty/chromium
to match the structure of the source tarball.

I can confirm that qt5-qtwebengine now compiles for x86.


GrantM11235:
The CI timed out, which makes sense since it took over two hours to compile on my desktop.

PR #6174, opened 2019-01-28
testing/redo: fix check failing due to missing /dev/tty

Ikke:
On the builders, the checks failed because they tried to open /dev/tty,
which does not exists.

Add a patch that continues when /dev/tty is missing.

Re-enable redo


clandmeter:
Did you report this upstream?

clandmeter:
PS, next time if something doesn't build and you fix it you do not have to bump pkgrel.

PR #6183, opened 2019-01-29
main/busybox: add program to get hw-provided clocksource on s390x

tuan-hoang1:
References:
qemu/include/qemu/timer.h
qemu/include/hw/s390x/tod.h

Thanks to David Hildenbrand.


tuan-hoang1:
@ncopa

PR #6209, opened 2019-01-31
main/linux-vanilla: Don't build VBOXGUEST module.

gena01:
* This fixes a conflict with the virtualbox guest additions package.
  See virtualbox-guest-modules package


clandmeter:
Should we not instead prefer the kernel module?

gena01:
I posted a thread on the devel mailing list about this. But to summarize here:
The kernel module is much much smaller than the full guest-modules built package. It's also missing a ton of symbols and API.

When you have virtualbox-guest-modules and virtualbox-guest-additions then "VBoxService" can't start and crashes on boot. It tries to load "vboxsf" kernel module that depends on "vboxguest". If the vboxguest is the one provided by the kernel then I get a ton of "missing symbol" errors as well and vboxsf can't be loaded.

rnalrd:
Looks like we want the bloated module. Any objection in merging this @clandmeter @ncopa?

PR #6212, opened 2019-02-01
community/composer: upgrade to 1.8.3

andypost:
Ref https://github.com/composer/composer/releases


andypost:
@skwashd it works fine in my CI

skwashd:
@andypost given their recent track record I now give composer 2 business days before trying out one of their releases.

PR #6221, opened 2019-02-03
main/linux-vanilla: Enable CONFIG_HW_RANDOM_VIRTIO for ppc64le

DDoSolitary:
This module has been enabled on all other architectures except ppc64le. It adds support for VirtIO's hardware RNG device, which is important when running on a virtual machine.


ncopa:
fixed with 7efbae67cd467b5a5f98b3b634fd1f82e9fdb7e7

PR #6226, opened 2019-02-03
community/php7-pecl-ssh2: fix build and add check()

andypost:
Closes #9958

Extension ported to support php 7.x but no release, so using huge patch

PS https://bugs.alpinelinux.org/issues/9958

### DIFF
--- a/community/php7-pecl-ssh2/APKBUILD
+++ b/community/php7-pecl-ssh2/APKBUILD
@@ -3,14 +3,16 @@
 pkgname=php7-pecl-ssh2
 _pkgreal=ssh2
 pkgver=1.1.2
-pkgrel=3
+pkgrel=4
 pkgdesc="PHP extension provide bindings for the libssh2 library - PECL"
 url="https://pecl.php.net/package/ssh2"
 arch="all"
 license="PHP-3.01"
+depends="php7-common"
 makedepends="php7-dev autoconf libssh2-dev re2c"
-options="!check" # Depends on setup of user, cert & running ssh server
-source="$pkgname-$pkgver.tgz::https://pecl.php.net/get/$_pkgreal-$pkgver.tgz"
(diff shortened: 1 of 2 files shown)


andypost:
> Error relocating /mnt/community/php7-pecl-ssh2/src/ssh2-1.1.2/modules/ssh2.so: php_ssh2_parse_fopen_modes: symbol not found

That function is inline
```
./src/ssh2-1.1.2/ssh2_sftp.c:inline unsigned long php_ssh2_parse_fopen_modes(char *openmode) {
```
any idea why that could happen?

andypost:
The problem in `libssh2-dev` liked against libressl
```
# apk add libssh2-dev
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
(1/8) Installing libssh2 (1.8.0-r4)
(2/8) Installing libressl2.7-libcrypto (2.7.4-r2)
(3/8) Installing libressl2.7-libssl (2.7.4-r2)
(4/8) Installing libressl2.7-libtls (2.7.4-r2)
(5/8) Installing pkgconf (1.6.0-r0)
(6/8) Installing libressl-dev (2.7.4-r2)
(7/8) Installing zlib-dev (1.2.11-r1)
(8/8) Installing libssh2-dev (1.8.0-r4)
Executing busybox-1.29.3-r10.trigger
OK: 26 MiB in 22 packages
```

andypost:
Meanwhile it is strange and needs @ncopa  1f32c763e2384b800532d15aa32b6aea91d0e44e

andypost:
@ncopa I think we should move the package to testing because it's outdated, purely supported and not linked right against current libssh2

andypost:
Gonna commit this, the inline should be removed http://clang.llvm.org/compatibility.html#inline

PR #6230, opened 2019-02-05
testing/cloudflared: new aport

joeholden:
Just opening this so I remember, can leave it on hold for now...

I left LICENSE as unknown until the redistribution stance is clarified at https://github.com/cloudflare/cloudflared/issues/53

It could probably also do with stdout/err being directed to syslog.

### DIFF
--- a/testing/cloudflared/APKBUILD
+++ b/testing/cloudflared/APKBUILD
@@ -1,0 +1,58 @@
+# Contributor: Joe Holden <jwh@zorins.us>
+# Maintainer: Joe Holden <jwh@zorins.us>
+pkgname=cloudflared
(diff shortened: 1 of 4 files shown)


joeholden:
LICENSE added to -doc, doesn't look like they're interested in clarifying the license but I don't see anything that prohibits redistribution (IANAL)

Ikke:
@joeholden Unless they provide a proper open license, this should be considered non-free software, and we cannot redistribute the binary. That means this could only be added under non-free.

joeholden:
Since they're not interested in answering license questions at all either on their repo or via support tickets, close this

PR #6237, opened 2019-02-06
community/tomcat-native: Remove dependency to java

bratkartoffel:
The library itself doesn't depend on java, it's only used to compile this aport.

Fixes https://bugs.alpinelinux.org/issues/9740

### DIFF
--- a/community/tomcat-native/APKBUILD
+++ b/community/tomcat-native/APKBUILD
@@ -2,12 +2,12 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=tomcat-native
 pkgver=1.2.21
-pkgrel=0
+pkgrel=1
 pkgdesc="Native resources optional component for Apache Tomcat"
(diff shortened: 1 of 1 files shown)


bratkartoffel:
The package provides no tests, so i've added a !check.
Thanks for reviewing, @maxwell-k

community/tomcat-native/APKBUILD
```diff
@@ -2,12 +2,12 @@
...
 pkgname=tomcat-native
 pkgver=1.2.21
-pkgrel=0
+pkgrel=1
 pkgdesc="Native resources optional component for Apache Tomcat"
 url="https://tomcat.apache.org/native-doc/"
 arch="all"
 license="Apache-2.0"
-depends="openjdk8-jre-base"
+options="!check"
```
maxwell-k:
```suggestion
options="!check"  # no tests provided
```

bratkartoffel: added a comment, thanks for reviewing

PR #6243, opened 2019-02-06
testing/bup: new aport

adamruzicka:
and dependency

TODO:
- [ ] implement `check`

### DIFF
--- a/testing/py-xattr/APKBUILD
+++ b/testing/py-xattr/APKBUILD
@@ -13,13 +13,11 @@ source="https://pyxattr.k1024.org/downloads/pyxattr-$pkgver.tar.gz"
 builddir="$srcdir/$_pkgname-$pkgver"
 
 build() {
-	cd "$builddir"
 	python2 setup.py build
 	python3 setup.py build
 }
 
 check() {
-	cd "$builddir"
 	python2 setup.py test
 	python3 setup.py test
 }
--- a/testing/py-xattr/APKBUILD
+++ b/testing/py-xattr/APKBUILD
@@ -6,25 +6,22 @@ pkgrel=0
 pkgdesc="A python module for manipulating extended attributes"
 url="https://pyxattr.k1024.org/"
 arch="all"
-license="GPL"
-depends=""
-depends_dev=""
+license="LGPL-2.1-or-later"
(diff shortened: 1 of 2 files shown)


testing/py-xattr/APKBUILD
```diff
@@ -0,0 +1,28 @@
+# Maintainer: Adam Ruzicka <a.ruzicka@outlook.com>
+pkgname=py-xattr
+_pkgname=pyxattr
+pkgver=0.6.0
+pkgrel=0
+pkgdesc="A python module for manipulating extended attributes"
+url="https://pyxattr.k1024.org/"
+arch="all"
+license="GPL"
```
tanertas: Alpine uses spdx license format. Please pick proper one from https://spdx.org/licenses/

testing/py-xattr/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends="python2"
+depends_dev=""
+makedepends="python2-dev attr-dev"
+install=""
+subpackages=""
```
tanertas: You may remove unused variables like three above.

testing/py-xattr/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+source="https://pyxattr.k1024.org/downloads/pyxattr-$pkgver.tar.gz"
+
+_builddir="$srcdir"/$_pkgname-$pkgver
```
tanertas: No need custom _builddir here. You just could override $builddir and use everywhere when need it.

testing/py-xattr/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+build() {
+	cd "$_builddir"
+	python2 setup.py build || return 1
```
tanertas: " || return 1" is not used anymore. You may discard all of them.

Ikke:
Checks are currently failing. You might want to rebase this so that you get drone integration as well.

adamruzicka:
I'm afraid python 3 support is in the works but not done yet (see https://github.com/bup/bup/commit/d30215975cb04ed75f027006defa5b35c3f6a3ac). So bup itself needs python2 and therefore it needs py-xattr for python2 as well

PR #6244, opened 2019-02-06
main/cython: upgrade to 0.29.6, add support for python2 and python3 packages, remove dev subpackage, move to community, remove libplist from testing as duplicate aport

OTLabs:
Upgrade to 0.29.6.
Add support for python2 and python3 packages.
Remove dev subpackage.
Move cython to community.
Remove libplist from testing (duplicate aport, already in community).


andypost:
needs @itoffshore review

clandmeter:
@Ikke you are the more experienced python user, does it make sense to rename cython?

clandmeter:
Also, is cython ever a runtime dependency? I see most packages pull in both cython and cython-dev in makedepends.

OTLabs:
@clandmeter I have removed dev subpackages.

OTLabs:
Should I create/rename to cython and cython2? Or cython3?

OTLabs:
Arch Linux uses cython and cython2.
https://www.archlinux.org/packages/community/x86_64/cython/

Ikke:
My inclination this case is to go for cython(2|3) in this case rather than py(2|3)-cython. This is not a packaged python package/module, so I don't think this needs to be prefixed with `py`.

clandmeter:
I would keep cython name and introduce two subpackages called cython2|3 and you can use the same install_if logic as in our python example. This way you can just depend on cython and the correct cythonX will be installed with it. Nothing will be broken in depending packages so no need for replaces or similar.

https://wiki.alpinelinux.org/wiki/Python_package_policies#Python_2_.26_3

clandmeter:
If you remove the cython-dev sub-package we will need to update all aports that depends on cython-dev. You should probably bundle them into this PR.

You should probably include something like the following to not break aports that are maintained outside of aports tree.

```
# compatible fix for removing dev subpkg from apkbuild
# cython is not a library but a compiler and does not need a dev subpackage
provides=cython-dev
```

packages that have cython-dev

```
community/libplist
community/py-gevent
community/py-numpy
community/py-opengl-accelerate
community/py-rencode
community/py-scipy
community/py3-aiohttp-gns3
community/xpra
testing/ceph
testing/libplist
testing/py-cassandra-driver
testing/py-h5py
testing/py-pdal
```

OTLabs:
I am reaching time limit on Travis CI: `The job exceeded the maximum time limit for jobs, and has been terminated.`

OTLabs:
Looks like Drone picks up old `cython` for all builds so there is no way to build packages correctly as dev part is "missing" (the dev subpackage is now part of main package, so when the old `cython` is used it misses significant dev part, mostly all C files):
`... Installing cython (0.29.2-r0)` , should be 0.29.6

Is it possible to teach Drone to use new `cython` package?

OTLabs:
Travis CI picks up old package also: `Installing cython (0.29.2-r0)`.

clandmeter:
Looks like we are bumping into https://github.com/alpinelinux/abuild/pull/53

as discussed on IRC

- move cython to community (will fix community builds)
- remove libplist from testing (duplicate aport)

This is starting to look good! :)

community/cython/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+# Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
+pkgname=cython
+pkgver=0.29.6
+pkgrel=0
+pkgdesc="Cython is an optimising static compiler for both the Python & the extended Cython programming languages."
+url="https://cython.org"
+arch="all"
+license="Apache-2.0"
+_py2_deps="python2-dev"
+_py3_deps="python3-dev"
```
clandmeter: Why did you define these two vars? I see them only used 1 time.

community/cython/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+options="!check" # several tests fail
+makedepends="py2-setuptools py3-setuptools $_py2_deps $_py3_deps"
+subpackages="${pkgname}2:_py2 ${pkgname}3:_py3 $pkgname-doc"
```
clandmeter: The use of variables should be limited to places where there is a chance of a value to change. In this case i don't think pkgname will ever change (we already concluded pkgname should stay same). It will add complexity to read it, even if it's just minimal.

clandmeter:
Looks like it all builds except the ones from testing which is due to pulling in older cython.
I also did some nitpicking regarding those pkgname variable, its up to you if you want to change it or not.

For the rest LGTM!

community/cython/APKBUILD
OTLabs: I used the "Python package policies" and copy-pasted the example. I have removed these two variables.

community/cython/APKBUILD
OTLabs: I have replaced the `pkgname` variable with its explicit value in whole file.

OTLabs:
Thank you for all your kind input. Hope I have addressed all your comments and suggestions in full. I am in favor of nitpicking. Please, let me know if there is anything else to be done to make this PR better.

PR #6247, opened 2019-02-07
community/libcouchbase: update to 2.10.3

nathanejohnson:


andypost:
```
-usr/lib/libcouchbase.so.2.0.61
+usr/lib/libcouchbase.so.2.0.65
```
it means needs to bump all dependencies

at least 
* perl-couchbase
* php7-pecl-couchbase

nathanejohnson:
@andypost so what do I need to do then?  do I just increment pkgrel on those two dependent packages?  wouldn't that need to happen after this is merged?

PR #6251, opened 2019-02-07
testing/py-pandas: new aport

cgiraldo:
http://pandas.pydata.org/
Powerful data structures for data analysis, time series, and statistics
testing/

### DIFF
--- a/testing/py-pandas/APKBUILD
+++ b/testing/py-pandas/APKBUILD
@@ -33,10 +33,6 @@ package() {
 		"$pkgdir"/usr/share/licenses/$pkgname/LICENSE
 }
 
-package() {
-	mkdir -p "$pkgdir"
-}
-
 _py() {
 	local python="$1"
 	pkgdesc="$pkgdesc ${python#python}"
--- a/testing/py-pandas/APKBUILD
+++ b/testing/py-pandas/APKBUILD
@@ -1,0 +1,63 @@
+# Contributor: Carlos Giraldo <cgiraldo@gradiant.org>
+# Maintainer: Carlos Giraldo <cgiraldo@gradiant.org>
+pkgname=py-pandas
+_pkgname=pandas
+pkgver=0.24.1
+pkgrel=0
+pkgdesc="Powerful data structures for data analysis, time series, and statistics"
+url="http://pandas.pydata.org/"
+arch="all"
+license="BSD"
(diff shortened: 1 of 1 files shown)


cgiraldo:
Please, 
Review current testing/py-pandas of the repo, because it gives an ERROR: py-pandas-0.23.4-r0: BAD signature when installing.

Also the APK of the repo does not generate independent py2-pandas and py3-pandas packages.

The  APK of this PR generate py2-pandas and py3-pandas, so we can install pandas for python3 without depending on python2.

ncopa:
Do we really need python2 support? I don't want add python2 packages unless specifically needed.

cgiraldo:
That's my point. This APKBUILD provides both variants as py2- and py3- subpackages, so you can add only the corresponding pandas package and its dependencis for python3 or python2.

The APKBUILD I find in the current Alpine testing repository provides a single package that installs python2 with it, and I think is not a desired behavior.

testing/py-pandas/APKBUILD
```diff
@@ -0,0 +1,63 @@
...
+
+	local python; for python in python2 python3; do
+		$python setup.py install --prefix=/usr --root="$pkgdir"
+	done
+
+	install -D -m 644 LICENSE \
+		"$pkgdir"/usr/share/licenses/$pkgname/LICENSE
+}
+
+package() {
```
Ikke: Package function is defined twice

cgiraldo: Removed duplication.

cgiraldo:
I check the alpine package:
[https://pkgs.alpinelinux.org/package/edge/testing/x86_64/py-pandas](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/py-pandas)

And the package does not present subpackages py3-pandas nor py2-pandas. 

But when build the package locally, It generates all three packages py-pandas py2-pandas and py3-pandas packages.

Maybe a problem with the APKBUILD definition?

Ikke:
@cgiraldo Yes, ncopa looked into it earlier. `subpkgarch` is an internal variable and should not be used in APKBUILDs. This is what is preventing from the subpackages to be added to the repos, causing it to be rebuild over and over.

Ikke:
I'll push a fix for this (verifying first whether it works).

Ikke:
The fix has been pushed and it now builds correctly.

PR #6252, opened 2019-02-07
Gdal with geos

abellgithub:
GDAL without GEOS support leaves a bunch of vector functionality turned off.


ncopa:
this seems to be fixed with 90c43756afc2687d098acd7bac280b1d6f6c4715

PR #6255, opened 2019-02-07
community/tcllib: new aport

tcely:
https://core.tcl.tk/tcllib
Tcl Standard Library

This is data for Tcl (Tool Command Language) that passes its tests.

### DIFF
--- a/community/tcllib/APKBUILD
+++ b/community/tcllib/APKBUILD
@@ -1,0 +1,53 @@
+# Contributor: tcely <tcllib+aports@tcely.33mail.com>
+# Maintainer: tcely <tcllib+aports@tcely.33mail.com>
+pkgname=tcllib
+pkgver=1.19
+pkgrel=0
+pkgdesc="Tcl Standard Library"
+url="https://core.tcl.tk/tcllib"
+arch="noarch"
+license="TCL"
+depends="tcl"
+depends_dev=""
+makedepends="$depends_dev tcl-dev"
+install=""
+subpackages="$pkgname-doc"
+source="https://core.tcl.tk/$pkgname/uv/$pkgname-$pkgver.tar.xz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+	cd "$builddir"
+	./configure \
+		--build="$CBUILD" \
+		--host="$CHOST" \
+		--prefix=/usr \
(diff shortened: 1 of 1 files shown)


tcely:
@clandmeter Setting USER should not be needed, but DroneCI failed check without it.

Ikke:
New aports should start in testing, not directly in community.

tcely:
@Ikke NACK. Don't be pedantic. I'm not going to change this to `testing` just to immediately move it to `community` so this can be used in `checkdepends`.

Ikke:
@ncopa Your opinion?

Ikke:
@tcely FYI, I even adhere to this myself as well. I never add new packages directly to community, but first add them in testing, and when everything is alright, move it to community. This is something ncopa reaffirmed to me as policy.

tcely:
@ikke you are welcome to add this to testing yourself. Get back to me when you decide it is worthy of community and it becomes useful. If you feel so strongly about gatekeeping then you do this pointless work and leave me out of it!

This hiearchy of repositories is ridiculous in my opinion and is a large source of problems in this project. I will not be participating in the pointless exercise of guiding this process!

ncopa:
I guess we need to write a FAQ entry for this.

The original idea with testing was to give people a chance to test packages early. A user would ask for a package, a developer would build the package but would not test if it actually worked. So it was pushed to testing repository so the user who requested it could test if it actually ran.

This also helped us with the problem "it works when i build it locally, but not when its built with the official build servers".

With this in mind, it has made it easier to merge patches quicker, because we don't necessarily need to review to careful of testing stuff. When we move it to community or main, we do a second, more careful review and need to verify that it actually works. We want avoid that users of `edge/community` gets clearly broken stuff.

It does happen that we make exception to this, for example in case some package already in `community` introduce a new dependency.

Instead of expect that the already overworked patch reviewers does the digging work to figure out what this is all about you could have mentioned that this is a check dependency for `fossil` in the commit message, and then added the commit that adds it to the `community/fossil`. That would have helped the reviewer to see that this is a case where it make sense to make an exception.

tcely:
@ncopa Please do write this out. It may eliminate the contradictory requirements that seem to be coming from nowhere.

tcely:
@Ikke A better approach would have been to ask a question rather than to declare this is wrong.

PR #6261, opened 2019-02-08
community/qt5-qtwebengine: remove patches which are already upstream

PureTryOut:
I somehow missed the fact that these patches got pushed upstream into 5.12.1, so now the build failed. This PR fixes the build by just removing the patches again.


PureTryOut:
Woops, @rnalrd already found this and did it himself. Sorry for the trouble!

PR #6264, opened 2019-02-08
main/curl: fix missing static library symbols

j8r:
Linking to `libcurl.a` results of missing symbols.
There is the same problem with `libssh2.a`.

### DIFF
--- a/main/curl/APKBUILD
+++ b/main/curl/APKBUILD
@@ -13,7 +13,7 @@ depends="ca-certificates"
 depends_dev="openssl-dev libssh2-dev nghttp2-dev zlib-dev"
 checkdepends="python2"
 makedepends="$depends_dev autoconf automake groff libtool perl"
-subpackages="$pkgname-dbg $pkgname-doc $pkgname-dev libcurl"
+subpackages="$pkgname-static $pkgname-dbg $pkgname-doc $pkgname-dev libcurl"
 source="https://curl.haxx.se/download/$pkgname-$pkgver.tar.xz"
 builddir="$srcdir/$pkgname-$pkgver"
 
@@ -112,6 +112,12 @@ libcurl() {
 
 	mkdir -p "$subpkgdir"/usr
 	mv "$pkgdir"/usr/lib "$subpkgdir"/usr
+}
+
+static() {
+	pkgdesc="Curl static library"
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/*.a "$subpkgdir"/usr/lib/
 }
 
 sha512sums="1629ba154691bf9d936e0bce69ec8fb54991a40d34bc16ffdfb117f91e3faa93164154fc9ae9043e963955862e69515018673b7239f2fd625684a59cdd1db81c  curl-7.64.1.tar.xz"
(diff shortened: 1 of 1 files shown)


brentp:
I am getting errors like "undefined reference to `nghttp2_http2_strerror'" for anything related to nghttp2 when trying to statically link to libcurl.a. Does this resolve that too? I see there is no static lib built for nghttp2 in current package, only the shared lib.

j8r:
I'm also facing this issue. This is a first step, this doesn't solve the missing `.a` archive for the `nghttp2` package.

j8r:
Do we want a new `curl-static`, or is it fine to keep the `libcurl.a` archive inside `curl-dev`?

5paceToast:
@j8r this has been recently mentioned in irc.
@ncopa seems to want to have an automatic `-static` subpackage, likely with an `install_if` somewhere (ala current `-openrc`).

On that basis, I'd recommend making a new `curl-static` in the interim (I don't know if and when the automatic package will be made).

tcely:
Do we want to remove the static function now that `abuild` handles this?

5paceToast:
This is in main/ and I don't know if this apkbuild could get backported.
If we don't expect anything of the sort, that would probably be ok.

main/curl/APKBUILD
```diff
@@ -82,7 +82,7 @@ builddir="$srcdir/$pkgname-$pkgver"
 build() {
 	cd "$builddir"
 	autoreconf -vif
-	./configure \
+	./configure LDFLAGS=-static PKG_CONFIG='pkg-config --static' \
```
ncopa: It looks like this will enforce static linking of curl itself. I don't mind that we ship a static library of curl in a subpackage, but we will not build curl static for all of our users.

j8r: No? I got:
```
ldd usr/bin/curl
        /lib/ld-musl-x86_64.so.1 (0x7f9c22c9d000)
        libssl.so.1.1 => /lib/libssl.so.1.1 (0x7f9c22b70000)
        libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0x7f9c228f2000)
        libz.so.1 => /lib/libz.so.1 (0x7f9c226db000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f9c22c9d000)
```

j8r:
After testing it again, it already works the static linking already works by changing `--with-nghttp2` to `--without-nghttp2` and trying to link against an application.
I will rather works on `nghttp2-static` then

j8r:
After testing it again, the static linking already works by changing `--with-nghttp2` to `--without-nghttp2` and linking against an application.
I will rather works on a fix for nghttp2-static then.

j8r:
After testing it again, the static linking already works by changing `--with-nghttp2` to `--without-nghttp2` and linking against an application.
I will rather works on a fix for nghttp2-static then.
Sorry for the annoyance.

j8r:
I got this information https://github.com/curl/curl/issues/2516#issuecomment-383545545.
> - if you dont define `NGHTTP2_STATICLIB` when building libcurl, then a shared
    libcurl is built, and static libcurl is built that relies on shared nghttp2
> - if you do define `NGHTTP2_STATICLIB` when building libcurl, then a shared
    libcurl is built that relies on static nghttp, and static libcurl is built

That's why @brentp and me have this undefined reference when trying to static link - static libcurl depends on shared nghttp2.
@ncopa I am not very sure what to do, I guess we need to compile curl twice - one with `-DNGHTTP2_STATICLIB` for `curl-static` and one without for `curl-dev`.

PR #6266, opened 2019-02-09
testing/libmustache: upgrade to 0.5.0

andypost:
Dependency for https://github.com/alpinelinux/aports/pull/6219


andypost:
@renoirb are you fine with it?

andypost:
```
-usr/lib/libmustache.so.4
-usr/lib/libmustache.so.4.0.0
+usr/lib/libmustache.so.5
+usr/lib/libmustache.so.5.0.0
```
needs to bump dependencies

andypost:
As only php7-mustache depends on it I will do it in https://github.com/alpinelinux/aports/pull/6219

PR #6270, opened 2019-02-10
main/openssh: Fix CVE-2018-20685, CVE-2019-6109, CVE-2019-6111

bratkartoffel:
Fix PR fixes some vulnerabilities in openssh, following description was copied from debian DSA-4387-1:

CVE-2018-20685

    Due to improper directory name validation, the scp client allows servers to
    modify permissions of the target directory by using empty or dot directory
    name.

CVE-2019-6109

    Due to missing character encoding in the progress display, the object name
    can be used to manipulate the client output, for example to employ ANSI
    codes to hide additional files being transferred.

CVE-2019-6111

    Due to scp client insufficient input validation in path names sent by
    server, a malicious server can do arbitrary file overwrites in target
    directory. If the recursive (-r) option is provided, the server can also
    manipulate subdirectories as well.
    .
    The check added in this version can lead to regression if the client and
    the server have differences in wildcard expansion rules. If the server is
    trusted for that purpose, the check can be disabled with a new -T option to
    the scp client.

Please also backport these patches to the stable branches.

### DIFF
--- a/main/openssh/APKBUILD
+++ b/main/openssh/APKBUILD
@@ -4,7 +4,7 @@
 pkgname=openssh
 pkgver=7.9_p1
 _myver=${pkgver%_*}${pkgver#*_}
-pkgrel=2
+pkgrel=3
 pkgdesc="Port of OpenBSD's free SSH release"
 url="https://www.openssh.org/portable.html"
 arch="all"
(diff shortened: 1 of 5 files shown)


bratkartoffel:
Fixed with e231e6bed510c2304f37093f0eadb5708da9728e, thanks

PR #6271, opened 2019-02-10
community/wine: upgrade to 4.0.1

TBK:
https://www.winehq.org/announce/4.0.1


TBK:
@Ikke Could I get you to review this?

TBK:
@ncopa Could this be merged before 3.10?

community/wine/APKBUILD
```diff
@@ -5,23 +5,21 @@
...
 _pkgver=${pkgver/_/-}
 pkgrel=1
 pkgdesc="A compatibility layer for running Windows programs"
 url="https://www.winehq.org"
 arch="x86 x86_64"
 license="LGPL-2.0-or-later"
-# As of 2.0.3 most of the tests fails
-options="!check"
-subpackages="$pkgname-libs $pkgname-dev $pkgname-doc"
+options="!check" # As of 2.0.3 most of the tests fails
```
Ikke: Should we find out why these tests are failing and at least report them upstream?

TBK: I have tried to find the cause and if there are any related bugs reported, but the errors codes are not exactly easy to get a clue off and are often very ambiguous.

I have enabled check() so you can see the CI output - https://cloud.drone.io/alpinelinux/aports/5520

TBK: I dug around and it turns out that Wine does not even pass the test suite on "Glibc linux" - http://test.winehq.org/data/

Though it seems worse on musl. I think we or somebody needs to collaborate with WineHQ to work on improving musl support and get the necessary patches merged. Would also be great if WineHQ could build weekly builds for musl.

maxice8:
ping

maxice8:
any progress ?

TBK:
On my end not at the moment.

PR #6282, opened 2019-02-10
community/rng-tools: upgrade to 6.6

TBK:
https://github.com/nhorman/rng-tools/releases


andypost:
fails on s390x https://build.alpinelinux.org/buildlogs/build-edge-ppc64le/community/rng-tools/rng-tools-6.6-r0.log

Probably because of AES

PR #6283, opened 2019-02-10
community/libmbim: upgrade to 1.18.0

TBK:
100% backwards compatible - https://abi-laboratory.pro/index.php?view=timeline&l=libmbim


andypost:
As well as #6284 it should be a part of modem manager upgrade
```
 usr/lib/libmbim-glib.so.4
-usr/lib/libmbim-glib.so.4.3.0
+usr/lib/libmbim-glib.so.4.4.0
```

TBK:
I have combined all 3 PRs into one - #6285

PR #6292, opened 2019-02-11
community/remmina: upgrade to 1.3.2

TBK:


TBK:
Still trying to figure out what package that is pulling in libressl and causing:
```
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: warning:
libcrypto.so.1.1, needed by /usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../lib/libssh.so, may conflict with libcrypto.so.43
```

I am compiling it without issues locally and think it might be related to some sort of cache issue with Travis.

PR #6293, opened 2019-02-11
main/grep: upgrade to 3.3

TBK:


main/grep/APKBUILD
```diff
@@ -23,7 +22,6 @@ prepare() {
 }
 
 build() {
-	cd "$builddir"
```
tcely: Why are you removing the `cd` command here?

TBK: Because I thought it was acceptable from after the release of AL3.9. If that is not the case I will revert it.

tcely:
If it is, that would be news to me. Can you point.me to the reason why this CD is no longer needed?

ncopa:
Not needed after those commits in abuild:
https://git.alpinelinux.org/abuild/commit/?id=2fe29d5829c0973ace1db350141b3c810ac888a7
https://git.alpinelinux.org/abuild/commit/?id=324a57592e5e8f0196bb1bd082c278d82673af8c

ncopa:
We should probably wait with removing the `cd "$builddir"`  til the documentation is updated and mentions this.

TBK:
I have been thinking of ways to introduce people to the ABUILD format. One would be to make an interactive mode for newapkbuild (-i) which takes you through each line with a short explanation and a link/man mention for more info.

The same could be done for online docs in form of a sample where you can click the lines and a tooltip would show up explaining its function.

@ncopa I have been out of the loop for a while. Do you mean the wiki or also other places?

ncopa:
I mean the wiki, yes

ncopa:
actually, i mean the [APKBUILD.5](https://git.alpinelinux.org/abuild/tree/APKBUILD.5)

nmeum:
I agree with @ncopa we should keep stylistic changes to a minimum until we have some kind of documented style guide. I applied this without the stylistic changes.

PR #6298, opened 2019-02-11
main/openssl: Fix openssl secfix version number

mirskifa:
I have raised the PR against master, but this change also needs to be ported to the https://github.com/alpinelinux/aports/tree/3.9-stable branch (and potentially others?).  If I should raise separate PRs for those, please let me know.


rnalrd:
Merged to edge&v3.9
thanks

PR #6305, opened 2019-02-11
community/crystal: upgrade to 0.27.2

carlhoerberg:


carlhoerberg:
A new prebuilt crystal is required, 0.27.1 or later.

carlhoerberg:
Ok, now I'm kind of stuck: 

```
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: _main.o: in function `*Errno::new<String>:Errno':
main_module:(.text+0x3d24): undefined reference to `errno'```

**Comment by carlhoerberg** [comment]
I guess we need to build each version up to 0.27.2 and upload the snapshots of each version to: https://dev.alpinelinux.org/archive/crystal/

**Comment by carlhoerberg** [comment]
Maybe @jirutka or @ncopa can help guide me?

**Comment by TBK** [comment]
Please rename the PR to `community/crystal: upgrade to 0.27.2`

**Comment by carlhoerberg** [comment]
related to https://github.com/crystal-lang/crystal/issues/7196

**Comment by straight-shoota** [comment]
The Linux builds provided via Github releases have `x86_64-linux-gnu` target baked in. For compiling simple programs, it would be enough to specify linker arguments, but unfortunately, you can't configure the target used for running macros. That's why the generic linux binary won't work with musl

One solution is to use the generic linux build to cross compile for `x86_64-linux-musl` from a gnu system and link it on Alpine. I suppose that's why the previous releases used custom binaries hosted on dev.alpinelinux.org instead of the official release binaries. Alternatively, you should be able to build 0.27.2 with the 0.27.0 compiler.

A quick and dirty patch would be to simply replace the string `x86_64-linux-gnu` with `x86_64-linux-musl` in the compiler binary. It occurs only once and can simply be switched out.

I hope we can resolve crystal-lang/crystal#7196 for 0.28.0 so that only one linux binary can target both gnu and musl.

**Comment by carlhoerberg** [comment]
@jirutka need 0.27.0 and then 0.27.1 to compile 0.27.2. How do one get access to be able to upload them to https://dev.alpinelinux.org/archive/crystal/ ?

**Comment by carlhoerberg** [comment]
@jirutka prebuilds (snapshots of) crystal 0.27.0 for x64 and aarch64 needs to be uploaded to https://dev.alpinelinux.org/archive/crystal/ but I don't know how to go about that.

File: community/crystal/APKBUILD [hunk 1]
```diff
@@ -1,9 +1,9 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=crystal
-pkgver=0.27.0
+pkgver=0.27.2
 pkgrel=0
-_bootver=0.26.0
+_bootver=0.27.0
 _llvmver=5
```
straight-shoota: I think you should be able to use llvm 6. There are some issues with 7, though.

carlhoerberg: is llvm 6 available for alpine? couldn't see it here: https://pkgs.alpinelinux.org/packages?name=llvm6*&branch=edge

straight-shoota: Oh, I simply assumed it was. Nevermind then.

PR #6306, opened 2019-02-11
main/bison: upgrade to 3.3.2

TBK:


main/bison/APKBUILD
```diff
@@ -15,7 +15,6 @@ source="https://ftp.gnu.org/gnu/bison/$pkgname-$pkgver.tar.xz"
 subpackages="$pkgname-doc"
 
 build() {
-	cd "$builddir"
```
Ikke: ncopa requested to keep these statements when they are present at least for aports in `main` to reduce merge conflicts when backporting changes.

TBK: Fixed.

PR #6275, opened 2019-02-10
testing/php7-pecl-vips: new aport

TBK:
Ping @andypost please take a look.

PR #6309, opened 2019-02-11
main/dpkg: upgrade to 1.9.6, clarify license & fix dependency

TBK:


main/dpkg/APKBUILD
```diff
@@ -1,17 +1,16 @@
...
-pkgrel=1
+pkgver=1.19.4
+pkgrel=0
 pkgdesc="The Debian Package Manager"
 url="http://packages.debian.org/dpkg"
 arch="all"
-license="GPL"
-depends="xz"
+license="GPL-2.0-or-later BSD-2-Clause"
+depends="xz tar"
```
oliv3r: i'd probably sort the depends alphabetically, and put the tar before the xz...

main/dpkg/APKBUILD
```diff
@@ -1,17 +1,16 @@
...
 options="!check"
-makedepends="bzip2-dev zlib-dev perl linux-headers"
+makedepends="bzip2-dev linux-headers perl zlib-dev"
```
oliv3r: ironically you are sorting these :)

main/dpkg/APKBUILD
```diff
@@ -1,17 +1,16 @@
...
 subpackages="$pkgname-doc $pkgname-dev"
 source="http://ftp.de.debian.org/debian/pool/main/d/$pkgname/${pkgname}_$pkgver.tar.xz"
-builddir="$srcdir"/$pkgname-$pkgver
```
oliv3r: if you remove this var; won't the 'cd $builddir' fail below in prepare?

main/dpkg/APKBUILD
```diff
@@ -23,7 +22,6 @@ prepare() {
 }
 
 build() {
-	cd "$builddir"
```
oliv3r: Hmm, isn't this the 'normal' pattern of APKBUILD files? Why would we want to stray from the normal?

main/dpkg/APKBUILD
TBK: I forgot first time around. Done now.

main/dpkg/APKBUILD
TBK: No, since "builddir="$srcdir"/$pkgname-$pkgver" is the default path. From AL3.9 forward it is acceptable to remove. I can not find the statement atm. I got it pointed out in https://github.com/alpinelinux/aports/pull/6275#discussion_r255364603

Thanks for the feedback. It made me have another look at the prepare function.

I turnouts that prepare was completely unnecessary since the dpkg team had made code changes to deal with cases where ncursesw is not used. Plus sed is problematic since it does not give a failed notification.

TBK:
@rnalrd Could I get you to review this PR?

PR #6317, opened 2019-02-12
community/docker: upgrade to 18.09.2

tomalok:
A new version of docker was released to address https://nvd.nist.gov/vuln/detail/CVE-2019-5736, which actually is about `runc` (which we package/update separately).  There are only a few other too-minor-to-mention commits between this and 18.09.1.

Probably don't need to backport to [3.9] branch, because backporting `runc` should be sufficient.  I don't know offhand if we want to backport to older Alpine versions?

@ncopa @andypost

### DIFF
--- a/community/docker/APKBUILD
+++ b/community/docker/APKBUILD
@@ -2,8 +2,8 @@
 # Maintainer: Jake Buchholz <tomalok@gmail.com>
 
 pkgname=docker
-pkgver=18.09.1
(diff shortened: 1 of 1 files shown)


andypost:
I think better to wait @ncopa opinion

tomalok:
Edge `docker` should probably continue to follow the current version (which is what this PR is for).

`runc` was already updated for 3.9-stable, so we should be okay with regards to the CVE there.

Question is whether we want to attempt a backport of the fix or upgrade docker for 3.8, 3.7, and 3.6 -- it's more difficult because `runc` wasn't its own package back then.

tomalok:
superceded by #6530

PR #6330, opened 2019-02-14
testing/py-pandas new aport

sgerrand:
💁 Packages the Python Data Analysis Library.

http://pandas.pydata.org/


ncopa:
I had to revert this because it does not create the expected subpackages for some reason and as a result it  causes the builders try to rebuild it every git push. Sorry.

ncopa:
It is also a duplicate #6251

sgerrand:
😞 Thanks for letting me know about the issue. I'll fix the subpackages build today.

sgerrand:
As background, I've been maintaining [an Alpine package for pandas](https://github.com/sgerrand/alpine-pkg-py-pandas) for ~ 2 years. This is the first of hopefully further packages that I'd like to bring into aports.

sgerrand:
Found the issue: the architecture should have been set as "all" rather than "noarch".

cgiraldo:
Hi! I submitted the #6251 pandas aport. 
I didn't find any pandas aports at the time I pull request #6251.

If this PR is finally merged, please, remove "python" from the depends line in order to not depend on python2 for the py3-pandas.

APK line should be:

`depends="py-dateutil py-numpy py-tz" `

Ikke:
The other PR has been merged, so this can be closed.

PR #6331, opened 2019-02-14
main/apache2: Create /run folder for apache2 on install

bratkartoffel:
The directory /run/apache2 is only created when running in openrc (see
apache2.initd). When installing on a non-openrc system (e.g. docker),
this folder is missing and apache2 refuses to start.

Closes https://bugs.alpinelinux.org/issues/9982

### DIFF
--- a/main/apache2/APKBUILD
+++ b/main/apache2/APKBUILD
@@ -3,7 +3,7 @@
 pkgname=apache2
 _pkgreal=httpd
 pkgver=2.4.38
-pkgrel=1
+pkgrel=2
 pkgdesc="A high performance Unix-based HTTP server"
 url="https://httpd.apache.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


kunkku:
It would be better to create the directory in the `package` function of APKBUILD, in order to get it properly removed on uninstallation.

bratkartoffel:
As the /run directory is non-persistent this shouldn't matter.
Also "apk del" only removes the folder if it's empty. I think that there may be other files dangling in there.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

ncopa:
I prefer creating directories from APKBUILD. There is no reason to do it from install script.

bratkartoffel:
Ok, I'll update this PR later after work.

bratkartoffel:
Added the directory to the package() task

PR #6334, opened 2019-02-14
testing/gdal: gdal should enable geos

morlay:
`geos`, `proj4` and `gdal` is used together frequently.
`proj4` could be added `proj4-dev`. but `geos` have to be enabled when compiling.


bratkartoffel:
Please increment the `pkgrel`

PR #6345, opened 2019-02-15
community/erlang: Fix compilation against musl >=1.1.21

bratkartoffel:
The latest update in musl removed some internal wrappers around
sigaction, which erlang needs. Erlang tries to patch them, so it can do
other fancy stuff before calling the real sigaction.

Without this PR all erlang applications failed to start with the following
message: "dlsym: Resource temporarily unavailable"

Other changes:
* APKBUILD modernized
* Made the build parallel (erlang supports this)
* Some subpackages (which only contain .beam files) are platfrom independant, so specify them as "noarch"
* check() function implemented, but disabled as the tests fail on my machine

Fixes https://bugs.alpinelinux.org/issues/9983 and all other packages requiring erlang.
Backporting **not** needed as alpine 3.9 is still on musl 1.1.20

### DIFF
--- a/community/erlang/APKBUILD
+++ b/community/erlang/APKBUILD
@@ -5,68 +5,64 @@
 pkgname=erlang
 pkgver=21.2.1
 _srcver=$pkgver
-pkgrel=0
+pkgrel=1
 pkgdesc="General-purpose programming language and runtime environment"
 url="http://www.erlang.org/"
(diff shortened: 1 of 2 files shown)


bratkartoffel:
This is somehow applied with 52e70f912af535cecaccc74f2ff45af487cf7101, thanks!

PR #6346, opened 2019-02-15
main/linux-vanilla: enable CONFIG_PINCTRL_BAYTRAIL=y

joeholden:
This fixes the card reader on some of the Atom tablet SoCs


ncopa:
why not compile it as module? `CONFIG_PINCTRL_BAYTRAIL=m`

ncopa:
> why not compile it as module? `CONFIG_PINCTRL_BAYTRAIL=m`

Becuase that is apparently not an option:
````
Intel Baytrail GPIO pin control (PINCTRL_BAYTRAIL) [N/y/?] (NEW) 
````

Ok, lets go for `CONFIG_PINCTRL_BAYTRAIL=y` then.

joeholden:
Yeah I thought I could get away with that too, but for some reason that particular platform code cannot be built as a module, it would be nice if it was...

PR #6351, opened 2019-02-15
community/libbluray: upgrade to 1.1.0

prspkt:


andypost:
ABI changed - needs rebuild dependent packages
```
-usr/lib/libbluray.so.2.0.2
+usr/lib/libbluray.so.2.1.0
```

prspkt:
The potentially affected pkgs are `kodi` `vlc` and `handbrake`, but I think in terms of [compatibility](https://abi-laboratory.pro/index.php?view=timeline&l=libbluray) we may OK.

PR #6354, opened 2019-02-16
community/mpv: enable wayland

tanertas:
main/libva(-glx): upgrade to 2.4.0, enable wayland
main/libva-intel-driver: upgrade to 2.3.0, enable wayland

### DIFF
--- a/community/mpv/APKBUILD
+++ b/community/mpv/APKBUILD
@@ -5,21 +5,49 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=mpv
 pkgver=0.29.1
-pkgrel=1
+pkgrel=2
 pkgdesc="Video player based on MPlayer/mplayer2"
 url="https://mpv.io/"
 arch="all"
 license="GPL-2.0-or-later LGPL-2.0-or-later"
-makedepends="libxxf86dga-dev libxv-dev libmad-dev lame-dev libao-dev
-	libtheora-dev xvidcore-dev zlib-dev sdl-dev freetype-dev
-	x264-dev faac-dev ttf-dejavu libxvmc-dev alsa-lib-dev libva-dev
-	mesa-dev ffmpeg-dev libass-dev perl lua5.2-dev libvdpau-dev python2
-	libxext-dev libxinerama-dev libxscrnsaver-dev py-docutils libxrandr-dev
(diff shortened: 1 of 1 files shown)


andypost:
Probably fixed with 198e5582

tanertas:
> Probably fixed with [198e558](https://github.com/alpinelinux/aports/commit/198e5582eff680a73286d15d300040902cd54e2b)

I have updated this PR accordingly.

tanertas:
@ncopa @Ikke can anyone take a look at this?

Ikke:
@tanertas Could you rebase this onto the latest master? Then we will get drone CI as well. Thanks.

community/mpv/APKBUILD
```diff
@@ -5,21 +5,49 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=mpv
 pkgver=0.29.1
-pkgrel=1
+pkgrel=2
```
Ikke: Are there any function changes, or is it just style changes? In the case of the latter, a `pkgrel` bump is not necessary (no need to rebuild the packages)

Ikke: Or is it necessary to rebuild because of dependencies?

tanertas: yes, rebuild needed due to the new dependencies.

ncopa:
I would have preferred to have each package in separate commit (but still a single PR). Would also have preferred to use `--enable-wayland` to indicate that wayland is explicitly requested, instead of relying on the autodetection.

tanertas:
@ncopa @Ikke  can you check this again?

tanertas:
@ncopa @ikke @andypost  I don't know why this PR closed. I just updated it recently.

Ikke:
Because I pushed it?

tanertas:
Ok, you can postpone this in favor of #7270 and #7272 .Those ones brings meson build system and newer libva.

tanertas:
I removed libva* related commits. Requires #7270 and #7272.

tanertas:
> Ok, you can postpone this in favor of #7270 and #7272 .Those ones brings meson build system and newer libva.

I updated the PR. You can remove R-main and S-hold tags. Can be committed along with #7270 and #7272.

PR #6355, opened 2019-02-16
community/rng-tools: upgrade to 6.7

andypost:
Follow-up to https://github.com/alpinelinux/aports/pull/6282 according https://github.com/nhorman/rng-tools/issues/46


andypost:
Now it needs another dependency
```
checking for libp11... no
configure: error: libp11 is required
```

PR #6358, opened 2019-02-16
community/s-nail: Move from testing

bratkartoffel:
This PR moves the s-nail aport from testing to community. I think this
aport is stable enough to provide stable releases.

Closes https://bugs.alpinelinux.org/issues/9919

Please also backport this change to 3.9

### DIFF
--- a/community/s-nail/APKBUILD
+++ b/community/s-nail/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Steffen Nurpmeso <steffen@sdaoden.eu>
 pkgname=s-nail
 pkgver=14.9.11
-pkgrel=3
+pkgrel=4
 pkgdesc="SysV mail/BSD Mail/POSIX mailx: send and receive Internet mail"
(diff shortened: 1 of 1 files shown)


community/s-nail/APKBUILD
```diff
@@ -15,7 +15,8 @@ subpackages="$pkgname-doc"
 source="https://ftp.sdaoden.eu/s-nail-$pkgver.tar.xz"
 builddir="$srcdir/"$pkgname-$pkgver
 
-build() {
+prepare() {
```
Ikke: Is there a reason you are moving this to prepare, rather than leaving this in build?

bratkartoffel: I've done this for testing the build more quickly, but i'll revert this change.

Ikke:
@bratkartoffel I've applied this to master (edge). To have it backported, please open an issues at https://bugs.alpinelinux.org

PR #6378, opened 2019-02-16
main/busybox-initscripts: Fix startup of syslog

bratkartoffel:
The upgrade to busybox 1.30.0 removed the "-Z" switch from busybox
syslog, replaced by "-t" (see commit 397f0cd902eaadce3b74e1e3021e941d9bb6ba72)

Currently syslog cannot start due to unknown arguments.


bratkartoffel:
Has already been applied with 1e0a736522e5301941563c0c1ab623626a6fedad, thanks

PR #6382, opened 2019-02-17
main/ansible: upgrade to 2.7.8

Lunik:
https://github.com/ansible/ansible/blob/stable-2.7/changelogs/CHANGELOG-v2.7.rst

### DIFF
--- a/main/ansible/APKBUILD
+++ b/main/ansible/APKBUILD
@@ -3,7 +3,7 @@
 # Contributor: Takuya Noguchi <takninnovationresearch@gmail.com>
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=ansible
-pkgver=2.7.0
+pkgver=2.7.7
 pkgrel=1
 pkgdesc="A configuration-management, deployment, task-execution, and multinode orchestration framework"
 url="https://ansible.com"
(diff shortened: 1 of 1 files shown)


main/ansible/APKBUILD
```diff
@@ -38,9 +38,9 @@ package() {
 
 	mkdir -p "$pkgdir"/usr/share/doc/$pkgname/examples/
 	install -m644 examples/* \
-		"$pkgdir"/usr/share/doc/$pkgname/examples/
+		"$pkgdir"/usr/share/doc/$pkgname/examples/ || true
```
andypost: that looks strange

andypost:
2,7,8 is out https://github.com/ansible/ansible/blob/stable-2.7/changelogs/CHANGELOG-v2.7.rst#release-summary

main/ansible/APKBUILD
Lunik: this is strange...
`install` command ignore subdirectories and the return non zero code.
This make the build fail.

Actually, we don't realy need what's inside `examples` but since it was there before, I didn't want to remove it.
So what should we do ?

Lunik:
Also, since there is ansible version 2.8 comming next month [ROADMAP 2.8](https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_8.html) should we wait ?

andypost:
btw 2.7.9 is out makes sense to upgrade to it at least

nmeum:
I will close this as it is superseded by #6763

PR #6390, opened 2019-02-18
testing/mpdecimal: fix ppc64le bld break by adding update_config_guess

mksully22:
On ppc64le build failing with:
configure: error: cannot guess build type; you must specify one

Adding update_config_guess will fix this


mksully22:
http://dup.pw/aports/fe500751 duplicates this and has been integrated so I will close this.

PR #6392, opened 2019-02-18
main/libjpeg-turbo: upgrade to 2.0.2

jbergstroem:
This brings a few changes to the build as a result of upstream switching to cmake.

I removed the patches and references to it in `APKBUILD`. Wasn't sure how to tackle that it referenced "security", but both patches are landed so they're not needed.


jbergstroem:
Guessing this may require some downstream dependency testing. Are there continuous builds for this?

main/libjpeg-turbo/APKBUILD
```diff
@@ -1,52 +1,43 @@
...
-source="https://downloads.sourceforge.net/libjpeg-turbo/libjpeg-turbo-$pkgver.tar.gz
-	0001-tjLoadImage-Fix-FPE-triggered-by-malformed-BMP.patch
-	CVE-2018-11813.patch
-	"
-
-# secfixes:
-#   1.5.3-r3:
-#   - CVE-2018-11813
-#   1.5.3-r2:
-#   - CVE-2018-1152
```
clandmeter: Why are you removing secfixes metadata?

jbergstroem: I couldn't find documentation (looked [here](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference) among other things) on how to cover cases which patches were no longer in the repo. These should stay permanently?

jbergstroem:
Looks like someone else landed a similar bump after this PR was opened. Closing.

PR #6398, opened 2019-02-19
testing/gn: new aport

sassman:
https://gn.googlesource.com/gn/+/refs/heads/master/README.md
GN is a meta-build system that generates build files for Ninja.

### DIFF
--- a/testing/gn/APKBUILD
+++ b/testing/gn/APKBUILD
@@ -1,4 +1,3 @@
-# Contributor:
 # Maintainer: Sven Assmann <sven@d34dl0ck.me>
 pkgname=gn
 pkgver=1530
--- a/testing/gn/APKBUILD
+++ b/testing/gn/APKBUILD
@@ -4,7 +4,7 @@ pkgname=gn
 pkgver=1530
 pkgrel=0
 pkgdesc="GN is a meta-build system that generates build files for Ninja."
-url="https://gn.googlesource.com/gn/+/refs/heads/master/README.md"
+url="https://gn.googlesource.com/gn/+/1ab6fa2c/README.md"
 arch="all"
 license="BSD-3-Clause"
 depends="libgcc musl"
@@ -25,10 +25,6 @@ prepare() {
 build() {
     cd "$builddir/git"
     ninja -C out -j 8
-}
-
-check() {
-    cd "$builddir"
 }
 
 package() {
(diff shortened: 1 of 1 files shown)


testing/gn/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+arch="all"
+license="BSD-3-Clause"
+depends="libgcc musl"
+makedepends="python2 ninja git clang"
+# source="$pkgname-$pkgver.tar.gz::https://gn.googlesource.com/gn/+archive/refs/heads/master.tar.gz"
+giturl="https://gn.googlesource.com/gn"
+builddir="$srcdir"
+
+fetch() {
+    git clone https://gn.googlesource.com/gn "$builddir/git"
```
Ikke: This is not deterministic. Whenever they push commits to master, it would use a different commit. What is the reason you don't use something like https://gn.googlesource.com/gn/+archive/1ab6fa2cab7ec64840db720a56018ca8939329f9.tar.gz as the source so that you know you get the same source every time?

testing/gn/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor:
+# Maintainer: Sven Assmann <sven@d34dl0ck.me>
+pkgname=gn
+pkgver=1530
```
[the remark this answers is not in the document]
sassman: unfortunately they don't tag yet. But I will change the checkout so, that this commit is pinned for this package here.

testing/gn/APKBUILD
sassman: well the tar.gz cannot be used, because their build script needs git, so that it can generate a header file that contains the commit hash that it was build on. 
But as said above, I will make sure I'll pin the commit hash.

testing/gn/APKBUILD
```diff
@@ -0,0 +1,36 @@
+# Contributor:
```
andypost: please remove this useless line

testing/gn/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+    ninja -C out -j 8
+}
+
+check() {
+    cd "$builddir"
+}
+
+package() {
+    install -m755 -D "$builddir/git/out/$pkgname" "$pkgdir/usr/bin/$pkgname"
+}
```
andypost: nitpick, please add new line at the end

testing/gn/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+    "$builddir/git/build/gen.py"
+}
+
+build() {
+    cd "$builddir/git"
+    ninja -C out -j 8
+}
+
+check() {
+    cd "$builddir"
```
andypost: check should do something useful

sassman:
@andypost all your feedback is considered.

PR #6403, opened 2019-02-19
testing/gdal: gdal should enable geos

morlay:
`geos`, `proj4` and `gdal` is used together frequently.
`proj4` could be added `proj4-dev`. but `geos` have to be enabled when compiling.


Ikke:
@morlay The consequence of this is that gdal can only be available on x86(_64), because geos is only available on those platforms. Are you aware of this?

morlay:
Yes. I's good for me. 
But other people may need other platform.
We could try to enable crossing compiling for geos.

PR #6412, opened 2019-02-19
testing/php7-tideways_xhprof: fix x86

andypost:


andypost:
Testing of https://github.com/tideways/php-xhprof-extension/pull/74

PR #6426, opened 2019-02-20
community/glm workaround for upstreambug #832, to allow building kicad

stef:
There is  a bug in glm(https://github.com/g-truc/glm/issues/832) that confuses gcc versions and C++11 support with C++14 support, which prohibits to compile kicad (https://bugs.launchpad.net/kicad/+bug/1804030). This patch reverts the change that introduces this problem.


stef:
upstream reports that this bug has been fixed:
> This issue should be fixed in the master branch for GLM 0.9.9.4 release.
https://github.com/g-truc/glm/issues/832#issuecomment-468612087

PR #6265, opened 2019-02-09
main/dhcp: improve

tcely:
Fixes https://bugs.alpinelinux.org/issues/10161
Fixes https://bugs.alpinelinux.org/issues/10024

Resolves alpinelinux/aports#6265
Resolves alpinelinux/aports#6435

PR #6435, opened 2019-02-21
 main/dhcp: fix missing /var/run/dhcp for dhcrelay

liske:
This fixes issue [#10024](https://bugs.alpinelinux.org/issues/10024).


main/dhcp/dhcrelay.initd
```diff
@@ -20,6 +20,8 @@ start() {
 		IFACES="${IFACES} -i ${i}"
 	done
 
+	checkpath -d -o dhcp:dhcp /var/run/dhcp
```
ncopa: I prefer long options in scripts as they are more readable

liske: @ncopa I've updated the PR to contain long options.

liske:
Can we get this PR merged?

tcely:
> Can we get this PR merged?

I would like to see #6265 merged instead. I agree this needs to be fixed, but merging this first will leave the other fixes out and cause more work for me.

PR #6442, opened 2019-02-22
community/mmc-utils: new aport

oliv3r:
mmc-utils sadly has no tags or releases, there is only the git version.

MMC Utils are useful when working with /dev/mmcblk devices, so mostly useful (embedded) arm

### DIFF
--- a/testing/mmc-utils/APKBUILD
+++ b/testing/mmc-utils/APKBUILD
@@ -1,0 +1,30 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname=mmc-utils
+pkgver="0_git$(date +%Y%m%d)"
+pkgrel=0
+pkgdesc="Configure MMC storage devices from userspace."
+url="http://pyropus.ca/software/memtester/"
+arch="all"
+license="GPL-2.0"
+depends=""
+makedepends="linux-headers"
+options="!check" # No checks available
+subpackages="${pkgname}-doc"
+_githash="aef913e31b659462fe6b9320d241676cba97f67b"
+source="https://git.kernel.org/pub/scm/linux/kernel/git/cjb/mmc-utils.git/snapshot/mmc-utils-${_githash}.tar.gz"
+
+build()
+{
+	cd "${srcdir}/mmc-utils-${_githash}"
+	make
+}
+
+package()
+{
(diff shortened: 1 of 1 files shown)


oliv3r:
I wasn't quite sure how to deal with a git-only package. Looking at a few existing ones, I took inspiration there.

Ikke:
> \>\>\> ERROR: mmc-utils: mmc-utils-0_git20190221.tar.gz is missing in checksums

You need to run abuild checksum to generate the checksums

oliv3r:
> > >>> ERROR: mmc-utils: mmc-utils-0_git20190221.tar.gz is missing in checksums
> 
> You need to run abuild checksum to generate the checksums

Thanks, but you haven't looked at the APKBUILD itself :)

There exists no tar.gz, or rather, the tar.gz is uploaded via/using the snapshot methodology as can be seen in other APKBUILDS. The repo has no releases. The only alternative would be to download a hashed zip, which I do not see anyone doing, they all seem to follow this approach... which works for updating, but the first time, seems to be well that's what I mentioned, I have no idea :)

Ikke:
@oliv3r When there is a hashed archive available, it's prefered and simpler to use that (no need to create a snapshot manually).

Current master would be https://git.kernel.org/pub/scm/linux/kernel/git/cjb/mmc-utils.git/snapshot/mmc-utils-aef913e31b659462fe6b9320d241676cba97f67b.tar.gz

I see that the last known release is 4.3.0, so the version would then be `4.3.0_git20181024`

Why did you call this `mmc-utils` btw?

oliv3r:
I think your mixing my two pr's. But if hash version is preferred ill update the pr. The wiki should probably reflect this as well however.

On February 27, 2019 10:06:10 PM GMT+01:00, Ikke <notifications@github.com> wrote:
>@oliv3r When there is a hashed archive available, it's prefered and
>simpler to use that (no need to create a snapshot manually).
>
>Current master would be
>https://git.kernel.org/pub/scm/linux/kernel/git/cjb/mmc-utils.git/snapshot/mmc-utils-aef913e31b659462fe6b9320d241676cba97f67b.tar.gz
>
>I see that the last known release is 4.3.0, so the version would then
>be `4.3.0_git20181024`
>
>Why did you call this `mmc-utils` btw?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/alpinelinux/aports/pull/6442#issuecomment-468030357

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #6443, opened 2019-02-22
community/memtester: new aport 4.3.0

oliv3r:
While memtester hasn't been updated in quite a while, it is a very common tool to do memory testing from Linux.


Ikke:
Usually new aports start in testing, they don't directly go into community

oliv3r:
Moved ;)

PR #6444, opened 2019-02-22
scripts: add armv7 support to images

oliv3r:
To get the ball to roll on starting with true armv7 releases, I started to modify the build scripts to generate armv7 builds!

### DIFF
--- a/scripts/mkimage.sh
+++ b/scripts/mkimage.sh
@@ -22,7 +22,7 @@ set -e
 all_sections=""
 all_profiles=""
 all_checksums="sha256 sha512"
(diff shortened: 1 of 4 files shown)


oliv3r:
Thanks for the merge, but was hoping we could start the discussion what else is needed to get this moving forward ;) I'll try to see if i can spot other items on the list of course, but the whole build system is a little to get into, as it's scatterd a little (scripts in various places mostly)

oliv3r:
I see we now have minirootfs for 3.9.1 and 3.9.2, is there a way we can retroactively also generate the 3.9.0 minirootfs?

PR #6446, opened 2019-02-23
testing/uboot-tools: upgrade and promote from testing to community

fboudra:


fboudra:
@clandmeter any chance to get it reviewed? poking since you're listed as contributor of the package.

Ikke:
@fboudra Is there an http(s) source? Or is it only available via FTP? Travis currently fails to connect to ftp.

fboudra:
@Ikke looks like http://ftp.denx.de/pub/u-boot/ works even if it isn't documented on the website. updating.

community/uboot-tools/APKBUILD
```diff
@@ -12,7 +12,7 @@ depends_dev=""
 makedepends="$depends_dev bison flex openssl-dev linux-headers sdl-dev"
 install=""
 subpackages=""
-source="ftp://ftp.denx.de/pub/u-boot/u-boot-$pkgver.tar.bz2"
+source="http://ftp.denx.de/pub/u-boot/$pkgname-$pkgver.tar.bz2"
```
tcely: `$pkgname` is `uboot-tools` and this URL needs `u-boot`

community/uboot-tools/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
-# Maintainer:
+# Maintainer: Fathi Boudra <fathi.boudra@linaro.org>
 pkgname=uboot-tools
```
tcely: Consider `_pkgname=u-boot` so that you can use `$_pkgname` in source URL.

community/uboot-tools/APKBUILD
Ikke: Yeah, it does not make sense to use $pkgname here. If we decide to rename the package, the url would still remain the same.

Ikke:
@fboudra I've pushed the changes, so this PR can be closed.

It's failing on s390x and ppc64le because it does not support efi sandboxing:

> error: #error "sandbox EFI support is only supported on ARM and x86"

Do you agree diisabling it on those platforms?

fboudra:
@Ikke yes, it should be disabled on those platforms

Ikke:
@fboudra Done. Feel free to close this PR now.

fboudra:
@Ikke Thansk!

PR #6467, opened 2019-02-24
testing/openjdk9: new aport

hjaekel:
New aport OpenJDK 9. This is a prerequisite for #6468 as OpenJDK 9 is needed as a bootstrap JDK for building OpenJDK 10. 

As OpenJDK 9 does not receive security updates from upstream, this aport should be deleted as soon as OpenJDK 10 is successfully build on all platforms.

I could only test this build on x86_64, I hope that it builds on all other platforms.

PR #6468, opened 2019-02-24
testing/openjdk10: new aport

hjaekel:
New aport OpenJDK 10. This is a prerequisite for #6469 as OpenJDK 10 is needed as a bootstrap JDK for building OpenJDK 11. This PR depends on #6467.

As OpenJDK 10 does not receive security updates from upstream, this aport should be deleted as soon as OpenJDK 11 is successfully build on all platforms.

I could only test this build on x86_64, I hope that it builds on all other platforms.


Ikke:
@hjaekel I'm closing this in favor of #6500 as they have some extra patches. Thanks for your contribution though. As you suggested, openjdk10 will be moved to unmaintained as soon as openjdk11 is bootstrapped.

PR #6470, opened 2019-02-24
Testing/update sopel 6.6.3

Ikke:
This version requires geoip2, replacing geoip as dependency.


testing/py3-maxminddb/APKBUILD
```diff
@@ -0,0 +1,44 @@
+# Contributor: Kevin Daudt <kdaudt@alpinelinux.org>
+# Maintainer: Kevin Daudt <kdaudt@alpinelinux.org>
+pkgname=py3-maxminddb
```
andypost: how that related to upgrade?

Ikke: I forgot to commit one package, which I just pushed.

sopel -> py3-geoip2 -> py3-maxminddb

PR #6474, opened 2019-02-25
main/irqbalance: make numa aware

HRio:


HRio:
@rnalrd please check

ref: https://github.com/alpinelinux/aports/pull/4350

PR #6484, opened 2019-02-26
testing/libguestfs: new aport

fboudra:
Closes: https://bugs.alpinelinux.org/issues/1792

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/testing/libguestfs/APKBUILD
+++ b/testing/libguestfs/APKBUILD
@@ -1,0 +1,118 @@
+# Contributor:
+# Maintainer: Fathi Boudra <fathi.boudra@linaro.org>
+pkgname=libguestfs
+pkgver=1.40.2
+pkgrel=0
+pkgdesc="tools for accessing and modifying virtual machine disk images"
+url="http://libguestfs.org/"
+arch="x86_64"
+license="GPL2"
+depends="cdrkit curl gptfdisk qemu-img qemu-system-x86_64"
+makedepends="
+ acl-dev
+ augeas-dev
+ bash
+ bison
+ cdrkit
+ coreutils
+ cpio
+ file-dev
+ flex
+ fuse-dev
+ gawk
+ gettext-dev
+ hivex-dev
+ jansson-dev
+ gperf
+ gptfdisk
+ libcap-dev
+ libconfig-dev
+ libselinux-dev
+ libselinux-utils
+ libtirpc-dev
+ libvirt-dev
+ libxml2-dev
+ libxml2-utils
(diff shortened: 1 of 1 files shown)


testing/libguestfs/APKBUILD
```diff
@@ -0,0 +1,114 @@
+# Contributor:
+# Maintainer: Fathi Boudra <fathi.boudra@linaro.org>
+pkgname=libguestfs
+pkgver=1.40.2
+pkgrel=0
+pkgdesc="libguestfs"
```
Ikke: Could you improve this description? It should be more then just the package name repeated.

testing/libguestfs/APKBUILD
```diff
@@ -0,0 +1,114 @@
...
+		--disable-lua \
+		--disable-golang \
+		--disable-gobject \
+		--without-java
+	# Building index-parse.c and index-scan.c by hand works around a race
+	# condition in the autotools cruft, where two other more copies of yacc
+	# race with each other, resulting in a corrupted file.
+	make -j1 -C builder index-parse.c
+	make -j1 -C builder index-scan.c
+	make -j$(nproc) V=1 INSTALLDIRS=vendor LTLIBINTL=-lintl
```
Ikke: abuild will use the amount of processes specified in /etc/abuild.conf, you should not specify it here (imagine someone not wanting to use all CPUs when building).

testing/libguestfs/APKBUILD
```diff
@@ -0,0 +1,114 @@
...
+}
+
+check() {
+	cd "$builddir"
+	: make quickcheck
```
Ikke: This is a noop, so effectively no tests are run. Either run tests, or provide !check to options with a comment why checks are skipped.

fboudra:
@Ikke Thanks for the review. All the comments have been addressed.

Wrt tests, I commented but kept check() for reference purpose if someone is interested to run them locally. I'd like to get it fixed by adding supermin dependency (https://github.com/fboudra/aports/commit/6f08b79c3e683823b0817388f13a913a68898a78), however it requires more work since the testsuite isn't distro-agnostic and Alpine isn't a platform officially supported by supermin.
Note: apk package manager support can certainly be contributed to supermin and we'll eventually be able to create Alpine based supermin appliance ;) Unfortunately, it's out of my current scope.

fboudra:
FYI supermin submitted as https://github.com/alpinelinux/aports/pull/6505

PR #6498, opened 2019-02-27
community/x265: upgrade to 3.0

prspkt:


prspkt:
x86 build for `handbrake` failure is due to CI building with `-DARCH_X86_64` instead of `-DARCH_X86_32`. This doesn't happen on the builders.
xpra fails because as of writing, https://xpra.org/ is down.

PR #6504, opened 2019-02-28
testing/nginx-naxsi: upgrade to 1.16.0

J0WI:
@itoffshore I think naxsi itself would be stable enough to merge into the `nginx` package.
What about the other modules?


itoffshore:
@J0WI I personally only use the additional `sysguard` module & will update it to the latest version from tengine as it now has [`sysguard_cpu` added](https://github.com/alibaba/tengine/blob/d271c477f9f8dd18411145853536cc947cc74e0d/CHANGES.te).

The `ngx-cache_purge` module [is stable & funded](http://labs.frickle.com/nginx_ngx_cache_purge/).

I would also like the `anonymise.patch` included if naxsi is moved.

testing/nginx-naxsi/APKBUILD
```diff
@@ -5,8 +5,8 @@
 
 pkgname=nginx-naxsi
 _pkgname=nginx
-pkgver=1.15.5
-pkgrel=1
+pkgver=1.15.9
```
andypost: release 1.15.12 is out

J0WI: rebased

PR #6513, opened 2019-03-01
main/libtirpc: upgrade to 1.1.4

prspkt:


clandmeter:
@prspkt I guess you already found out, but the clone errors from drone are related to histories that are more than 50 commits old (the current clone depth). Rebasing master into your feature branch should solve it. I may add a fix for this later when, nore sure yet. Could be that others also bump into this issue, remind them if needed. Thanks!

prspkt:
> @prspkt I guess you already found out, but the clone errors from drone are related to histories that are more than 50 commits old (the current clone depth). Rebasing master into your feature branch should solve it.

@clandmeter Yes I noticed this behavior in the drone configuration and i think it's fine. A simple rebase is all that's needed.

> @prspkt Could be that others also bump into this issue, remind them if needed. Thanks!

Yep, will do. Thanks for the heads up!

PR #6515, opened 2019-03-01
main/ser2net: upgrade to 3.5.1 release

fboudra:
Upgrade to 3.5.1 release:
 * bump pkver to 3.5.1, reset pkrel to 0
 * use $pkname whenever possible
 * drop TIOCSRS485.patch - not needed. The configure checks
   for TIOCSRS485 and set HAVE_DECL_TIOCSRS485 accordingly
 * get rid of prepare() - not needed anymore
 * use $builddir instead of $_builddir
 * use one-liner for install command calls

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>


main/ser2net/APKBUILD
```diff
@@ -1,35 +1,25 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=ser2net
-pkgver=3.4
-pkgrel=1
+pkgver=3.5.1
+pkgrel=0
 pkgdesc="A proxy that allows telnet/tcp connections to be made to serial ports"
-url="https://sourceforge.net/projects/ser2net"
+url="https://sourceforge.net/projects/$pkgname"
```
ncopa: I prefer to not user `$pkgname` in `url` as I don't think it adds much value. I also often use the `url` to look for new version or find how to report upstream bugs and then i just click on the url. If there is a `$pkgname` there then I will have to copy/paste it to the browser and manually expand it.

main/ser2net/APKBUILD
```diff
@@ -43,25 +33,13 @@ build() {
 }
 
 package() {
-	cd "$_builddir"
+	cd "$builddir"
 	make DESTDIR="$pkgdir" install || return 1
```
ncopa: while you are at it, maybe also remove all the  ` || return 1` from the APKBUILD?

fboudra:
@ncopa Thanks for the review. All the comments have been addressed.

main/ser2net/APKBUILD
```diff
@@ -12,56 +12,33 @@ depends_dev=""
...
-prepare() {
-	local i
-	cd "$_builddir"
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
-}
+source="https://sourceforge.net/projects/$pkgname/files/$pkgname/$pkgname-$pkgver.tar.gz
```
prspkt: please set the `source` variable to: `https://sourceforge.net/projects/ser2net/files/ser2net/$pkgname-$pkgver.tar.gz`

fboudra: for curiosity, what the rationale behind this request? For the comment about url field, I get it, it's convenient for a copy/paste. But for the source field, we already use variables, so what's the reason to use it only for the tarball name?

prspkt: It makes it easier to select the link up to the variables `https://sourceforge.net/projects/ser2net/files/ser2net/` and click go to, instead of copy-pasting the link, replacing the variables and then clicking ok, go etc.
It helps if you want to take a quick look at the distfiles.

fboudra: Thanks for giving some more input on your rationale.

PR #6519, opened 2019-03-01
community/go: update to 1.12.1

J0WI:


nmeum:
Build fails, I upgraded it locally without including the patch and in that case it build fine.

J0WI:
Looks even worse without the patch

nmeum:
Well…fetch fails now so it doesn't even attempt to build it. No idea why though.

nmeum:
Could you rebase your branch against master? This should make the fetch stage pass.

J0WI:
This fails on 32bit arches again.

nmeum:
That's good to know because I only tested it on x86_64 would be probably be a good idea to investigate why this tests passed with go 1.11 but no longer pass with go 1.12. That being sad the current check() for this package is pretty hacky especially since the testsuite is not really intended to be run with a pie default buildmode.

nmeum:
Superseded by 9130ec9f2a50285c35b4cd060788d797cc38b497

PR #6522, opened 2019-03-02
testing/kea: Removed .initd extension from service files

valentinesd:


clandmeter:
Please correct the commit message and bump pkgrel, else it will never be rebuild.

clandmeter:
Nice thats better, thanks!

There seems to be more issues with this apkbuild. I think I will clean it up a bit more.

PR #5609, opened 2018-11-11
non-free/amd-ucode: remove aport

J0WI:
Seems to be covered by `main/linux-firmware`

PR #6523, opened 2019-03-02
Linux firmware

oliv3r:
Start packaging the linux firmware tree.

In true alpine fashion, we don't want one mega archive containing all the firmware. We start with two packages.

For radeon, we only package all radeon firmware, as I know that one card actually can take various bin files, this is years old knowledge, so if someone can debunk that (and thus split the package further) lets do that.

The broadcom tree is big and a little messy, so lets start with atleast one known combination. Btw, the txt files are nvram configuration files, so the intention is to keep them all in the same package (related to the firmware). So in this case, this is not the firmware only for the rasp-pi; it is currently just the conly configuration.


J0WI:
See also:
#5610
#5609

oliv3r:
It is, I did look for it; but come to think of it, I only looked in non-free. Oh and I looked at pkgs.alpinelinux.org for any firmware files but didn't find anything!

Sorry for the noise ...

PR #6526, opened 2019-03-02
testing/cjose: new aport

jbergstroem:
cjose is a C library implementing the Javascript Object Signing and Encryption. It depends on jansson and openssl.

Since this is my first "brand new" aport I have a few q's I'd like some guidance on:

1. Who do I add as maintainer? (I assume I'm the contributor?)
2. `-doc` doesn't seem to package correctly and I'm not sure why. It correctly installs into `$DESTDIR/usr/share/cjose/html` but the corresponding `-doc` package is empty.
3. ~the main package also seems to grow an empty `$DESTDIR/usr/share/cjose` folder.~ edit: not according to CI

Note: this library is used by [apache traffic server](https://trafficserver.apache.org/), which I am working to support in Alpine Linux (as of now, only 9.0.x/HEAD). If you're curious you can follow progress [in this Dockerfile](https://github.com/jbergstroem/ts-alpine).

### DIFF
--- a/testing/cjose/APKBUILD
+++ b/testing/cjose/APKBUILD
@@ -1,0 +1,45 @@
+# Contributor: Johan Bergström <bugs@bergstroem.nu>
+# Maintainer: Johan Bergström <bugs@bergstroem.nu>
+pkgname=cjose
+pkgver=0.6.1
+pkgrel=0
+pkgdesc="A C library implementing the Javascript Object Signing and Encryption"
+url="https://github.com/cisco/cjose"
+arch="all"
+license="MIT"
+depends="jansson openssl"
+makedepends="check-dev coreutils doxygen jansson-dev openssl-dev perl"
+install=""
+subpackages="$pkgname-dev $pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/cisco/$pkgname/archive/$pkgver.tar.gz"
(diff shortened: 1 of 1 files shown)


Ikke:
1. You would be the maintainer, given that you commit to keeping this package up to date and working
2. The [default_doc][0] function only looks in `/usr/share/{doc,man,info,html,sgml,licenses,gtk-doc,ri,help}`, so `/usr/share/cjose` is not moved by default. You can fix this by overriding `doc()`, calling `default_doc`, and then moving `$pkgdir/usr/share/cjose` to `$subpkgdir/usr/share/cjose`.

[0]:https://git.alpinelinux.org/abuild/tree/abuild.in#n1599

jbergstroem:
@ikke said:
> You would be the maintainer, given that you commit to keeping this package up to date and working

Done.

> The default_doc function only looks in `/usr/share/{doc,man,info,html,sgml,licenses,gtk-doc,ri,help}, so /usr/share/cjose` is not moved by default. You can fix this by overriding `doc()`, calling `default_doc`, and then moving `$pkgdir/usr/share/cjose` to `$subpkgdir/usr/share/cjose`.

The problem was actually that the doxygen generation was called via `make install`  - which can be fine - but not installed into `DESTDIR`. Likely upstream build system bug. Will look into it separately.

jbergstroem:
Rebased!

PR #6528, opened 2019-03-02
testing/php7-pecl-grpc: remove incorrect dependency

prspkt:
@andypost Make dependency on protobuf-dev is not required?


andypost:
Good catch, meanwhile there new release

prspkt:
> Good catch, meanwhile there new release
Thanks @andypost !

PR #6530, opened 2019-03-03
community/docker: udpate to 18.09.3

tomalok:
https://github.com/docker/docker-ce/releases/tag/v18.09.3

The more important fixes in this version:
* When copying existing folder, ignore xattr set errors when the target filesystem doesn't support xattr. docker/engine#135
* Graphdriver: fix device mode not being detected if character-device bit is set. docker/engine#160
* Fix nil pointer derefence on failure to connect to containerd. docker/engine#162
* Delete stale containerd object on start failure. docker/engine#154

### DIFF
--- a/community/docker/APKBUILD
+++ b/community/docker/APKBUILD
@@ -2,8 +2,8 @@
 # Maintainer: Jake Buchholz <tomalok@gmail.com>
 
 pkgname=docker
-pkgver=18.09.1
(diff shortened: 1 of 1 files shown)


andypost:
Is there related containerd upgrade?

tomalok:
As of now, `containerd` is still at 1.2.4.  There are a few additional commits to `runc` since commit 6635b4f0c6af3810594d2770f662f34ddc15b40d, but `containerd` is still referencing that one.

PR #6545, opened 2019-03-05
testing/stress-ng: New aport

oliv3r:

### DIFF
--- a/testing/stress-ng/APKBUILD
+++ b/testing/stress-ng/APKBUILD
@@ -1,0 +1,60 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname=stress-ng
+pkgver=0.09.54
+pkgrel=0
+pkgdesc="stress-ng will stress test a computer system in various selectable ways."
+url="https://kernel.ubuntu.com/~cking/stress-ng/"
+arch="all"
+license="GPL-2.0-or-later"
+depends="
+	libc6-compat
+	libgcrypt
+"
+makedepends="
+	attr-dev
+	keyutils-dev
+	libaio-dev
+	libbsd-dev
+	libcap-dev
+	libgcrypt-dev
+	linux-headers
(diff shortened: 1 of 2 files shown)


testing/stress-ng/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+	libgcrypt-dev
+	linux-headers
+	lksctp-tools-dev
+	zlib-dev
+"
+options="!check" # Tests are unreliable with qemu and take a long time
+subpackages="${pkgname}-doc ${pkgname}-examples"
+source="
+	https://kernel.ubuntu.com/~cking/tarballs/${pkgname}/${pkgname}-0.09.54.tar.xz
+
```
Ikke: redundant blank line

testing/stress-ng/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+doc()
+{
+	mkdir -p "${subpkgdir}/usr/share/man/"
+	mv "${pkgdir}/usr/share/man/man1/" "${subpkgdir}/usr/share/man/"
+	rm -r "${pkgdir}/usr/share/man/"
+	rmdir "${pkgdir}/usr/share" || true
+}
+
+examples()
```
Ikke: What is the reason you opted for a separate examples subpackage?

testing/stress-ng/APKBUILD
oliv3r: actually, it is the separator between the source and the patch. I saw a lot of packages making that distinction which, for readability, makes sense.

testing/stress-ng/APKBUILD
oliv3r: Also, I noticed some packages doing that. And in the alpine sense, it makes sense to do so. The goal is to have a minimal amount of packages/clutter installed. For containers/embedded, I totally see that users want only the package. For a desktop (where space isn't a concern) having all the docs is not too bad. The examples you really only ever want is when you are experimenting. So we can drop it all together of course, but I figured, why not?

PR #6546, opened 2019-03-05
testing/mmc-utils: Fix a few small details to the APKBUILD

oliv3r:
There are some copy-paste mistakes in the pkgver and url among others.

This patch addresses those small oversights.

### DIFF
--- a/testing/mmc-utils/APKBUILD
+++ b/testing/mmc-utils/APKBUILD
@@ -1,10 +1,10 @@
 # Contributor: Olliver Schinagl <oliver@schinagl.nl>
 # Maintainer: Olliver Schinagl <oliver@schinagl.nl>
 pkgname=mmc-utils
-pkgver="4.3.0_git20181024"
-pkgrel=0
+pkgver="0_git20181024"
+pkgrel=1
 pkgdesc="Configure MMC storage devices from userspace."
-url="http://pyropus.ca/software/memtester/"
+url="https://git.kernel.org/pub/scm/linux/kernel/git/cjb/mmc-utils.git"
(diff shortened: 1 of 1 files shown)


oliv3r:
Any updates on this?

testing/mmc-utils/APKBUILD
```diff
@@ -1,30 +1,30 @@
 # Contributor: Olliver Schinagl <oliver@schinagl.nl>
 # Maintainer: Olliver Schinagl <oliver@schinagl.nl>
 pkgname=mmc-utils
-pkgver="4.3.0_git20181024"
-pkgrel=0
+pkgver="0_git$(date +%Y%m%d)"
```
Ikke: The package version should be static. If the same package with the same source is built, it should have the same version. Now it will get a new version if it's built on a different day.

oliv3r: Ah I think this is a leftover from when I didn't store the commitish, but just pulled master. So each time you build, you pull a new master, so the date would be new, everytime you build, you get a new hash e.g. new version.

Would be much nicer if apk would support commitish in its version field ...

Ikke: The version field is used to determine whether a package should be updated. A random hash is not beneficial for that.

oliv3r: yes of course, but that does not make it less useful :p you want the date to be incremental, rather then the hash; that's completely understandable but the date also only tells you so much, anyway, that would be a generic alpine discussion, for a different thread. I've updated the PR as requested ;)

PR #6553, opened 2019-03-06
community/knot: upgrade to 2.8.0

tcely:
Also, patch `knot-resolver` in #6558 to not break when building against `knot-2.8.0`.

~~Depends on #6554~~

For https://bugs.alpinelinux.org/issues/10066


tcely:
Shouldn't TravisCI be able to make use of packages it just built for later builds?

PR #6562, opened 2019-03-06
main/openssl: security update to patch CVE-2019-1543

J0WI:
This needs to be backported to 3.9. 3.8 and below are not affected.
https://www.openssl.org/news/secadv/20190306.txt


J0WI:
This has been fixed in 4bb02b09003b63e5aa9c4d9be68117e15e663c00, but it is not yet backported to 3.9.

PR #2757, opened 2017-11-17
testing/cockroach: new aport

vincentmele:
https://www.cockroachlabs.com
An open source, survivable, strongly consistent, scale-out SQL database.

PR #6564, opened 2019-03-06
testing/cockroach: new aport

joeholden:
Currently x86_64 only, requires further fixes for other architectures however a new release is due relatively soon.

### DIFF
--- a/testing/cockroach/APKBUILD
+++ b/testing/cockroach/APKBUILD
@@ -1,0 +1,38 @@
+# Contributor: Joe Holden <jwh@zorins.us>
+# Maintainer: Joe Holden <jwh@zorins.us>
+pkgname=cockroach
+pkgver=2.1.6
+pkgrel=0
+pkgdesc="The open source, cloud-native SQL database."
+url="https://cockroachlabs.com"
+arch="x86_64"
+license="Apache-2.0"
+makedepends="go bash cmake autoconf linux-headers ncurses-dev"
+install="$pkgname.pre-install"
+pkgusers="cockroach"
+pkggroups="cockroach"
+subpackages="$pkgname-openrc"
+options="!check"
+source="${pkgname}-v${pkgver}.tar.gz::https://binaries.cockroachdb.com/${pkgname}-v${pkgver}.src.tgz
+	cockroach.initd
+	cockroach.confd
+	01_no_githooks.patch"
+
+builddir=${srcdir}/${pkgname}-v${pkgver}
+
+build() {
+	cd ${builddir}
+	make CXXFLAGS="-D_BSD_SOURCE" CFLAGS="-D_BSD_SOURCE" TAGS="stdmalloc" BUILDTYPE="release" buildoss
+}
+
+package() {
+	cd ${builddir}
+
+	install -m755 -D ${srcdir}/${pkgname}.initd ${pkgdir}/etc/init.d/${pkgname}
+	install -m644 -D ${srcdir}/${pkgname}.confd ${pkgdir}/etc/conf.d/${pkgname}
+	install -m750 -o ${pkgusers} -g ${pkggroups} -D src/github.com/cockroachdb/${pkgname}/${pkgname}oss ${pkgdir}/usr/sbin/${pkgname}
+}
+sha512sums="13ced601cc840a13fa705d9678e1413635f23a5734a6a5876109d080f23f7d94418c5378e7e0d31ac24201019ee20d52a3d30391d640293dce1ecaae799bc548  cockroach-v2.1.6.tar.gz
+94264601b5b8516d87072017d40ba293d17320c706ad7a2e3c49f1af2b1030071e33cf3353adce26c634d284d0be4e41f2f9a9749c5071e9f0921ed03429f4cf  cockroach.initd
+7039d17eb3c251d941bc73af2264c9618b59f8301165764371bf1539c46366d35496b2c00ea134731fd1ca9a0b362f9ba331b03aa62339f75e14504af9eb44f9  cockroach.confd
+fe611479ba32a01a1ee7770170c54570cb4ac558fb756c44c90f9e31d59a5da74af81adf1e100e723d3956aea96748955919e5b000ee91d8f8fcb26cc678556f  01_no_githooks.patch"
--- a/testing/cockroach/cockroach.confd
+++ b/testing/cockroach/cockroach.confd
@@ -1,0 +1,1 @@
+cockroach_opts="start --insecure --store=/var/lib/cockroach"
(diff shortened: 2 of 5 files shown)


testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+makedepends="go bash cmake autoconf linux-headers ncurses-dev"
+install="$pkgname.pre-install"
+pkgusers="$pkgname"
+pkggroups="$pkgname"
+subpackages="$pkgname-openrc"
+options="!check"
+source="${pkgname}-${pkgver}.tar.gz::https://binaries.cockroachdb.com/${pkgname}-${_pkgver}.src.tgz
+        cockroach.initd
+        cockroach.confd
+	01_no_githooks.patch"
```
TBK: Issue with the indentation (should be tab).

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Contributor: Joe Holden <jwh@zorins.us>
+# Maintainer: Joe Holden <jwh@zorins.us>
+pkgname=cockroach
+pkgver=2.1.5
+_pkgver=v2.1.5
+pkgrel=1
+pkgdesc="The open source, cloud-native SQL database."
+url="https://cockroachlabs.com"
+arch="x86_64"
+license="APL2"
```
TBK: I am guessing the license is Apache-2.0. Please use the SPDX naming scheme https://spdx.org/licenses/

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,42 @@
+# Contributor: Joe Holden <jwh@zorins.us>
+# Maintainer: Joe Holden <jwh@zorins.us>
+pkgname=cockroach
+pkgver=2.1.5
+_pkgver=v2.1.5
```
TBK: No need for two version variables. Just add the v to the source url:
```
$pkgname-$pkgver.tar.gz::https://binaries.cockroachdb.com/$pkgname-v$pkgver.src.tgz
```

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+subpackages="$pkgname-openrc"
+options="!check"
+source="${pkgname}-${pkgver}.tar.gz::https://binaries.cockroachdb.com/${pkgname}-${_pkgver}.src.tgz
+        cockroach.initd
+        cockroach.confd
+	01_no_githooks.patch"
+
+prepare() {
+	cd ${srcdir}/${pkgname}-${_pkgver}
+	patch -p1 -N -i ${srcdir}/01_no_githooks.patch
```
TBK: Drop these two lines and replace it with:

```
prepare(){
	default_prepare
}
```

Use the builddir variable instead of `cd ${srcdir}/${pkgname}-${_pkgver}` - https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#builddir

TBK:
Related stalled aport PRs:
https://github.com/alpinelinux/aports/pull/2757
https://github.com/alpinelinux/aports/pull/1183

testing/cockroach/APKBUILD
joeholden: Theres a commit missing here, prepare() was previously a mess... I was going to leave it until the new release is out but it doesn't look like they're as close as I though they were - I'll push the correct commit in a bit.

TBK: Was half asleep when I wrote my comment. You can drop prepare(), it is run by default and only explicitly needed when you need to do some extra stuff.

testing/cockroach/APKBUILD
joeholden: Ah, just noticed that - was left over from pre-release testing, I'll nuke that too

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+options="!check"
+source="${pkgname}-v${pkgver}.tar.gz::https://binaries.cockroachdb.com/${pkgname}-v${pkgver}.src.tgz
+        cockroach.initd
+        cockroach.confd
+	01_no_githooks.patch"
+
+builddir=${srcdir}/${pkgname}-v${pkgver}
+
+build() {
+	cd ${srcdir}/${pkgname}-v${pkgver}
```
TBK: should be cd "$builddir"

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+	make CXXFLAGS="-D_BSD_SOURCE" CFLAGS="-D_BSD_SOURCE" TAGS="stdmalloc" BUILDTYPE="release" buildoss
+}
+
+package() {
+	cd ${srcdir}/${pkgname}-v${pkgver}
```
TBK: should be `cd "$builddir"`

joeholden: I just mentioned this on IRC, for simple situations abuild should really figure it out on it's own if the builddir matches the source name, builddir is only there to appease abuild and honestly this PR is already about 40 commits (before squashing) to many

testing/cockroach/APKBUILD
joeholden: Indeed it should, fixed

testing/cockroach/APKBUILD
joeholden: Yup, changed

testing/cockroach/APKBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgname=cockroach
+pkgver=2.1.6
+pkgrel=0
+pkgdesc="The open source, cloud-native SQL database."
+url="https://cockroachlabs.com"
+arch="x86_64"
+license="Apache-2.0"
+makedepends="go bash cmake autoconf linux-headers ncurses-dev"
+install="$pkgname.pre-install"
+pkgusers="$pkgname"
```
Ikke: I'm not sure if it makes sense to let this depend on the package name. If the package would be renamed, or someone creates a specialized version of the package, it should probably still use the same user/group.

As the user/group are static in the pre-install script anyway, it makes sense to make them static here as well.w

testing/cockroach/cockroach.confd
```diff
@@ -0,0 +1 @@
+cockroach_opts="start --insecure --store=/var/lib/cockroach"
```
Ikke: `start` is not really optional, is it? I think that can better be moved to the initd `command_args`.

testing/cockroach/APKBUILD
joeholden: Updated in my last push

testing/cockroach/cockroach.confd
joeholden: Well, I put it there rather than command_args so that the 'init' command for initialising a cluster rather than starting a standalone node can easily be substituted without touching init scripts.

Ikke: Alright, makes sense.

PR #6565, opened 2019-03-06
main/linux-vanilla: enable CONFIG_CRYPTO_USER config settings

mksully22:
CONFIG_CRYPTO_USER* kernel build options should be enabled for ppc64le (like all other architectures). Without these settings the testing/ell package will fail the test-uuid test because the call to 
socket(PF_ALG, SOCK_SEQPACKET | SOCK_CLOEXEC, 0); will fail with errno=97 (EAFNOSUPPORT)


mksully22:
Travis checks failed due to time limit.

PR #6568, opened 2019-03-06
testing/tuptime: Add install line to APKBUILD

rfrail3:
Up until now, the .pre-install and .post-install scripts aren't executed, so the package didn't work. This fix the issue adding the line on APKBUILD.


ncopa:
you need bump `pkgrel` or the package will not be rebuilt.

rfrail3:
Thanks @ncopa it continues on #6587

PR #6584, opened 2019-03-08
community/salt: upgrade to 2019.2

J0WI:


Ikke:
@J0WI I noticed that this package does not have a maintainer. Are you willing to maintain this package?

Ikke:
abuild gives quite some warnings:

> \>>> WARNING: salt-master*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc
\>>> WARNING: salt: APKBUILD does not run any tests!

So it needs to be modernized.

J0WI:
Fixed in 0b7c193a2f49aa4605403e307421eb663b777364

PR #6587, opened 2019-03-08
testing/tuptime: right .pre-install execution

rfrail3:
Proper .pre-install execution

### DIFF
--- a/testing/tuptime/APKBUILD
+++ b/testing/tuptime/APKBUILD
@@ -2,18 +2,18 @@
 # Maintainer: Roberto Oliveira <robertoguimaraes8@gmail.com>
 pkgname=tuptime
 pkgver=3.4.2
-pkgrel=2
+pkgrel=3
 pkgdesc="Report the historical and statistical real time of the system"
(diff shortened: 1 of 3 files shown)


Ikke:
@rfrail3 In the future, it's not necessary to create a new pull request to update it. You can either rewrite / amend the current commit and force push it, or if you're not comfortable with that, just add a new commit with the fix, and push that to you branch and the PR will automatically be updated.

testing/tuptime/APKBUILD
```diff
@@ -2,18 +2,18 @@
...
 pkgname=tuptime
 pkgver=3.4.2
-pkgrel=2
+pkgrel=3
 pkgdesc="Report the historical and statistical real time of the system"
 url="https://github.com/rfrail3/tuptime"
 arch="noarch"
 license="GPL-2.0-only"
 depends="python3"
+install="$pkgname-openrc.pre-install"
```
Ikke: I'll fix this, but abuild warns:
> \>>> WARNING: tuptime: You should not have $install in source

PR #6595, opened 2019-03-08
testing/libzookeeper: fix ppc64le build error by disabling Werror

mksully22:
With gcc 8.2 ppc64le build encounters error:
src/zookeeper.c: In function 'format_endpoint_info':
src/zookeeper.c:3506:21: error: '%d' directive writing between 1 and 5 bytes into a region of size between 0 and 127 [-Werror=format-overflow=]
     sprintf(buf,"%s:%d",addrstr,ntohs(port));

Disabling Werror during compile will fix.

### DIFF
--- a/testing/libzookeeper/APKBUILD
+++ b/testing/libzookeeper/APKBUILD
@@ -2,14 +2,15 @@
 # Maintainer: Maciej Klak <klak.maciej@gmail.com>
 pkgname=libzookeeper
 pkgver=3.4.13
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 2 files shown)


Ikke:
Is that an actual fix, or just silencing the error?

mksully22:
just silencing the error

mksully22:
Instead of silencing the impacts of the Warning on current gcc levels an update to provide an actual fix is provided.

Ikke:
I like that a lot better, thanks.

PR #6596, opened 2019-03-09
Postfix: Added ICU in order to enable UTF8 support.

tatodorov:
Added icu-dev to the list of make dependencies which is a prerequisite for UTF8 support in Postfix.
This change resolves the following warning message:

warning: smtputf8_enable is true, but EAI support is not compiled in


tatodorov:
Already implemented in PR #6631

PR #6604, opened 2019-03-09
main/mini_httpd: upgrade to 1.30

TBK:
@ncopa fixes CVE-2018-18778


nmeum:
> ACME mini_httpd before 1.30 lets remote users read arbitrary files.

Will merge the upgrade to fix this in edge, but the CVE fix should probably be backported to the stable branches.

PR #6607, opened 2019-03-10
testing/amiitool: new aport

TBK:


testing/amiitool/APKBUILD
```diff
@@ -0,0 +1,25 @@
+# Contributor: TBK <alpine@jjtc.eu>
+# Maintainer: TBK <alpine@jjtc.eu>
+pkgname=amiitool
+pkgver=r2
```
Ikke: Sadly this does not work as expected with apk, as it treats it as a postfix (like a pre-release). So it's best to just use `2` as the version and then prepend the 'r' where necessary.

TBK: Roger, I already suggested upstream to switch to semver.

Will fix asap.

Ikke: You forgot to put the 'r' back in the url, it now 404's

TBK: doh, too many balls in the air. I have made the corrections.

PR #6608, opened 2019-03-10
 testing/sdfat: new aport

TBK:
https://github.com/TBK/kernel-sdfat
FAT12/16/32(VFAT)/64(exFAT) filesytems kernel module

### DIFF
--- a/testing/sdfat/APKBUILD
+++ b/testing/sdfat/APKBUILD
@@ -1,0 +1,63 @@
+# Maintainer: TBK <alpine@jjtc.eu>
+# Contributor: TBK <alpine@jjtc.eu>
+
+# when changing _ver we *must* bump _rel
+_name=sdfat
+_ver=2.1.8.2
+_rel=0
+
+_flavor=${FLAVOR:-vanilla}
+_kpkg=linux-$_flavor
+_kver=4.19.48
+_krel=0
+
+_kpkgver="$_kver-r$_krel"
+_kabi="$_kver-$_krel-$_flavor"
+
+pkgname=$_name-$_flavor
+pkgver=$_kver
+pkgrel=$(( $_krel + $_rel ))
+
+pkgdesc="FAT12/16/32(VFAT)/64(exFAT) filesytems kernel module"
+url="https://github.com/TBK/kernel-sdfat"
+arch="all"
+license="GPL-2.0-or-later"
+options="!check" # no test suite
+depends="$_kpkg=$_kpkgver util-linux"
+makedepends="$_kpkg-dev=$_kpkgver linux-headers"
+install_if="$_kpkg=$_kpkgver $_name"
+source="$_name-$_ver.tar.gz::https://github.com/TBK/kernel-sdfat/archive/$_ver.tar.gz
+	mount.exfat
+	"
+builddir="$srcdir/kernel-$_name-$_ver"
+
+prepare() {
+	default_prepare
+	local _kapkbuild=../../main/linux-$_flavor/APKBUILD
+	if [ -f $_kapkbuild ]; then
+		(	. $_kapkbuild
+			pkgname=$_name-$_flavor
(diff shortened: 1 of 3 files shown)


TBK:
CIs fails because the pkgs are build in the wrong order. I do not know how to fix it...

TBK:
Commit order does nothing :/

Ikke:
@TBK 

Trying to build sdfat in an lxc container gives the following error:

> make[1]: *** /lib/modules/4.9.65/build: No such file or directory.  Stop.

So it seems to depend on a linux-$flavor package (which is not installed in an lxc contian

TBK:
> 
> 
> @TBK
> 
> Trying to build sdfat in an lxc container gives the following error:
> 
> > make[1]: *** /lib/modules/4.9.65/build: No such file or directory.  Stop.
> 
> So it seems to depend on a linux-$flavor package (which is not installed in an lxc contian

I made the exfat-utils-sdfat depend on the sdfat package:
```
depends="sdfat-${FLAVOR:-vanilla}"
```

Ikke:
Yes, so I try to build sdfat first, but that one already fails..

TBK:
The template I used for the sdfat package is a copy of the one used for all of the kernel module packages. 

Maybe you could get vanilla - https://pkgs.alpinelinux.org/packages?name=linux-*&branch=edge&arch=x86_64 and set FLAVOR=vanilla as an env variable.

TBK:
@ncopa Could I get you to review this?

Ikke:
I tried it again, but it still does not work. It's looking for `lib/modules/$(uname -r)/`, which does not exist. Two issues I see:

1. The modules path is `$(uname -r)-<rel>-<flavor>`.
2. The running kernel may be different from the installed modules. Especially in the build system, where the installed linux-<flavor> version may be different from the kernel running on the host. In my case, `uname -r` returns '4.9.65', while `/lib/modules/` contains '4.19.41-0-vanilla'

Hope this helps a bit.

TBK:
@Ikke It builds cleanly now. So could I get you to review it again?

testing/sdfat/APKBUILD
```diff
@@ -0,0 +1,63 @@
...
+
+pkgname=$_name-$_flavor
+pkgver=$_kver
+pkgrel=$(( $_krel + $_rel ))
+
+pkgdesc="FAT12/16/32(VFAT)/64(exFAT) filesytems kernel module"
+url="https://github.com/TBK/kernel-sdfat"
+arch="all"
+license="GPL-2.0-or-later"
+options="!check"
```
tcely: Please add a comment about why `!check` is used.

Ikke:
@tbk It works for me with this patch, to match the current available kernel version:

```diff
diff --git a/testing/sdfat/APKBUILD b/testing/sdfat/APKBUILD
index 2de954ea14..d80e24ad65 100644
--- a/testing/sdfat/APKBUILD
+++ b/testing/sdfat/APKBUILD
@@ -8,7 +8,7 @@ _rel=0

 _flavor=${FLAVOR:-vanilla}
 _kpkg=linux-$_flavor
-_kver=4.19.45
+_kver=4.19.48
 _krel=0

 _kpkgver="$_kver-r$_krel"
```

Does that mean we need to make sure this gets updated along with any kernel upgrades?

TBK:
> @TBK It works for me with this patch, to match the current available kernel version:

@Ikke Thanks. I apparently missed the last kernel bump. Now fixed.

> Does that mean we need to make sure this gets updated along with any kernel upgrades?

Yes, just as we do with all other kernel modules - zfs, wireguard... - https://pkgs.alpinelinux.org/packages?name=*-vanilla&branch=edge&arch=x86_64

Ikke:
The kernel has been updated in the mean time to `4.19.49`, so I updated the commit to use that version.

TBK:
@Ikke thanks :+1:

PR #6619, opened 2019-03-10
community/quassel: upgrade to 0.13.1

J0WI:


Ikke:
@J0WI This also needs an -openrc subpkg

J0WI:
Are the necessary steps somewhere documented?

Ikke:
@J0WI 

There is a default_openrc function, so usually it's just enough to declare `$pkgname-openrc` as a subpackage. But as this is a meta package with a core and common components, I was not sure if the default dependencies would make sense.

J0WI:
Fixed in 2c1d00d4f2e00bd37f8e35b7aea97d9f2fd5ead9

PR #6620, opened 2019-03-10
main/postfix: upgrade to 3.4.2

lexsca:
From postfix-announce mailing list:

>[An on-line version of this announcement will be available at
>http://www.postfix.org/announcements/postfix-3.4.2.html]
>
>Postfix stable release 3.4.2 is available.
>  * DANE trust anchor file support was broken after the Postfix 3.4
>    TLS library overhaul. Fix by Scott Kitterman.
>
>  * LINUX5 is supported, based on sanity checks with a Rawhide
>    prerelease.
>
>  * Fixes for two null pointer reads in warnings for errors that
>    never happen (in code from 7 and 18 years ago, respectively).
>
>The plan is to maintain zero code difference between the stable
>release and the development release, until there have been no
>bug-of-the-week patches for much more than several weeks.

### DIFF
--- a/main/postfix/APKBUILD
+++ b/main/postfix/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=postfix
-pkgver=3.4.1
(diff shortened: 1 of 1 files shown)


lexsca:
the digest changed?!  closing this for now :(

PR #6631, opened 2019-03-11
main/postfix: upgrade to 3.4.3

lexsca:
ref [http://www.postfix.org/announcements/postfix-3.4.3.html](http://www.postfix.org/announcements/postfix-3.4.3.html)

this also includes `icu-dev` lib to support `smtputf8_enable` as in PR #6596


lexsca:
this may be obsolete because of PR #6631

andypost:
Closing for #6676

PR #6632, opened 2019-03-11
main/openrc: use tod as hwclock on s390x

tuan-hoang1:


main/openrc/use-tod-as-clocksource-for-s390x.patch
```diff
@@ -0,0 +1,37 @@
...
++	: $(( retval += $? ))
+ 
+ 	eend $retval "Failed to set the system clock"
+ 
+@@ -131,7 +133,7 @@ start()
+ stop()
+ {
+ 	# Don't tweak the hardware clock on LiveCD halt.
+-	[ -n "$CDBOOT" ] && return 0
++	[ -n "$CDBOOT" ] || [ "$(uname -m)" = "s390x" ] && return 0
```
ncopa: Can we change this to an `if ...; then ...` statement instead? Its more readable.

ncopa:
I think this should be suggested for upstream openrc. https://github.com/OpenRC/openrc

tuan-hoang1:
Okay. Will do. Thanks

tuan-hoang1:
As I comment in openrc thread, I just figured out s390x needs neither hwclock nor swclock.

hwclock is enabled already and it returns immediately on s390x, which is harmless.

If the hwclock.init message is annoying enough, we will disable hwclock in the future :)
Closing.

PR #6636, opened 2019-03-11
community/fstrm: upgrade to 0.5.0

tcely:


Ikke:
@tcely You enabled s390x, but it's still failing to build there (2 tests failing)

Any reason you enabled s390x?

tcely:
Blind hope that the new version would pass.

PR #3321, opened 2018-02-15
testing/dnsdist: new aport

tcely:
https://dnsdist.org
dnsdist is a highly DNS-, DoS-, and abuse-aware loadbalancer.

This is a reworked version of #2755

Depends on #3965

PR #6638, opened 2019-03-11
 community/dnsdist: move from testing

tcely:
Reported the armhf test failures / slowness upstream: https://github.com/PowerDNS/pdns/issues/7590


community/dnsdist/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: Vince Mele <vmele@inoc.com>
-# Maintainer: Vince Mele <vmele@inoc.com>
```
Ikke: Did you contact Vince about this?

tcely: Not since the first PR they submitted. See https://github.com/alpinelinux/aports/pull/3321

tcely:
Would it be possible to get `test-suite.log` from `edge-armhf` so that I can see why the test failed?

Ikke:
@tcely Is this ready to be reviewed / pushed?

tcely:
@Ikke Yes.

I would prefer the DroneCI build didn't time out while testing, but since we didn't get an answer about why the test is slow and I can't increase the time out I can't do more.

This was supposed to be a simple move. Exposing armhf test slowness is beside the point.

andypost:
I takes 10mins in median so looks good

tcely:
@nmeum @Ikke @andypost 

Can this be merged now?

PR #6652, opened 2019-03-12
main/jsoncpp: add static library

j8r:
I am not sure if we create `libjsoncpp-static`. `libjsoncpp.a` weights ~420000 KB


j8r:
Any update?

ncopa:
we need to bump pkgrel or the builder will not build new package of it. I think it would also make sense to have it as a `jsoncpp-static` subpackage.

ncopa:
I improved it a bit by use `--default-library both` to avoid build it twice in different build dirs. 4b77b1489cc2fb839d9c3de341f3037cc3a1e758

PR #6656, opened 2019-03-13
main/zfs: import pool before swap

HRio:
Just like for lvm, we need to import the zfs pool before swap is started


HRio:
@ncopa thanks, now fixed.

HRio:
@ncopa now merged upstream https://github.com/zfsonlinux/zfs/commit/c742bf1e68be582fb2db029e38768f8450d2609d

PR #6658, opened 2019-03-13
main/dovecot: Upgrade Dovecot to 2.3.5 and Pigeonhole to 0.5.5.

tatodorov:
References:
https://dovecot.org/download.html
https://pigeonhole.dovecot.org/download.html


tatodorov:
The Travis build fails due to "The job exceeded the maximum log length, and has been terminated.".
According to the description on the following link, the job will be terminated if the log exceeds 4 MB.

https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded

When building the package on my machine, the log size is 4.1 MB and it is not due to repeated exception.
It is just more verbose.

J0WI:
This is already fixed with b37a739e81ca9a8962a75c6a3675c5a20b86ae11

PR #6618, opened 2019-03-10
main/postgresql: upgrade to 11.2

J0WI:
https://www.postgresql.org/about/news/1920/

PR #6659, opened 2019-03-13
main/postgresql: Upgrade to PostgreSQL 11.2

tatodorov:
Reference:
https://www.postgresql.org/ftp/source/v11.2/


tatodorov:
The Travis build fails due to "The job exceeded the maximum log length, and has been terminated.".
According to the description on the following link, the job will be terminated if the log exceeds 4 MB.

https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded

This is not caused by an exception.
The logging is too verbose.

andypost:
it upgrades dovecot so one commit should be removed and duplicates https://github.com/alpinelinux/aports/pull/6618

PR #6660, opened 2019-03-13
main/pgpool: upgrade to 3.7.9

J0WI:


tcely:
http://www.pgpool.net/docs/latest/en/html/release-3-7-9.html

J0WI:
Do you know what BSD license they use?
https://pgpool.net/mediawiki/index.php/pgpool-II_License

maxice8:
It doesn't appear to be any license, i think you should use custom:

The 2 closest matching licenses are 0BSD and MIT both of which have Warranty disclaimers that this one does not have.

PR #6676, opened 2019-03-15
main/postfix: upgrade to 3.4.4

lexsca:
ref [http://www.postfix.org/announcements/postfix-3.4.4.html](http://www.postfix.org/announcements/postfix-3.4.4.html)

this also includes `icu-dev` lib to support `smtputf8_enable` as in PR #6596


lexsca:
the 3.4 train of postfix had a rough start, but this version seems stable.  any objections to getting it merged?

lexsca:
spoke too soon.  closing this pr in favor of #6894 with postfix 3.4.5. hopefully this is the last new postfix for a while.

PR #6711, opened 2019-03-16
main/cryptsetup: upgrade to 2.1.0

nmeum:
ChangeLog: https://gitlab.com/cryptsetup/cryptsetup/blob/master/docs/v2.1.0-ReleaseNotes

The most important change probably is the following:

> Cryptsetup 2.1 version uses a new on-disk LUKS2 format as the default LUKS format and increases default LUKS2 header size.

### DIFF
--- a/main/cryptsetup/APKBUILD
+++ b/main/cryptsetup/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=cryptsetup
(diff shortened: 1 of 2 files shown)


nmeum:
@ncopa any chance you could take a look at this soonish? I've been using this with my LUKS2 based setup for the past month and didn't ran into any issues. The only reason I haven't merged this myself yet is that I don't use LUKS1.

The changelog also states

> When upgrading a stable distribution, please use configure option --with-default-luks-format=LUKS1 to maintain backward compatibility.

Didn't do that though as I don't think that Alpine Linux Edge would qualify as a "stable distribution. Not sure what your opinion on this is.

PR #6716, opened 2019-03-17
main/alpine-baselayout: Properly initialize TZ

5paceToast:
Patch 1 in fixing setup-alpine's TZ handling.
Second patch is against alpine-conf.


5paceToast:
alpine-conf PR is alpinelinux/alpine-conf#21

5paceToast:
I would rather keep it, for instances of people that do not use alpine-conf and may not (yet) be aware that docs exist.
The file is self-documenting in a way.

ncopa:
i bumping pkgver and removing the last line in commit message. Thanks!

PR #6719, opened 2019-03-17
[WIP] community/go: use GOFLAGS instead of default-buildmode-pie.patch

nmeum:
This is an attempt at removing `default-buildmode-pie.patch` from our go aport while still building all go binaries as PIEs by default on the builders. This should make less tests of the testsuite fail and should ease upgrading go in the future. Additionally, it fixes the `loadinternal: cannot find runtime/cgo` warning currently occasionally emitted by the current version of our go compiler.

Removing `default-buildmode-pie.patch` has been achieved by setting the the `GOFLAGS` environment variable in `/etc/abuild.conf` to `GOFLAGS=-buildmode=pie`. This work similar to `CFLAGS`.

Drawbacks:

* The go binary itself is no longer a PIE
* Software compiled by users is no longer compiled with `-buildmode=pie` unless the users explicitly adds `GOFLAGS=-buildmode=pie` to the environment as well

TODO:

* [ ] Enable tests on x86
* [x] Upgrade `community/go` to 1.12.1


kaey:
1. You can build go with pie if you pass `GO_LDFLAGS=-buildmode=pie` to make.bash
2. instead of patching abuild it's better to send a pr to it directly https://gitlab.alpinelinux.org/alpine/abuild
3. Not allowing APKBUILD maintainers to modify GOFLAGS needs a lint rule in https://gitlab.alpinelinux.org/Leo/atools
4. Support for drone can be dropped is I understand, as project is switching to gitlab

nmeum:
> You can build go with pie if you pass `GO_LDFLAGS=-buildmode=pie` to make.bash

Thanks, good to know.

> Instead of patching abuild it's better to send a pr to it directly

Indeed, but this specific patch needs to be backported.

> Not allowing APKBUILD maintainers to modify GOFLAGS needs a lint rule in https://gitlab.alpinelinux.org/Leo/atools

atools isn't an official linting tool, but as soon as I get around to start working on this again it will be good to add indeed.

> Support for drone can be dropped is I understand, as project is switching to gitlab

This is currently being discussed, see https://lists.alpinelinux.org/~alpine/devel/%3C3OHWOOASIXH2T.20R4V2YVU7P8Y%408pit.net%3E

---

Also closing this (horribly outdated), will continue working on it locally.

PR #6723, opened 2019-03-17
testing/parso: new aport

Ikke:


Ikke:
Parso is required as checkdepends for py-jedi. The test-suite for parso requires tox, but tox does not run at the moment because py-virtualenv is missing as a dependency.

Ikke:
The ifrst commit is required to get the 2nd commit to work, but CI doesn't pick that up.

testing/py-parso/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends="python3"
+source="https://files.pythonhosted.org/packages/source/p/parso/parso-$pkgver.tar.gz"
+builddir="$srcdir/parso-$pkgver"
+
+build() {
+	python3 setup.py build
+}
+
+check() {
+	tox -e py36 # This should match the current python version
```
clandmeter: You could try something like so it will work on upcoming 3.7 upgrade.
```
local _pyver=$(python3 -c 'import sys; print(str(sys.version_info[0])+str(sys.version_info[1]))')
tox -e py$_pyver
```

Ikke: Thanks, was indeed thinking about how to solve that.

Ikke:
#6740 has been pushed. Once the package is available, the CI tests for this PR should pass.

clandmeter:
Is this now fixed? CI doesn't want to build due to branch is out of date.

andypost:
> pkg_resources.DistributionNotFound: The 'six<2,>=1.0.0' distribution was not found and is required by tox

Ikke:
@clandmeter It's still waiting for #6759, which I found out later.

andypost:
@ikke please rebase

PR #6724, opened 2019-03-17
community/go: upgrade to 1.12.1

clstb:
https://golang.org/doc/devel/release.html#go1.12.minor

### DIFF
--- a/community/go/APKBUILD
+++ b/community/go/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=go
-pkgver=1.11.5
+pkgver=1.12.1
 pkgrel=0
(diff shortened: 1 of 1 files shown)


andypost:
armfhf/armv7 fails on test
```
# plugin.test
/home/buildozer/drone/aports/community/go/src/go/pkg/tool/linux_arm/link: running gcc failed: exit status 1
collect2: fatal error: cannot find 'ld' compilation terminated. 
```

clstb:
Thanks, I am aware. Will investigate soon.

ncopa:
this is a duplicate of #6519

nmeum:
The problem seems to be that go uses gold ld on ARM and ARM64. See https://github.com/golang/go/blob/master/src/cmd/link/internal/ld/lib.go#L1228-L1247 gold is packaged in a binutils subpackage and not installed by default. Adding binutils-gold to checkdepends fixes this.

See: #6838 

The only remaining question is whether binutils-gold should also be added to depends instead of checkdepends on ARM*.

PR #4516, opened 2018-06-15
Bulk change: Update urls to https using HTTPS Everywhere

J0WI:
Not sure if you accept bulk changes, but this can be reproduced by using [this script](https://github.com/EFForg/https-everywhere/tree/4117544090de90167069ecfab1a85dc1eac33303/utils/rewriter), applying [this patch](https://github.com/EFForg/https-everywhere/pull/15747) and removing [this line](https://github.com/EFForg/https-everywhere/blob/4117544090de90167069ecfab1a85dc1eac33303/utils/rewriter/rewriter.js#L78).

PR #6732, opened 2019-03-18
Add secfixes comments

J0WI:
Please backport the security fixes in `main/file` to stable branches.


main/file/APKBUILD
```diff
@@ -4,13 +4,21 @@ pkgname=file
 pkgver=5.36
 pkgrel=0
 pkgdesc="File type identification utility"
-url="http://www.darwinsys.com/file"
+url="https://www.darwinsys.com/file"
```
ncopa: with this change we need bump pkgrel, because `apk info --webpage` will give different result.

J0WI: Isn't this updated anyway with the next change?

ncopa:
i squashed the commits. thanks!

main/file/APKBUILD
J0WI: btw: would be great if you could give a feedback here:
https://github.com/alpinelinux/aports/pull/4516#issuecomment-429306674

PR #6735, opened 2019-03-18
community/nx-libs: fix build break by FontEncDirectory return of const char *

mksully22:
Build fails with error:
encparse.c:789:1: error: conflicting types for 'FontEncDirectory'
 FontEncDirectory()
 ^~~~~~~~~~~~~~~~
In file included from encparse.c:62:
/usr/include/X11/fonts/fontenc.h:112:13: note: previous declaration of 'FontEncDirectory' was here
 const char *FontEncDirectory(void);

In the 1.1.4 version of libfontenc-dev that provides /usr/include/X11/fonts/fontenc.h a change was introduced to  have FontEncDirectory return a const char *:
- char *FontEncDirectory(void);
+ const char *FontEncDirectory(void);

This patch modifies nx-libs to be consistent with this new calling convention.

### DIFF
--- a/community/nx-libs/APKBUILD
+++ b/community/nx-libs/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=nx-libs
 pkgver=3.5.0.32
-pkgrel=0
+pkgrel=1
 pkgdesc="NoMachine libraries (redistributed by x2go)"
 url="http://x2go.org"
 arch="all"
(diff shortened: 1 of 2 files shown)


community/nx-libs/APKBUILD
```diff
@@ -15,6 +15,7 @@ install=""
 subpackages="$pkgname-dev $pkgname-doc"
 source="http://code.x2go.org/releases/source/$pkgname/$pkgname-$pkgver-full.tar.gz
 	xf86bigfont.patch
+	fix-const-char.patch
```
ncopa: whitespace damage. space before EOL. I will fix it for you.

PR #6736, opened 2019-03-18
[3.9] main/zfs: import pool before swap

HRio:
Just like for lvm, we need to import the zfs pool before swap is started

(cherry picked from commit f4f23af2e0ed83d17000d9cb19695540cc0d485c)


HRio:
@ncopa I think we should backport this, what do you think?

PR #6742, opened 2019-03-18
commmunity/glm: Fixed constexpr with GCC

mksully22:
community/gource and community/logstalgia fail to build, generating errors of type:
/usr/include/glm/detail/type_vec2.hpp:90:40: error: 'constexpr const T& glm::vec<2, T, Q>::operator[](glm::vec<2, T, Q>::GLM_FUNC_DECL GLM_CONSTEXPR T const& operator[](length_type i) const;

As discussed in glm issue: https://github.com/g-truc/glm/issues/832
a fix has been committed into glm master branch: https://github.com/g-truc/glm/commit/68c7e7e50b934cd265251a0660096f1415647bbb

Until glm version 0.9.9.4 is released this patch can be backported onto 0.9.9.3 to allow gource and logstaligia packages to build.

### DIFF
--- a/community/glm/APKBUILD
+++ b/community/glm/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=glm
 pkgver=0.9.9.3
-pkgrel=1
+pkgrel=2
 pkgdesc="C++ mathematics library for graphics programming"
 url="http://glm.g-truc.net/"
 arch="noarch"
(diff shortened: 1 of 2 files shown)


Ikke:
For the record, g-truc/glm@68c7e7e seems to be the commit that breaks it, and the included patch reverts that commit

Ikke:
Patch didn't apply because the project uses crlf as line ending. I used unix2dos on the patch, and added `patch_args="--binary -p1"`

PR #6751, opened 2019-03-19
main/musl: correct FADV macro on s390x

tuan-hoang1:
Copy from upstream


tuan-hoang1:
@ncopa: I forgot this patch, which goes with : https://github.com/alpinelinux/aports/pull/6728

PR #6729, opened 2019-03-18
[3.9] main/strace: fix SIGNAL_FRAMESIZE on s390x

tuan-hoang1:
copy from commit 6d2ed715191752850b4907496c13cca72e98bfe5

PR #6752, opened 2019-03-19
[3.9] main/musl: correct FADV macro on s390x

tuan-hoang1:
Copy from upstream


tuan-hoang1:
This PR goes along with https://github.com/alpinelinux/aports/pull/6729

PR #6756, opened 2019-03-19
testing/llvm6: fix test failure on ppc64le and re-enable

mksully22:


J0WI:
@mksully22 why have you already closed this? Is your patch working?

mksully22:
@JOWI in https://github.com/alpinelinux/aports/commit/27b0e49a86ff773833fe463edff6964f26eacbd4#commitcomment-32825298 ncopa indicated that he had the fix figured out.  Also in more extensive testing I haven't been able to get the baseline failure to reproduce 100% of the time on my multiple ppc64le (p8,p9,qemu,docker) setups consistency. So even though I've never gotten a failure with the patch I can't be 100% sure...

J0WI:
It's not yet committed. I think the patch is fine since the same approach is used in llvm7.

PR #6759, opened 2019-03-19
main/py-tox: add py-six as a dependency

Ikke:
py-six is required by py-tox:

> pkg_resources.DistributionNotFound: The 'six<2,>=1.0.0' distribution was not
> found and is required by tox


Ikke:
This should be the last fix for py-tox, I've tested it in a clean docker container and this works

> $ apk add py3-tox
> $ tox

PR #6762, opened 2019-03-19
main/nginx: add new package for nginx-mod-http-geoip2

aptalca:
This PR creates a new package for nginx-mod-http-geoip2 that has a dependency on package `libmaxminddb`.
Nginx is also compiled with support for the dynamic module.

The package source is https://github.com/leev/ngx_http_geoip2_module

### DIFF
--- a/main/nginx/APKBUILD
+++ b/main/nginx/APKBUILD
@@ -15,7 +15,7 @@ pkgname=nginx
 # NOTE: Upgrade only to even-numbered versions (e.g. 1.14.z, 1.16.z)!
 # Odd-numbered versions are mainline (development) versions.
 pkgver=1.14.2
-pkgrel=1
+pkgrel=2
 # Revision of nginx-tests to use for check().
 _tests_hgrev=d6daf03478ad
 _njs_ver=0.2.0
@@ -24,8 +24,8 @@ url="http://www.nginx.org/en"
 arch="all"
 license="BSD-2-Clause"
 depends=""
-makedepends="linux-headers gd-dev geoip-dev libxml2-dev libxslt-dev
-	openssl-dev paxmark pcre-dev perl-dev pkgconf zlib-dev"
+makedepends="linux-headers gd-dev geoip-dev libmaxminddb-dev libxml2-dev
+	libxslt-dev openssl-dev paxmark pcre-dev perl-dev pkgconf zlib-dev"
 checkdepends="gd perl perl-fcgi perl-io-socket-ssl perl-net-ssleay
 	perl-protocol-websocket tzdata uwsgi-python"
 pkgusers="nginx"
@@ -120,6 +120,8 @@ _add_module "rtmp" "v1.2.1" "https://github.com/arut/nginx-rtmp-module"
 _rtmp_provides="$pkgname-rtmp"  # for backward compatibility
 
 _add_module "http-vod" "1.24" "https://github.com/kaltura/nginx-vod-module"
+
+_add_module "http-geoip2" "3.2" "https://github.com/leev/ngx_http_geoip2_module"
 
 prepare() {
 	local file; for file in $source; do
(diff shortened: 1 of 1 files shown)


aptalca:
@maxwell-k thanks, commits squashed as requested

maxwell-k:
@aptalca: that's great thank you, at the moment our Drone CI is failing. If you please rebase that commit on the current master the CI should pass.

The Drone config has changed recently and it now applies a PR differently, the current approach is at:
https://github.com/alpinelinux/alpine-drone-ci/blob/master/files/usr/local/bin/build.sh#L81

Thanks for your help.

aptalca:
@maxwell-k thanks for the heads up and the pointers. All tests are successful now.

main/nginx/APKBUILD
```diff
@@ -121,6 +121,10 @@ _rtmp_provides="$pkgname-rtmp"  # for backward compatibility
 
 _add_module "http-vod" "1.24" "https://github.com/kaltura/nginx-vod-module"
 
+_add_module "http-geoip2" "3.2" "https://github.com/leev/ngx_http_geoip2_module"
+_http_geoip2_depends="libmaxminddb"
```
clandmeter: Why are you depending on libmaxmindb?
abuild will automatically scan elf files for dependencies.

aptalca: I was not aware of that. I noticed the other module packages listing their dependencies, so added the dependency here. If it's not needed, I can remove and squash/rebase. Please confirm.

clandmeter: FYR, you can try to build it with abuild -rK (dont forget to cleanup build deps afterwards) and check the contents of

```
pkg/.control.$subpkgname/.PKGINFO
pkg/.control.nginx-mod-http-geoip2/.PKGINFO
```
This should already include the depends to libmaxinddb.
Explicit depends on libs is not needed in normal cases.

aptalca: Yup, you're right. I locally built the package without the dependency listed and `.PKGINFO` lists the libmaxminddb.so and installing the package also installs libmaxminddb.

However, digging through the PKGINFO files, I realized that libmaxminddb is now also "auto detected" as a dependency for the main nginx package. No idea how that happened. libmaxminddb-dev is installed as a build dependency and the module is supposed to be built as a dynamic module, same as the geoip-dev package that is installed as a build dependency for the geoip1 module

fabled: `checkpkg` says that the main package gets:
``` diff
+usr/lib/nginx/modules/ngx_stream_geoip2_module.so
```

So the automatic package split is not moving all modules to the subpackage. The above file that is left in the main package is introducing the dependency to libmaxminddb.so.

aptalca: @fabled thanks so much for pointing me in the right direction. I'll get that figured out shortly

PeterDaveHello:
Hello, thanks for adding this package, just wondering if there is anything wrong that I can't simply use it? Got no such problem with `nginx-mod-http-geoip`, thank you.

```
$ docker run --rm -it alpine:3.10 sh
/ # apk add nginx nginx-mod-http-geoip2
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
(1/8) Installing pcre (8.43-r0)
(2/8) Installing nginx (1.16.1-r0)
Executing nginx-1.16.1-r0.pre-install
(3/8) Installing ca-certificates (20190108-r0)
(4/8) Installing nghttp2-libs (1.39.2-r0)
(5/8) Installing libcurl (7.66.0-r0)
(6/8) Installing curl (7.66.0-r0)
(7/8) Installing libmaxminddb (1.3.2-r0)
(8/8) Installing nginx-mod-http-geoip2 (1.16.1-r0)
Executing busybox-1.30.1-r2.trigger
Executing ca-certificates-20190108-r0.trigger
OK: 9 MiB in 22 packages
/ # nginx -t
nginx: [emerg] dlopen() "/var/lib/nginx/modules/ngx_stream_geoip2_module.so" failed (Error relocating /var/lib/nginx/modules/ngx_stream_geoip2_module.so: ngx_stream_add_variable: symbol not found) in /etc/nginx/modules/http_geoip2.conf:1
nginx: configuration file /etc/nginx/nginx.conf test failed
```

PeterDaveHello:
Okay I saw #9234 & #9351

PR #6763, opened 2019-03-20
main/ansible: upgrade to 2.7.9

prspkt:
supersedes #6382 (2.7.8)

### DIFF
--- a/main/ansible/APKBUILD
+++ b/main/ansible/APKBUILD
@@ -9,9 +9,8 @@ pkgdesc="A configuration-management, deployment, task-execution, and multinode o
 url="https://ansible.com"
 arch="noarch"
 license="GPL-3.0-or-later"
-_py=py3
-depends="python3 $_py-yaml $_py-paramiko $_py-jinja2 $_py-markupsafe $_py-crypto"
-makedepends="python2-dev py-setuptools"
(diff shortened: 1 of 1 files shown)


J0WI:
This will also fix CVE-2018-16876.

prspkt:
CVE-2018-16876 was addressed in [v2.7.5](https://github.com/ansible/ansible/blob/stable-2.7/changelogs/CHANGELOG-v2.7.rst#v2-7-5).

J0WI:
Yes, but 2.7.1-2.7.8 do not exist in Alpine Linux.

prspkt:
That is correct. We never shipped the version/revision of `ansible` that fixed the particular CVE and thus we don't make an entry in the `secfixes` log.

The `secfixes` section is not a CVE tracker for the ansible project. It is a CVE tracker for the ansible package that ships as part of Alpine Linux.

andypost:
I think better to add aggregated secfixes cos they are all packaged by-passed versions

andypost:
What @fabaff thinks about it?

nmeum:
> aI think better to add ggregated secfixes cos they are all packaged by-passed versions

:+1: in my experience this is also what we do with other packages.

nmeum:
Took the liberty of adding that CVE to secfixes, used `cp -r` instead of `mv` and pushed it.
Thanks!

PR #6764, opened 2019-03-20
DroneCI: deeper fetch

tcely:


andypost:
Why so deep to fetch?

tcely:
@andypost To avoid build failures when there were more than 50 commits to master.

As it stands, drone fails to merge if the PR merge base is the 51st commit in master.

clandmeter:
Im not sure i want to increase it, i think this is a basic workflow issue.

1. This will add a delay to fetch more data from github which is not needed at the time of the build.
2. This will make branches with less recent history (not recently rebased) build while contributor should rebase when pushing new content to his feature branch.

This rebase rule should be added to our developers documentation /cc @5paceToast 

I think the notification should also be more sane @nmeum already mentioned this on IRC, now its very unclear for the submitter what is going on. 
The problem with current implementation is that its using drone's own fetch logic which does not provide any sane error messaging except the one generated from git. So the only proper way to fix this imho is to disable the drone clone option and do that ourselves and add some logic to it.

tcely:
@clandmeter If you plan to write your own clone logic, then you should consider fetching the PR merge  instead of head, then you can avoid all of this mess entirely.

andypost:
I think 50 is reasonable amount of commits - it is not easy task to review them all

andypost:
Btw kinda "rebaser" bot may help just check mergability once a week

tcely:
@andypost 
> I think 50 is reasonable amount of commits - it is not easy task to review them all

We aren't discussing how many commits are on the feature branch. It could have only one. However, if more than fifty commits have occurred on master since that branch was created then the DroneCI merge fails.

```
fatal: refusing to merge unrelated histories
```

clandmeter:
@tcely is right, the history is purely to make merging possible, its the limit of commits to fetch from aports master, not feature branch.
The issue we try to solve here is to get a list of changed aports and we can calculate this in multiple ways. Having complete trees(branches) local you can always generate a new diff between them and this is exactly what we need to see which aports have changed.
We could however fetch master with 1 as depth (we only fetch the tip) and then we manually apply the PR diff which is available via github web interface, so we let github create the diff(erence) instead of us.

tcely:
@clandmeter I'd like to see this merged until a better clone/fetch is ready.

Constant rebasing is a lot to ask of contributors especially without any compelling reason for it.

clandmeter:
This should be fixed in 067a50cc7cbc23025324c32b3a1235d6501c4af1

PR #6765, opened 2019-03-20
testing/vault: upgrade to 1.1.0

mtneug:
Upgrade to newly released 1.1.0.

### DIFF
--- a/testing/vault/APKBUILD
+++ b/testing/vault/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Christian Kampka <christian@kampka.net>
 # Maintainer: Gennady Feldman <gena01@gmail.com>
 pkgname=vault
-pkgver=1.0.3
+pkgver=1.1.0
 pkgrel=0
 pkgdesc="Vault is a tool for securely accessing secrets"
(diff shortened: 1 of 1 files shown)


Ikke:
@mtneug Do you think this package is stable enough to move to community? Means that go-bindata-assetfs would also need to be moved.

Ikke:
So the maintainer apparently disappeared and now they setup an unmaintained copy, looking for a maintainer.

Do you think this project is stable enough that it would not need a lot of updates for it to still work?

mtneug:
I think it is very likely it will continue to work. The purpose of go-bindata is to generate Go code to embed non-go files. Go 1.x [promises backwards compatibility](https://golang.org/doc/go1compat). The generated files as well as the code of go-bindata itself therefore will continue to be valid Go code. Additionally, since projects like Vault and Traefik relay on it, I assume there will be some fixes if incompatibilities arise.

Ikke:
Alright, sounds good. I will apply this update. Can you prepare a new PR to move these 3 packages to community?

mtneug:
Yes, I will do that.

PR #6766, opened 2019-03-20
community/wine_gecko: modernize

J0WI:
Not sure if this should be moved to unmaintained?

### DIFF
--- a/community/wine_gecko/APKBUILD
+++ b/community/wine_gecko/APKBUILD
@@ -3,12 +3,12 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=wine_gecko
 pkgver=2.47
-pkgrel=0
+pkgrel=1
 pkgdesc="Wine's built-in replacement for Microsoft's Internet Explorer"
-url="http://wiki.winehq.org/Gecko"
+url="https://wiki.winehq.org/Gecko"
 arch="x86 x86_64"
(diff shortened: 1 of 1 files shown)


Ikke:
@J0WI What is the reason this should be moved to unmaintained?

J0WI:
There was no progress over the last ~3 years and I think it's very insecure if the browser engine has not been updated since then.
https://sourceforge.net/p/wine/wine-gecko/ci/master/tree/

Ikke:
So it's more because upstream is unmaintained, not the package itself?

J0WI:
> So it's more because upstream is unmaintained, not the package itself?

Yes

community/wine_gecko/APKBUILD
```diff
@@ -3,30 +3,22 @@
...
 pkgver=2.47
-pkgrel=0
+pkgrel=1
 pkgdesc="Wine's built-in replacement for Microsoft's Internet Explorer"
-url="http://wiki.winehq.org/Gecko"
+url="https://wiki.winehq.org/Gecko"
 arch="x86 x86_64"
 license="MPL"
-options="!archcheck"
+options="!archcheck !check"
```
andypost: makes sense to mention why no tests possible

J0WI: Done

andypost:
Commited as 5c9217e4

PR #6679, opened 2019-03-15
testing/php7-pecl-swoole: Bump Swoole to 4.3.1

petk:
Ref https://pecl.php.net/package-changelog.php?package=swoole&release=4.3.1

PR #6774, opened 2019-03-20
testing/php7-pecl-swoole: upgrade to 4.3.1

andypost:
Closes https://github.com/alpinelinux/aports/pull/6679


andypost:
still works only x86-64(

PR #6787, opened 2019-03-21
Move Vault to community

mtneug:
As discussed in #6765, this moves Vault, go-bindata and go-bindata-assetfs to community.


andypost:
@mtneug I think you need to set yourself as maintainer for `go-bindata` because it looks weird to add a community package without it

PR #6792, opened 2019-03-21
community/unit: upgrade to 1.8.0

andypost:
Ref https://unit.nginx.org/CHANGES.txt


andypost:
@misery release numbers now semver

PR #6797, opened 2019-03-21
community/lld: upgrade to 7.0.1

J0WI:


ncopa:
this was fixed with a84ab695096337553ae79182143730853331e0e5

ncopa:
I think we should rebase this and update commit message. The use of extranal lit looks better than we did in a84ab69.

community/lld/APKBUILD
```diff
@@ -33,6 +33,7 @@ build() {
 		-DCMAKE_INSTALL_PREFIX=/usr \
 		\
 		-DCMAKE_SKIP_INSTALL_RPATH=ON \
+		-DLLVM_EXTERNAL_LIT=/usr/lib/llvm7/bin/lit \
```
ncopa: Can we also use somthing like  `-DLLVM_EXTERNAL_LIT=/usr/lib/llvm${pkgver%%.*}/bin/lit` here?

J0WI: Tests are not passing as expected. Any hints?

J0WI:
a84ab695096337553ae79182143730853331e0e5

PR #6800, opened 2019-03-22
testing/skim: new aport

5paceToast:
https://github.com/lotabout/skim
A fuzzy finder in rust.
Extremely similar to fzf, but has better parallelism and memory safety
characteristics.


5paceToast:
Tests are failing because apparently cargo/rust is still restricted to x86_64-only...
This kind of seems strange, since rust claims to have musl-based non-x86_64 targets as tier 2, and working, but I'll restrict it for now to make CI happy.

Ikke:
Removed empty variables and marked tmux subpkg as noarch as suggested by abuild.

PR #6802, opened 2019-03-22
WIP: main/nfs-utils: do not run statd as root

HRio:


main/nfs-utils/nfs-utils.post-upgrade
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+ver_new="$1"
+ver_old="$2"
+
+if [ "$(apk version -t "$ver_old" '2.3.2-r2')" = '<' ]; then
+	# migrate owner from root: -> statd:
+	chown statd: /var/lib/nfs \
```
HRio: @ncopa how to handle that rpc.statd should not be running when doing this? are we "allowed" to restart processes during upgrade?

nmeum: Generally speaking I would *personally* prefer avoiding automigration magic in post install scripts and instead print a message informing the user of the necessary migration steps.

tcely: Given that the old process is running as root, does this chown affect it at all?

main/nfs-utils/nfs-utils.pre-install
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+adduser -S -D -h /var/lib/nfs -s /sbin/nologin -G nogroup -g statd statd 2>/dev/null
```
tcely: What created the `statd` group?

main/nfs-utils/APKBUILD
```diff
@@ -55,6 +56,7 @@ build() {
 		--mandir=/usr/share/man \
 		--infodir=/usr/share/info \
 		--without-tcp-wrappers \
+		--with-statduser=statd \
```
tcely: Does adding `pkgusers` make sense too?

PR #6816, opened 2019-03-23
main/linux-vanilla: use 1280 for stack frame warning on 64-bit

zx2c4:
The goal of upstream Linux is to reduce the stackframe size on 32-bit to
1024 and on 64-bit to 1280, inline with how gcc generally works. While
this hasn't been achieved yet everywhere and in all configurations, the
Alpine status quo of 1024 on 64-bit is something that doesn't have plans
to happen. Given that the intent was to be conservative, we raise this
to 1280, rather than something large like 2048.

Reference: https://lore.kernel.org/lkml/CAK8P3a0CS3QzEKEV5==qj8hUYgW+q2v1f13jA+s0TjQd8kYXFA@mail.gmail.com/


ncopa:
what about ppc64le and s390x?

zx2c4:
Added s390x. ppc64le is already at 2048, I assume for reasons.

ncopa:
fixed with dae185e9767fcff8165a00d4710ce5144f6d4cf5

PR #6825, opened 2019-03-24
[WIP] testing/lightdm* -> community/

tanertas:
lightdm* : move from testing
lightdm-gtk-greeter: upgrade to 2.0.6


tanertas:
Both CI fails because it can't find lightm dependency which have to be in community repo. This is already done with this PR but CI can't cope it.

Ikke:
`lightdm` depends on `accountsservice`, which is still in testing. Packages in community cannot depend on packages in testing (same for packages in main that cannot depend on community or testing). So if you want to move lightdm to community, accountsservice should as well (preferably in the same PR)

tcely:
This needs to be rebased

PR #6826, opened 2019-03-24
testing/php7-pecl-swoole: add arm patch to build aarch64 and arm*

andypost:
WIP


andypost:
with boost it fails on x86
```
aports/testing/php7-pecl-swoole/src/swoole-4.3.1/include/context.h:93:21:  error: 'fcontext_t' in namespace 'boost::context' does not name a type
```

PR #6831, opened 2019-03-24
github: update the stalebot message

Ikke:


clandmeter:
@Ikke I think the idea was to make this message included in .github/state.md and adjust the actual stale message to link to this for more information. Maybe include the rebase info into the actual stale message.

ncopa:
IMHO, the message needs to be significantly more friendly. The bot is stupid and in many cases it is us that is blocking due to not enough resources to review the PR. Threatening to close an issue when it is our fault is rude.

I think we should try include an explanation why we use the stupid bot in the first place to make it clear that it is (likely) not the contributors fault.

I would like to have the tone in the message to be along the lines: 

> Sorry we are overloaded and in our desperation we have configured this stupid bot. We would appreciate if you could help us by rebase this PR.
>
> If you no longer care about it, then feel free to not do anything and it will be autoclosed in 2 weeks. Thank you!

ncopa:
> @Ikke I think the idea was to make this message included in .github/state.md and adjust the actual stale message to link to this for more information. Maybe include the rebase info into the actual stale message.

The idea with a link to a longer explanation was to be able to communicate why we have the bot and how contributors can help us. It would be best if we can communicate that in a friendly way so short that we dont need to have it in separate file.

Ikke:
Ok, so just the bot message, clear and concise explaining what the situation is and what actions could be taken.

Ikke:
@ncopa I've updated the message, let me know what you think.

PR #6833, opened 2019-03-25
Services needing entropy

HRio:
Related to https://bugs.alpinelinux.org/issues/10025


HRio:
main/haveged fails on x86 (on drone), Ok on the others.

HRio:
I have also started work on entropy broker for systems that can not use rng-tools or haveged:
https://github.com/alpinelinux/aports/commit/6880dfb6e909b5cb5d73e770f5a0f4227515b610
The work is currently a work in progress that only builds for x86/x86_64 and it only provides an openrc script for the broker service. It needs rc scripts for the client (that one to declare `provide entropy`) as well as the server processes that will provide entropy to the broker service.

HRio:
@ncopa and @clandmeter can you help in reviewing this? thanks

PR #6836, opened 2019-03-25
main/dpkg: fix broken source link by upgrade to 1.19.5

mksully22:


nmeum:
Will merge this, but in the long run it would be nice to have a source URL which doesn't constantly change. E.g. you can't currently fetch the source for the dpkg aport on the `3.9-stable` branch because it is no longer available.

PR #6846, opened 2019-03-26
Php protobuf - rename, upgrade and add check

andypost:
https://github.com/protocolbuffers/protobuf/releases/tag/v3.7.1


andypost:
@TBK looks makes sense to upgrade other protobuf packages

PR #6851, opened 2019-03-27
Perl perlmagick for review only

timlegge:
@ncopa as you are the maintainer of the ImageMagick aport it would likely be a good idea for you to review this perlmagick aport.  If it looks okay I will pull out the change to the ImageMagick version and squash the commits

This requires a version upgrade of ImageMagick (included in this pull request but likely should not be) as the ImageMagick version cannot be downloaded anymore.

This was a fair amount of trial and error and you may know of a better approach.  I had thought of simply adding it to the ImageMagick but PerlMagic is only really needed if the user needs it.

The check is a hack because the tests provided in the Makefile don't work for me.  Even the docs say expect some of them to fail.

### DIFF
--- a/testing/perlmagick/APKBUILD
+++ b/testing/perlmagick/APKBUILD
@@ -1,0 +1,82 @@
+# Contributor: Timothy Legge <timlegge@gmail.com>
+# Maintainer: Timothy Legge <timlegge@gmail.com>
+# Natanael Copa <ncopa@alpinelinux.org>
+pkgname=perlmagick
+pkgver=7.0.8.36
+pkgrel=0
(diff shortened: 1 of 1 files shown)


timlegge:
@ncopa the build fails here with a >> ERROR: perlmagick*: Has /home/... in rpath
I struggled quite a while with this till I realized the current git version of abuild works fine.  Not really sure how to work around that

timlegge:
looks like removing the RPATH works.  Need to test with Foswiki

timlegge:
Closing will send new version

timlegge:
Closing

timlegge:
Closin

PR #6848, opened 2019-03-26
main/protobuf: upgrade to 3.8.0

andypost:
Ref https://github.com/protocolbuffers/protobuf/releases
https://abi-laboratory.pro/?view=timeline&l=protobuf
Tests https://github.com/google/googletest/releases

php https://github.com/alpinelinux/aports/pull/6846
py https://github.com/alpinelinux/aports/pull/6847

PR #6852, opened 2019-03-27
Upgrade boost rebuild

tcely:
~~Depends on https://github.com/alpinelinux/aports/pull/4657~~

CI is useless for community / testing since those will only use the published boost version, not the upgraded version that was built in main. See https://github.com/alpinelinux/abuild/pull/53

### DIFF
--- a/main/boost/APKBUILD
+++ b/main/boost/APKBUILD
@@ -71,8 +71,9 @@ _libs="
 	wserialization
 	"
 for _lib in $_libs; do
-	case "$_options_carch " in # traing space is there to match last word
-	*"--without-${_lib} ") continue;;
+	# the trailing space in the below case is intentional so that we can match the last word
+	case "${_options_carch//$'\n'/ } " in
+		*"--without-${_lib} "*) continue ;;
 	esac
 	subpackages="$subpackages ${pkgname}-${_lib}:_boostlib"
 done; unset -v _lib
--- a/main/boost/APKBUILD
+++ b/main/boost/APKBUILD
@@ -96,7 +96,7 @@ _set_options() {
 		runtime-link=shared
 		link=shared,static
 		cflags=-fno-strict-aliasing
-		--layout=system
+		--layout=tagged-1.66
 		-q
 		-j${JOBS:-2}
 		$_options_carch
--- a/community/abiword/APKBUILD
+++ b/community/abiword/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=abiword
 pkgver=3.0.2
-pkgrel=4
+pkgrel=5
 pkgdesc="A fully-featured word processor"
 url="http://www.abisource.com"
 arch="all"
--- a/community/asio/APKBUILD
+++ b/community/asio/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=asio
 pkgver=1.13.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Cross-platform C++ library for network programming"
 url="https://think-async.com"
 arch="all"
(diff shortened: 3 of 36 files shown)


andypost:
FYI most of rebuilds are the same for https://github.com/alpinelinux/aports/pull/6848

ncopa:
lucene++ fails to build:
````
CMake Error at /usr/share/cmake/Modules/FindBoost.cmake:2132 (message):
  Unable to find the requested Boost libraries.                        
                                                                                                                             
  Boost version: 1.69.0                                                    
                                                                   
  Boost include path: /usr/include    
                                                                                                                             
  Could not find the following Boost libraries:                                                                              
                                                                                                                             
          boost_date_time                                                                                                    
          boost_filesystem               
          boost_iostreams                                 
          boost_regex                                                                                                        
          boost_system                                   
          boost_thread                                    
                                                                       
  Some (but not all) of the required Boost libraries were found.  You may    
  need to install these additional Boost libraries.  Alternatively, set                                                      
  BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT
  to the location of Boost.                                                                                                  
  CMakeLists.txt:61 (find_package)                                                                                           
````

**Comment by ncopa** [comment]
> lucene++ fails to build:

I suspect that it may be related to that it now includes the architecture in library name:
````diff
--- filelist-boost-thread-old
+++ filelist-boost-thread
@@ -1,4 +1,4 @@
 .PKGINFO                                
 usr/                                 
 usr/lib/
-usr/lib/libboost_thread-mt.so.1.67.0
+usr/lib/libboost_thread-mt-x64.so.1.69.0
````

File: main/boost/APKBUILD [hunk 1]
```diff
@@ -46,27 +46,35 @@ _libs="
...
 	wave
 	wserialization
 	"
 for _lib in $_libs; do
-	case "$_options_carch" in
-		*$_lib*) continue;;
-	esac
+	for _option in $_options_carch; do
+		case "$_option" in --without-${_lib}) continue ;; esac
+	done
```
ncopa: I think the logic here is broken. There is a nested loop here and the `continue` statement there will only continue the the inner loop.

tcely:
lucene++ does not appear to be used by anything. Maybe it should just be removed rather than patched.

main/boost/APKBUILD
tcely: I concur. Moving the assignment into the inner loop looks to be what we want.

tcely:
> > lucene++ fails to build:
> 
> I suspect that it may be related to that it now includes the architecture in library name:

https://gitlab.kitware.com/cmake/cmake/merge_requests/2568

ncopa:
> > > lucene++ fails to build:
> > 
> > 
> > I suspect that it may be related to that it now includes the architecture in library name:
> 
> https://gitlab.kitware.com/cmake/cmake/merge_requests/2568

I believe this is related:
https://gitlab.kitware.com/cmake/cmake/issues/18553

main/boost/APKBUILD
```diff
@@ -96,7 +96,7 @@ _set_options() {
 		runtime-link=shared
 		link=shared,static
 		cflags=-fno-strict-aliasing
-		--layout=system
+		--layout=tagged-1.66
```
ncopa: Why would we want to do this at this point? `--layout=system` makes things much simpler for us.

main/boost/APKBUILD
```diff
@@ -71,8 +71,9 @@ _libs="
 	wserialization
 	"
 for _lib in $_libs; do
-	case "$_options_carch " in # traing space is there to match last word
-	*"--without-${_lib} ") continue;;
+	# the trailing space in the below case is intentional so that we can match the last word
+	case "${_options_carch//$'\n'/ } " in
+		*"--without-${_lib} "*) continue ;;
```
ncopa: The only useful thing in this commit is the introduced `*` which is completely hidden in the useless changes. And commit message does not explain anything about why this change is needed.

However, the fact that builders has not choked over the bug this fixes, may indicate that we no longer need exclude things for s390x, so we can probably remove the `_options_carch` and the `case` conditional completely.

ncopa:
fixed with 6895c7b3edf045aea12c75e47f60356ae604f152

PR #6853, opened 2019-03-27
main/nginx: upgrade njs submodule to 0.3.1

andypost:
Ref http://nginx.org/en/docs/njs/changes.html#njs0.3.0


TBK:
You need a newer version of the test suite - https://hg.nginx.org/nginx-tests/rev/f4ae08adc23f

andypost:
@TBK any idea how to select neede3d commit from tests? latest one fails tests

TBK:
upstream_ip_hash_ipv6.t should skip with "skipped: no inet6 support" but for some reason it does not.

I can't immediately see if it is a issue with the test code. Either disable it since CI's atm generally do not support IPv6 or it could be due to some missing perl dependency.
https://hg.nginx.org/nginx-tests/file/tip/upstream_ip_hash_ipv6.t

andypost:
btw patch no longer needed and applies wrong, looks tests needs upgrade to https://hg.nginx.org/nginx-tests/rev/f4ae08adc23f (locally it passes tests), I will revert to this if latest without patch will not pass

TBK:
@andypost Sorry, I did not answer your question.

I do not know the right approach to determine the correct commit to use. To my knowledge the nginx team does not announce/tag or in anyway indicate which version of the test suite that matches a certain version of nginx, njs...

My approach so far I has been to use trial and error. Maybe @jirutka can enlighten us.

andypost:
Removed ipv6 test added in http://hg.nginx.org/nginx-tests/rev/f7f387086278 and tests started to pass

PR #6855, opened 2019-03-27
main/zstd: enable armhf

tcely:


tcely:
Still a bus error during check.

PR #6856, opened 2019-03-27
Upgrade boost lucene++

tcely:


ncopa:
This looks like a duplicate of #6852 and #4657. and it looks like all 3 PRs have commits with exact same subject of commit message. I will have to go through each commit again and compare it with the corresponding commit in the other PRs and verify if there is any difference in the commit or if commit is identical. If there is a difference then I will need to try figure out why and if I have commented it in some of the other PRs.

Then I will probably need to do all the above once again for the other two PRs.

Sorry that I cannot process this fast enough.

tcely:
This is a draft I put together to debug boost. Please don't bother.

PR #6857, opened 2019-03-27
testing/entropybroker: new aport

HRio:


testing/entropybroker/APKBUILD
```diff
@@ -0,0 +1,69 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+# Maintainer:
```
Ikke: This package needs a maintainer. Can you maintain it yourself?

HRio: I usually start with `Maintainer` as blank, when I am trying to get a better feeling of the package while working on improving it in testing/, if somebody else steps up during that time and would like to maintain it I am fine with that. 

For instance did like that for swaywm and related packages. Then just before it was in a good enough state for community/ move, Drew Devault  said he could maintain it.

So if possible I would like to have some runtime and stabilizing a new package in testing/ before I decide to adopt a package as maintainer, during that time I also usually test how it works to get patches into upstream.

PR #6862, opened 2019-03-28
Ginja chris patch 1

GinjaChris:
community/h2o: amend init script


andypost:
@GinjaChris you're doing something wrong - your PR is not against master
PS: https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md

PR #6863, opened 2019-03-28
Update h2o.initd

GinjaChris:
community/h2o: remove "-m master" from init.d file, runs as default "-w worker" instead


andypost:
it needs to bump pkgrel cos changing file

andypost:
Also commit message should be like `community/h2o: fix whatever in initd ` https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request-pr

PR #6865, opened 2019-03-28
testing/rtrlib: new aport

cfra:
https://github.com/rtrlib/rtrlib
An open-source C implementation of the RPKI/Router Protocol client

### DIFF
--- a/testing/rtrlib/APKBUILD
+++ b/testing/rtrlib/APKBUILD
@@ -1,0 +1,39 @@
+# Maintainer: Christian Franke <nobody@nowhere.ws>
+pkgname=rtrlib
+pkgver=0.6.3
+pkgrel=0
+pkgdesc="An open-source C implementation of the RPKI/Router Protocol client"
+url="https://github.com/rtrlib/rtrlib"
+arch="all"
+license="MIT"
+makedepends="cmake cmocka-dev doxygen graphviz libssh-dev"
+subpackages="$pkgname-dev $pkgname-doc"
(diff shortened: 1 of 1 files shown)


testing/rtrlib/APKBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Christian Franke <nobody@nowhere.ws>
+pkgname=rtrlib
+pkgver=0.6.3
+pkgrel=0
+pkgdesc="An open-source C implementation of the RPKI/Router Protocol client"
+url="https://github.com/rtrlib/rtrlib"
+arch="all"
+license="MIT"
+depends=""
```
andypost: no reason to keep empty lines

testing/rtrlib/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+makedepends="cmake cmocka-dev doxygen graphviz libssh-dev"
+install=""
+subpackages="$pkgname-dev $pkgname-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/rtrlib/$pkgname/archive/v$pkgver.tar.gz"
```
andypost: better to remove `$pkgname` from URL to easy navigate to source

testing/rtrlib/APKBUILD
cfra: Good point, thank you for pointing it out. :pray:

testing/rtrlib/APKBUILD
cfra: Done.

cfra:
@andypost I think I have addressed both of your coments.

Is there anything else which I should do to move this PR further along?

PR #6794, opened 2019-03-21
 testing/zimwriterfs: new aport

OTLabs:
https://openzim.org/
Console tool to create ZIM files

PR #6866, opened 2019-03-28
testing/libzim: moving gtest from makedepends to checkdepends, skip big memory test

OTLabs:
There is no need in package rebuild as the gtest is used during check() only.

### DIFF
--- a/testing/libzim/APKBUILD
+++ b/testing/libzim/APKBUILD
@@ -8,7 +8,8 @@ url="https://openzim.org/"
 arch="all"
 license="GPL-2.0"
 makedepends="meson zlib-dev xz-dev libexecinfo libexecinfo-dev icu-dev xapian-core-dev 
-gtest gtest-dev python3-dev"
+	     gtest-dev python3-dev"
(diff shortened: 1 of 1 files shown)


OTLabs:
1 of 5 tests (cluster test) is failing in Drone for aarch64, same time other arm architectures are passing this test.

Ikke:
Tried restarting the ci build, but same thing happens again, timeout on test 1.

OTLabs:
Something similar has happened during initial commit #6794: Drone had no problems, but Travis do.

tcely:
It's a bit odd that `gtest-dev` does not depend on `gtest` since without it some shared libraries will be missing.

I created https://github.com/alpinelinux/aports/pull/6877 to fix that.

OTLabs:
This test takes unto 12GB of memory, https://github.com/openzim/libzim/issues/147.

Will disable it with `SKIP_BIG_MEMORY_TEST=1 ninja test`.

Ikke:
@OTLabs Nice find!

I will look at this when I have time.

PR #6868, opened 2019-03-28
testing/frr: Update to 7.0

cfra:
Insert myself as a maintainer. I have not heard from Arthur Jones for
quite some time. Given that there has been no update to the package
since August 2018 although upstream released two new major versions in
the meantime, this package seems abandoned to me, so I hope this is
acceptable.

FRR should build and run fine on all architectures, so the package
`arch` should not be limited to x86_64.

While a small part of FRR's source is provided under the terms of the
LGPL license, the resulting binary can only be distributed under the
terms of the GPL, version 2.0 or (at the users option) any later
version. Update the `license` field to reflect this.

Starting from version 6.0, FRR requires libyang, so add it as a
build time dependency.

FRR has modules for integrating SNMP and RPKI. Add their dependencies
as build time dependencies and package them as subpackages.

Add `noarch` to those subpackages which do not contain any architecture
specific files.

FRR changed its statup script to  `/usr/lib/frr/frrinit.sh` - propagate this
change to the OpenRC wrapper script.

`inet_ntop` in musl has been fixed to be standards compliant, so the
patch `make-check-bypass-isisd-tlv.patch` is no longer needed.

According to upstream the headers and shared libraries are not for
consumption by external programs, so they should not be packaged
as a `-dev` package.

Depends on PR  #6865


Ikke:
This depends on `rtrlib-dev`. which apparently does not exist yet.

cfra:
@Ikke you are correct. For adding this package, I opened #6865.
This is also why this PR is currently a draft, I intended to make it a real PR once #6865 has made it in.

cfra:
Promoting this from draft status to real PR as #6865 has been merged (https://github.com/alpinelinux/aports/commit/1db72d3cef32520fd5b1818a5fb79126ac683b3d )

Ikke:
@cfra Can you rebase this onto the latest master?

cfra:
There seems to be a regression, because the rebased version fails in `make check`. Possibly this is due to a newer libyang version. I have opened https://github.com/FRRouting/frr/issues/4208 to investigate this.
Will update this PR afterwards.

Ikke:
> LGTM

This is still broken atm, so I'd wait before shipping it :-)

cfra:
You are indeed correct that this is still broken. There is a version 7.0.1 of FRR which improves things slightly, but it still breaks with libyang 1.0r2.
FRR's position seems to be that this should be fixed once libyang 1.0r3 gets released.
Once that happens, I am planning to publish an update of libyang to 1.0r3 and update this
PR once the new libyang release is in.

Ikke:
Yes, I verified that those tickets were still open.

TBK:
`cd $builddir` can be removed from build, check and package.

Ikke:
@cfra From the upstream ticket:

> This should be fixed already since FRR now supports both libyang >= 0.16-r3 and libyang >= 1.0.8. 

So should this be fixed now?

cfra:
I think this is not fixed in 7.0. However it might be fixed in FRR 7.1.

I will close this pull request and create a new one for FRR 7.1.

PR #6870, opened 2019-03-28
community/imagemagick6: security upgrade to 6.9.10.37

J0WI:
See also https://github.com/alpinelinux/aports/pull/6869

PR #6871, opened 2019-03-28
community/ruby-rmagick: upgrade to 3.0.0

J0WI:


Ikke:
@J0WI If this depends on #6870, it should be part of the same PR. If this can be applied separately, then then can remain separate.

J0WI:
It does not depend on #6870.

community/ruby-rmagick/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=ruby-rmagick
 _gemname=rmagick
-pkgver=2.16.0
+pkgver=3.0.0
 _pkgver=${pkgver//./-}
 pkgrel=2
```
andypost: needs to bump to 0

andypost:
Fixed with c81b4acc

PR #6877, opened 2019-03-28
main/gtest: -dev should install libraries too

tcely:
Using *.pc files is difficult when the libraries are not installed.

### DIFF
--- a/main/gtest/APKBUILD
+++ b/main/gtest/APKBUILD
@@ -9,7 +9,7 @@ options="!check"  # googletest-death-test-test hangs
 license="BSD-3-Clause"
 depends=""
 depends_dev="${pkgname}=${pkgver}-r${pkgrel} gmock=${pkgver}-r${pkgrel} cmake"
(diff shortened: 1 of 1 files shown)


tcely:
Please assign to @fcolista

main/gtest/APKBUILD
```diff
@@ -1,14 +1,15 @@
...
+pkgrel=1
 pkgdesc="Google Test - C++ testing utility based on the xUnit framework (like JUnit)"
 url="https://github.com/google/googletest"
 arch="all"
 options="!check"  # googletest-death-test-test hangs
 license="BSD-3-Clause"
 depends=""
-makedepends="python2-dev cmake"
+depends_dev="${pkgname}=${pkgver}-r${pkgrel} gmock=${pkgver}-r${pkgrel} cmake"
+makedepends="$depends_dev python2-dev"
```
andypost: maybe move it to py3 same time?

tcely: That's a good thought.

PR #6881, opened 2019-03-28
main/tzdata: upgrade to 2019a

J0WI:
https://mm.icann.org/pipermail/tz-announce/2019-March/000055.html

### DIFF
--- a/main/tzdata/APKBUILD
+++ b/main/tzdata/APKBUILD
@@ -2,8 +2,8 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=tzdata
-pkgver=2018i
-_tzcodever=2018i
+pkgver=2019a
+_tzcodever=2019a
 _ptzver=0.5
 pkgrel=0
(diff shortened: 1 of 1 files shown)


J0WI:
Please consider backporting this package. Alpine 3.6 is still on 2017a. This is critical when validating security certificates.

ncopa:
> Please consider backporting this package. Alpine 3.6 is still on 2017a. This is critical when validating security certificates.

It would be good if you could file an issue on https://bugs.alpinelinux.org for things that needs to be backported. I normally don't read all github comments before tagging new releases, but I do look over the issues on bugs.a.o.

J0WI:
> It would be good if you could file an issue on https://bugs.alpinelinux.org for things that needs to be backported.

Unfortunately the ability to create issues as guest have been disabled recently. I'm not convinced to register there too.

tcely:
I created https://bugs.alpinelinux.org/issues/10202 for this PR.

tcely:
It looks like at least `edge` has been updated in: https://github.com/alpinelinux/aports/commit/65c2c2ffbfdbb3663a6e9ec476183a5c90467c62#diff-e39254f7a81874241c6826b4ee193488

PR #6885, opened 2019-03-29
testing/traefik: upgrade to 1.7.12 and enable x86 and arm

andypost:
Ref https://github.com/containous/traefik/releases/tag/v1.7.12

### DIFF
--- a/testing/traefik/APKBUILD
+++ b/testing/traefik/APKBUILD
@@ -1,11 +1,11 @@
 # Contributor: Joe Holden <jwh@zorins.us>
 # Maintainer: Joe Holden <jwh@zorins.us>
 pkgname=traefik
-pkgver=1.7.11
+pkgver=1.7.12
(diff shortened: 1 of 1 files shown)


tcely:
The tests run in `check` don't appear to be 32-bit safe. Did this get reported to upstream?

andypost:
I found no reports, and gonna dig first, but @tcely is right about x64 mention in fails

andypost:
@tcely I see that 32bits missing and no idea how it can be solved( please help to report upstream, maybe there's a way to bypass x64 tests on 32 arch

TBK:
Duplicate PR https://github.com/alpinelinux/aports/pull/8029 upgraded with commit https://github.com/alpinelinux/aports/commit/f7ff1b606d040ecd6db56a1dbe861e0a7094cf3b

This PR still has value in enabling more archs.

PR #6894, opened 2019-04-01
main/postfix: upgrade to 3.4.5

lexsca:
ref [https://de.postfix.org/ftpmirror/official/postfix-3.4.5.RELEASE_NOTES](https://de.postfix.org/ftpmirror/official/postfix-3.4.5.RELEASE_NOTES)

~~this also includes `icu-dev` lib to support `smtputf8_enable` as in PR #6596~~

### DIFF
--- a/main/postfix/APKBUILD
+++ b/main/postfix/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=postfix
-pkgver=3.4.1
+pkgver=3.4.5
 pkgrel=0
 pkgdesc="Secure and fast drop-in replacement for Sendmail (MTA)"
 url="http://www.postfix.org/"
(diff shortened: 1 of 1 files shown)


ncopa:
This will increase size with ~32MB due to the icu dependency. I am not sure if we care about that or not?

Previously discussed here:
https://bugs.alpinelinux.org/issues/5089
https://patchwork.alpinelinux.org/patch/1012/

lexsca:
> This will increase size with ~32MB due to the icu dependency. I am not sure if we care about that or not?
> 
> Previously discussed here:
> https://bugs.alpinelinux.org/issues/5089
> https://patchwork.alpinelinux.org/patch/1012/

fair point. happy to remove it and let folks at added it as a separate aport if needed.

tcely:
> This will increase size with ~32MB due to the icu dependency. I am not sure if we care about that or not?

I personally don't care. Adding ~32 MB to the installed size for `postfix` is not ideal, but that's like two emails with attachments. It does not matter in the least. Anyone who is installing this to serve it's purpose is not going to be limiting the storage space available.

ncopa:
Still feels wrong to bloat things up that much for that small feature.

Looks like it only needs `ucasemap_utf8FoldCase`, `uidna_nameToASCII_UTF8` and `uidna_nameToUnicodeUTF8`.

There is also a `libidnkitlite` library. Might be possible to use that.

lexsca:
> Still feels wrong to bloat things up that much for that small feature.

removed `icu-dev`, as it's not critical to me and adds significant bloat.

PR #6898, opened 2019-04-01
testing/perl-perlmagic: new aport

timlegge:
@ncopa as you are the maintainer of the ImageMagick aport it would likely be a good idea for you to review this perlmagick aport. 

It needs ImageMagick 7.0.8.36 packages to be updated as well.

### DIFF
--- a/testing/perlmagick/APKBUILD
+++ b/testing/perlmagick/APKBUILD
@@ -1,0 +1,78 @@
+# Contributor: Timothy Legge <timlegge@gmail.com>
+# Maintainer: Timothy Legge <timlegge@gmail.com>
+pkgname=perlmagick
+pkgver=7.0.8.39
+pkgrel=0
+_pkgver=${pkgver%.*}-${pkgver##*.}
+_abiver=7
+pkgdesc="Perl Module for Image Magick"
+url="https://www.imagemagick.org/"
+arch="all"
+license="ImageMagick"
+options="libtool !checkroot"
+depends="imagemagick imagemagick-libs imagemagick-c++"
+makedepends="perl-dev libtool chrpath"
+subpackages="$pkgname-doc"
+source="https://www.imagemagick.org/download/releases/ImageMagick-$_pkgver.tar.xz"
+builddir="$srcdir/ImageMagick-${_pkgver}"
+
+build() {
+	cd "$builddir"
+	# fix doc dir, Gentoo bug 91911
+	sed -i -e \
+		's:DOCUMENTATION_PATH="${DATA_DIR}/doc/${DOCUMENTATION_RELATIVE_PATH}":DOCUMENTATION_PATH="/usr/share/doc/imagemagick":g' \
+		configure
+	./configure \
+		--build=$CBUILD \
+		--host=$CHOST \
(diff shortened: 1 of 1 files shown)


Ikke:
Please adjust the commit message to the [aports style][0], for example: 

> testing/perl-permagic: new aport

[0]:https://github.com/alpinelinux/aports/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request-pr

testing/perlmagick/APKBUILD
```diff
@@ -0,0 +1,82 @@
...
+	chrpath -d ./blib/arch/auto/Image/Magick/Q16HDRI/Q16HDRI.so
+}
+
+check() {
+	cd "$builddir/PerlMagick"
+	if ! [ -e blib/lib/Image/Magick.pm ]; then
+		error "Module does not exist Magick.pm"
+		return 1
+	fi
+        if ! [ -e blib/lib/Image/Magick/Q16HDRI.pm ]; then
```
Ikke: Inconsistent whitespace. Please use tabs for indentation.

testing/perlmagick/APKBUILD
```diff
@@ -0,0 +1,82 @@
...
+_pkgver=${pkgver%.*}-${pkgver##*.}
+_abiver=7
+pkgdesc="Perl Module for Image Magick"
+url="https://www.imagemagick.org/"
+arch="all"
+license="ImageMagick"
+options="libtool !checkroot"
+depends="imagemagick imagemagick-libs imagemagick-c++"
+makedepends="perl-dev libtool chrpath"
+checkdepends=""
```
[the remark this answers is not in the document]
timlegge: Will Do

testing/perlmagick/APKBUILD
timlegge: oops.  Will do

timlegge:
will squash the changes and send a force push

Ikke:
@timlegge Why does ImageMagick need to be updated? This seems a new package. And if that's the cause it would be best to do it in the same PR, so it can be applied in one go.

Thanks.

timlegge:
> @timlegge Why does ImageMagick need to be updated? This seems a new package. And if that's the cause it would be best to do it in the same PR, so it can be applied in one go.

perlmagick is the perl modules to use ImageMagick via perl.  It essentially requires adding the --with-perl to the configure for ImageMagick and doing the make of the perl modules and perlmagick install.  I did not see a way to do that with the ImageMagick (like a sub package).  Many distributions consider it to be a separate installation package you may want imagemagick or both imagemagick and perlmagick

It does however require the imagemagic and the permagick imagemagick versions to be the same.

If you can think of alternatives let me know.

timlegge:
I will likely close this PR as I have figured out how to add it to the main/imagemagick package

timlegge:
I have created PR7078 https://github.com/alpinelinux/aports/pull/7078 which would appear to be the best way to address this package for perlmagick

Ikke:
Would it make sense to close this PR then?

timlegge:
yes.  closing

PR #6911, opened 2019-04-02
main/dovecot: add warning Fatal: Dovecot version mismatch

itoffshore:
* fixes: https://bugs.archlinux.org/task/57782


ncopa:
i dont think upgrades should trigger restart of services. Its the responsibility of sysadmin who does the upgrade.

itoffshore:
Debian restart the service in `dovecot-core.postinst`
Changed the `PR` to show a warning

PR #6913, opened 2019-04-02
main/xen: upgrade to 4.12

HRio:
Upstream renamed /usr/lib/fs -> /usr/lib/xenfsimage

Drop patches included upstream


HRio:
the normal travis problem with "big" packages "The job exceeded the maximum log length, and has been terminated."

PR #6936, opened 2019-04-04
main/libcanberra: adds support for pulseaudio

xsteadfastx:


Ikke:
This needs a `pkgrel` bump.

xsteadfastx:
> This needs a `pkgrel` bump.

how could i have missed that....

xsteadfastx:
but why does it complain that it cant find `pulseaudio-dev`? its in the repos.

Ikke:
Because this package is in main, while `pulseaudio-dev` is in community. Packages in main cannot depend on packages in community.

xsteadfastx:
what can i do about it? pulseaudio is completly in community but libcanberra needs pulseaudio support that i can build some pulseaudio tool packages like `pavucontrol`.

maybe a metapackage of libcanberra like `libcanberra-pulse`? is it possible to have this in community for example if the rest is in main?

xsteadfastx:
i read there is a `depends_dev`. but i think there is no `depends_*` for any subpackage name or?

Ikke:
You cannot have subpackages in a different repository. The only way to solve this is by either move pulseaudio to main or libcanberra to community.

xsteadfastx:
im thinking about doing a full new `libcanberra-pulse` aport. which has just the pulse files and depend on the `libcanberra` package. but that would duplicate alot of stuff.

Ikke:
The plan is to move libgnome to community (with libcanberra as well). I'm currently verifying what would need to move to community in order to be able to do that.

xsteadfastx:
here is another try. its a own package with only the pulse part. what do you think?

https://github.com/xsteadfastx/aports/blob/libcanberra-pulse2/testing/libcanberra-pulse/APKBUILD

Ikke:
I tested it locally by installing directly into `$pkgdir`, and I don't see any rpath reference in the .so with `readelf -d`.

I had to fix the download link because you used `$pkgname` the first time, which does not exist.

The license is also not valid [SPDX][] identifier, it should probably be `LGPL-2.1-or-later`.

I defer the decision whether this is a good approach to others.

[SPDX]:https://spdx.org/licenses/

xsteadfastx:
@Ikke do you mean with make install? or just grabbing the libcanberra-pulse.so? that last didnt worked for me... abuild even throws a error because of the rpath. i know this is ugly but i really want to get pavucontrol into alpine.

Ikke:
@xsteadfastx I mean that I ran abuild without the adjustment and it still does not complain for me.

xsteadfastx:
> @xsteadfastx I mean that I ran abuild without the adjustment and it still does not complain for me.

maybe we talk about different things... but i tested it on some boxes and with the seperated APKBUILD file (the one im posted and not the one in the PR) does not set the RPATH only if i run make install first. ok i dont know... maybe i just dont get this things.

maxice8:
ping

xsteadfastx:
>     * for commit [3ccbfdd](https://github.com/alpinelinux/aports/commit/3ccbfdda22be46e3ef863072b8dcdbaed21c64d9):
> 
> 
> ```
> Use present tense (adds -> add)
> ```
> 
>     * for file main/libcanberra/APKBUILD:
> 
> 
> ```
> LGPL-2.1+ should be LGPL-2.1-or-later
> empty variables can be removed for brevity
> ```

done... but still... the major problem is that `pulseaudio-dev` is not in main and it doesnt build because of that.

maxice8:
That requires either moving libcanberra to community/ where pulseaudio is or moving pulseaudio to main where libcanberra is

Ikke:
For the record, libcanberra has been moved to community and pulse support has been added as well.

xsteadfastx:
oh cool :) brilliant

PR #6937, opened 2019-04-04
main/unbound: use dns-root-hints

tcely:


tcely:
@rnalrd This is the last resolver using its own hints file.

rnalrd:
Thanks

Will review it tomorrow unless someone else will do it first. I'm currently out of office.


Il giorno 15 apr 2019, 17:18, alle ore 17:18, tcely <notifications@github.com> ha scritto:

main/unbound/APKBUILD
```diff
@@ -77,20 +78,29 @@ package() {
...
 libs() {
 	pkgdesc="unbound shared libraries"
+	depends="$depends_libs"
 
 	mkdir -p "$subpkgdir"/usr/lib
 	mv "$pkgdir"/usr/lib/lib*.so.* "$subpkgdir"/usr/lib/
 }
 
+openrc() {
+	depends="$depends_openrc"
```
rnalrd: I don't understand why is necessary a custom openrc(). Also $depends_openrc is uninitialized/empty.

main/unbound/APKBUILD
```diff
@@ -69,10 +67,9 @@ package() {
 
 	cd "$pkgdir"
 
-	install -Dm755 "$srcdir"/update-unbound-root-hints \
-		./etc/periodic/monthly/update-unbound-root-hints
-
-	install -m644 -D "$srcdir"/root.hints ./etc/unbound/root.hints
+	mkdir -p ./etc/unbound
+	rm -f ./etc/unbound/root.hints
+	ln -s ../../usr/share/dns-root-hints/named.root ./etc/unbound/root.hints
```
rnalrd: Why not pointing to an absolute path?

tcely:
> pkgrel should be bumped

It was. What am I missing?

rnalrd:
This and last commit does bump pkgrel:

https://github.com/alpinelinux/aports/pull/6937/commits/53baea25220fda5d590c69c37bf194b3f61a221b

rnalrd:
Anyway, it doesn't matter anymore, @HRio commit was merged and pkgrel bumped.
Please see the remaining comments.

main/unbound/APKBUILD
tcely: > cleanup depends for subpackages

depends get inherited unless reset.

main/unbound/APKBUILD
tcely: An absolute path only works when the target exists in exactly one location. This way the link works in chroot or under $srcdir too.

tcely:
@clandmeter This patching approach is fairly fragile.

https://cloud.drone.io/alpinelinux/aports/1561

tcely:
@rnalrd changes done

main/unbound/APKBUILD
```diff
@@ -10,7 +10,8 @@ arch="all"
 license="BSD-3-Clause"
 depends="dns-root-hints dnssec-root"
 depends_dev="expat-dev"
-makedepends="$depends_dev libevent-dev openssl-dev python2-dev swig linux-headers"
+depends_migrate="/bin/sh apk-tools dns-root-hints openrc"
```
rnalrd: Afaik this not an official APKBUILD var, hence I would prefix it with "_".

main/unbound/APKBUILD
rnalrd: Right even though so far many if not all openrc packages just inherit depends from main pkg.

I'd rather use 'depends=""'. Using an uninitialized variable that looks an official APKBUILD variable it's kinda misleading.

tcely: It should not. In the future it will not. See https://github.com/alpinelinux/abuild/commit/801578a06a0ee602f3533bb0184c0b58b96e3fa1#diff-88af1a35b35ecf6afcb7687f5afc7bbc

main/unbound/APKBUILD
tcely: I would not. This would complicate efforts to automatically handle `depends_*` variables when calculating depends in abuild.

PR #6939, opened 2019-04-04
main/dns-root-hints: upgrade to 2019073000

tcely:
Follows on from https://github.com/alpinelinux/aports/pull/5950#issuecomment-479919563


tcely:
@rnalrd Please explain the hold label you added.

tcely:
This package is now out of date.

2019070301

ncopa:
We need to store a copy of `named.root`, or it will not be possible to rebuild the package when upstream updates it.

Maybe we can re-fetch it and compare in the check function or something?

ncopa:
I think this should be updated via cron job, not via apk package. We need provide package to be able to boostrap it though.

rnalrd:
"snapshot" does that.

I did not bother to upgrade because even though version upstream changed, there's just changes in whitespace for named.root.


Il giorno 24 lug 2019, 10:48, alle ore 10:48, Natanael Copa <notifications@github.com> ha scritto:

tcely:
> We need to be able to rebuild the package even if new version is available from upstream, so we need to keep a copy of the stale data.

Why? If this is actually something we need, where do we archive old sources that are no longer published?
This is not a unique case, any software that only publishes the current version would be the same.

PR #6946, opened 2019-04-04
main/linux-vanilla: support xen_pvh boot via grub

HRio:
Patches from Debian Buster (backport from Linux 4.20)

### DIFF
--- a/main/linux-vanilla/APKBUILD
+++ b/main/linux-vanilla/APKBUILD
@@ -7,7 +7,7 @@ case $pkgver in
 	*.*.*)	_kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=0
+pkgrel=1
 pkgdesc="Linux vanilla kernel"
 url="http://kernel.org"
 depends="mkinitfs"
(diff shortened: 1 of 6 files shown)


HRio:
Tested both linux-vanilla and linux-virt from xen_pvh grub.

ncopa:
if we start add patches to our kernel, can we really call it "vanilla" then?

ncopa:
I think the proper thing to do is to ask upstream backport those for linux-4.19.y

HRio:
> if we start add patches to our kernel, can we really call it "vanilla" then?

OK, understand.

HRio:
@jgross1 what would you say, are the "RSDP address" patches LTS 4.19 "material" or should we wait for the next LTS? i.e. do you think it could be accepted for stable backport?

jgross1:
On 05/04/2019 17:45, Henrik Riomar wrote:
> @jgross1 <https://github.com/jgross1> what would you say, are the "RSDP
> address" patches LTS 4.19 "material" or should we wait for the next LTS?
> i.e. do you think it could be accepted for stable backport?

You'd need to ask the stable maintainer (Greg), not me. :-)

Basically they add a feature, so in theory they are no backport
candidates. But maybe Greg will take them nevertheless, who knows?


Juergen

HRio:
> You'd need to ask the stable maintainer (Greg), not me. :-) Basically they add a feature, so in theory they are no backport candidates.

Thanks.

Found some rules now after after some more searching for how an end user could possibly request a Linux LTS backport :smile: 

https://www.kernel.org/doc/html/v4.19/process/stable-kernel-rules.html seems they are way too big for the stable series.

PR #6948, opened 2019-04-04
WIP testing/howard-bc: bump to 1.2.6

gavinhoward:
This version includes support for locales, so I also added a `$pkgname-lang` subpackage. I also removed a warning.

### DIFF
--- a/testing/howard-bc/APKBUILD
+++ b/testing/howard-bc/APKBUILD
@@ -2,19 +2,19 @@
 # Maintainer: Gavin D. Howard <yzena.tech@gmail.com>
 pkgname=howard-bc
 _pkgname=bc
-subpackages="$pkgname-doc"
-pkgver=1.1.4
+pkgver=1.2.6
 pkgrel=0
 pkgdesc="POSIX bc with GNU extensions"
 url="https://github.com/gavinhoward/bc"
 arch="all"
 license="BSD-2-Clause"
+subpackages="$pkgname-doc"
 source="$pkgname-$pkgver.tar.xz::https://github.com/gavinhoward/$_pkgname/releases/download/$pkgver/$_pkgname-$pkgver.tar.xz"
 builddir="$srcdir/$_pkgname-$pkgver"
 
 build() {
 	cd "$builddir"
-	PREFIX=/usr DESTDIR="$pkgdir" EXECSUFFIX=-howard ./configure.sh -G
+	PREFIX=/usr DESTDIR="$pkgdir" EXECSUFFIX=-howard ./configure.sh -GN
 	make
 }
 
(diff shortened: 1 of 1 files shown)


testing/howard-bc/APKBUILD
```diff
@@ -32,11 +32,13 @@ doc() {
 	# Doc files
 	install -Dm644 "$builddir"/LICENSE.md \
 			"$subpkgdir"/usr/share/doc/$pkgname/LICENSE.md
+
+	rm -rf "$pkgdir"/usr/share/man/
```
andypost: why it removes? I guess  packaging should do the trick

gavinhoward: This was the line that removed a warning. The warning said

```
>>> WARNING: howard-bc*: Found /usr/share/man but package name doesn't end with -doc
```

That seemed the right thing to do since it already has a `*-doc` package, but you know better.

testing/howard-bc/APKBUILD
```diff
@@ -2,8 +2,8 @@
 # Maintainer: Gavin D. Howard <yzena.tech@gmail.com>
 pkgname=howard-bc
 _pkgname=bc
-subpackages="$pkgname-doc"
-pkgver=1.1.4
+subpackages="$pkgname-doc $pkgname-lang"
```
prspkt: please place `subpackages` between `license` and `source`

testing/howard-bc/APKBUILD
prspkt: please remove the custom `doc` function

testing/howard-bc/APKBUILD
```diff
@@ -32,11 +32,13 @@ doc() {
 	# Doc files
 	install -Dm644 "$builddir"/LICENSE.md \
 			"$subpkgdir"/usr/share/doc/$pkgname/LICENSE.md
+
+	rm -rf "$pkgdir"/usr/share/man/
 }
 
 package() {
 	cd "$builddir"
 	make install
```
prspkt:
```suggestion
	make install
	
	install -Dm644 LICENSE.md \
		"$pkgdir"/usr/share/doc/$pkgname/LICENSE.md
```

gavinhoward:
I will make the suggested changes. However, I found a problem with the build system: `gencat` can't be run offline because its files are [not guaranteed to be portable][1]. Unfortunately, Alpine does not have `gencat` (it is provided by glibc of all things...), so I may have to remove the `*-lang` subpackage.

Until the problem is fixed, I am going to label this a WIP.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/gencat.html#tag_20_51_16

testing/howard-bc/APKBUILD
tcely: Is this removing empty directories or are some files not being moved to the `-doc` package?

gavinhoward: I think it is an empty directory.

testing/howard-bc/APKBUILD
gavinhoward: Should be resolved by the recent changes.

gavinhoward:
Ready for more review.

PR #6949, opened 2019-04-04
testing/perl-time-parsedate replaces main/perl-time-modules

timlegge:
According to https://metacpan.org/changes/distribution/Time-ParseDate#L22 the Time::Modules was renamed to Time::Parsedate in 2013.0916 2013-09-16.  This allows the nwr perl-time-parsedate to replace tthe old perl-time-modules version that has not been updated since 2013


Ikke:
This would need a `pkgrel` bump to actually do something.

PR #6951, opened 2019-04-05
community/bareos: Fix segfaults on startup and job run

bratkartoffel:
This PR fixes two undefined behaviours with pthreads leading to
segfaults.

Closes Bug #10156
https://bugs.alpinelinux.org/issues/10156
Additionally opened upstream issue: https://bugs.bareos.org/view.php?id=1073


bratkartoffel:
bareos-dir now starts up and I can connect with bconsole.
Commands seem to work, but there are still segfaults when a job starts.

PR #6952, opened 2019-04-05
testing/hiawatha: add letsencrypt subpkg / helper

itoffshore:
* moves the letsencrypt `php` scripts to a subpkg for running outside
  containers

* adds a wrapper script to detect a `php` interpreter

### DIFF
--- a/testing/hiawatha/APKBUILD
+++ b/testing/hiawatha/APKBUILD
@@ -1,18 +1,21 @@
 # Maintainer: Kurt Marasco <celilo@lavabit.com>
 # Contributor: Pascal Ernster <aur at hardfalcon dot net>
+# Contributor: Stuart Cardall <developer at it-offshore dot co.uk>
 pkgname=hiawatha
 pkgver=10.9
-pkgrel=0
+pkgrel=1
 pkgdesc='Secure and advanced webserver'
 url='https://www.hiawatha-webserver.org/'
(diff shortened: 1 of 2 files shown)


testing/hiawatha/letsencrypt.helper
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+phplist="php5 php7 php8"
```
andypost: Only php7 is supported, no reason to have 5 which is eol, and 8 which still in debates

testing/hiawatha/APKBUILD
```diff
@@ -48,6 +51,20 @@ package() {
     "$pkgdir"/usr/share/doc/hiawatha/hiawatha.conf.sample
 }
 
+letsencrypt() {
+  subpkgarch=noarch
```
andypost: It needs to point php7 as dependency

testing/hiawatha/letsencrypt.helper
itoffshore: Until `suhosin` [supports `php7`](https://github.com/sektioneins/suhosin7) - I use `php5`

itoffshore: fixed

Ikke:
@andypost Is this ready to be merged?

PR #6972, opened 2019-04-07
community/gimp: upgrade to 2.10.10

J0WI:


community/gegl/APKBUILD
```diff
@@ -27,12 +27,12 @@ build() {
 
 check() {
 	cd "$builddir"
-	make check
+	#make check
```
rnalrd: If for whatever reasons tests can't be run, it's preferable that are disabled via "options=" and adding a comments why.
Thanks!

J0WI: Comitted accidentally, reverted now.

rnalrd:
Merged

PR #6986, opened 2019-04-08
main/libical: fix db.h: No such file or directory

ollieparanoid:
Make libical-dev depend on db-dev, so db.h is available when building
against libical-dev.

Fix this error when building programs that depend on libical-dev:

```
In file included from /home/pmos/build/src/kcalcore-18.12.3/src/icalformat.cpp:45:
/usr/include/libical/icalss.h:492:10: fatal error: db.h: No such file or directory
 #include <db.h>
          ^~~~~~
```

---

I have verified that this fixes building kcalcore:
https://gitlab.com/postmarketOS/pmaports/issues/229

Regression from https://github.com/alpinelinux/aports/commit/e6ee35d11eca7f4d046efe106cc7d5c19d8c76b3 (CC @brebs-gh, @rnalrd)

### DIFF
--- a/main/libical/APKBUILD
+++ b/main/libical/APKBUILD
@@ -1,14 +1,15 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=libical
 pkgver=3.0.4
-pkgrel=0
+pkgrel=1
 pkgdesc="Reference implementation of the iCalendar format"
 url="https://libical.github.io/libical/"
 arch="all"
 license="LGPL MPL"
 depends=""
+depends_dev="db-dev"
 # tzdata is needed for tests
-makedepends="db-dev perl cmake glib-dev icu-dev libxml2-dev tzdata"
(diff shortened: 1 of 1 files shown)


main/libical/APKBUILD
```diff
@@ -38,4 +38,12 @@ package() {
 	make DESTDIR="$pkgdir" install
 }
 
+# libical-dev needs db.h from db-dev, so depend on it
+# /usr/include/libical/icalss.h:492:10: fatal error: db.h: No such file or directory
+dev() {
+	default_dev
+	depends="$depends db-dev"
```
ncopa: `depends="$depends ..."` is problematic in split functions and should be avoided.
See http://lists.alpinelinux.org/alpine-devel/6582.html

Instead do:
````sh
depends_dev="db-dev"
makedepends="$depends_dev ..."
````

ollieparanoid:
@ncopa: thanks for the review, updated. Do you want to merge it?

PR #6996, opened 2019-04-09
testing/libyang: fix byte order on s390x

tuan-hoang1:
Copy upstream commits

### DIFF
--- a/testing/libyang/APKBUILD
+++ b/testing/libyang/APKBUILD
@@ -3,21 +3,27 @@
 pkgname=libyang
 pkgver=1.0_p2
 _realver=${pkgver/_p/-r}
-pkgrel=0
+pkgrel=1
 pkgdesc="YANG data modelling language parser and toolkit"
 url="https://github.com/CESNET/libyang"
-arch="all !s390x"
+arch="all"
 license="BSD-3-Clause-Clear"
 depends=""
 makedepends="bison cmake cmocka-dev flex pcre-dev"
 install=""
 subpackages="$pkgname-dev $pkgname-doc"
-source="${pkgname}-${pkgver}.tar.gz::https://github.com/CESNET/$pkgname/archive/v$_realver.tar.gz"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/CESNET/$pkgname/archive/v$_realver.tar.gz
+	5b3c324097fa3e122edb774213aeafd6eba137f9.patch
+	09e7159e4f91a39f3d19ba6590d558ab027a8a84.patch
+	"
 builddir="$srcdir/$pkgname-$_realver/build"
 
-build() {
+prepare() {
 	mkdir -p "$builddir"
-	cd "$builddir"
+	builddir="$builddir/.." default_prepare
+}
+
+build() {
 	if [ "$CBUILD" != "$CHOST" ]; then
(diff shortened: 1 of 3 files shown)


testing/libyang/APKBUILD
```diff
@@ -3,21 +3,24 @@
...
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/CESNET/$pkgname/archive/v$_realver.tar.gz
+	5b3c324097fa3e122edb774213aeafd6eba137f9.patch
+	09e7159e4f91a39f3d19ba6590d558ab027a8a84.patch
+	"
+builddir="$srcdir/$pkgname-$_realver"
 
 build() {
-	mkdir -p "$builddir"
-	cd "$builddir"
+	mkdir -p "$builddir/build"
```
Ikke: Would it make sense to do the `mkdir` in a prepare function and then change the builddir to this dir so you don't have to repeat that all the time?

andypost: This change is really out of scope of fix, I also see no reason in dir change

tuan-hoang1: @Ikke : I also tried that option. But `package()` function does not respect the `$builddir` change while `check()` does. So I also need to `cd "$builddir"/build" in `package()`.

tuan-hoang1: @andypost : with addition `.patch` files, `$builddir` needs to be `$srcdir/$pkgname-$pkgver` instead of `$srcdir/$pkgname-$pkgver/build`.

Ikke: @tuan-hoang1 This patch works for me:

```diff
diff --git a/testing/libyang/APKBUILD b/testing/libyang/APKBUILD
index e940e66579..d13dec4d50 100644
--- a/testing/libyang/APKBUILD
+++ b/testing/libyang/APKBUILD
@@ -16,11 +16,15 @@ source="${pkgname}-${pkgver}.tar.gz::https://github.com/CESNET/$pkgname/archive/
        5b3c324097fa3e122edb774213aeafd6eba137f9.patch
        09e7159e4f91a39f3d19ba6590d558ab027a8a84.patch
        "
-builddir="$srcdir/$pkgname-$_realver"
+builddir="$srcdir/$pkgname-$_realver/build"
+
+prepare() {
+       mkdir -p "$builddir"
+
+       builddir="$builddir/.." default_prepare
+}

 build() {
-       mkdir -p "$builddir/build"
-       cd "$builddir/build"
        if [ "$CBUILD" != "$CHOST" ]; then
                CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
        fi
@@ -38,12 +42,10 @@ build() {
 }

 check() {
-       cd "$builddir/build"
        make test
 }

 package() {
-       cd "$builddir/build"
        make DESTDIR="$pkgdir" install
 }
```

tuan-hoang1: I'm surprised that we allow this : `builddir="$builddir/.." default_prepare` but if you say okay then I'm okay. Upating the patch ... Many thanks.

Ikke: Not sure if desired to do it that way. It was more a quick fix to get it at least working.

It's basically a choice, either do a `cd` in each function or override `builddir` for `default_prepare`.

PR #6997, opened 2019-04-09
community/rspamd: upgrade to 1.9.1

TBK:


TBK:
@bratkartoffel https://github.com/rspamd/rspamd/blob/master/contrib/DEPENDENCY_INFO.md and Rspamd itself is under Apache-2.0.

bratkartoffel:
The DEPENDENCY_INFO.md lists the licences of the dependencies, not the package itself. Thus I think that only Apache-2.0 should be listed as licence in APKBUILD.

TBK:
Ask @jirutka the reasoning.

jirutka:
IIRC these dependencies are bundled with rspamd, so they have to be included in the license info.

PR #7003, opened 2019-04-09
community/openjdk7: security upgrade to 7.211.2.6.17

J0WI:


community/openjdk7/icedtea-jdk-fix-freetype-detection.patch
```diff
@@ -0,0 +1,20 @@
+--- openjdk/jdk/make/tools/freetypecheck/freetypecheck.c.orig
```
bratkartoffel: I think the first two lines are wrong. Shouldn't it be different in prefix, not suffix?
e.g.:
```
--- old/openjdk/jdk/make/tools/freetypecheck/freetypecheck.c
+++ new/openjdk/jdk/make/tools/freetypecheck/freetypecheck.c
```

community/openjdk7/APKBUILD
```diff
@@ -20,12 +21,12 @@ makedepends="bash findutils tar zip file paxmark gawk util-linux libxslt
...
 	libxrender-dev alsa-lib-dev freetype-dev fontconfig-dev
 	gtk+2.0-dev krb5-dev attr-dev pcsc-lite-dev lksctp-tools-dev
-	libxcomposite-dev nss-static"
+	libxcomposite-dev nss-static xz"
 install=""
 
 # upstream binary versions
 RHINO_VER=1.7.7.2
-ANT_VER=1.9.11
+ANT_VER=1.9.14
```
bratkartoffel: Is there any particular reason to increment the antversion too?

J0WI: Is there any particular reason to use the old version? :stuck_out_tongue_winking_eye:

bratkartoffel: Except for the "never change a running (build) system"? None

J0WI: Done.

J0WI:
Fixed in 6759502eb830da958c990d6f74189d91562ac2d2

PR #7014, opened 2019-04-10
testing/rpm2cpio: new aport

TBK:


tcely:
A `check` function that downloads a RPM and checks that `cpio -it` can list the output of this script seems like a good addition to me.

TBK:
@tcely sounds like a good idea. I try to find a suitable small rpm and make the check function.

tcely:
I'd suggest something like http://mirror.centos.org/centos/7/os/x86_64/Packages/centos-bookmarks-7-1.el7.noarch.rpm

testing/rpm2cpio/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+pkgrel=0
+pkgdesc="Convert .rpm files to cpio format"
+url="https://www.freshports.org/archivers/rpm2cpio/"
+arch="noarch"
+license="BSD-2-Clause"
+options="!check" # No test suite
+depends="libarchive-tools"
+source="rpm2cpio.sh"
+
+build() {
```
Ikke: This `build` function can be removed when it does nothing.

TBK: @Ikke Fixed.

PR #7016, opened 2019-04-10
main/git: minor APKBUILD improvements

oliv3r:
Two small improvements to git

### DIFF
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -20,6 +20,7 @@ depends=""
 # we need tcl and tk to be built before git due to git-gui and gitk
 makedepends="zlib-dev openssl-dev curl-dev expat-dev perl-dev python2-dev
 	pcre2-dev asciidoc xmlto perl-error tcl tk"
+options="!check" # Check is currently failing on t3900-i18n-commit.sh ISO-2022-JP.txt
 # note that order matters
 subpackages="$pkgname-doc
 	$pkgname-bash-completion:completion:noarch
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -11,7 +11,7 @@
 #   - CVE-2017-1000117
 pkgname=git
 pkgver=2.21.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Distributed version control system"
 url="https://www.git-scm.com"
 arch="all"
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -41,7 +41,6 @@ subpackages="$pkgname-doc
 	$pkgname-diff-highlight:diff_highlight:noarch
 	"
 source="https://www.kernel.org/pub/software/scm/git/git-$pkgver.tar.xz
-	bb-tar.patch
 	git-daemon.initd
 	git-daemon.confd
 	"
@@ -267,6 +266,5 @@ _perl_config() {
 }
 
 sha512sums="83f57c3950a07f6773a3aea66611d22daba0e5599e5d8f0751a16f6fdbeab0f3844d942a39a5642051212df99d1d4513253c36829b1454b4f0977cc6026fd973  git-2.21.0.tar.xz
-85767b5e03137008d6a96199e769e3979f75d83603ac8cb13a3481a915005637409a4fd94e0720da2ec6cd1124f35eba7cf20109a94816c4b4898a81fbc46bd2  bb-tar.patch
 89528cdd14c51fd568aa61cf6c5eae08ea0844e59f9af9292da5fc6c268261f4166017d002d494400945e248df6b844e2f9f9cd2d9345d516983f5a110e4c42a  git-daemon.initd
 fbf1f425206a76e2a8f82342537ed939ff7e623d644c086ca2ced5f69b36734695f9f80ebda1728f75a94d6cd2fcb71bf845b64239368caab418e4d368c141ec  git-daemon.confd"
--- a/main/git/APKBUILD
+++ b/main/git/APKBUILD
@@ -32,6 +32,7 @@ subpackages="$pkgname-doc
 	$pkgname-cvs::noarch
 	$pkgname-p4::noarch
 	$pkgname-daemon
+	$pkgname-daemon-openrc:daemon_openrc
 	$pkgname-gitweb::noarch
 	$pkgname-subtree::noarch
 	$pkgname-subtree-doc:subtree_doc:noarch
@@ -158,8 +159,15 @@ daemon() {
 		"$pkgdir"/$_gitcoredir/git-http-backend \
 		"$pkgdir"/$_gitcoredir/git-shell \
 		"$subpkgdir"/$_gitcoredir/
+}
 
-	mv "$pkgdir"/etc "$subpkgdir"/
+daemon_openrc() {
+	pkgdesc="Git protocol daemon"
+
+	default_openrc
+
+	depends="git-daemon=${pkgver}-r${pkgrel} ${depends_openrc}"
+	install_if="openrc ${pkgname}-daemon=${pkgver}-r${pkgrel}"
 }
 
 gitweb() {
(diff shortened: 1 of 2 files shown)


main/git/APKBUILD
```diff
@@ -32,6 +32,7 @@ subpackages="$pkgname-doc
 	$pkgname-cvs::noarch
 	$pkgname-p4::noarch
 	$pkgname-daemon
+	$pkgname-daemon-openrc:daemon_openrc:noarch
```
tcely: noarch should not be needed here.

main/git/APKBUILD
```diff
@@ -166,8 +166,16 @@ daemon() {
 		"$pkgdir"/$_gitcoredir/git-http-backend \
 		"$pkgdir"/$_gitcoredir/git-shell \
 		"$subpkgdir"/$_gitcoredir/
+}
+
+daemon_openrc() {
+	pkgdesc="Git protocol daemon (OpenRC init scripts)"
+	depends="git-daemon=${pkgver}-r${pkgrel} ${depends_openrc}"
```
tcely: The depends on git-daemon does not make sense. This should be installed automatically after git-daemon.

main/git/APKBUILD
```diff
@@ -166,8 +166,16 @@ daemon() {
...
+	install_if="openrc ${pkgname}-daemon=${pkgver}-r${pkgrel}"
+	replaces="git"
```
tcely: Why would this replace git?!

main/git/APKBUILD
```diff
@@ -166,8 +166,16 @@ daemon() {
...
 
-	mv "$pkgdir"/etc "$subpkgdir"/
+	mkdir -p "${subpkgdir}/etc"
+	mv "${pkgdir}/etc" "${subpkgdir}/"
```
tcely: Using default_openrc is a better idea.

main/git/APKBUILD
```diff
@@ -11,7 +11,7 @@
 #   - CVE-2017-1000117
 pkgname=git
 pkgver=2.21.0
-pkgrel=0
+pkgrel=2
```
tcely: Why skip over 1?

main/git/APKBUILD
tcely: Let me be more precise. Why would git-daemon-openrc ever replace git?

git has all the command binaries and -openrc should only have conf and init data files.

main/git/APKBUILD
oliv3r: alright, good point

main/git/APKBUILD
oliv3r: Well if as a user I install git-daemon-openrc; I would expect it to pull in git-daemon right? and openrc, as i want an openrc enabled git-daemon so i need both.
however, if I install git-daemon, I only want to pull in git-daemon-openrc if openrc is already part of my installation.
From other APKBUILDS I gather this to be what is needed?

main/git/APKBUILD
oliv3r: why would git-daemon replace git? :p I wasn't sure on this one, I wasn't quite sure on the logic in the other cases, as the same holds for git-daemon. So either this was intended, or it's wrong for most/all other cases as well; and if that is so, I'll happily add a patch to this series to remove the other cases

main/git/APKBUILD
oliv3r: but doesn't work sadly. Because the openrc package is specifically for git-daemon, abuild doesn't seem to pick up on it. It may be that it only works for the base package? I'm happy to try again if you are convinced this should work with default_openrc. Would an alternative solution work where instead of using _default_openrc I call the external actual function? I may have not tried that combination.

main/git/APKBUILD
oliv3r: since this is a series of patches, each commit ups the version; if you prefer to have a final bump as a separate commit that works too; chronologically this makes more sense of course; but I've noticed git history for the aports are not that atomic/chronological, so either is fine with me.

main/git/APKBUILD
tcely: -openrc like -doc isn't typically meant to be installed by hand. So the direction of I'm installing -doc so install the origin package too isn't setup.

If you installed/install docs all the installed software triggers a bunch of -doc subpackages to be installed too. Similarly for the openrc package in this case.

main/git/APKBUILD
tcely: I'd be happy to look over other cases too.

For now, please remove that replaces line.

main/git/APKBUILD
tcely: The detection for the subpackage is handled by providing the daemon_openrc function. You can call the default_openrc function from within your function to use the logic from abuild instead of moving etc yourself.

main/git/APKBUILD
tcely: I won't nit about which commit has the pkgrel bump. I was just curious why you incremented by more than one.

main/git/APKBUILD
oliv3r: I get that, but as a user I have few options:
* install git-daemon (whilst openrc is already installed) -> i get git-daemon-openrc
* install git-daemon-openrc, which pulls in all dependencies (openrc, git-daemon)
* install git-daemon (whilst openrc is NOT installed) -> I get neither git-daemon-openrc nor openrc

I was under the impression that this solution would do exactly that; so where would this be going wrong then (ignoring the 'not using default_openrc case for the moment)?

main/git/APKBUILD
oliv3r: Ok; i'll remove it here, and add a patch removing it for the other cases as well.

main/git/APKBUILD
oliv3r: ah, right; so keep the function, but call default_openrc; that makes sense.

main/git/APKBUILD
oliv3r: ah right, so each patch in the series bimps it by one; but i'm happy to have a bump patch to complete the series to bump it just by one.

main/git/APKBUILD
tcely: I believe you have those cases correct.

The second one (install git-daemon-openrc)  is not something that is typically handled as it is the opposite direction of how we expect these dependencies to work.

oliv3r: well both scenario's should work now then :)

we do indeed and should expect what you suggest. A user will more then often have openrc installed and thus when installing the daemon will pull in the -openrc package.

In the container world; only the daemon bit will be installed, so without openrc no problem.

I think this is accomplished now, and we'd only go out of our way, to not support the use case where someone lazy just pulls daemon-openrc. It should 'just work' (tm) imo. Even if it is not expected.

oliv3r:
@tcely I adjusted the use of replace; still not 100% if it still does what it should do. It looks and seems sensible to me ...

main/git/APKBUILD
```diff
@@ -159,21 +153,26 @@ p4() {
...
 	mkdir -p "$subpkgdir"/$_gitcoredir
 	mv "$pkgdir"/$_gitcoredir/git-daemon \
 		"$pkgdir"/$_gitcoredir/git-http-backend \
 		"$pkgdir"/$_gitcoredir/git-shell \
 		"$subpkgdir"/$_gitcoredir/
+}
+
+daemon_openrc() {
+	pkgdesc="Git protocol daemon"
+	depends="git-daemon=${pkgver}-r${pkgrel} ${depends_openrc}"
```
tcely: You should move depends below the call to default_openrc so that it is not reset by that function.

oliv3r: ah yes, missed that one!

main/git/APKBUILD
```diff
@@ -159,21 +153,26 @@ p4() {
...
 	mv "$pkgdir"/$_gitcoredir/git-daemon \
 		"$pkgdir"/$_gitcoredir/git-http-backend \
 		"$pkgdir"/$_gitcoredir/git-shell \
 		"$subpkgdir"/$_gitcoredir/
+}
+
+daemon_openrc() {
+	pkgdesc="Git protocol daemon"
+
+	default_openrc
```
tcely: `rmdir "$pkgdr"/etc || :`

oliv3r:
@tcely done; test and merging time then :)

main/git/APKBUILD
clandmeter: Why is this rmdir needed?

tcely: It's not strictly necessary. Removing the empty `etc` directory is just cleaner.

PR #7017, opened 2019-04-10
community/mtd-utils: Promote to community status

oliv3r:
The mtd-utils seem use, updates and are in a good shape. Lets put them
in the community repo.

### DIFF
--- a/community/mtd-utils/APKBUILD
+++ b/community/mtd-utils/APKBUILD
@@ -5,31 +5,33 @@ pkgrel=0
 pkgdesc="Utilities for handling MTD devices, and for dealing with FTL, NFTL JFFS2, etc."
 url="http://www.linux-mtd.infradead.org/"
 arch="all"
-license="GPL-2.0"
-makedepends="linux-headers lzo-dev util-linux-dev zlib-dev acl-dev"
-subpackages="$pkgname-doc"
-source="ftp://ftp.infradead.org/pub/mtd-utils/$pkgname-$pkgver.tar.bz2"
+license="GPL-2.0-only"
+makedepends="
+	acl-dev
+	linux-headers
+	lzo-dev
+	util-linux-dev
+	zlib-dev
+"
+subpackages="${pkgname}-doc"
+source="
+	ftp://ftp.infradead.org/pub/mtd-utils/${pkgname}-${pkgver}.tar.bz2
+"
 
-builddir="$srcdir"/$pkgname-$pkgver
-
-build() {
-	cd "$builddir"
+build()
+{
 	./configure \
-	            --build=$CBUILD \
-	            --host=$CHOST \
-	            --prefix=/usr \
-	            --sysconfdir=/etc \
-	            --mandir=/usr/share/man \
-		|| return 1
-	make \
-		|| return 1
+	            --build="${CBUILD}" \
+	            --host="${CHOST}" \
+	            --mandir="/usr/share/man" \
+	            --prefix="/usr" \
+	            --sysconfdir="/etc"
+	make
 }
 
-package() {
-	cd "$builddir"
-	make install -j1 \
-	     DESTDIR="$pkgdir" \
-		|| return 1
+package()
+{
+	make DESTDIR="${pkgdir}" install
 }
 
 md5sums="70b80fccdfc1b536a91deefb91f4e2be  mtd-utils-2.0.0.tar.bz2"
--- a/community/mtd-utils/APKBUILD
+++ b/community/mtd-utils/APKBUILD
@@ -1,6 +1,6 @@
 # Maintainer: Olliver Schinagl <oliver@schinagl.nl>
 pkgname=mtd-utils
-pkgver=2.0.0
+pkgver=2.1.0
 pkgrel=0
 pkgdesc="Utilities for handling MTD devices, and for dealing with FTL, NFTL JFFS2, etc."
 url="http://www.linux-mtd.infradead.org/"
@@ -8,18 +8,28 @@ arch="all"
 license="GPL-2.0-only"
 makedepends="
 	acl-dev
+	autoconf
+	automake
+	libtool
 	linux-headers
 	lzo-dev
+	openssl-dev
 	util-linux-dev
 	zlib-dev
 "
 subpackages="${pkgname}-doc"
+_githash="1dba7944fe18978415a3ffc43932359a36b99b25"
 source="
-	ftp://ftp.infradead.org/pub/mtd-utils/${pkgname}-${pkgver}.tar.bz2
+	${pkgname}-${pkgver}.tar.gz::http://git.infradead.org/mtd-utils.git/snapshot/${_githash}.tar.gz
+	0001-unittests-test_lib-Include-proper-header-for-_IOC_SI.patch
+	0002-unittests-libmtd_test-Include-fcntl-header.patch
+	0003-unittests-Define-the-use-of-_GNU_SOURCE.patch
 "
+builddir="${srcdir}/${pkgname}-$(echo "${_githash}" | cut -c1-7)"
 
 build()
 {
+	./autogen.sh
 	./configure \
 	            --build="${CBUILD}" \
 	            --host="${CHOST}" \
@@ -34,6 +44,7 @@ package()
 	make DESTDIR="${pkgdir}" install
 }
 
-md5sums="70b80fccdfc1b536a91deefb91f4e2be  mtd-utils-2.0.0.tar.bz2"
-sha256sums="20a377aa86af4c102e930b0d0bb9859e3a10aeb16445f6a1273989b206a717c8  mtd-utils-2.0.0.tar.bz2"
-sha512sums="52d3fab58d254fd672f53b3f1868b19ca572a8f257fc86519d7afc7979776afa37c0f21d63cc9e9cdc2e342f1cfb2d077b07847f18f6484dc660077ee347f2fb  mtd-utils-2.0.0.tar.bz2"
+sha512sums="bb64b6cbdaee3c731ed1dc136064507fbfffe8a45879b0c4024a2b46d320875252f42af8dc268b30247ebb81a7a89b951e19a896bb415de38877c45b031e1081  mtd-utils-2.1.0.tar.gz
+70013f50fad529e94733cd86349d4efe4d99c5febd7c407ab3db0c2995fd6be0524b3ebc59537ff13a4be9a8f63100468a9faa8590ea4d84f79995398b0fd7d9  0001-unittests-test_lib-Include-proper-header-for-_IOC_SI.patch
+eb6d72e2810218c90044f562ec2d58543f42066ca4719ec5c1199dfdebeacacbde931b59ad12a39df7834fb2b8cef240bed839840c11b3cac9761dd44e63bca8  0002-unittests-libmtd_test-Include-fcntl-header.patch
+e29c485a96b959c26f39a0a05a78e2a562f2bda0a2acc7ed532b6ebda46139f9e4c88b4df0f9dc13051dd5aa50ad60f82880f3da412b59d46fca527a88a30f5d  0003-unittests-Define-the-use-of-_GNU_SOURCE.patch"
--- a/community/mtd-utils/APKBUILD
+++ b/community/mtd-utils/APKBUILD
@@ -1,4 +1,4 @@
-# Maintainer:
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
 pkgname=mtd-utils
 pkgver=2.0.0
 pkgrel=0
--- a/community/mtd-utils/APKBUILD
+++ b/community/mtd-utils/APKBUILD
@@ -10,6 +10,7 @@ makedepends="
 	acl-dev
 	autoconf
 	automake
+	cmocka-dev
 	libtool
 	linux-headers
 	lzo-dev
@@ -32,11 +33,18 @@ build()
 	./autogen.sh
 	./configure \
 	            --build="${CBUILD}" \
+	            --enable-tests \
+	            --enable-unit-tests \
 	            --host="${CHOST}" \
 	            --mandir="/usr/share/man" \
 	            --prefix="/usr" \
 	            --sysconfdir="/etc"
 	make
+}
+
+check()
+{
+	make check
 }
 
 package()
--- a/community/mtd-utils/APKBUILD
+++ b/community/mtd-utils/APKBUILD
@@ -4,7 +4,7 @@ pkgver=2.1.0
 pkgrel=0
 pkgdesc="Utilities for handling MTD devices, and for dealing with FTL, NFTL JFFS2, etc."
 url="http://www.linux-mtd.infradead.org/"
-arch="all"
+arch="noarch"
 license="GPL-2.0-only"
 makedepends="
 	acl-dev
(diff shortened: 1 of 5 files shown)


oliv3r:
I have no idea why the build is failing, but it's missing to find some travis shell file, so that's not from the APKBUILD

rnalrd:
There's no maintainer declared, would you take it?
Thanks!

rnalrd:
Also please add the repo before the package name in the commit message: "community/mtd-utils: ..."

tcely:
@oliv3r The TravisCI build failed trying to fetch the source tarball. Is there perhaps a better mirror (maybe with HTTPS) we could switch to?

For the commit message I'd recommend: `community/mtd-utils: move from testing`

oliv3r:
> There's no maintainer declared, would you take it?
> Thanks!

sure thing; as such, I aso bumped the version

oliv3r:
> Also please add the repo before the package name in the commit message: "community/mtd-utils: ..."

done; it just automatically filled this in from the githook before.

oliv3r:
> @oliv3r The TravisCI build failed trying to fetch the source tarball. Is there perhaps a better mirror (maybe with HTTPS) we could switch to?

Sadly, there is no official mirror, the cgit instance doesn't support snapshots nor https and there aren't any official mirrors.

Is there some retry mechanism we can trigger? (I understand this is super annoying of course)
> 
> For the commit message I'd recommend: `community/mtd-utils: move from testing`

done

tcely:
The source is <1MiB, maybe we just commit it in this repository until there is a better solution.

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+license="GPL-2.0"
+makedepends="
+	acl-dev
+	linux-headers
+	lzo-dev
+	openssl-dev
+	util-linux-dev
+	zlib-dev
+"
+subpackages="${pkgname}-doc"
```
TBK:
```suggestion
subpackages="$pkgname-doc"
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+arch="all"
+license="GPL-2.0"
+makedepends="
+	acl-dev
+	linux-headers
+	lzo-dev
+	openssl-dev
+	util-linux-dev
+	zlib-dev
+"
```
TBK:
```suggestion
    "
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+subpackages="${pkgname}-doc"
+source="ftp://ftp.infradead.org/pub/mtd-utils/${pkgname}-${pkgver}.tar.bz2"
```
TBK:
```suggestion
source="ftp://ftp.infradead.org/pub/mtd-utils/$pkgname-$pkgver.tar.bz2"
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	            --host="${CHOST}" \
+	            --prefix="/usr" \
+	            --sysconfdir="/etc" \
+	            --mandir="/usr/share/man" \
+		|| return 1
+	make \
+		|| return 1
+}
+
+package()
```
TBK:
```suggestion
package() {
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+{
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	make install -j1 \
```
TBK:
```suggestion
	make -j1 DESTDIR="$pkgdir" install -j1
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	     DESTDIR="${pkgdir}" \
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+		|| return 1
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	linux-headers
+	lzo-dev
+	openssl-dev
+	util-linux-dev
+	zlib-dev
+"
+subpackages="${pkgname}-doc"
+source="ftp://ftp.infradead.org/pub/mtd-utils/${pkgname}-${pkgver}.tar.bz2"
+
+build()
```
TBK:
```suggestion
build() {
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+{
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	./configure \
+	            --build="${CBUILD}" \
+	            --host="${CHOST}" \
+	            --prefix="/usr" \
+	            --sysconfdir="/etc" \
+	            --mandir="/usr/share/man" \
```
TBK:
```suggestion
	            --mandir="/usr/share/man"
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+		|| return 1
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	make \
```
TBK:
```suggestion
	make
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+		|| return 1
```
TBK:
```suggestion

```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	util-linux-dev
+	zlib-dev
+"
+subpackages="${pkgname}-doc"
+source="ftp://ftp.infradead.org/pub/mtd-utils/${pkgname}-${pkgver}.tar.bz2"
+
+build()
+{
+	./configure \
+	            --build="${CBUILD}" \
```
TBK:
```suggestion
	            --build="$CBUILD" \
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
...
+	            --host="${CHOST}" \
```
TBK:
```suggestion
	            --host="$CHOST" \
```

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,40 @@
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname=mtd-utils
+pkgver=2.1.0
+pkgrel=0
+pkgdesc="Utilities for handling MTD devices, and for dealing with FTL, NFTL JFFS2, etc."
+url="http://www.linux-mtd.infradead.org/"
+arch="all"
+license="GPL-2.0"
```
TBK:
```suggestion
license="GPL-2.0-or-later"
```

community/mtd-utils/APKBUILD
tcely: Was the duplicated `-j1` intended?

community/mtd-utils/APKBUILD
oliv3r: I know this is the 'style' used throughout a lot of APKBUILD's. But it's very poor style and is heavily frowned uppon :) so lets atleast make it proper where we are touching things anyway.

Proper shell variable usage is:
Use {} where possible, to clearly indicate the start/end of your variable; quote all strings.

community/mtd-utils/APKBUILD
oliv3r: can't change the license; the license been the same since 1991 and it's not or later from what I can see

community/mtd-utils/APKBUILD
oliv3r: copy paste thing whilst moving the install from in between to the end I suspect

community/mtd-utils/APKBUILD
oliv3r: I don't get that; would you write:
```sh
if () {
    do_stuff
      }
```

tcely: Just as you don't put the other continued lines in the first column I don't expect to see the ending quote there.

community/mtd-utils/APKBUILD
TBK: gpl-2 is not a SPDX license - https://spdx.org/licenses/

community/mtd-utils/APKBUILD
TBK: Was a mistake, both cases can properly be removed.

community/mtd-utils/APKBUILD
TBK: Only the following SPDX identifiers are acceptable for a GNU-GPL license:
GPL-3.0-or-later
GPL-3.0-only
GPL-2.0-or-later
GPL-2.0-only
GPL-1.0-or-later
GPL-1.0-only

GPL-2.0, is not. I looked at mtd-utils' COPYING file and it is the GPL-2.0-or-later variant.

community/mtd-utils/APKBUILD
TBK: I am all for "better styling", when the new org. structure is settled there should be plenty opportunity to make a working group which goal is to make/improve the style and improve the APKBUILD format.

community/mtd-utils/APKBUILD
Ikke: See the [spdx licence list][0] for reference

[0]:https://spdx.org/licenses/

community/mtd-utils/APKBUILD
oliv3r: > gpl-2 is not a SPDX license identifier - https://spdx.org/licenses/
> 
> https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license

Yes, I understand what you mean now. Actually, no you are wrong; but you are right.
the -only and -later monikers are for SDPX v3 (or some version) whereas currently we use the v2 (or was it v1?) SPDX license identifier.

However, I completely agree with you in that we should use the proper identifier. I was not aware that there was the -only variant, I thought GPL-v2 IS the -only variant. My bad.

oliv3r:
So fixed the license, and split the package into smaller subpackages.

Anything else?

tcely:
@oliv3r The TravisCI failure is the same problem fetching the source as I mentioned before.

oliv3r:
> @oliv3r The TravisCI failure is the same problem fetching the source as I mentioned before.
> 
> ```
> >>> mtd-utils: Fetching ftp://ftp.infradead.org/pub/mtd-utils/mtd-utils-2.1.0.tar.bz2
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
>   0  608k    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
> curl: (19) RETR response: 425
> >>> ERROR: mtd-utils: fetch failed
> ```

yeah I just realized it is the same repo; i thought I saw it on one of my other PR's as well. So yeah either mirror it, or can we add a retry loop for 'bad connections' in the travis script? I mean retrying 3 times shouldn't be a bad idea. Though RETR sounds like a retry error to begin with? I've been building this repo a few times, and didn't have that much trouble; maybe travis doesn't like ftp links?

tcely:
> yeah I just realized it is the same repo; i thought I saw it on one of my other PR's as well. So yeah either mirror it, or can we add a retry loop for 'bad connections' in the travis script? I mean retrying 3 times shouldn't be a bad idea. Though RETR sounds like a retry error to begin with? I've been building this repo a few times, and didn't have that much trouble; maybe travis doesn't like ftp links?

https://en.wikipedia.org/wiki/List_of_FTP_commands
`RETR` in FTP is for retrieve not retry.

I don't know why TravisCI is consistently failing with FTP for this file. The best option I came up with was to just check the source into this repository.

oliv3r:
So whoever wants to decide on what to do with the travis issue; Either we fix travis to retry (ftp) urls a few times before giving up; or we clone the repo as part of the APKBUILD. I personally prefer fixing travis and keeping the APKBUILD clean; but if needed be, can modify the the APKBUILD and make it pull the repo instead.

tcely:
I don't believe a retry is going to help. abuild-fetch already tries both curl and wget. If you want to try it, you can define a fetch function in the APKBUILD file.

community/mtd-utils/APKBUILD
```diff
@@ -0,0 +1,122 @@
...
+	mkdir -p "${subpkgdir}/usr/sbin"
+	find "${pkgdir}/usr/sbin/" -type f \(\
+	     -iname '*ubi*' -o \
+	     -iname 'lsmtd' -o -iname 'mtdinfo' \)\
+	     -exec mv "{}" "${subpkgdir}/usr/sbin/" +
+}
+
+package()
+{
+	depends="${subpackages##*-doc}"
```
Ikke: Only non-style issue I have is that this relies on the doc package being listed as first in the subpackages list, which might be non-obvious for others and can hide subpackages that are mentioned before.

I think the best way to handle this is to have a separate variable containing all the subpackages the main package depends on, and then include that both in the subpackages and depends variable.

oliv3r: fair enough; to be fair, I didnt' even properly test it either, so shame on me here. I just typed it; and wanted to check if it would remove it from any random spot. (and for any -doc package)

So if you prefer to have _subpackages I'll do that instead.

oliv3r:
> I don't believe a retry is going to help. abuild-fetch already tries both curl and wget. If you want to try it, you can define a fetch function in the APKBUILD file.

In that case, I think travis/curl/wget somehow always fail to get ftp files? Is there a maintainer/admin that can verify check this (or just put my claim to shame). If this is not fixable in the infra structure; i'll add a fetch function which will do the git-clone instead.

oliv3r:
> I don't believe a retry is going to help. abuild-fetch already tries both curl and wget. If you want to try it, you can define a fetch function in the APKBUILD file.

In that case, I think travis/curl/wget somehow always fail to get ftp files? Is there a maintainer/admin that can verify check this (or just put my claim to shame). If this is not fixable in the infra structure; i'll add a fetch function which will do the git-clone instead (which I still think is a little ugly, but if its the only way ...)

Ikke:
I've heard earlier that travis had issues with FTP, though a preliminary search seems to indicate that it should be supported. @clandmeter any idea?

oliv3r:
So I solved it differently, While we can't pull tags as snapshots from infradead, we can actually pull commitish based snapshots. So I changed the build to pull in the tag-hash-snapshot. This does add the complexity of having to run autogen.sh and generate the configure file, so in theory, it could deviate from upstream.

Upside is I can use the patches directly I submitted to upstream.

Ikke:
Merged as fff74d8119, c520aa0a98, 9a92e00856, 6c10903577, 0c08f698c1. Closing task

PR #7021, opened 2019-04-10
Upgrade Ceph to latest version and move to community repo

iggy:
**Summary**

* Move Ceph to community (along with `oath-toolkit` and `leveldb`)
* Upgrade to latest ceph release (14.1.0 -> 14.2.0)
* Add back radosgw


Ikke:
@iggy Please adjust the commit message titles to match the aports standard. That means they should have a `<reponame>/<packagename>: ` prefix.

Thanks.

tcely:
@iggy Because the prefix has the destination, I find it's helpful to mention the source of the move too.

For example: `community/ceph: move from testing`

Then I would put additional details in the commit message body.

PR #7023, opened 2019-04-10
community/cython: symlink cython to cython3

OTLabs:
- rename binaries to cython2 and cython3
- choose cython3 as the default binary

### DIFF
--- a/community/cython/APKBUILD
+++ b/community/cython/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=cython
 pkgver=0.29.6
-pkgrel=0
+pkgrel=1
 pkgdesc="Cython is an optimising static compiler for both the Python & the extended Cython programming languages."
 url="https://cython.org"
 arch="all"
@@ -45,11 +45,20 @@ package() {
 _py2() {
 	cd "$builddir"-py2
 	_py python2
+
+	local f; for f in cygdb cython cythonize; do
+	  mv "$subpkgdir"/usr/bin/$f "$subpkgdir"/usr/bin/${f}2
+	done
 }
 
 _py3() {
 	cd "$builddir"
 	_py python3
+
+	local f; for f in cygdb cython cythonize; do
+	  mv "$subpkgdir"/usr/bin/$f "$subpkgdir"/usr/bin/${f}3
+	  ln -s ${f}3 "$subpkgdir"/usr/bin/$f
+	done
 }
 
 _py() {


tcely:
I'd suggest `community/cython: symlink cython to cython3` then a blank line, then:

```
- rename binaries to cython2 and cython3
- choose cython3 as the default binary
```

commit log lines should be short and the longer stuff can go into the message body.

community/cython/APKBUILD
```diff
@@ -45,11 +45,20 @@ package() {
 _py2() {
 	cd "$builddir"-py2
 	_py python2
+
```
tcely: Using `local f` is good practice to avoid leaking the `f` variable into later executions.

community/cython/APKBUILD
```diff
@@ -45,11 +45,20 @@ package() {
...
+	for f in cygdb cython cythonize; do
+	  mv "$subpkgdir"/usr/bin/$f "$subpkgdir"/usr/bin/${f}2
+	done
 }
 
 _py3() {
 	cd "$builddir"
 	_py python3
+
```
tcely: The same `local f` improvement as above can be used here.

andypost:
@itoffshore are you ok with changes?

community/cython/APKBUILD
OTLabs: Implemented.

community/cython/APKBUILD
OTLabs: Implemented.

OTLabs:
@tcely I have fully addressed all your comments and suggestions. Thank you!

tcely:
@OTLabs You're welcome.

Editing the title and message of this PR would be helpful too.

OTLabs:
@tcely Auch, just forgot. Done! Thank you again!

itoffshore:
> @itoffshore are you ok with changes?

@andypost - yes all ok

nmeum:
This broke libplist a705e59b07b8d3c4816be2e3609187c072b7836e might have broken other packages as well.

OTLabs:
@nmeum The recommendation I have received on IRC was to use explicit versioning in packages, e.g. if package uses `python3` all scripts should be patched to use `python3` instead of `python`. Please, correct me if I am wrong.

The cython package was updated in 2 steps. This change was a second one. During first step all dependent packages were checked to build well. I am really sorry to miss that this change could broke some packages.

nmeum:
Sure, I agree with that. Just wanted to point out that this problem exists.

> I am really sorry to miss that this change could broke some packages.

Sure, no problem stuff like that happens. My comment was just intended as a heads-up.

OTLabs:
Thank you!

PR #7031, opened 2019-04-11
[3.7] main/gnutls: security upgrade to 3.6.7

J0WI:


rnalrd:
It needs nettle-3.4.1, not sure we want to upgrade from 3.3.

J0WI:
https://abi-laboratory.pro/?view=timeline&l=nettle

ncopa:
Yes this fails with:
````
checking for NETTLE... no
configure: error: 
  ***
  *** Libnettle 3.4.1 was not found.

>>> ERROR: gnutls: all failed
````

ncopa:
Do you think you could report the exact CVEs we need to fix? nettle upgrade seems risky in stable branch. We may need disable arm neon for `armhf` and `armv7` if we update.

ncopa:
> Do you think you could report the exact CVEs we need to fix? nettle upgrade seems risky in stable branch. We may need disable arm neon for `armhf` and `armv7` if we update.

Never mind. its:
````
#     - CVE-2019-3836
#     - CVE-2019-3829
````

J0WI:
The patches do not apply cleanly for previous versions.

maxice8:
Merged

PR #7033, opened 2019-04-11
main/supervisor: upgrade to 4.0.1

J0WI:
http://supervisord.org/changes.html#id2

### DIFF
--- a/main/supervisor/APKBUILD
+++ b/main/supervisor/APKBUILD
@@ -1,14 +1,15 @@
 # Contributor: Peter Bui <pnutzh4x0r@gmail.com>
 # Maintainer:
 pkgname=supervisor
-pkgver=3.3.5
+pkgver=4.0.1
 pkgrel=0
 pkgdesc="A system for controlling process state under UNIX"
 url="http://supervisord.org"
 arch="noarch"
 license="BSD"
-depends="python2 py-meld3 py-setuptools"
+depends="python py-meld3 py-setuptools"
 makedepends="$depends_dev"
+subpackages="$pkgname-openrc"
 source="$pkgname-$pkgver.tar.gz::https://github.com/Supervisor/$pkgname/archive/$pkgver.tar.gz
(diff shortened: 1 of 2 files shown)


TBK:
Please add a subpackage for openrc.

main/supervisor/APKBUILD
```diff
@@ -1,14 +1,15 @@
 # Contributor: Peter Bui <pnutzh4x0r@gmail.com>
 # Maintainer:
 pkgname=supervisor
-pkgver=3.3.5
+pkgver=4.0.1
 pkgrel=0
 pkgdesc="A system for controlling process state under UNIX"
 url="http://supervisord.org"
 arch="noarch"
 license="BSD"
```
bratkartoffel: It seems they moved to a "custom" licence:
https://github.com/Supervisor/supervisor/blob/master/COPYRIGHT.txt

main/supervisor/APKBUILD
```diff
@@ -1,14 +1,15 @@
...
-depends="python2 py-meld3 py-setuptools"
+depends="python py-meld3 py-setuptools"
```
ncopa: It does not support `python3`? I think we should explicit depend on the python version number

main/supervisor/APKBUILD
andypost: No idea how to add custom license(

main/supervisor/APKBUILD
TBK: It does, v4 added py3 support.

andypost: It supports py3 since 4.0 http://supervisord.org/changes.html#id3 that's is why I update dependency as well

andypost:
btw 4.0.2 update with the same questions https://github.com/alpinelinux/aports/pull/7132

tcely:
Please migrate any changes you want to keep to #7132 then close this.

PR #7061, opened 2019-04-12
[3.9] community/php7: security upgrade to 7.2.17 

J0WI:


andypost:
Thanks for filing it but no reason in intermediate commits (upgrade to 7.2.15 & 16)

rnalrd:
As pointed out by @andypost there are no relevant commits, including bumps and rebuilds ported from edge.
Also the secfixes comments should refer to the version of the alpine package actually built. 7.2.16 was never built in Alpine.

Thanks

PR #7074, opened 2019-04-13
community/emacs upgrade to 26.2

stef:


stef:
This is strange, travis fails because it cannot download the emacs sources? and drone fails because there seems to be some ASLR or exec-shield mitigation that is enabled on the host. on a private alpine based vm none of these problems appear to happen.

regarding the ASLR/exec-shield issue emacs writes this:
> *** Segfault during 'make'
> 
> If Emacs segfaults when 'make' executes one of these commands:
> 
>   LC_ALL=C ./temacs -batch -l loadup bootstrap
>   LC_ALL=C ./temacs -batch -l loadup dump
> 
> the problem may be due to inadequate workarounds for address space
> layout randomization (ASLR), an operating system feature that
> randomizes the virtual address space of a process.  ASLR is commonly
> enabled in Linux and NetBSD kernels, and is intended to deter exploits
> of pointer-related bugs in applications.  If ASLR is enabled, the
> command:
> 
>    cat /proc/sys/kernel/randomize_va_space  # GNU/Linux
>    sysctl security.pax.aslr.global          # NetBSD
> 
> outputs a nonzero value.
> 
> These segfaults should not occur on most modern systems, because the
> Emacs build procedure uses the command 'setfattr' or 'paxctl' to mark
> the Emacs executable as requiring non-randomized address space, and
> Emacs uses the 'personality' system call to disable address space
> randomization when dumping.  However, older kernels may not support
> 'setfattr', 'paxctl', or 'personality', and newer Linux kernels have a
> secure computing mode (seccomp) that can be configured to disable the
> 'personality' call.
> 
> It may be possible to work around the 'personality' problem in a newer
> Linux kernel by configuring seccomp to allow the 'personality' call.
> For example, if you are building Emacs under Docker, you can run the
> Docker container with a security profile that allows 'personality' by
> using Docker's --security-opt option with an appropriate profile; see
> <https://docs.docker.com/engine/security/seccomp/>.
> 
> To work around the ASLR problem in either an older or a newer kernel,
> you can temporarily disable the feature while building Emacs.  On
> GNU/Linux you can do so using the following command (as root).
> 
>     echo 0 > /proc/sys/kernel/randomize_va_space
> 
> You can re-enable the feature when you are done, by echoing the
> original value back to the file.  NetBSD uses a different command,
> e.g., 'sysctl -w security.pax.aslr.global=0'.
> 
> Alternatively, you can try using the 'setarch' command when building
> temacs like this, where -R disables address space randomization:
> 
>     setarch $(uname -m) -R make
> 
> ASLR is not the only problem that can break Emacs dumping.  Another
> issue is that in Red Hat Linux kernels, Exec-shield is enabled by
> default, and this creates a different memory layout.  Emacs should
> handle this at build time, but if this fails the following
> instructions may be useful.  Exec-shield is enabled on your system if
> 
>     cat /proc/sys/kernel/exec-shield
> 
> prints a nonzero value.  You can temporarily disable it as follows:
> 
>     echo 0 > /proc/sys/kernel/exec-shield
> 
> As with randomize_va_space, you can re-enable Exec-shield when you are
> done, by echoing the original value back to the file.

TBK:
Please rename commit and PR to: ` community/emacs: upgrade to 26.2`

TBK:
You forgot the ":" in the PR and commit name:

```
 community/emacs upgrade to 26.2
                ^
 community/emacs: upgrade to 26.2
```

Try to apply this patch https://debbugs.gnu.org/cgi/bugreport.cgi?msg=76;bug=24682;filename=0001-Port-to-Ubuntu-16.10-which-needs-gcc-nopie.patch;att=1 from https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24682 and see if it solves the bootstrap-emacs issue.

Ikke:
@stef It builds for me locally as well, so I will push this, but could you still try the patch that TBK pointed to?

stef:
that bugreport is from 2016, and we have been building emacs without problems without this patch previously.

PR #7081, opened 2019-04-14
testing/debconf: new aport

TBK:


testing/debconf/APKBUILD
```diff
@@ -0,0 +1,55 @@
...
+depends="perl"
+makedepends="findutils po4a"
+subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch"
+source="https://deb.debian.org/debian/pool/main/d/debconf/debconf_$pkgver.tar.xz
+	fix-makefile_destdir.patch
+	"
+builddir="$srcdir/$pkgname"
+
+build() {
+	cd "$builddir"
```
Ikke: These are redundant nowadays, since abuild automatically cd's to `$builddir`

TBK: @Ikke fixed. I have expanded my Makefile patch so there are two new subpkgs to mirror the Debian offers. The only difference is that Debians' debconf-utils contains the utils related man pages, I package them in debconf-doc.

I will upstream the patch.

PR #7085, opened 2019-04-15
Verify downloads with GPG

tcely:
Makes use of https://github.com/alpinelinux/abuild/pull/57 to verify GPG signatures before generating checksums.


Ikke:
As this will introduce new global variables, this should only be merged once the mentioned PR is accepted and the global variables are documented in the [APKBUILD reference](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference).

PR #7089, opened 2019-04-15
 main/poppler: upgrade to 0.77.0

TBK:
ABI 100% backwards compatible https://abi-laboratory.pro/index.php?view=timeline&l=poppler

```
Release 0.77.0:
        core:
         * Fix crash on signature handling. Issue #766
         * Fix small memory leak in SignatureHandler::getCertificateInfo
         * Splash: Restrict filling of overlapping boxes. Issue #750
         * Fix crash on malformed files

        qt5:
         * Fix optional content handling with exclusive layers

        cpp:
         * Make render_page thread-safe

        utils:
         * pdfsig: Fix small memory leak
         * pdftotext: Fix typo in manpage

Release 0.76.1:
        core:
         * Make the mul tables be calculated at compile time with constexpr.
         * splash: Fix compile with SPLASH_CMYK enabled
         * Some typo fixing in error messages

        qt5:
         * Fix regression in annotation handling

        build system:
         * Fix some typos in build system output and comments

Release 0.76.0:
        core:
         * Fix regression on case-insensitive search. Issue #743
         * Remove GooList, use std::vector instead
         * Fix radiobutton reporting wrong state. Issue #159
         * Handle UTF16-LE strings
         * Don't error out if there's no DA in FreeText annotation
         * cairo: Compute correct coverage values for box filter.
         * cairo: Constrain number of cycles in rescale filter. 
         * Read more fields from ViewerPreferences
         * Introduce and use Ref::INVALID
         * Fix crashes in broken files
         * Fix mismatched free/delete
         * Add missing include guards

        utils:
         * pdftohtml: Properly initialize HtmlOutputDev::page to avoid SIGSEGV upon error exit. Issue #742

Release 0.75.0:
        core:
         * Fix rendering of some annotations
         * Fix crashes in broken files
         * Small internal code improvements

        cpp:
         * Improve documentation
         * tests: Add showing version information to poppler-dump

        utils:
         * pdfattach: new util
         * pdftohtml: add -dataurls parameter
         * pdftoppm: add -sep and -forcenum parameters
         * pdftohtml: make singleHtml and stout not mutually exclusive
         * pdfsig: fix use after free

Release 0.74.0:
        core:
         * Remove support for obsolete systems. Issue #709
         * Include timezone in timeToDateString()
         * Fix/silence some warnings
         * Fix issues with broken files

        build system:
         * Fix linking in FreeBSD
         * Fix fseeko configure check on Android for API level < 24
         * Remove unused MacroPushRequiredVars.cmake

        qt5:
         * Add API that lazily builds an outline by wrapping the internal objects
         * Demo: Use new API to build Table Of Contents lazily

        glib:
         * Improve documentation
         * Fix cast from 'GTime *' (aka 'int *') to 'time_t *' (aka 'long *')

        utils:
         * pdfsig: add -nssdir option

        cpp:
         * Add a way to get all the named destinations in a document.

Release 0.73.0:
        core:
         * Fix regression reading some encrypted files. Issue #690
         * Add X509CertificateInfo classes
         * Add new 'IgnoreDiacritics' option to ::findText(). Issue #637
         * Open files with CLOEXEC flag set
         * Remove Gulong, Guint, Gushort, Guchar typedefs
         * Fix handling of some broken files.

        cpp:
         * Make initialization of globalParams threadsafe
         * Fix page::text_list encoding issue
         * Improve handling of UTF-16 by considering Endianess
         * Add API to specify a custom data directory

        qt5:
         * Expose X509CertificateInfo
         * Add the possibility of getting version
         * Add new 'IgnoreDiacritics' search flag. Issue #637
         * Make initialization of globalParams threadsafe
         * ArthurOutputDev: Remove all Splash code usage

        glib:
         * add new 'POPPLER_FIND_IGNORE_DIACRITICS' find flag. Issue #637
         * Fix named destinations. Issue #631
         * Make PrintScaling preference available in API. Bug #92779

        build system:
         * Rename ENABLE_XPDF_HEADERS to ENABLE_UNSTABLE_API_ABI_HEADERS
         * support enabling NSS on mingw
         * Windows: only set SOVERSION for shared libs

Release 0.72.0:
        core:
         * Fix checkbox lacking AP not bein able to change state. Issue #655
         * Draw line annotation endings (arrow, circle, ...)
         * cairo: Don't use UNIQUE_ID for PS output, to avoid using PS memory on cairo >= 1.5.10
         * Be more stubborn looking for a nssdb. Issue #669
         * GooString::fromInt: Repair the return value.
         * Minor performance improvements
         * Avoid cycles in PDF parsing
         * Stream::makeFilter: Fix memory leak
         * Fix various issues with malformed files
         * Rename GooString::getCString to GooString::c_str
         * Regenerate UnicodeDecompTables.h from python 3.7.1

        utils:
         * pdfdetach: Check for valid embedded file before trying to save it. Issue #661
         * pdfdetach: Check for valid file name of embedded file before using it to determine save path. Issue #660
         * Fix typos in utils.

        glib:
         * Fix missing PopplerAttachment destructor call
         * Support getting form widget additional actions.
         * docs: Small improvements

        qt5:
         * Internally compile with -DQT_NO_SIGNALS_SLOTS_KEYWORDS
```

### DIFF
--- a/community/atril/APKBUILD
+++ b/community/atril/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Alan Lacerda <alacerda@alpinelinux.org>
 pkgname=atril
 pkgver=1.22.0
-pkgrel=0
+pkgrel=1
 pkgdesc="A document viewer for MATE"
 url="https://github.com/mate-desktop/atril"
 arch="all"
--- a/community/claws-mail/APKBUILD
+++ b/community/claws-mail/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=claws-mail
 pkgver=3.17.3
-pkgrel=6
+pkgrel=7
 pkgdesc="A GTK+ based e-mail client."
 url="https://www.claws-mail.org"
 arch="all"
--- a/community/cups-filters/APKBUILD
+++ b/community/cups-filters/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=cups-filters
 pkgver=1.24.0
-pkgrel=0
+pkgrel=1
 pkgdesc="OpenPrinting CUPS filters and backends"
 url="https://wiki.linuxfoundation.org/openprinting/cups-filters"
 arch="all"
--- a/community/diff-pdf/APKBUILD
+++ b/community/diff-pdf/APKBUILD
@@ -4,7 +4,7 @@
 pkgname=diff-pdf
 pkgver=0.2_git20170816
 _gitrev=48416f3ed9085db77cce22f836ef4456bbaacccc
-pkgrel=1
+pkgrel=2
 pkgdesc="A simple tool for visually comparing two PDF files"
 url="https://vslavik.github.io/diff-pdf/"
 arch="all"
--- a/community/evince/APKBUILD
+++ b/community/evince/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: 
 pkgname=evince
 pkgver=3.32.0
-pkgrel=0
+pkgrel=1
 pkgdesc="simple document viewer for GTK+"
 url="http://projects.gnome.org/evince/"
 arch="all"
--- a/community/gimp/APKBUILD
+++ b/community/gimp/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=gimp
 pkgver=2.10.10
-pkgrel=0
+pkgrel=1
 pkgdesc="GNU Image Manipulation Program"
 url="https://www.gimp.org/"
 arch="all"
--- a/community/inkscape/APKBUILD
+++ b/community/inkscape/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=inkscape
 pkgver=0.92.4
-pkgrel=1
+pkgrel=2
 pkgdesc="A vector-based drawing program - svg compliant"
 url="https://inkscape.org/"
 arch="all"
@@ -27,7 +27,9 @@ makedepends="
 depends="desktop-file-utils"
 checkdepends="bash py-lxml py-numpy>=1.14.3-r1"
 subpackages="$pkgname-doc $pkgname-lang $pkgname-view"
-source="https://launchpad.net/inkscape/${pkgver%.*}.x/$pkgver/+download/inkscape-$pkgver.tar.bz2"
+source="https://launchpad.net/inkscape/${pkgver%.*}.x/$pkgver/+download/inkscape-$pkgver.tar.bz2
+	poppler-076.patch 
+	"
 options="!check" # cxxtest hangs at least on x86_64
 builddir="$srcdir/$pkgname-$pkgver"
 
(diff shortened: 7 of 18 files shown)


TBK:
> The job exceeded the maximum time limit for jobs, and has been terminated.

Travis CI is too slow... 63% of the last packages :(

Ikke:
Would these packages need to be rebuilt as well, as they depend on libpoppler?

* cups-filters
* poppler-glib
* poppler-utils
* inkscape-view
* inkscape
* texlive
* libreoffice-draw

libreoffice had to be rebuilt the other day because libpoppler was updated but it sitll depended on the older soname.

abuild does not report a soname difference, but this seems to have `so:libpoppler.so.86`, while these packages in edge depend on `so:libpoppler.so.82`

TBK:
According to the ABI report from https://abi-laboratory.pro/index.php?view=timeline&l=poppler it should not be a problem.

The previous poppler upgrade jumped from 0.56.0 to 0.71.0 where there where 7 symbols removed from the qt .so file + it moved from qt4 to qt5 - https://abi-laboratory.pro/index.php?view=objects_report&l=poppler&v1=0.61.1&v2=0.62.0

The other day I stumped upon what I would describe as a issue with the way that abuild handles .so files which can also effect result in all the dependent packages having to be rebuild to link against the newer version. Which is what you describe.

I have not dug deeper into this so I do not know if it is by mistake or design, but at the moment the non version tagged symbolic link .so file gets packaged in many cases in the dev subpackage of the aport.

There are quite a few https://pkgs.alpinelinux.org/contents?page=1&file=%2A.so&arch=x86_64&branch=edge&name=%2A-dev

Example with poppler:
Main pkg got the tagged so files- https://pkgs.alpinelinux.org/contents?branch=edge&name=poppler&arch=aarch64&repo=main
Dev subpkg got the non tagged - https://pkgs.alpinelinux.org/contents?file=*.so&path=&name=poppler-dev&branch=edge&repo=main&arch=aarch64
Glibc subpkg https://pkgs.alpinelinux.org/contents?branch=edge&name=poppler-glib&arch=aarch64&repo=main

So if I am not mistaken a lib/program that is build against a specific tagged .so.x.x.x file and not the non tagged symbolic link, an aport upgrade where a symbolic link for the previous soname version is not provided will break the dynamically link.

rnalrd:
It seems that at least inkscape needs patching with newer poppler:

https://github.com/void-linux/void-packages/blob/master/srcpkgs/inkscape/patches/poppler-072.patch

It would be nice to verify before merging.

Thanks

maxice8:
there is also a new patch for poppler-076 https://github.com/void-linux/void-packages/blob/7f8e3d9bae54aa6b34b2f755cfec2fa3129b994c/srcpkgs/inkscape/patches/poppler-076.patch

I did the poppler-0.76 update on Void Linux, any packages both Void and Alpine have are patched to work with poppler-0.76

maxice8:
0.76.1 is out

TBK:
I will bump and fix hopefully later today, in the process of reconfiguring my workstation.

TBK:
Neither Travis nor Drone can build this:

> Drone - warning: maximum output exceeded
> Travis - The job exceeded the maximum time limit for jobs, and has been terminated.

Ikke:
Probably due to the amount of packages involved.

ncopa:
cups-filters fails to build:
````
filter/pdftoraster.cxx:42:10: fatal error: goo/GooString.h: No such file or directory                                                            
 #include "goo/GooString.h"                                                                                                                      
          ^~~~~~~~~~~~~~~~~                                                                                                                   
compilation terminated.     
````

TBK:
> cups-filters fails to build:
> 
> ```
> filter/pdftoraster.cxx:42:10: fatal error: goo/GooString.h: No such file or directory                                                            
>  #include "goo/GooString.h"                                                                                                                      
>           ^~~~~~~~~~~~~~~~~                                                                                                                   
> compilation terminated.     
> ```

Missed that ENABLE_XPDF_HEADERS had been replaced by ENABLE_UNSTABLE_API_ABI_HEADERS.  After fixing it I build cups-filters locally without any issue.

TBK:
@ncopa It should be ready to go. I would appreciate it if it would get merged soon so I do not have to keep rebasing.

rnalrd:
@ncopa @TBK we are in the process of releasing v3.10, do we want to include it or not?

rnalrd:
At this point I think we should merge it right after 3.10 release.
Thanks

TBK:
@rnalrd I agree. I will wait for 3.10 to be release and then rebase.

Ikke:
@TBK 3.10 has been released in the mean time :-)

PR #7099, opened 2019-04-15
[3.7] main/imagemagick: security upgrade to 7.0.8.39 

J0WI:
Retry of #6982. Includes `pkgrel` bump for packages depending on ImageMagick.
cc @rnalrd


rnalrd:
This is something we don't usually do: version bumps in stable branches that requires rebuilding packages. Thank kinda defeats the concept of "stable" branch.
Thanks

PR #7115, opened 2019-04-17
testing/docker-compose: Fix missing deps

bratkartoffel:
This closes Bug #10282
https://bugs.alpinelinux.org/issues/10282

### DIFF
--- a/community/py-jsonschema/APKBUILD
+++ b/community/py-jsonschema/APKBUILD
@@ -3,11 +3,12 @@
 pkgname=py-jsonschema
 _pkgname=jsonschema
 pkgver=3.0.1
-pkgrel=1
+pkgrel=2
 pkgdesc="An implementation of JSON Schema validation for Python"
 url="https://github.com/Julian/jsonschema"
 arch="noarch"
 license="MIT"
+depends="py-pyrsistent py-attrs"
 makedepends="python3-dev python2-dev py-setuptools"
 subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
 source="https://files.pythonhosted.org/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
@@ -26,10 +27,12 @@ package() {
 
 _py2() {
 	replaces="$pkgname"
+	depends="${depends/py-/py2-}"
 	_py python2
 }
 
 _py3() {
+	depends="${depends/py-/py3-}"
 	_py python3
 }
 
--- a/testing/docker-compose/APKBUILD
+++ b/testing/docker-compose/APKBUILD
@@ -1,10 +1,10 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=docker-compose
-pkgver=1.23.2
-pkgrel=1
+pkgver=1.24.0
+pkgrel=0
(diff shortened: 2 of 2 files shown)


tcely:
Please add running `docker-compose version` to check to catch this kind of problem in the future.

tcely:
https://docs.docker.com/release-notes/docker-compose/

This can be upgraded to the latest version.

tcely:
Moving docker-compose to community should help with CI failures.

bratkartoffel:
I think we should not mix too many actions into one PR and focus on making it work again.
We can upgrade / move this aport later.

community/py-pyrsistent/APKBUILD
```diff
@@ -0,0 +1,53 @@
...
+check() {
+        cd "$builddir"
+        python3 setup.py test
+}
+
+package() {
+	mkdir -p "$pkgdir"
+}
+
+_py2() {
```
Ikke: Do you need python2 support here? Alpine LInux is going to drop python2 support in the near future. docker-compose is using python, so for that it's not needed.

bratkartoffel: You're right, I just added python2 support as the library provides it.
Will remove python2 support and rebase this evening.

Ikke:
py3-pyrsistent has just been pushed (I didn't realize it was part of this PR), so it can be dropped from this PR.

Also notice that you're adding the packages directly to community, while they should start in testing (and as docker-compose is in testing, there is no need for them to go directly to community).

tcely:
See https://github.com/alpinelinux/aports/pull/7233

bratkartoffel:
> Also notice that you're adding the packages directly to community, while they should start in testing (and as docker-compose is in testing, there is no need for them to go directly to community).

py-pyrsistent is a dependency of community/py-jsonschema, so it also has to be in community, hasn't it?

community/py-pyrsistent/APKBUILD
bratkartoffel: As this aport is a dependency of py-jsonschema this means that we have to drop python2 support there too. I'm unsure if we should do this now

Ikke:
@bratkartoffel Ah, you are right, then I'll move pyrsistent to community first.

Ikke:
py3-pyrsistent has been [moved][commit] to community

[commit]:https://github.com/alpinelinux/aports/commit/3b978a8

tomalok:
It should also be noted that the latest release of `docker-compose` is at [1.24.0](https://github.com/docker/compose/releases/tag/1.24.0).

I fiddled around with trying to get the tests working a day or two ago -- got them to go further by adding `python3-dev` to `checkdepends="..."`.  After the test suite printed "running build_ext", it tried to execute `docker-compose test` (an unknown subcommand) and failed.

mtneug:
Just a note: the dependencies are not only missing for `docker-compose`, but e.g. for `cloud-init` as well.

Ikke:
Dependencies issues are fixed now. 

It appears you forgot to `apk checksum` after the last update.

Ikke:
I fixed the remaining issues and pushed it. Note that I removed the simple `docker-compose -v` check, as:

1. Just running version does often not assert that it's actually working
2. Running `docker-compose` in the source directory is not even the same as the installed version. `docker-compose -v` ran just fine in the srcdir, but installed by setuptools, it complains about paramiko missing.

PR #7119, opened 2019-04-18
testing/git-secret: update to 0.2.5

invidian:
* drop passhprase input patch, since it's included in upstream
* rebase remaining patches

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>


testing/git-secret/APKBUILD
```diff
@@ -4,7 +4,7 @@
 # Note: git-secret is very badly designed, I'll replace this abuild once
 #   I find or write something better.
 pkgname=git-secret
-pkgver=0.2.2
+pkgver=0.2.5
 pkgrel=1
```
maxwell-k: The pkgrel can be reset to zero because you have moved to a new pkgver

```suggestion
pkgrel=0
```

maxwell-k:
This PR looks promising, at the moment our Drone CI is failing. If you please rebase that commit on the current master the CI should pass.

The Drone config has changed recently and it now applies a PR differently, the current approach is at:
https://github.com/alpinelinux/alpine-drone-ci/blob/master/files/usr/local/bin/build.sh#L81

Thanks for your help.

testing/git-secret/APKBUILD
invidian: Good to know, thanks for pointing out! I'm coming from Arch Linux, where all packages starts with `pkgrel=1`.

invidian:
Thanks for the review, I rebased it and now CI is passing :)

PR #7120, opened 2019-04-18
community/pdns-recursor: enable s390x

tcely:
boost-context is now available

### DIFF
--- a/community/pdns-recursor/APKBUILD
+++ b/community/pdns-recursor/APKBUILD
@@ -2,11 +2,10 @@
 # Maintainer: tcely <pdns-recursor+aports@tcely.33mail.com>
 pkgname=pdns-recursor
 pkgver=4.1.12
-pkgrel=1
+pkgrel=2
 pkgdesc="PowerDNS Recursive Server"
 url="https://www.powerdns.com/"
-# s390x: missing boost-context
-arch="all !s390x"
+arch="all"
 license="GPL-2.0-or-later"
 depends="dns-root-hints"
 depends_dev=""


community/pdns-recursor/APKBUILD
```diff
@@ -2,11 +2,10 @@
 # Maintainer: tcely <pdns-recursor+aports@tcely.33mail.com>
 pkgname=pdns-recursor
 pkgver=4.1.12
-pkgrel=1
+pkgrel=2
```
andypost: I think no reason to bump, there's no package for s390x so it will be added but other arches could remain as is

tcely: I'd rather bump than have different "source" versions in the same pkgver/pkgrel descriptor.

Ikke:
It still fails to build on s390x:

<pre>
/usr/lib/gcc/s390x-alpine-linux-musl/8.3.0/../../../../s390x-alpine-linux-musl/bin/ld: mtasker_context.o: in function `pdns_swapcontext(pdns_ucontext_t&, pdns_ucontext_t const&)':
mtasker_context.cc:(.text+0xe8): undefined reference to `jump_fcontext'
/usr/lib/gcc/s390x-alpine-linux-musl/8.3.0/../../../../s390x-alpine-linux-musl/bin/ld: mtasker_context.o: in function `pdns_makecontext(pdns_ucontext_t&, boost::function<void ()>&)':
mtasker_context.cc:(.text+0x1ea): undefined reference to `make_fcontext'
/usr/lib/gcc/s390x-alpine-linux-musl/8.3.0/../../../../s390x-alpine-linux-musl/bin/ld: mtasker_context.cc:(.text+0x20e): undefined reference to `jump_fcontext'
/usr/lib/gcc/s390x-alpine-linux-musl/8.3.0/../../../../s390x-alpine-linux-musl/bin/ld: mtasker_context.o: in function `threadWrapper':
mtasker_context.cc:(.text+0x27e): undefined reference to `jump_fcontext'
/usr/lib/gcc/s390x-alpine-linux-musl/8.3.0/../../../../s390x-alpine-linux-musl/bin/ld: mtasker_context.cc:(.text+0x382): undefined reference to `jump_fcontext'
</pre>

PR #7126, opened 2019-04-18
main/libtool: do not disable static

tcely:
Fixes https://bugs.alpinelinux.org/issues/10301

### DIFF
--- a/main/libtool/APKBUILD
+++ b/main/libtool/APKBUILD
@@ -25,8 +25,8 @@ build() {
 	./configure \
 		--build=$CBUILD \
 		--host=$CHOST \
-		--prefix=/usr \
-		--disable-static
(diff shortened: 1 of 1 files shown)


TBK:
Would it be an idea to make a subpkg for the static lib?

tcely:
@TBK It's an idea. I'm thinking it might be a better idea to install both libraries with `libltdl` instead.

ncopa:
I don't think a static `libltdl` makes any sense. And I think there is no good reason to use libltdl at all nowdays. It is just a wrapper around `dlopen`, which sort of defeats the concept of static linking.

I also don't  think https://bugs.alpinelinux.org/issues/10301 makes any sense. I believe libtool is currently capable of building static libraries.

PR #6753, opened 2019-03-19
main/libuv: upgrade to 1.28.0

OTLabs:
https://github.com/libuv/libuv/releases 1.28.0

PR #7131, opened 2019-04-19
main/libuv: upgrade to 1.28.0

andypost:
Ref https://github.com/libuv/libuv/releases

https://abi-laboratory.pro/?view=timeline&l=libuv


tcely:
See #6753

PR #7143, opened 2019-04-20
main/raspberrypi: upgrade to 0.20190416

innerspacepilot:
Updated version

### DIFF
--- a/main/raspberrypi/APKBUILD
+++ b/main/raspberrypi/APKBUILD
@@ -1,8 +1,8 @@
 # Maintainer: Timo Teräs <timo.teras@iki.fi>
 pkgname=raspberrypi
-pkgver=0.20181212
-_commitid=7cbfbd38d9824535164f93a1d32c81a33a00ca31
-pkgrel=1
+pkgver=0.20190416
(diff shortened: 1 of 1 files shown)


main/raspberrypi/APKBUILD
```diff
@@ -44,6 +44,7 @@ package() {
 	# nuke the unwanted stuff
 	rm -rf "$pkgdir"/opt/vc/src
 	rm -rf "$pkgdir"/opt/vc/lib/*.a
+	cp -r "$pkgdir"/opt/vc/etc "$pkgdir"
```
fabled: I don't understand why the etc config is copied. Does it fix something? Should it then be moved instead?

innerspacepilot: This is the reason https://github.com/raspberrypi/userland/commit/e5803f2c986cbf8c919c60278b3231dcdf4271a6

innerspacepilot: Seems like moving it would be the right way.
I'll make a change

innerspacepilot:
fixed

innerspacepilot:
@fabled, Is it ok now?

PR #7144, opened 2019-04-20
main/raspberrypi: fix components failing

innerspacepilot:
added "--no-as-needed" linker flag to make mmal vc.ril.* components stop failing.
Description here: https://github.com/raspberrypi/userland/issues/178


Ikke:
Hello innerspacepilot,

For this to be effective, you need to bump the `pkgrel` variable, in this case from 0 to 1. That will trigger the builders to rebuild the package.

Kind regards, Kevin.

innerspacepilot:
Hello, Kevin.
pkgrel fixed.
Thanks.

PR #85, opened 2016-05-23
/testing/keepassx: Update to 0.4.4

AmatCoder:
Fix CVE-2015-8378

PR #7149, opened 2019-04-20
community/py-execnet: upgrade to 1.6.0

dimon222:
1.6.0 (2019-03-31)
------------------

* ``execnet`` no longer supports Python 2.6 and 3.3 (#85). Users of those Python versions
  using a recent enough ``pip`` should not be affected, as ``pip`` will only install
  ``1.5.0`` for them.

* Update test suite to support ``pytest>4``.


Ikke:
This was merged as 1996bb4, closing PR.

PR #7159, opened 2019-04-20
main/ghostscript: upgrade to 9.27

andypost:
Ref https://www.ghostscript.com/doc/9.27/News.htm


J0WI:
I tried to modernize this locally and get rid of the ` || return 1` pattern, but for some reason I wasn't able to built it afterwards.

andypost:
It fails only on travis
```
In file included from ./base/memory_.h:23,
                 from ./obj/gsmd5.h:1,
                 from ./obj/gsmd5.c:56:
./base/std.h:25:10: fatal error: arch.h: No such file or directory
 #include "arch.h"
          ^~~~~~~~
compilation terminated.
```

PR #7163, opened 2019-04-20
main/samba: upgrade to 4.10.2

kr0k0:
Upgrade also talloc, tevent, cmocka, tdb and ldb to required/latest version.
Remove python2, as it will be EOL january next year.

### DIFF
--- a/main/cmocka/APKBUILD
+++ b/main/cmocka/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=cmocka
-pkgver=1.1.1
-pkgrel=1
+pkgver=1.1.5
+pkgrel=0
 pkgdesc="An elegant unit testing framework for C with support for mock objects"
 url="https://cmocka.org/"
 arch="all"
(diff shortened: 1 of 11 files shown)


Ikke:
CI fails with the following error:

> ERROR: System library pyldb-util.cpython-37m-i386-linux-gnu of version 1.5.4 not found, and bundling disabled

kr0k0:
Fix ldb test on 32bit arch with a patch which fedora uses.

Fix samba build with removing of include of "unistd.h" in "$builddir/lib/replace/replace.h", as it was also not included in past samba versions. Maybe you have a better solution.
Error was:
source4/heimdal/lib/asn1/asn1_err.c:47:23: error: 'link' redeclared as different kind of symbol
 static struct et_list link = { 0, 0 };
                              ^~~~
In file included from /usr/include/fortify/unistd.h:22,
                 from ../../lib/replace/../replace/replace.h:166,
                 from ../../source4/heimdal_build/config.h:10,
                 from source4/heimdal/lib/asn1/asn1_err.c:1:
/usr/include/unistd.h:56:5: note: previous declaration of 'link' was here
 int link(const char *, const char *);
     ^~~~

travis-ci build failed because of source download error:
curl: (16) Error in the HTTP2 framing layer
 Maybe you can restart the build.

Please review and merge if it's ok for you.

rnalrd:
The samba makedepends on dbus is strictly needed?

kr0k0:
Not strictly, but without the snapper vfs module isn't built.
Should I put it in a subpackage?

rnalrd:
I have not investigated but if it saves pulling in multiple dependencies when installing samba or if it saves a significative amount of space (several hundreds KBs or more) probably it's worthy to move it into a subpkg.

ncopa:
looks like samba depends on `cups` who depends on `avahi` who depends on `dbus` so the dbus libs will be pulled in regardless

rnalrd:
Then I have no objections in merging this. It would be nice to have this in 3.10.

PR #7166, opened 2019-04-21
main/ncurses: Update to 6.1_p20190420

Nemykal:
Previous version 404s upstream


ncopa:
I wonder if we can use patches from https://invisible-mirror.net/archives/ncurses/6.1/ instead, which does not seem to be removed.

PR #7169, opened 2019-04-21
testing/perl-moosex-types: new aport

timlegge:


Ikke:
@timlegge I apparently applied this too soon, the [build][] fails with:

> The following REQUIRED prerequisites were not satisfied: 
> Module::Build::Tiny is not installed (version '0.034' required)

[build]:https://build.alpinelinux.org/buildlogs/build-edge-ppc64le/testing/perl-moosex-types/perl-moosex-types-0.50-r0.log

Could you take a look at that?

Ikke:
I temporarily disabled the checks to let the builders continue. Let me know when you have a fix.

timlegge:
should be fixed now.  I have apkbuild-cpan installed which has perl-module-build-tiny as a dependency

Thanks

PR #7196, opened 2019-04-22
testing/zola: new aport

maxice8:


Ikke:
This depends on cargo, which is currently broken in edge.

PR #7199, opened 2019-04-22
testing/plata-theme: new aport

maxice8:


Ikke:
Drone `armhf` failed because it took longer then 1h.

PR #7201, opened 2019-04-22
main/python3: -dev depends on py3-setuptools

tcely:
A package using -dev will likely have setup.py that should be used.


tcely:
@ncopa I am missing where the problem comes up.

PR #7202, opened 2019-04-22
community/rust: rebuild for libgit2

tcely:


tcely:
This is a thornier problem than I first thought. I don't see a way around the dependency problems while still linking against the newer `libgit2` version.

PR #7203, opened 2019-04-22
community/abiword: modify plugin function names to avoid ash conflict

mksully22:
APKBUILD includes instructions to build plugin routines,
for _i in $_plugins; do
        eval "plugin_${_i}() { _do_plugin $_i; }"
done

One of the plugin names is "command" which results in a script function being created:
command() { 
_do_plugin command; 
}

This command() created to support plugin handling conflicts with the builtin command() function supplied by the ash interpreter. This results in an error during build when abuild's default_doc() issues a
local gzip=$(command -v pigz || echo gzip)

This fix renames the plugin handling functions to add a prefix, avoiding the conflicts

### DIFF
--- a/community/abiword/APKBUILD
+++ b/community/abiword/APKBUILD
@@ -26,7 +26,7 @@ _plugins="applix babelfish bmp clarisworks collab docbook command eml epub \
 	xslfo"
 
 for _i in $_plugins; do
-	subpackages="$subpackages $pkgname-plugin-$_i:$_i"
+	subpackages="$subpackages $pkgname-plugin-$_i:plugin_${_i}"
 done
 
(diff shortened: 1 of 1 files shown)


community/abiword/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=abiword
 pkgver=3.0.2
-pkgrel=4
+pkgrel=5
```
Ikke: Is a `pkgrel` bump needed in this case? Is there an issue with the existing abiword-plugin-command package?

mksully22: Thanks Ikke. Removed pkgrel bump

PR #7207, opened 2019-04-22
community/cargo-static: split off from community/rust

Cogitri:
This allows us to have a static cargo version for bootstrapping (e.g. to not
break when libgit is upgraded) while still offering the users a shared one.

### DIFF
--- a/testing/drone-cli/APKBUILD
+++ b/testing/drone-cli/APKBUILD
@@ -1,13 +1,13 @@
 # Contributor: Carlo Landmeter <clandmeter@alpinelinux.org>
-# Maintainer:
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=drone-cli
-pkgver=1.0.6
+pkgver=1.1.0
 pkgrel=0
-pkgdesc="Drone CLI" 
+pkgdesc="Drone CLI"
 url="https://github.com/drone/drone-cli"
 arch="all"
 license="Apache-2.0"
-options="!check"
+options="!check" # No tests to run
 makedepends="go"
 source="drone-cli-$pkgver.tar.gz::https://github.com/drone/drone-cli/archive/v$pkgver.tar.gz"
 builddir="$srcdir/src/github.com/drone/$pkgname"
@@ -20,7 +20,6 @@ prepare() {
 }
 
 build() {
-	cd "$builddir"
 	export GOPATH="$srcdir"
 	go build -ldflags "-X main.version=${pkgver}" -o bin/drone ./drone
 }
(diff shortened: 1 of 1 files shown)


community/cargo-static/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+arch="x86_64"
+license="Apache-2.0 MIT"
+depends="rust"
+makedepends="
+	curl-dev
+	openssl-dev
+	cargo-bootstrap
+	"
+source="$pkgname-$pkgver.tar.gz::https://github.com/rust-lang/cargo/archive/${pkgver}.tar.gz
+	https://alpha.de.repo.voidlinux.org/distfiles/cargo-0.33.0-x86_64-unknown-linux-musl.tar.xz
```
Cogitri: We will be able to remove this once this has been built once and instead use cargo-bootstrap which is provided by this package.

Cogitri:
Hum, I can build this locally, I'm not quite sure why it tries to pull in cargo (which needs old libgit2) here?

community/cargo-static/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+pkgrel=1
+pkgdesc="Rust package manager - static version for bootstrapping"
+url="http://www.rust-lang.org"
+arch="x86_64"
+license="Apache-2.0 MIT"
+depends="rust"
+makedepends="
+	curl-dev
+	openssl-dev
+	cargo-bootstrap
```
tcely: Remove `cargo-bootstrap` from `makedepends` when boot strapping.

Cogitri: Done

community/cargo-static/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+pkgrel=2
+pkgdesc="Rust package manager - static version for bootstrapping"
+url="http://www.rust-lang.org"
+arch="x86_64"
+license="Apache-2.0 MIT"
+depends="rust"
+makedepends="
+	curl-dev
+	openssl-dev
+	cargo
```
Cogitri: After bootstrapping this can be cargo-bootstrap.

Cogitri:
Hum, not quite sure why CI tries to build rust too, I did rebase.

tcely:
> Hum, not quite sure why CI tries to build rust too, I did rebase.

It's building `rust` because you modified that `APKBUILD` on this branch. Remove those changes and it won't build `rust` again as part of the CI.

Cogitri:
Ah, alright, on Void CI would only rebuild on pkgrel changes.

Since cargo-bootstrap is on 1.31 right now and the new one will be on 0.32 for now, do I have to do something special for the new one to be pulled in, or should that work fine after the rust rebuild (which should remove cargo-bootstrap)?

PR #7210, opened 2019-04-23
main/py-urllib3: upgrade to 1.25.1

dimon222:
1.25.1 (2019-04-24)
-------------------

* Add support for Google's ``Brotli`` package.

* Upgrade bundled rfc3986 to v1.3.1

1.25 (2019-04-22)
-----------------

* Require and validate certificates by default when using HTTPS

* Upgraded ``urllib3.utils.parse_url()`` to be RFC 3986 compliant.

* Added support for ``key_password`` for ``HTTPSConnectionPool`` to use
  encrypted ``key_file`` without creating your own ``SSLContext`` object.

* Add TLSv1.3 support to CPython, pyOpenSSL, and SecureTransport ``SSLContext``
  implementations.

* Switched the default multipart header encoder from RFC 2231 to HTML 5 working draft.

* Fixed issue where OpenSSL would block if an encrypted client private key was
  given and no password was given. Instead an ``SSLError`` is raised.

* Added support for Brotli content encoding. It is enabled automatically if
  ``brotlipy`` package is installed which can be requested with
  ``urllib3[brotli]`` extra.

* Drop ciphers using DSS key exchange from default TLS cipher suites.
  Improve default ciphers when using SecureTransport.

* Implemented a more efficient ``HTTPResponse.__iter__()`` method.

### DIFF
--- a/main/py-urllib3/APKBUILD
+++ b/main/py-urllib3/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=py-urllib3
 _pkgname=${pkgname/py-/}
-pkgver=1.24.2
+pkgver=1.25.1
 pkgrel=0
 pkgdesc="HTTP library with thread-safe connection pooling, file post, and more"
 url="https://github.com/shazow/urllib3"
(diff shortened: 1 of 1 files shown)


dimon222:
Bumped to 1.25.1 that came out today

clandmeter:
Have you verified this doesn't break any deps on this package ie:

- py-botocore
- py-elasticsearch

dimon222:
@clandmeter sadly, nope

According to their reqs, seems botocore/requests require specifically <1.25. Maybe this PR can be kept open until they upgrade it.
https://github.com/boto/botocore/blob/develop/setup.py#L45

py-elasticsearch doesn't depend on urllib3 directly because it's transitive dependency, but it uses requests that is still not upgraded. So obv cascade affected

clandmeter:
This is already in aports via 07527e9196c53c99c60e07f470dd5a8178c8de60

PR #7212, opened 2019-04-23
testing/telegram-desktop: update to 1.6.7

maxice8:


maxice8:
CI Failing because it can't take the changes to xxhash (same PR) into effect.

Ikke:
Why do we need to include all these cmake files ourselves?

maxice8:
On Sat, May 11, 2019 at 07:58:55AM -0700, Kevin Daudt wrote:
> Why do we need to include all these cmake files ourselves?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7212#issuecomment-491518075

They aren't upstream as far as i know, they are present in Void Linux's packaging.

This PR is also outdated to 1.7.0 and the maintainer has a simpler patch on patchwork that keeps the drawbacks of more static linking (not using system libraries) and gyp

Ikke:
Do you want to update this to 1.7.0 now, or will you do that later in a new PR?

maxice8:
On Sat, May 11, 2019 at 09:45:06AM -0700, Kevin Daudt wrote:
> Do you want to update this to 1.7.0 now, or will you do that later in a new PR?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7212#issuecomment-491526423

The maintainer of the package has a patch on patchwork, my work here is done

PR #7215, opened 2019-04-23
testing/kodi-rpi: new aport

innerspacepilot:
Hello!
This is kodi for raspberry pi.
armv7 only at this time.


ncopa:
Why cannot we simply enable armv7 for `community/kodi`? Why does this need to be a separate package?

innerspacepilot:
Because it uses raspberry firmware for building and working.
It has different dependencies and, for example, it won't build if mesa-dev is installed.
It will be easier to maintain, because common APKBUILD will be really complicated.

Cogitri:
Sorry, but a 75K LOC downstream behemoth isn't really something we want in our repos.

PR #7219, opened 2019-04-23
main/fish: update to 3.0.2

maxice8:


mtneug:
This changes at least the sysconf directory from `/etc` (probably more paths?):
```
> echo $__fish_sysconf_dir
/usr/etc/fish
```
Scripts in `/etc/fish/conf.d/` no longer load. Was this intended?

maxice8:
No, that slipped because i don't use anything in /etc/fish/conf.d, made #7298

PR #7220, opened 2019-04-23
main/s390-tools: fix bash scripts

tuan-hoang1:
Install bash as runtime dependency to run zconf scripts


main/s390-tools/bash.patch
```diff
@@ -0,0 +1,73 @@
...
++++ b/zconf/chccwdev
+@@ -429,9 +429,9 @@ for BUSID in $BUSIDLIST; do
+ 				exit 1
+ 			fi
+ 			if [ "$FORCE" != "" ]; then
+-				echo $FORCE 2>&1 >$SYSPATH/$ONLINEATTR | PrintError
++				/bin/echo $FORCE 2>&1 >$SYSPATH/$ONLINEATTR | PrintError
+ 			else
+-				echo $ONLINE 2>&1 > $SYSPATH/$ONLINEATTR | PrintError
++				/bin/echo $ONLINE 2>&1 > $SYSPATH/$ONLINEATTR | PrintError
```
tcely: Goodness. Those are some odd redirections!

Why is `/bin/echo` better here anyway?

tuan-hoang1:
@ncopa: this one is ready to go. Thanks.

main/s390-tools/bash.patch
tuan-hoang1: Not really odd. Redirecting stderr to stdout (your terminal) so any error from echoing to that /sys path would be caught by `PrintError` function. It could be rewritten as `echo $ONLINE > $SYSPATH/$ONLINEATTR |& PrintError`, if I understand the code correctly :D

PR #7222, opened 2019-04-23
testing/tracker: new aport

Cogitri:


testing/tracker/APKBUILD
```diff
@@ -0,0 +1,50 @@
...
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--buildtype=release \
+		-Dsystemd_user_services=no \
+		. output
+	ninja -C output
+}
+
+check() {
+	cd "$builddir"
```
maxice8: needs to do actual tests

Cogitri: Forgot to push latest version, my bad.

testing/tracker/APKBUILD
```diff
@@ -0,0 +1,68 @@
...
+	ninja -C output test
+}
+
+package() {
+	cd "$builddir"
+	DESTDIR="$pkgdir" ninja -C output install
+}
+
+libs() {
```
Ikke: Because libs is already [defined][libs], this is redundant.

[libs]:https://git.alpinelinux.org/abuild/tree/abuild.in#n1746

Cogitri:
Thank you for the review :)

Cogitri:
Fixed.

PR #7121, opened 2019-04-18
main/nginx: upgrade lua-nginx-module to 0.10.14

frenck:
https://github.com/openresty/lua-nginx-module/releases/tag/v0.10.14

The most important change in this release is the added support for aarch64.
Ref: https://github.com/openresty/lua-nginx-module/commit/7286812116940216344ade33722c49ae47037605

PR #7223, opened 2019-04-23
main/nginx: upgrade to 1.16.0

J0WI:
Fixes #6853 and #7121


J0WI:
I think the test fails due the lack of an IPv6 interface.

andypost:
One test needs removal https://github.com/alpinelinux/aports/pull/6853

J0WI:
Rebased and merged with #6853.

main/nginx/APKBUILD
```diff
@@ -94,14 +93,14 @@ _add_module "http-fancyindex" "v0.4.3" "https://github.com/aperezdc/ngx-fancyind
 _add_module "http-headers-more" "v0.33" "https://github.com/openresty/headers-more-nginx-module"
 _http_headers_more_so="ngx_http_headers_more_filter_module.so"
 
-_add_module "http-lua" "v0.10.13" "https://github.com/openresty/lua-nginx-module"
+_add_module "http-lua" "v0.10.14" "https://github.com/openresty/lua-nginx-module"
```
andypost: Closes https://github.com/alpinelinux/aports/pull/7121

main/nginx/APKBUILD
```diff
@@ -14,13 +14,13 @@
...
 # Odd-numbered versions are mainline (development) versions.
-pkgver=1.14.2
-pkgrel=1
+pkgver=1.16.0
+pkgrel=0
 # Revision of nginx-tests to use for check().
-_tests_hgrev=d6daf03478ad
-_njs_ver=0.2.0
+_tests_hgrev=2be630357aa7
+_njs_ver=0.3.1
```
andypost: Closes https://github.com/alpinelinux/aports/pull/6853

jirutka:
Merged in fba754600f083dd392d15572d860305908eb15ff, thanks!

Ikke:
Build [fails on s390x][s390x] due to lack of luajit.

[s390x]:https://build.alpinelinux.org/buildlogs/build-edge-s390x/main/nginx/nginx-1.16.0-r0.log

J0WI:
Turns out that this condition does not work anymore:
https://github.com/alpinelinux/aports/blob/fba754600f083dd392d15572d860305908eb15ff/main/nginx/APKBUILD#L51-L55

Ikke:
It installs lua5.1-dev, so that's not the issue:

> (31/48) Installing lua5.1-dev (5.1.5-r7)

Ikke:
ncopa suggested disabling this module on s390x if we cannot prevent it from using luajit, something like

```diff
diff --git a/main/nginx/APKBUILD b/main/nginx/APKBUILD
index 8dadad4ea1..57218ce2b1 100644
--- a/main/nginx/APKBUILD
+++ b/main/nginx/APKBUILD
@@ -93,9 +93,14 @@ _add_module "http-fancyindex" "v0.4.3" "https://github.com/aperezdc/ngx-fancyind
 _add_module "http-headers-more" "v0.33" "https://github.com/openresty/headers-more-nginx-module"
 _http_headers_more_so="ngx_http_headers_more_filter_module.so"
 
-_add_module "http-lua" "v0.10.15" "https://github.com/openresty/lua-nginx-module"
-_http_lua_depends="$pkgname-mod-devel-kit"
-_http_lua_provides="$pkgname-lua"  # for backward compatibility
+case "$CARCH" in
+       s390x) ;;
+       *)
+               _add_module "http-lua" "v0.10.15" "https://github.com/openresty/lua-nginx-module"
+               _http_lua_depends="$pkgname-mod-devel-kit"
+               _http_lua_provides="$pkgname-lua"  # for backward compatibility
+       ;;
+esac
```

J0WI:
Can we also try `--without-pcre`?
https://nginx.org/en/docs/ngx_core_module.html#pcre_jit

Ikke:
I'm not sure that's related to luajit, pcre has it's on jit compiler.

jirutka:
Nginx Lua module dropped support for PUC Lua, now it supports only LuaJIT.

What about just disabling nginx package on s390x…?

J0WI:
The whole pkg or just the module? :worried:

andypost:
I suppose disable module is enough

J0WI:
Patches are available to port luajit: https://github.com/LuaJIT/LuaJIT/pull/395

ncopa:
> Nginx Lua module dropped support for PUC Lua, now it supports only LuaJIT.

They dropped support for PUC Lua in a patch release? I don't think we can drop PUC Lua support like that for any architecture in stable branch.

> What about just disabling nginx package on s390x…?

I don't think we can simply remove a package in a stable branch (3.9-stable) in one architecture (s390x) just because it fixes an issue in another (aarch64). nginx is probably one of the few packages that people on s390x actually use (in contrary to desktop packages like xfce and similar).

ncopa:
Ok, it looks like only luajit has been used for all arches except s390x. Dropping the Lua module for s390x is probably the best option here.

PR #7231, opened 2019-04-23
testing/knot-resolver4: disable on s390x

tcely:
LuaJIT is not yet available.


testing/knot-resolver4/APKBUILD
```diff
@@ -3,10 +3,10 @@
 _pkgname="knot-resolver"
 pkgname="${_pkgname}4"
 pkgver=4.0.0
-pkgrel=0
+pkgrel=1
```
Ikke: `pkgrel` bump is not necessary in this case, there is no need to trigger a rebuild.

PR #7237, opened 2019-04-23
testing/jitterentropy-rngd: new aport

HRio:


testing/jitterentropy-rngd/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+# Maintainer:
+pkgname=jitterentropy-rngd
+pkgver=1.0.8
+pkgrel=0
+pkgdesc="CPU Jitter RNG Daemon"
+url="http://www.chronox.de/jent.html"
+arch="all"
+license="GPL-2.0 BSD"
```
maxice8: BSD isn't spdx and is also ambiguous

GPL-2.0 needs -or-later or -only suffix as per SPDX 3.0

testing/jitterentropy-rngd/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+makedepends="linux-headers"
+options="!check" # no test suite
+subpackages="$pkgname-doc $pkgname-openrc"
+source="jitterentropy-rngd-$pkgver.tar.gz::https://github.com/smuellerDD/jitterentropy-rngd/archive/v$pkgver.tar.gz
+	$pkgname.initd
+	"
+builddir="$srcdir/jitterentropy-rngd-$pkgver"
+
+build() {
+	cd "$builddir"
```
Ikke: `cd "$builddir"` is obsolete nowadays and can be removed (as abuild does that automatically).

testing/jitterentropy-rngd/APKBUILD
HRio: thanks, is `"GPL-2.0-only BSD-3-Clause"` OK? when its GPL or BSD (not GPL+BSD)

testing/jitterentropy-rngd/APKBUILD
HRio: OK, any news on when `newapkbuild` will stop adding it?

testing/jitterentropy-rngd/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Henrik Riomar <henrik.riomar@gmail.com>
+# Maintainer:
```
Ikke: You don't want to maintain this?

testing/jitterentropy-rngd/APKBUILD
Ikke: I submitted a PR to remove it from the template.

testing/jitterentropy-rngd/APKBUILD
HRio: I am adding it as something to test for people that report issues with`haveged`, somebody mentioned it on irc the other day I think. If this package does not work better then `haveged` in their systems, we can drop `jitterentropy-rngd` package.

testing/jitterentropy-rngd/APKBUILD
HRio: thanks, will update this PR.

testing/jitterentropy-rngd/APKBUILD
maxice8: check the source code that are under GPL-2.0- , they will have a 

`this is licensed under GNU GPL 2, **or later**` or `this is licensed under GNU GPL 2 **only**`

HRio: there is no later for gpl in the sources.

My question is, whats the proper form for license, when all files are dual licensed. I.e. its not some files bsd and some gpl. But all are dual.

HRio: I changed to "GPL-2.0-only BSD-3-Clause"

PR #7065, opened 2019-04-12
main/gawk: upgrade to 5.0.0

andypost:
https://www.mail-archive.com/info-gnu@gnu.org/msg02589.html

PR #7243, opened 2019-04-24
main/gawk: update to 5.0.0

maxice8:


andypost:
Trying to fix check in https://github.com/alpinelinux/aports/pull/7065

PR #7244, opened 2019-04-24
main/libwebp: update to 1.0.2

maxice8:


main/libwebp/APKBUILD
```diff
@@ -18,7 +18,6 @@ prepare() {
 }
 
 build() {
-	cd "$builddir"
```
ncopa: FWIW, removing those `cd "$builddir"` lines may require more work if/when backporting to stable branches, so in general, it would be nice to keep them for a while. At least in `main` repo.

maxice8:
On Wed, Apr 24, 2019 at 08:48:16AM +0000, Natanael Copa wrote:
> ncopa commented on this pull request.
> 
> 
> 
> > @@ -18,7 +18,6 @@ prepare() {
>  }
>  
>  build() {
> -	cd "$builddir"
> 
> FWIW, removing those `cd "$builddir"` lines may require more work if/when backporting to stable branches, so in general, it would be nice to keep them for a while. At least in `main` repo.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7244#pullrequestreview-229975205

Ack, cd '$builddir' lines won't be removed for packages in main/

PR #7245, opened 2019-04-24
community/dunst: update to 1.4.0

maxice8:


Ikke:
Some tests are failing:


> \* Suite suite_icon:
> FAIL test_get_pixbuf_from_file_tilde: pixbuf (test/icon.c:31) (418 ticks, 0.000 sec)
> FAIL test_get_pixbuf_from_file_absolute: pixbuf (test/icon.c:46) (51 ticks, 0.000 sec)
> PASS test_get_pixbuf_from_icon_invalid: (57 ticks, 0.000 sec)
> FAIL test_get_pixbuf_from_icon_both: SVG pixbuf hasn't precedence (test/icon.c:66) (638 ticks, 0.001 sec)
> FAIL test_get_pixbuf_from_icon_onlysvg: pixbuf (test/icon.c:75) (110 ticks, 0.000 sec)
> PASS test_get_pixbuf_from_icon_onlypng: (136 ticks, 0.000 sec)
> PASS test_get_pixbuf_from_icon_filename: (105 ticks, 0.000 sec)
> FAIL test_get_pixbuf_from_icon_fileuri: pixbuf (test/icon.c:108) (50 ticks, 0.000 sec)

maxice8:
disabled tests for now as they require dbus

Ikke:
@stwa Can you review?

stwa:
Looks good to me! I'm only concerned about the checks as they are disabled now. I can't reproduce any errors caused by dbus, even without a running dbus daemon. Are there any logs to see what happened?
If dbus really is an issue: I would prefer to disable the dbus test suite only and still run the other tests. Should be possible by creating a patch for `Makefile` or `test/dbus.c`.

main/gdk-pixbuf/APKBUILD
```diff
@@ -1,22 +1,20 @@
...
-pkgrel=2
+pkgrel=3
 pkgdesc="GTK+ image loading library"
 url="https://www.gtk.org/"
 arch="all"
-license="LGPL-2.0+"
-depends="shared-mime-info"
-depends_dev=""
+license="LGPL-2.0-or-later"
+depends="shared-mime-info librsvg"
```
maxice8: @ncopa is the addition of librsvg alright ? it is needed for programs that want to load svg files with gdk-pixbuf-loader, like dunst which uses it in its tests.

if it is added here librsvg it won't be necessary to add librsvg on a case by case basis but it will increase the dependencies of gdk-pixbuf.

PR #7261, opened 2019-04-24
main/upower: upgrade to 0.99.10

Cogitri:


main/upower/APKBUILD
```diff
@@ -1,17 +1,23 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=upower
-pkgver=0.99.7
+pkgver=0.99.10
+_distver="$(echo ${pkgver} | tr '.' '_')"
```
tcely:
```suggestion
_distver="${pkgver//\./_}"
```

Cogitri: Ah, APKBUILD warned me that that's a BASHism, isn't it?

tcely: You are correct that's not standard, but we use it all over aports anyway.

Cogitri: Alrighty then

main/upower/APKBUILD
```diff
@@ -1,17 +1,23 @@
...
 url="https://upower.freedesktop.org"
 arch="all"
 license="GPL-2.0-or-later"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-makedepends="linux-headers gtk+-dev libgudev-dev libusb-dev polkit-dev
-	dbus-glib-dev libxslt gobject-introspection-dev docbook-xsl"
-source="https://upower.freedesktop.org/releases/upower-$pkgver.tar.xz"
-# XXX: tests are broken on alpine for some reason, but not on adelie
-options="!check"
+makedepends="linux-headers gtk+-dev libgudev-dev libusb-dev polkit-dev gtk-doc
```
tcely: I am curious, what is `gtk-doc` for here?

main/upower/APKBUILD
tcely:
```suggestion
_pkgname="UPOWER_${pkgver//./_}"
```

main/upower/APKBUILD
```diff
@@ -1,17 +1,23 @@
...
+	dbus-glib-dev libxslt gobject-introspection-dev docbook-xsl autoconf automake
+	py3-dbus"
+source="https://gitlab.freedesktop.org/upower/upower/-/archive/UPOWER_${_distver}/upower-UPOWER_${_distver}.tar.bz2"
```
tcely:
```suggestion
source="https://gitlab.freedesktop.org/${pkgname}/${pkgname}/-/archive/${_pkgname}/${pkgname}-${_pkgname}.tar.bz2"
```

main/upower/APKBUILD
```diff
@@ -1,17 +1,23 @@
...
+builddir="$srcdir/$pkgname-UPOWER_${_distver}"
```
tcely:
```suggestion
builddir="${srcdir}/${pkgname}-${_pkgname}"
```

Cogitri: Uh? We don't seem to use curly braces for these elsewhere

main/upower/APKBUILD
Cogitri: I've been asked to write pkgname out here by @Ikke in another PR

main/upower/APKBUILD
Cogitri: Required for autogen.sh because we don't have a gtk-doc-am package or similar from what I can see

main/upower/APKBUILD
tcely: Yes, some Alpine Linux folks are anti braces. My advice is to always use them, it's a tiny bit less to track in your head when figuring out which characters will give you an unset variable.

main/upower/APKBUILD
tcely: You mean, not to use the variable? Did you get an explanation as to why?

main/upower/APKBUILD
Cogitri: I'm totally with you there, I rather place braces than debug weird errors, but maybe it'd make sense to implement some common guide line for this? I've been told in a review before to remove them, so it kind of sucks when I have to guess what "kind" of reviewer will look at my stuff :)

main/upower/APKBUILD
Cogitri: Uh..I can search for it tomorrow, not quite sure what PR it was in.

main/upower/APKBUILD
tcely: Interesting.

main/upower/APKBUILD
tcely: I totally agree with you. So many of the reviewer requests seem to come out of nowhere. We should certainly have them written down, but changes in this area have been tremendously slow in my experience.

main/upower/APKBUILD
Ikke: The upstream package name is separate from the aports package name, they can change for different reasons. Therefore it's not recommended to use `pkgname` in the source url and other places that is not controlled by the APKBUILD.

One usecase is when you would create a variant of this package (so with a different package name). Then suddenly the build breaks because the url is incorrect.

tcely: So, let's solve the usecase you're concerned about without violating DRY.

main/upower/APKBUILD
ncopa: I support this. In general we should try avoid using bashisms so we should try avoid `${var//search/replace}`, however, i prefer that bashism over using forks in global scope, which will slow down parsing of APKBUILDs in aports tree.

main/upower/APKBUILD
ncopa: I tend to prefer try avoid use variables in URLs because it makes it easier to copy the URL (or part of the URL) paste to browser in addition to the reason @_ikke_ mentioned. I think the value of those outweights the value of DRY.

main/upower/APKBUILD
ncopa: I prefer to only use braces where it makes sense to use them not everywhere. But I respect that others may disagree so I'm not gonna stand in the way if a package maintainer insists on braces everywhere in APKBUILDs he/she maintains.

ncopa:
Feel free to take over maintainership if you insist on using variables in using variables in URLs or use braces everywhere. Please respect maintainer's wishes otherwise.

Cogitri:
I didn't use braces and kept writing the pkgname out here since I'm not exactly interested in taking maintainership of upower here. Thanks for the reviews :)

PR #7274, opened 2019-04-24
testing/perl-crypt-des_ede3: new aport

timlegge:

PR #7279, opened 2019-04-25
testing/perl-convert-pem: new aport

timlegge:


timlegge:
> LGTM although i don't like having a 3k loc patch

Me either.  It basically deletes the outdated included Test Module and allows the default tests to run.  The patch process runs automatically but I suppose I could have deleted the offending directory in the prepare

maxice8:
IMO it is better to have that in prepare()

timlegge:
> IMO it is better to have that in prepare()

Will do, need to wait on some PRs to get dependencies anyway

maxice8:
You can have a single PR for multiple packages if one depend on another, just have them on separate commits

timlegge:
> You can have a single PR for multiple packages if one depend on another, just have them on separate commits

Wish I had known that several weeks ago :-(

Ikke:
`perl-crypt-des_ede3` is currently missing as dependency.

timlegge:
> `perl-crypt-des_ede3` is currently missing as dependency.
> 
> _edit:_
> I just pushed #7274

sorted out the dependency issues.  Thanks

PR #7281, opened 2019-04-25
testing/perl-crypt-smime: new aport

timlegge:


testing/perl-crypt-smime/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+# Contributor: Timothy Legge <timlegge@gmail.com>
+# Maintainer: Timothy Legge <timlegge@gmail.com>
+pkgname=perl-crypt-smime
+_pkgreal=Crypt-SMIME
+pkgver=0.25
+pkgrel=0
+pkgdesc="S/MIME message signing, verification, encryption and decryption"
+url="https://metacpan.org/release/Crypt-SMIME/"
+arch="all"
+license="Unknown"
```
maxice8: According to [wiki](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license), license unknown is not acceptable.

maxice8:
Please squash into a single commit

testing/perl-crypt-smime/APKBUILD
timlegge: Ah.  That's why apkbuild-cpan.in has a special case for not inserting the info if it unknown.  Fixing

PR #7282, opened 2019-04-25
py-pyrsistent: rename to py-pyrsistent and create py2- subpackage

maxice8:


tcely:
Is there a reason to add `py2-pyrsistent` when `python2` EOL is so near?

maxice8:
To fulfill https://bugs.alpinelinux.org/issues/10282 but in retrospect i'd rather only fix py3

PR #7284, opened 2019-04-25
main/dialog: update to 1.3.20190211

maxice8:
- Remove static library


main/dialog/APKBUILD
```diff
@@ -30,6 +30,9 @@ build() {
 package() {
 	cd "$builddir"
 	make DESTDIR="$pkgdir" install
+
+	# Get rid of static library
+	rm -rf "$pkgdir"/usr/lib
```
tcely: Wouldn't it be better to split this into a `-static` subpackage instead?

maxice8: ok

PR #7294, opened 2019-04-25
community/restic: needs SSH for remote repositories

tcely:


nmeum:
I would argue that this is an optional dependency and IIRC the policy was no to include optional dependencies in `$depends`.

tcely:
@nmeum some room for that argument does exist. Losing the ability to access storage for a backup program does not qualify as optional in my opinion.

clandmeter:
I guess if you are backing up you already need openssh installed so it will be available when you need to restore. Also where do you draw the line for the many storage solutions restic has support for? Let the people choose their weapon instead of deciding for them.

nmeum:
> Losing the ability to access storage for a backup program does not qualify as optional in my opinion.

I store some of my backups on external hard drives, no openssh required to access/create those.

tcely:
There are two external tools that `restic` uses: 1) `ssh` and 2) `rclone`.

I draw the line at breaking the expected user experience. I would depend on `rclone` too but it's not packaged. As for letting them choose, we are actively preventing that choice by not depending on `ssh`. Expecting users to go dig through the `restic` sources to figure out why the examples in the documentation (which they are most likely to start with) don't work is unreasonable.

We have a set of users who are happy with never using options to `find` that are not in the `busybox` applet. We don't go around breaking software that uses GNU `find`, we provide `findutils` and add it to `depends` when some other package is using it.

tcely:
> Since restic works perfectly fine without ssh I am _personally_ against this change and won't merge this.

I strongly disagree with your classification of `ssh` as optional. It is in no way optional to access my backups. The idea that parts of `restic` keep working (for you) does nothing to change this fact. In the current form, for my environment, this package installs broken (unfit to be used). Unless whoever is installing it has run into the problem before, and remembers the solution. Otherwise, yet another round of issue reporting then resolution imust be performed.

Ikke:
I feel this is more like an optional dependency then a hard dependency, but apk has no built-in concept of optional dependencies.

tcely:
Without the concept of optional dependencies, we either include the dependency or remain broken.

The feelings about the type of dependency `restic` has on the `ssh` command are really beside the point here. Especially since we don't have the type you feel would best fit.

I (as the user) chose to install `restic` (and everything it depends on). Your decision (as package developers) to omit a dependency because of feelings about the type of that dependency should not cause errors for me.

As it stands, a non-empty set of users' `restic` commands are failing until each user debugs the problem you created.

maxice8:
any progress on this discussion ?

tcely:
I'm not going to change my mind about this. #8248 is waiting now also.

When that is in `community` too, then I'll argue for depending on `ssh` and `rclone` as well.

PR #7295, opened 2019-04-25
main/rng-tools: move from community

HRio:


ncopa:
Could we have a short explanation on why we move it to main in commit message?

HRio:
> Could we have a short explanation on why we move it to main in commit message?

Agree.

What do you think about this

> Maintaining the system entropy pool with entropy pool with a HWRNG is quite important
on embedded devices (e.g. RPI3) and on virtual instances, due to that these systems
generate very little entropy on their own.

> Starting rng-tools in runlevel boot on a system with a hwrng avoids that the Linux
kernel "blocks" for a couple of minutes during early boot.

> rng-tools supports feeding the kernel entopy pool with random data from different
sources, among these /dev/hwrng and the Intel `rdrand` instruction.

HRio:
@ncopa rephrased the draft a bit and added a part about installer media. Now part of the git commit message.

PR #7299, opened 2019-04-25
main/libwebp: enable static libraries

maxice8:
fixes https://bugs.alpinelinux.org/issues/10260


maxice8:
Is there really a reason ? if you're linking against static libs which are one use only, might as well put them in -dev where it also can benefit from possible pkgconfig files telling how to properly link it via Requires.private.

Also i see no default -static function which would be nice.

PR #7301, opened 2019-04-25
main/curl: disable SSH support via libssh2

maxice8:
fixes #10222

See: https://bugs.alpinelinux.org/issues/10222 for more info

### DIFF
--- a/main/curl/APKBUILD
+++ b/main/curl/APKBUILD
@@ -4,13 +4,13 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=curl
 pkgver=7.64.1
-pkgrel=1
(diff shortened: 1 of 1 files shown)


main/curl/APKBUILD
```diff
@@ -90,7 +90,7 @@ build() {
 		--enable-unix-sockets \
 		--without-libidn \
 		--without-libidn2 \
-		--with-libssh2 \
+		--without-libssh2 \
```
nmeum: It would probably be a good idea to add a link the redmine bug in this line to ensure that this isn't enabled again in the near future.

maxice8: Applied, thanks for the suggestion

PR #7304, opened 2019-04-25
main/xfsprogs: update to 4.20.0

maxice8:


main/xfsprogs/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
 pkgname=xfsprogs
-pkgver=4.19.0
-pkgrel=1
+pkgver=4.20.0
+pkgrel=0
 pkgdesc="XFS filesystem utilities"
 url="http://xfs.org/index.php/Main_Page"
 arch="all"
-license="LGPL"
+license="LGPL-2.1-or-later"
```
clandmeter: Next time please mention if you do more than just bump, this so we know what to look for.

Thanks!

maxice8: Should be in title or can be in commit body ?

PR #7307, opened 2019-04-25
main/dbus-glib: update to 0.110

maxice8:


ncopa:
i was a bit too quick on the trigger here. One test fails:
````
FAIL: tests/test-30574
````

I have disabled tests to unblock the builders for now.

maxice8:
ack

PR #7309, opened 2019-04-25
community/py3-pyrsistent: don't require net in abuild

maxice8:


Ikke:
Suggested commit message:

```
community/py3-pyrsistent: don't require net in abuild

By using py3-pytest-runner, a network connection is no longer required for building
```

maxice8:
Updated, thank you for the commit suggestion.

PR #7310, opened 2019-04-25
testing/knot-resolver4: fix sigsegv in stats module init

tcely:


maxice8:
i think pkgrel needs to be bumped

maxice8:
squash all into a single commit

testing/knot-resolver4/stats-module-memset-upstreams.patch
```diff
@@ -0,0 +1,10 @@
+--- a/modules/stats/stats.c
++++ b/modules/stats/stats.c
+@@ -498,6 +498,7 @@
+ 	}
+ 	for (size_t i = 0; i < UPSTREAMS_COUNT; ++i) {
+ 		struct sockaddr *sa = (struct sockaddr *)&data->upstreams.q.at[i];
++		memset(sa, 0, sizeof(struct sockaddr));
```
Ikke: I'm by no means an expert, but why would it be necessary / make sense to zero-ize `sa` after assigning it a value? Could you not just ignore the assignment and immediately set sa to 0?

tcely: Assigning the sa pointer to poiont at the item in the `q` array can not be skipped. We want to zero-ize the `sockaddr` struct in the array before we attempt to set a field in that struct.

Ikke: Ok, thanks for the explanation.

tcely: You're welcome. As to why this seems to be needed for us and not others, I am not clear on that.

PR #7320, opened 2019-04-26
main/haveged: remove net dependency

tofurky:
haveged does not require any network connectivity to function, and this
errant dependency can cause the daemon to not start properly at boot.

### DIFF
--- a/main/haveged/APKBUILD
+++ b/main/haveged/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=haveged
 pkgver=1.9.4
-pkgrel=3
+pkgrel=4
 pkgdesc="Entropy harvesting daemon using CPU timings"
 url="http://www.issihosts.com/haveged/"
 arch="all"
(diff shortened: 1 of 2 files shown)


tofurky:
@jirutka please take a look if you can, thanks!

tomalok:
rng-tools' `rngd` init script says...
```
depend() {
	need localmount
	after urandom
	provide entropy
}
```
do the localmount and urandom dependencies make sense for `haveged` too?

tofurky:
hmm, not sure. https://github.com/alpinelinux/aports/blob/master/testing/jitterentropy-rngd/jitterentropy-rngd.initd has those as well.
on my system (postmarketos, following edge) urandom isn't enabled. localmount might be needed if a system has a separate /usr partition or something i guess.

HRio:
> on my system (postmarketos, following edge)

the `urandom` "service" needs a place to store the data from last run, so if postmarketos is running "diskless" i.e. from rams that could explain why its not enabled. If there is a good place to store the data (seed) for `urandom` that service is a good thing to have. (the place for the seed is configured in /etc/conf.d/urandom)

HRio:
> do the localmount and urandom dependencies make sense for `haveged` too?

I would say so yes.

tomalok:
@tofurky `after urandom` will only ensure that `haveged` will start after `urandom` *if* `urandom`is enabled.

tofurky:
> > on my system (postmarketos, following edge)
> 
> the `urandom` "service" needs a place to store the data from last run, so if postmarketos is running "diskless" i.e. from rams that could explain why its not enabled. If there is a good place to store the data (seed) for `urandom` that service is a good thing to have. (the place for the seed is configured in /etc/conf.d/urandom)

it's running from flash (f2fs). not sure why urandom is disabled. i don't see anything obvious in their image build scripts.

anyways i will push an updated PR in the next day or so, thanks.

tofurky:
i've modified the init script to add those other dependencies and push --force'd to my branch.

i did edit the commit message but left the PR as-is, please let me know if that should be edited to reflect the current commit, thanks!

PR #7326, opened 2019-04-26
main/powertop: update to 2.10

maxice8:


main/powertop/APKBUILD
```diff
@@ -24,11 +20,15 @@ build() {
 	make
 }
 
+check() {
+	# There is no testsuite, check the binary
+	./src/powertop --version
+}
+
```
clandmeter: We have had discussions about this before and we prefer not to use a simple test like this.
Check function should always use a testsuite or else disable it. We dont want to give the impression
we test our packages but instead just do a simple version check.

maxice8: Can you update https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#options !check entry to reflect this change in policy ?

maxice8: Updated. Thanks for the review.

ncopa:
Checksums needs to be fixed

ncopa:
This is a bug:
````
event-parse.c:5124:7: warning: assignment to 'const char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
   msg = strerror_r(errnum, buf, buflen);
       ^         
````

It seems to expect GNU libc `strerror_r` behavior. It needs to be fixed to use POSIX variant, which returns `0` for success and `-1` for fail.

PR #7328, opened 2019-04-26
community/py-watchdog: split into py2- and py3- packages

maxice8:
- py3- is required for testing py3-werkzeug

### DIFF
--- a/community/py-watchdog/APKBUILD
+++ b/community/py-watchdog/APKBUILD
@@ -2,37 +2,46 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=py-watchdog
 pkgver=0.9.0
-pkgrel=0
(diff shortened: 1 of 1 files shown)


Ikke:
Only py-sphinx-watchdog seems to depend on this package. It might be possible to switch to just python3.

maxice8:
Nothing depends on py-sphinx-watchdog, are there any pointers i can take to see if it is better to switch to python3 or provide both ?

prspkt:
@maxice8 [This](https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Python) will probably cover any missing points related to Python. You can use the last 2 commits from [here](https://github.com/prspkt/aports/commits/py-watchdog) to upgrade `py-watchdog` to python3 only.

maxice8:
Thanks

PR #7329, opened 2019-04-26
main/libmtp: upgrade to 1.1.16

maxice8:


maxice8:
@ncopa

ncopa:
seems like it fails to build on aarch64?

PR #7332, opened 2019-04-26
unmaintained/py-urlnorm: move from main

maxice8:
- Is python2 only (fails to build on python3)
- No maintainer
- Outdated


tcely:
The suggested message form for this kind of commit is:
`<DST>: move from <SRC`

Please update to use `unmaintained/py-urlnorm: move from main` instead.

maxice8:
On Fri, Apr 26, 2019 at 03:40:10PM -0700, tcely wrote:
> The suggested message form for this kind of commit is:
> `<DST>: move from <SRC`
> 
> Please update to use `unmaintained/py-urlnorm: move from main` instead.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7332#issuecomment-487222886

Updated

PR #7342, opened 2019-04-26
main/py-psycopg2: upgrade to 2.8.2

fabaff:


main/py-psycopg2/APKBUILD
```diff
@@ -2,8 +2,8 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=py-psycopg2
 _pkgname=psycopg2
-pkgver=2.7.7
-pkgrel=1
+pkgver=2.8.2
+pkgrel=0
 pkgdesc="PostgreSQL adapter for python"
 url="http://initd.org/psycopg"
 arch="all"
```
maxice8: license says it is LGPL-3.0-or-later.

 LGTM after the change is applied.

Ikke:
CI fails because it's missing a dependency on `py-setuptools`, which can be a `makedepends`, but in somecases even a normal `depends`.

maxice8:
Check packages that provide binaries for import of pkg_resources, those are part of py-setuptools, otherwise it can go in makedepends.

PR #7343, opened 2019-04-26
community/xkcdpass: upgrade to 1.17.3

fabaff:


maxice8:
identation inside check() function is wonky

PR #7344, opened 2019-04-26
community/py-markdown: upgrade to 3.1

fabaff:


community/py-markdown/APKBUILD
```diff
@@ -2,8 +2,8 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=py-markdown
 _pkgname=Markdown
-pkgver=3.0.1
-pkgrel=1
+pkgver=3.1
+pkgrel=0
 pkgdesc="A Python implementation of Markdown"
 url="http://pythonhosted.org/Markdown/"
 arch="noarch"
```
maxice8: BSD isn't an SPDX identifier and is ambiguous

maxice8: empty variables can be removed from brevity

maxice8: makedepends only need py-setuptools and py3-setuptools

andypost: I think this needs separate issue cos a lot of packages affected

PR #7347, opened 2019-04-26
community/chromium: fix sandbox crash with musl 1.1.22

nmeum:
Fixes #10257

**Don't merge yet, I am still building this locally**


ncopa:
i think this was fixed with b6ee7c5d8a5f54c54c344ee055d7e9a1af034fd6?

nmeum:
Ah indeed so address space access is sufficient. fair enough. will close the redmine issue then.

PR #7350, opened 2019-04-26
testing/k3s: new aport

OTLabs:
https://k3s.io
Lightweight Kubernetes. 5 less than k8s.

### DIFF
--- a/testing/k3s/APKBUILD
+++ b/testing/k3s/APKBUILD
@@ -1,0 +1,56 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=k3s
+pkgver=0.4.0
+pkgrel=0
+pkgdesc="Lightweigt Kubernetes. 5 less than k8s."
+url="https://k3s.io"
+arch="all"
+license="Apache-2.0"
+options="!check" # No test suite from upstream
+depends="containerd"
+makedepends="go linux-headers"
+install=""
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/rancher/k3s/archive/v$pkgver.tar.gz"
+builddir="$srcdir/src/github.com/rancher/$pkgname"
+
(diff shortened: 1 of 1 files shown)


Ikke:
Build fails with:

> /usr/bin/abuild: cd: line 2254: can't cd to /home/buildozer/aports/testing/k3s/pkg/k3s: No such file or directory

OTLabs:
@Ikke The packaging section was missed in commit, now it is fixed.

testing/k3s/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+
+	default_prepare
+}
+
+build() {
+	msg2 "Prepare GOPATH and GOBIN..."
+	export GOPATH="$srcdir"
+	export GOBIN="$GOPATH/bin"
+
+	msg2 "Building k3s..."
```
andypost: not sure it makes sense to keep it

OTLabs: There is no any output during build process of each of 3 targets and the build process takes some time. I used these msg2 to have a visual indication of build progress and this gives me an idea which target produces errors if any. That's the motto behind use of msg2. If you feel that removing them would be beneficial it would be my pleasure to incorporate these changes.

testing/k3s/APKBUILD
```diff
@@ -0,0 +1,67 @@
...
+	go build -o k3s
+
+	msg2 "Building kubectl..."
+	go build -o kubectl ./cmd/kubectl
+
+	msg2 "Building hypercube..."
+	go build -o hyperkube ./vendor/k8s.io/kubernetes/cmd/hyperkube
+}
+
+check() {
```
Ikke: The [policy][] officially changed that checks should use test suites from upstream. So just running the command in some way no longer suffices. So if no test suite is available from upstream, the package should be marked as such.

[policy]:https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#options

OTLabs: I have removed checks, thank you for bringing to my attention new corresponding policy to follow.

PR #7352, opened 2019-04-26
non-free/unifi: upgrade to 5.10.21

TBK:


Ikke:
This depends on mongodb, which has moved to non-free as well, and thus is not available in the repositories.

TBK:
I can remove the dependency. In production I build the pkg without it anyway and use a different container for mongodb. I was just wondering why Drone passes and Travis does not...

Ikke:
Because drone is not even trying to build it.

TBK:
I realized that. I was more why one tries to build and the other does not. Will try to go through the CI code and see if and find where the logic differs.

Second question, what is intended procedure for non-free, should it build or ignore?

Ikke:
@clandmeter ?

clandmeter:
I think we dont build for nonfree, but i guess we could add it.

PR #7354, opened 2019-04-26
Create no-response.yml

tcely:
https://probot.github.io/apps/no-response/

This is a feature @ncopa wanted when discussing improvements on the mailing list.
https://lists.alpinelinux.org/alpine-devel/6219.html

If using this label, we need one of the owners to add this bot and to merge this PR into the git repository.


Ikke:
This seems to do the same job as stalebot?

tcely:
https://probot.github.io/apps/stale/

Not exactly. They are similar but this one is more targeted in when the label is removed.

PR #7355, opened 2019-04-26
testing/ufw-extras: new aport

TBK:


testing/ufw-extras/APKBUILD
```diff
@@ -0,0 +1,21 @@
...
+arch="noarch"
+license="GPL-2.0-or-later"
+options="!check" # No test suite
+depends="ufw"
+source="$pkgname-$pkgver.tar.gz::https://github.com/xyproto/ufw-extras/archive/$pkgver.tar.gz"
+
+package() {
+	for f in "ufw-"*
+	do
+		install -Dm644 "$f" "$pkgdir/etc/ufw/applications.d/$(basename $f)"
```
maxice8: is the basename necessary ? the for loop is not transversing a directory.

also i think the more common usage of the for loop is with `; do` instead of the `do` in the next line

maxice8: otherwise LGTM

TBK: Thx for the review. I have changed the for loop syntax and removed basename.

testing/ufw-extras/APKBUILD
```diff
@@ -0,0 +1,21 @@
+# Contributor: TBK <alpine@jjtc.eu>
+# Maintainer:
```
Ikke: I gather you don't want to maintain this package?

TBK: I left i unfilled, because I felt it would be better if the person wanting to maintain ufw also claimed maintainership of this.

PR #7357, opened 2019-04-26
main/ninja: install ninja_syntax.py

tml:
Some software that wants to use the Python API of the ninja build system requires the ninja_syntax.py be installed alongside ninja itself.

### DIFF
--- a/main/ninja/APKBUILD
+++ b/main/ninja/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=ninja
 pkgver=1.9.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Small build system with a focus on speed"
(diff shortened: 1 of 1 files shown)


maxice8:
pkgrel needs to be bumped to affect changes to packages

tcely:
Please look into splitting this into a subpackage. Also, what happens when python3.7 is used?

main/ninja/APKBUILD
```diff
@@ -39,6 +39,9 @@ package() {
 
 	install -m644 -D misc/bash-completion \
 		"$pkgdir/usr/share/bash-completion/completions/ninja"
+
+	install -m644 -D misc/ninja_syntax.py \
+		"$pkgdir/usr/lib/python3.6/site-packages/ninja_syntax.py"
```
tcely: Please consider what happens when python3.7 is used.

We should not hard code the python version in the installation path.

PR #7360, opened 2019-04-26
Create no-response.yml

tcely:
https://probot.github.io/apps/no-response/

This is a feature @ncopa wanted when discussing improvements on the mailing list.
https://lists.alpinelinux.org/alpine-devel/6219.html

If using this label, we need one of the owners to add this bot and to merge this PR into the git repository.

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,9 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 33
+# Label requiring a response
+responseRequiredLabel: Response-Required
+
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+#closeComment: false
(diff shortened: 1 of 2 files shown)


Ikke:
@tcely If I read that request from ncopa, it appears what he wants is the opposite. Not closing when someone did not respond, but add a label when they do respond, so that we know that we need to do something now.

tcely:
This work flow is add a label when a response is required, and have the bot remove that label when the author responds. Closing is optional.

It makes sense to close PRs that go unresponsive in my opinion, but if it's decided we should keep them open that is fine too.

Ikke:
The problem is that a PRs can disappear in the noise when we requested changes and the user then responds. By removing the label in that case, you cannot find these PRs among all other PRs.

If we would add a label in that case, you can now filter the PRs where the user has responded. That's the idea at least.

tcely:
So perhaps we should fork this bot and change it to add a label when the response is seen.

tcely:
In light of https://github.com/probot/no-response/issues/14 we will be better served by our own implementation anyway.

PR #7362, opened 2019-04-26
main/haproxy: upgrade to 1.9.7 and fix license

andypost:
Ref https://www.mail-archive.com/haproxy@formilux.org/msg33410.html

http://www.haproxy.org/download/1.9/src/CHANGELOG

### DIFF
--- a/main/haproxy/APKBUILD
+++ b/main/haproxy/APKBUILD
@@ -1,20 +1,19 @@
 # Contributor: Jeff Bilyk <jbilyk@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=haproxy
-pkgver=1.9.6
+pkgver=1.9.7
 _pkgmajorver=${pkgver%.*}
 pkgrel=0
 pkgdesc="A TCP/HTTP reverse proxy for high availability environments"
-url="http://haproxy.1wt.eu"
+url="https://www.haproxy.org/"
 arch="all"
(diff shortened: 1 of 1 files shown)


main/haproxy/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Jeff Bilyk <jbilyk@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=haproxy
-pkgver=1.9.6
+pkgver=1.9.7
 _pkgmajorver=${pkgver%.*}
 pkgrel=0
 pkgdesc="A TCP/HTTP reverse proxy for high availability environments"
```
maxice8: license is 'GPL-2.0-or-later  LGPL-2.1-or-later'

maxice8: empty variables can be removed for brevity

andypost: License changed, according to http://www.haproxy.org/download/1.5/doc/LICENSE

maxice8:
Shouldn't it also have a $pkgname-openrc package because it provides an openrc service ?

main/haproxy/APKBUILD
```diff
@@ -1,13 +1,13 @@
...
 url="http://haproxy.1wt.eu"
```
arambau: The URL for HAProxy should be https://www.haproxy.org/.

PR #7366, opened 2019-04-27
main/postgresql: depend on readline, not libedit

5paceToast:
Fixes [bugs.a.o #10354](https://bugs.alpinelinux.org/issues/10354)

### DIFF
--- a/main/postgresql/APKBUILD
+++ b/main/postgresql/APKBUILD
@@ -1,9 +1,10 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 # Contributor: G.J.R. Timmer <gjr.timmer@gmail.com>
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
+# Contributor: Chloe Kudryavtsev <toast@toastin.space>
 pkgname=postgresql
 pkgver=11.2
-pkgrel=1
+pkgrel=2
 pkgdesc="A sophisticated object-relational DBMS"
 url="https://www.postgresql.org/"
 arch="all"
@@ -14,7 +15,7 @@ pkgusers="postgres"
 pkggroups="postgres"
 checkdepends="diffutils"
 depends_dev="openssl-dev"
-makedepends="$depends_dev libedit-dev zlib-dev libxml2-dev util-linux-dev
+makedepends="$depends_dev readline-dev zlib-dev libxml2-dev util-linux-dev
 	openldap-dev tcl-dev perl-dev python2-dev python3-dev"
 subpackages="$pkgname-contrib $pkgname-dev $pkgname-doc libpq $pkgname-libs
 	$pkgname-client $pkgname-pltcl


andypost:
Any pointer why libedit not preferable?

5paceToast:
@andypost according to bugs.a.o [#10354](https://bugs.alpinelinux.org/issues/10354) (and the linked [ML post](https://www.postgresql.org/message-id/opsdtodlpwg2z5qo%40relay.plus.net)) using libedit causes segmentation faults and adds unwanted strings into history.
This is likely some incompatibility in the API, and is fixed by using readline instead.

jirutka:
It sems that we cannot (shouldn’t) build PostgreSQL with readline due to incompatible licenses:

[@ncopa on #10354](https://bugs.alpinelinux.org/issues/10354#note-3):
> FYI, postgresql is BSD licensed and GNU readline is GPL (not LGPL).

Read also [Milan P. Stanić’s comment](https://bugs.alpinelinux.org/issues/10354#note-4):
> Arch and Void Linux builds with readline anyway (as I see in their build files)
and Debian uses Perl wrapper to overcome issue and preload readline when psql wrapper invoked.
What would be acceptable for Alpine, build with readline or 'stole' Perl wrapper from Debian.
(and this wrapper have some other useful features, btw)

That Debian’s perl script he mentioned is [this](https://sources.debian.org/src/postgresql-common/200+deb10u1/pg_wrapper/). As you can see, the interesting part for readline is just about setting `LD_PRELOAD`, so we can rewrite it to shell to avoid unnecessary dependency on Perl.

5paceToast:
I disagree with the Debian approach from a legal and philosophical standpoint.
1. PostgreSQL's liberal BSD license is compatible with the GPL (it does make the binary (derivative work) effectively GPL too, but it isn't a direct conflict).
2. In terms of the potential conflict (OpenSSL being potentially incompatible), I agree with RedHat's (Fedora's) lawyers: I think OpenSSL is subject to the system library exception.

For some extra discussion, please see the [related lwn article](https://lwn.net/Articles/428111/).
On the other hand, dynamic linking (and LD_LIBRARY_PATH and other ld.so companions in particular) is a known security hole (see, for example, [rpath+ldd privilage escalation](https://www.contextis.com/en/blog/linux-privilege-escalation-via-dynamically-linked-shared-object-library) and [CVE-2017-1000253](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-1000253)).

I would personally prefer the technologically "clean" solution, and allow RedHat's resources to defend us in the unlikely scenario the FSF suddenly decides this is not legitimate use.

There is also the additional challenge that if we wanted to compile against libedit, but still use readline in runtime, we would also need to find a way to get around abuild adding an automatic dependency to `so:libedit.so.0=0.0.60` - something I'm not sure is easily achievable.
I'm willing to explore that last option, but would ask for help on getting that done, in case you have any ideas.

vieira:
While I appreciate the philosophical and purity standpoints, I am still presented with sigsegv everytime I paste a long query in psql, and history does not work.

If even debian does not build against libedit maybe we should too put the more practical considerations above the philosophical ones.

ncopa:
as I understand, this is a bug in libedit?  I think our options are to either fix libedit or remove it completely. Since there are more apps using libedit I think we should first report it upstream, and then fix it.

vieira:
This is a regression from Alpine 3.8 / psql 10.8

PR #7372, opened 2019-04-27
main/uriparser: upgrade to 0.9.3

maxice8:
- Switch build system to cmake
- Fix license identifier
- enable tests but disable a failing test that causes segfaults
- split library into lib$pkgname


Ikke:
@fabaff Can you review this PR? Thanks.

PR #7373, opened 2019-04-27
testing/acme.sh: add missing openssl dependency

maxice8:
- Also removed superfluous return 0 statements

fixes #10355

### DIFF
--- a/testing/acme.sh/APKBUILD
+++ b/testing/acme.sh/APKBUILD
@@ -2,27 +2,17 @@
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=acme.sh
 pkgver=2.8.0
-pkgrel=0
+pkgrel=1
 pkgdesc="An ACME Shell script, an acme client alternative to certbot"
+options="!check" # No testsuite
 url="https://github.com/Neilpang/acme.sh"
 arch="noarch"
 license="GPL-3.0"
-depends="curl"
+depends="curl cmd:openssl"
 source="$pkgname-$pkgver.tar.gz::https://github.com/Neilpang/acme.sh/archive/$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$pkgver"
 
-build() {
-	cd "$builddir"
-	return 0
-}
-
-check() {
-	cd "$builddir"
-	return 0
-}
-
 package() {
-	cd "$builddir"
 	install -dm755 "$pkgdir"/usr/share/acme.sh/deploy
 	install -dm755 "$pkgdir"/usr/share/acme.sh/dnsapi
 	install -dm755 "$pkgdir"/usr/bin


maxice8:
@fcolista

testing/acme.sh/APKBUILD
```diff
@@ -2,23 +2,21 @@
...
 pkgname=acme.sh
 pkgver=2.8.0
-pkgrel=0
+pkgrel=1
 pkgdesc="An ACME Shell script, an acme client alternative to certbot"
 url="https://github.com/Neilpang/acme.sh"
 arch="noarch"
 license="GPL-3.0"
-depends="curl"
+depends="curl openssl"
```
Ikke: `cmd:openssl` might suffice in this case.

maxice8: done

testing/acme.sh/APKBUILD
```diff
@@ -2,23 +2,21 @@
...
 pkgname=acme.sh
 pkgver=2.8.0
-pkgrel=0
+pkgrel=1
 pkgdesc="An ACME Shell script, an acme client alternative to certbot"
 url="https://github.com/Neilpang/acme.sh"
 arch="noarch"
 license="GPL-3.0"
-depends="curl"
+depends="curl cmd:openssl"
```
TBK:
```suggestion
options="!check" # No test suite
depends="curl cmd:openssl"
```

testing/acme.sh/APKBUILD
```diff
@@ -2,23 +2,21 @@
...
 source="$pkgname-$pkgver.tar.gz::https://github.com/Neilpang/acme.sh/archive/$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$pkgver"
```
TBK:
```suggestion
```

testing/acme.sh/APKBUILD
```diff
@@ -2,23 +2,21 @@
...
 
 build() {
 	cd "$builddir"
-	return 0
 }
 
 check() {
 	cd "$builddir"
-	return 0
 }
```
TBK: You can remove both the build and check function since they do nothing.

testing/acme.sh/APKBUILD
Ikke: Why do you suggest removing `builddir`?

testing/acme.sh/APKBUILD
maxice8: ah, ok. Wasn't sure if they were needed or not

testing/acme.sh/APKBUILD
maxice8: ncopa asked me to keep them for packages in main/, should i keep them on all repos ?

TBK: I am not sure what the styling policy is atm. I thought it was "old main" repo pkgs that only was required to keep the builddir variable even though it is the default path to minimize backporting comparability issues. 

abuild sets builddir if not present in the apkbuild file to the default `$srcdir/$pkgname-$pkgver`: 
https://git.alpinelinux.org/abuild/commit/abuild.in?id=4f37c8efd306335a9c3ac6862b0984c965bf270f

cd "$builddir" is not required since abuild commit:
https://git.alpinelinux.org/abuild/commit/?id=2fe29d5829c0973ace1db350141b3c810ac888a7
https://git.alpinelinux.org/abuild/commit/?id=324a57592e5e8f0196bb1bd082c278d82673af8c

I personally find it easier to spot if something funky is going on with the builddir path if it is only included "when needed". It could be considered exploiting abuild or a matter of styling opinion.

testing/acme.sh/APKBUILD
Ikke: abuild defines all these functions. When you put them in in your APKBUILD, you just override them.

testing/acme.sh/APKBUILD
maxice8: so should i remove them only on non-main/ packages ?

Ikke: yes, because packages in main are more likely to get backported, you want to avoid making too many unnecessary changes there.

maxice8: ok, thank you

maxice8:
fixed

PR #7380, opened 2019-04-27
testing/ibus: install pkg-config file

Cogitri:


Ikke:
@ddevault Please review

testing/ibus/APKBUILD
```diff
@@ -23,9 +23,12 @@ source="
 install="$pkgname.post-install"
 builddir="$srcdir/$pkgname-$pkgver"
 
+prepare() {
+	NOCONFIGURE=1 ./autogen.sh
+}
```
Cogitri: Oops, just noticed that I should run default_prepare here so that the patch is applied and this doesn't need coreutils. Will fix in a bit.

ddevault:
LGTM but yeah you can skip coreutils

Cogitri:
Pushed fixed version

PR #7381, opened 2019-04-27
main/icu: upgrade to 64.2

J0WI:
A lot of packages needs to be rebuilt against this.

### DIFF
--- a/main/icu/APKBUILD
+++ b/main/icu/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Sergey Lukin <sergej.lukin@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=icu
-pkgver=63.1
+pkgver=64.2
 
 # convert x.y.z to x_y_z
 _ver=${pkgver//./_}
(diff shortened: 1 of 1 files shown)


maxice8:
According to `apk info -r icu-libs`

icu-libs-63.1-r0 is required by:
firefox-62.0.3-r4
qt5-qtbase-5.12.1-r1
slop-7.4-r0
maim-5.5.3-r0

maxice8:
missing python on depends apparently

J0WI:
Fixed.

ncopa:
This breaks ABI so we need rebuild everything that depends on:
````
so:libicudata.so.63
so:libicui18n.so.63
so:libicuio.so.63
so:libicutest.so.63
so:libicutu.so.63
so:libicuuc.so.63
````

ncopa:
@J0WI we would like to get in touch with you. can you please send an email to @clandmeter or me?

J0WI:
> This breaks ABI so we need rebuild everything that depends on:

Will you bump them after merging this? It would conflict with other PRs when I touch all packages here.

ncopa:
> > This breaks ABI so we need rebuild everything that depends on:
> 
> Will you bump them after merging this? It would conflict with other PRs when I touch all packages here.

Yeah, i think i can bump them. probably easiest that way.

ncopa:
fixed with 334ebffde9dec34becdd628ad56007699e98ea81

PR #7383, opened 2019-04-27
testing/perl-locale-msgfmt: new aport

timlegge:


testing/perl-locale-msgfmt/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+# Maintainer: Timothy Legge <timlegge@gmail.com>
+pkgname=perl-locale-msgfmt
+_pkgreal=Locale-Msgfmt
+pkgver=0.15
+pkgrel=0
+pkgdesc="Compile .po files to .mo files"
+url="https://metacpan.org/release/Locale-Msgfmt/"
+arch="noarch"
+license="GPL-1.0-or-later Artistic-1.0-Perl"
+depends="$cpandepends"
```
maxice8: is effectively an empty variable since it depends on a variable that doesn't exist

testing/perl-locale-msgfmt/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+makedepends="perl-dev $cpanmakedepends"
+checkdepends="$cpancheckdepends"
```
maxice8: is effectively an empty variable since it depends on a variable that doesn't exist

testing/perl-locale-msgfmt/APKBUILD
```diff
@@ -0,0 +1,35 @@
...
+pkgname=perl-locale-msgfmt
+_pkgreal=Locale-Msgfmt
+pkgver=0.15
+pkgrel=0
+pkgdesc="Compile .po files to .mo files"
+url="https://metacpan.org/release/Locale-Msgfmt/"
+arch="noarch"
+license="GPL-1.0-or-later Artistic-1.0-Perl"
+depends="$cpandepends"
+makedepends="perl-dev $cpanmakedepends"
```
maxice8: $cpanmakedepends reference can be removed

testing/perl-locale-msgfmt/APKBUILD
timlegge: wondered about that

PR #7385, opened 2019-04-27
main/perl-uri: upgrade to 1.76

timlegge:


Ikke:
This package has quite a [list][] of depending packages.

[list]:https://pkgs.alpinelinux.org/package/edge/main/x86/perl-uri

timlegge:
> This package has quite a [list](https://pkgs.alpinelinux.org/package/edge/main/x86/perl-uri) of depending packages.

Sorry, I had not looked but I am not surprised I guess..  It is likely something that can wait.  I noticed it was behind.

Ikke:
It depends on the backwards compatibility of this version. If it's a backwards compatible change, it can probably be merged, but otherwise the depending package might need to be updated at the same time.

timlegge:
It looks like 1.76 is essentially 1.74.  The changes in 1.75 were reverted by version 1.76:

https://metacpan.org/diff/file?target=OALDERS/URI-1.76/&source=ETHER%2FURI-1.74

So, essentially it is a version bump.  I was looking to upgrade it because a Foswiki install that has it as a dependency was complaining about the lower version in alpine than in CPAN.  I have no objection to leaving it as is and closing the PR but getting it to the latest version doesn't look like it will hurt anything either.

I will be sure to look closer on upgrade PRs in the future.

Tim

Ikke:
I don't think it's an issue to update to this version, as long as it's clear what the impact is.

PR #7386, opened 2019-04-27
testing/wine-development: new aport

TBK:


Ikke:
Consider adding `provides="wine"` so that anything that depends on wine will be able to use this package as well.

TBK:
Travis CI failed because:

> The job exceeded the maximum time limit for jobs, and has been terminated.

PR #7389, opened 2019-04-27
main/perl-xml-xpath: upgrade to 1.44

timlegge:


Ikke:
This package has 2 dependencies:

* perl-graphviz
* perl-xml-tidy

Is this version backwards compattible?

timlegge:
> This package has 2 depending packages:
> 
> * perl-graphviz
> * perl-xml-tidy
> 
> Is this version backwards compattible?

Hi

For perl-xml-tidy I doubt anyone other than me is currently using it as I added it just recently.  

That being said, I tested it against Net::SAML2 which is my use case and it appeared to be fine.  In addition, I rebuilt both perl-graphviz and perl-xml-tidy with the new version installed.  

As I checked out the issues on the https://rt.cpan.org/Public/Dist/Display.html?Name=XML-XPath some one has logged an issue but when I tested it with 1.42 and with 1.44 I got the same results.

As for graphviz I don't have any other way of testing it.

Tim

Ikke:
@timlegge Can you rebase this? This package has been moved to main as it's a dependency for libvirt.

timlegge:
> @timlegge Can you rebase this? This package has been moved to main as it's a dependency for libvirt.

@ncopa Done

PR #7395, opened 2019-04-28
community/libsoup: upgrade to 2.66.1

maxice8:
- Split libsoup-gnome and libsoup-gnome-dev
- Switch build system to meson
- Move sqlite-dev to makedepends as it is automatically found via pc:
- Don't build tests since it saves times and it is already !check
- Remove obsolete dependency on libgcrypt and gnutls


maxice8:
@ncopa

PR #7397, opened 2019-04-28
main/libdc1394: upgrade to 2.2.6

maxice8:
- Remove obsolete patch
- Fix license
- Modernize


ncopa:
````
589  -usr/lib/libdc1394.so.22
590	-usr/lib/libdc1394.so.22.2.1
591	+usr/lib/libdc1394.so.25
592	+usr/lib/libdc1394.so.25.0.0
````

I looks like this breaks ABI.

````
$ apk info --provides libdc1394
libdc1394-2.2.5-r1 provides:
so:libdc1394.so.22=22.2.1
````
Yup, we need to rebuild everything that depends on `so:libdc1394.so.22`:

````
$ apk search --origin --quiet --exact --rdepends so:libdc1394.so.22
libdc1394
gst-plugins-bad
opencv
vlc
````

PR #7401, opened 2019-04-28
community/mpdscribble: modernize

maxice8:
- Split $pkgname-openrc
- fix url
- fix license
- fix building
- adapt to modern style

### DIFF
--- a/community/mpdscribble/APKBUILD
+++ b/community/mpdscribble/APKBUILD
@@ -1,34 +1,23 @@
-# Contributor: 
+# Contributor: Leo <thinkabit.ukim@gmail.com>
 # Maintainer: 
 pkgname=mpdscribble
 pkgver=0.22
-pkgrel=1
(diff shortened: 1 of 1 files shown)


community/mpdscribble/APKBUILD
```diff
@@ -1,55 +1,37 @@
...
 prepare() {
-	local i
-	cd "$_builddir"
-	update_config_sub || return 1
-	for i in $source; do
-		case $i in
-		*.patch) msg $i; patch -p1 -i "$srcdir"/$i || return 1;;
-		esac
-	done
+	autoreconf -fi
```
prspkt:
```suggestion
	default_prepare
	autoreconf -fi
```
Even though there are no patches to be applied, we always call `default_prepare` when overriding the default function.

maxice8: done

PR #7404, opened 2019-04-28
testing/py-protobuf: upgrade to 3.7.1

andypost:
Ref https://pypi.org/project/protobuf/3.7.1/#history


maxice8:
Shouldn't this be updated together with protobuf ?

andypost:
at least php and py have no dependency

PR #7408, opened 2019-04-28
community/emacs: modernize

maxice8:
- Fix license
- Remove deprecated gconf dependency

### DIFF
--- a/community/emacs/APKBUILD
+++ b/community/emacs/APKBUILD
@@ -10,7 +10,7 @@ depends="emacs-nox"
 url="https://www.gnu.org/software/emacs/emacs.html"
 license="GPL-3.0"
 makedepends="autoconf automake linux-headers paxmark gawk
-	librsvg-dev giflib-dev libxpm-dev gtk+2.0-dev gconf-dev alsa-lib-dev
+	librsvg-dev giflib-dev libxpm-dev gtk+2.0-dev alsa-lib-dev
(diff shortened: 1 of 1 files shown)


Ikke:
Build fails due to segfault:


> Dumping under the name emacs
> **************************************************
> Warning: Your system has a gap between BSS and the
> heap (33088543 bytes). This usually means that exec-shield
> or something similar is in effect. The dump may
> fail because of this. See the section about
> exec-shield in etc/PROBLEMS for more information.
> **************************************************
> 6796416 of 16777216 static heap bytes used
> make[1]: *** [Makefile:740: bootstrap-emacs] Segmentation fault (core dumped)

community/emacs/APKBUILD
```diff
@@ -10,7 +10,7 @@ depends="emacs-nox"
 url="https://www.gnu.org/software/emacs/emacs.html"
 license="GPL-3.0"
```
TBK: SPDX identifier needed.

TBK:
Is the issue related to https://debbugs.gnu.org/cgi/bugreport.cgi?bug=23529

maxice8:
apparently yes.

maxice8:
No, works here, must be some configuration on CI-side ?

Ikke:
This built fine locally. After googling, it might have to do with either docker, or some kernel setting `/proc/sys/kernel/randomize_va_space`.

PR #7416, opened 2019-04-28
community/libgnome-keyring: modernize

maxice8:
- Fix license
- Remove unnecessary dependencies like gconf-dev and anything in
  depends_dev
- Use modern style


Cogitri:
Could I take on maintainership of this?

PR #7420, opened 2019-04-28
gnome: remove deprecated/obsolete technology

maxice8:
See body of each commit for what to look for.


main/gconf/APKBUILD
```diff
@@ -22,7 +20,10 @@ build() {
 		--build=$CBUILD \
 		--host=$CHOST \
 		--prefix=/usr \
-		--sysconfdir=/etc
+		--sysconfdir=/etc \
+		--disable-static \
```
tcely: Splitting out `-static` would be better.

maxice8: I don't think there is much value in splitting out static libraries of gnome stuff

ncopa:
Seems like py-libgnome still depends on libgnome?

ncopa:
what about `community/libgnome-keyring` which has `makedepends=libgnome-dev`?

maxice8:
> what about `community/libgnome-keyring` which has `makedepends=libgnome-dev`?

https://github.com/alpinelinux/aports/pull/7416

maxice8:
> Seems like py-libgnome still depends on libgnome?

https://github.com/alpinelinux/aports/pull/7418

PR #7421, opened 2019-04-28
community/libkeyfinder: modernize

maxice8:
- Switch to qt5, qt4 is EOL.
- Fix license


Ikke:
@ganwell Please review

PR #7433, opened 2019-04-29
main/perl-inline: fix

maxice8:
- Fix wrong license
- use $() instead of backticks ``

@rnalrd

### DIFF
--- a/main/perl-inline/APKBUILD
+++ b/main/perl-inline/APKBUILD
@@ -3,11 +3,11 @@
 pkgname=perl-inline
 _pkgname=Inline
 pkgver=0.83
-pkgrel=0
+pkgrel=1
 pkgdesc="Write Perl subroutines in other programming languages"
 url="http://search.cpan.org/~ingy/Inline/"
 arch="noarch"
-license="Artistic-1.0-Perl"
+license="Artistic-1.0-Perl GPL-1.0-or-later"
 depends="perl"
 makedepends="perl-dev"
 checkdepends="perl-test-warn"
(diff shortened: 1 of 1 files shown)


maxice8:
Most packages use both

rnalrd:
I know, I'm asking opinions on what is correct: i think we should pick one. Other non-perl packages list multiple licenses because different pieces of code use different licenses. This is not the case.

maxice8:
Well we can extend our usage of SPDX and make use of the `OR` operator

GPL-1.0-or-later OR Artistic-1.0-Perl

> Other non-perl packages list multiple licenses because different pieces of code use different licenses.

Not always true, see dual-licensing. I am in favor of  using SPDX's `OR` to show it. If they have different pieces of code they should use `AND`

main/perl-inline/APKBUILD
```diff
@@ -17,12 +17,12 @@ builddir="$srcdir"/$_pkgname-$pkgver
 
 prepare() {
 	default_prepare
-	export CFLAGS=`perl -MConfig -E 'say $Config{ccflags}'`
+	export CFLAGS="$(perl -MConfig -E 'say $Config{ccflags}')"
```
fabled: Currently `apkbuild-cpan` does not include quotation marks in the new templates. Perhaps use it to regenerate this? Or update also it to contain the quotes. In any case, expansion like this should not strictly need the quotes, and that's why I omitted them. Thoughts?

maxice8:
On Mon, May 27, 2019 at 11:59:37PM -0700, Timo Teräs wrote:
> fabled commented on this pull request.
> 
> 
> 
> > @@ -17,12 +17,12 @@ builddir="$srcdir"/$_pkgname-$pkgver
>  
>  prepare() {
>  	default_prepare
> -	export CFLAGS=`perl -MConfig -E 'say $Config{ccflags}'`
> +	export CFLAGS="$(perl -MConfig -E 'say $Config{ccflags}')"
> 
> Currently `apkbuild-cpan` does not include quotation marks in the new templates. Perhaps use it to regenerate this? Or update also it to contain the quotes. In any case, expansion like this should not strictly need the quotes, and that's why I omitted them. Thoughts?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7433#pullrequestreview-242477799

Fixed

PR #7434, opened 2019-04-29
community/dunst: update to 1.4.0

maxice8:
@ncopa


Ikke:
See also #7245

maxice8:
Test will fail as long as CI doesn't take changes in packages in the same PR into account

tcely:
> Test will fail as long as CI doesn't take changes in packages in the same PR into account

See https://github.com/alpinelinux/abuild/pull/53 for the reason why a package building in community is not using the packages previously built in the PR from main.

ncopa:
we need to verify that we don't introduce any circular deps here

PR #7436, opened 2019-04-29
testing/zola: upgrade to 0.7.0

maxice8:


TBK:
There is an issue with the bash-com... subpkg L36. The file is getting renamed to bash.fish instead of zola but who does not want to bash some fish :grin:

maxice8:
There are people that don't use fish for interactive usage ?  ¯\_(ツ)_/¯

PR #7444, opened 2019-04-30
testing/py3-websockets: modernize

maxice8:
- Switch to GitHub tarball which includes tests
- Enable check
- Fix license (BSD -> BSD-3-Clause)
- Remove superfluous dependency on py3-tox
- Use modern style


maxice8:
@fabaff

Ikke:
Test fails on aarch64:

> FAIL: test_keepalive_ping_stops_when_connection_closing (tests.test_protocol.ServerTests)
> \----------------------------------------------------------------------
> Traceback (most recent call last):
File "/home/buildozer/aports/testing/py3-websockets/src/websockets-7.0/tests/test_protocol.py", line 945, in test_keepalive_ping_stops_when_connection_closing
self.assertTrue(self.protocol.keepalive_ping_task.cancelled())
AssertionError: False is not true

maxice8:
ping

PR #7448, opened 2019-04-30
testing/py3-pystache: rename from py-pystache and modernize

maxice8:
- Enable tests
- Use modern style


Ikke:
Could you split up the move from the improvements in separate commits? That should lead to at least somewhat better diffs. Now it just looks like one package got removed and another added.

maxice8:
done

Ikke:
Thanks, that helped :-)

PR #7450, opened 2019-04-30
testing/crosstool-ng: upgrade to 1.24.0

OTLabs:
- http://crosstool-ng.github.io/2019/04/13/release-1.24.0.html 1.24.0
- Added new dependency to makedepends: gettext-dev
- Updated licenses info
- Added bash completion as separate package
- Source URL updated to official one

### DIFF
--- a/testing/crosstool-ng/APKBUILD
+++ b/testing/crosstool-ng/APKBUILD
@@ -1,15 +1,15 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=crosstool-ng
-pkgver=1.23.0
-pkgrel=1
+pkgver=1.24.0
+pkgrel=0
 pkgdesc="tool for building toolchains"
 url="http://crosstool-ng.org/"
-license="GPL"
+license="LGPL-2.1-or-later GPL-2.0-or-later CC-BY-SA-2.5"
 arch="x86 x86_64 ppc64le"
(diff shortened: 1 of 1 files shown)


testing/crosstool-ng/APKBUILD
```diff
@@ -1,13 +1,13 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=crosstool-ng
-pkgver=1.23.0
+pkgver=1.24.0
 pkgrel=1
 pkgdesc="tool for building toolchains"
 url="http://crosstool-ng.org/"
 license="GPL"
```
maxice8: GPL is not spdx and is ambiguous

OTLabs: Updated licenses info. Thank you for pointing out the need of this update!

testing/crosstool-ng/APKBUILD
```diff
@@ -25,6 +25,18 @@ build() {
 package() {
 	cd "$builddir"
 	make DESTDIR="$pkgdir" install
+
+	# bash completion has it's own package
+	rm -r "$pkgdir"/usr/share/bash-completion/
```
Ikke: It's perhaps a matter of preference, but isn't it easier to just move the files from `$pkgdir` to `$subpkgdir` rather than removing and 'reinstalling'?

OTLabs: I have implemented the move as it was suggested. Thank you!

PR #7451, opened 2019-04-30
community/opentracker: fix build break by adding include prefix

mksully22:
build fails with errors of this type: 
trackerlogic.c:16:10: fatal error: byte.h: No such file or directory
 #include "byte.h"

header files are part of libowfat-dev package and have been installed at /usr/include/libowfat
Adding PREFIX=/usr/include in APKBUILD before "make" invocation will allow path search for header files to work correctly.

### DIFF
--- a/community/opentracker/APKBUILD
+++ b/community/opentracker/APKBUILD
@@ -1,40 +1,31 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=opentracker
 pkgver=0_cvs20100625
-pkgrel=4
+pkgrel=5
 pkgdesc="an open and free bittorrent tracker"
 url="http://erdgeist.org/arts/software/opentracker/"
 arch="all"
 license="GPL"
-depends=
 makedepends="libowfat-dev zlib-dev"
 install="$pkgname.pre-install"
-subpackages=
 source="https://dev.alpinelinux.org/opentracker/opentracker-$pkgver.tar.gz
 	opentracker.initd"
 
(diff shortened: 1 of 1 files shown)


community/opentracker/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=opentracker
 pkgver=0_cvs20100625
-pkgrel=4
+pkgrel=5
 pkgdesc="an open and free bittorrent tracker"
 url="http://erdgeist.org/arts/software/opentracker/"
 arch="all"
```
maxice8: license is not an SPDX identifier and ambiguous

maxice8: empty variables can be removed for brevity

community/opentracker/APKBUILD
```diff
@@ -20,7 +20,7 @@ prepare() {
```
maxice8: _builddir can be renamed to builddir

community/opentracker/APKBUILD
```diff
@@ -20,7 +20,7 @@ prepare() {
 
 build() {
```
maxice8: the prepare function looks useless

community/opentracker/APKBUILD
```diff
@@ -20,7 +20,7 @@ prepare() {
...
 	cd "$_builddir"
-	make || return 1
+	PREFIX=/usr/include make || return 1
```
maxice8: no need for || return 1 anymore

PR #7456, opened 2019-04-30
community/openblas: upgrade to 0.3.6

sashkab:
This should resolve https://bugs.alpinelinux.org/issues/10379


sashkab:
Failure in travis is due to "The job exceeded the maximum log length, and has been terminated.", not sure how to handle this.

Ikke:
There's not a lot you can do about it. It's passing in drone.io, so I should not worry too much about it.

PR #7469, opened 2019-05-01
community/alsa-plugins: modernize

maxice8:
- Use modern style
- Fix url=
- Fix pulseaudio


maxice8:
Failure due to travis not liking ftp://

PR #7470, opened 2019-05-01
testing/otpclient: new aport

maxice8:
Highly secure and easy to use GTK+ OTP client (TOTP and HOTP)
https://github.com/paolostivanin/OTPClient


Ikke:
I get the same error when I build libcotp locally.

maxice8:
aarch64 fails because zbar is missing from aarch64

PR #7474, opened 2019-05-01
scripts/mkimg.*.sh: add haveged & rng-tools

HRio:
We need a way to generate entropy already at installation.

If not, things like https downloads or starting sshd for a remote
shell to the installer can block for a few minutes.

### DIFF
--- a/scripts/mkimg.base.sh
+++ b/scripts/mkimg.base.sh
@@ -297,7 +297,7 @@ profile_base() {
 	case "$ARCH" in
 	x86*) grub_mod="$grub_mod multiboot2 efi_uga";;
 	esac
-	apks="alpine-base alpine-mirrors busybox kbd-bkeymaps chrony e2fsprogs network-extras openssl openssh tzdata"
+	apks="alpine-base alpine-mirrors busybox kbd-bkeymaps chrony e2fsprogs haveged network-extras openssl openssh tzdata"
 	apkovl=
 	hostname="alpine"
 }
--- a/scripts/mkimg.standard.sh
+++ b/scripts/mkimg.standard.sh
@@ -44,7 +44,7 @@ profile_extended() {
 		iptables iputils irssi ldns-tools links
 		ncurses-terminfo net-snmp net-snmp-tools nrpe nsd
 		opennhrp openvpn openvswitch pingu ppp quagga
-		quagga-nhrp rpcbind sntpc socat ssmtp strongswan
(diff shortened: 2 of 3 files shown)


HRio:
@ncopa test built the xen iso after this change
```console
./apks/x86_64/rng-tools-6.7-r0.apk
./apks/x86_64/rng-tools-openrc-6.7-r0.apk
./apks/x86_64/haveged-1.9.4-r3.apk
./apks/x86_64/haveged-openrc-1.9.4-r3.apk
```

Ikke:
@HRio Could you add the motivation in the commit message for prosperity? Thanks

HRio:
> @HRio Could you add the motivation in the commit message for prosperity? Thanks

done.

ncopa:
do we need both haveged and rng-tools? Wouldnt one of them be enough?

ncopa:
I don't mind adding both to extended iso, but would prefer only add one for base.

HRio:
> I don't mind adding both to extended iso, but would prefer only add one for base.

On rpi3 and modern Intel/AMD CPUs most people would want rng-tools. haveged is for systems without a hwrng. So haveged works everywhere, but I guess people with the right hw, want "the real thing".

ncopa:
what we ship with iso is only what is needed to get the system installed. `rng-tools` seems to be smaller, but I guess we would need something that works everywhere? Also if you do have a hwrng, you could use the kernel's trust cpu feature? So it makes probably more sense to ship `haveged` only, and `rng-tools` with the extended iso.

HRio:
> Also if you do have a hwrng, you could use the kernel's trust cpu feature

the trust cpu option kernel is just for RDSEED/RDRAND CPU instruction, not for a "normal" hwrng device, like `/dev/hwrng` that it quite common on arm hardware. But rng-tools support both these types.

liqa:
Arguments are sorted in alphabetical order, would be nice to keep that way.

And do we care about the quality of the generated entropy in installation process? Because Peter Anvin warns against the HAVEGE algorithm @ https://lwn.net/Articles/525459/

HRio:
> Arguments are sorted in alphabetical order, would be nice to keep that way.
> 
> And do we care about the quality of the generated entropy in installation process? Because Peter Anvin warns against the HAVEGE algorithm @ https://lwn.net/Articles/525459/

Yes its not the best, I agree. This change just makes it available on the iso's for people that would like to use it.

For extended and xen we also make rng-tools available, so a hwrng can be used.

PR #7475, opened 2019-05-01
testing/browserpass: new aport

Cogitri:


testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgname=browserpass
+pkgver=3.0.6
+pkgrel=0
+pkgdesc="Browser extension for pass"
+url="https://github.com/browserpass/browserpass-native"
+arch="all"
+license="MIT"
+depends="pass"
+makedepends="go git"
+options="!check"
```
maxice8: Please write the reason out

testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/browserpass/browserpass-native/archive/$pkgver.tar.gz"
+builddir="$srcdir/github.com/browserpass/browserpass-native"
+
+prepare() {
+	mkdir -p "$srcdir"/github.com/browserpass
```
maxice8: you can use ${builddir%/*}

testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+	mkdir -p "${builddir%/*}"
+	mv "$srcdir/browserpass-native-$pkgver" "$builddir"
+	default_prepare
+	sed -i "s:%%replace%%:/usr/bin/browserpass-native:" \
+			browser-files/chromium-host.json browser-files/firefox-host.json
+}
+
+build() {
+	export GOPATH="$startdir"
+	go get -v -d
```
Ikke: I don't see any vendored libs or other kind of lock file. This means that dependencies can change any moment and the build is not reproducible, which is required for alpine packages.

maximbaz: Hi, I'm using "Go modules", the lock file is called `go.sum`, and Go team [claims them to be reproducible](https://github.com/golang/go/wiki/Modules#modules):

> Modules record precise dependency requirements and create reproducible builds.

testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+	go fix
+	go build
```
maximbaz: Just curious, why don't you use `make` to run all these commands, [like Arch Linux package does](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/browserpass)?

testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+}
+
+package() {
+	install -dm0755 "$pkgdir"/usr/bin \
+		"$pkgdir"/usr/lib/mozilla/native-messaging-hosts \
+		"$pkgdir"/etc/chromium/native-messaging-hosts \
+		"$pkgdir"/etc/chromium/policies/managed
```
maximbaz: Personally I would recommend you to follow the same approach [as I did for Arch Linux](https://www.archlinux.org/packages/?sort=&q=browserpass&maintainer=maximbaz&flagged=): don't bundle browser configs into native host, instead create separate packages for browser extensions, each of which is able to configure its browser for Browserpass.

- Right now you are configuring only two browsers, but more browsers exist and are supported
- Given that extension is handling sensitive data, dedicated packages for browser extensions will help users installing extension from PGP-signed sources, instead of Webstore.

But if you don't want this separation, then sure, keep these configs in, but at least also include Makefile, so people can [manually configure other browsers](https://github.com/browserpass/browserpass-native#configure-browsers).

testing/browserpass/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+	local targetname=com.github.browserpass.native.json
+
+	install -m644 browser-files/firefox-host.json "$pkgdir"/usr/lib/mozilla/native-messaging-hosts/$targetname
+
+	install -m644 browser-files/chromium-host.json "$pkgdir"/etc/chromium/native-messaging-hosts/$targetname
+	install -m644 browser-files/chromium-policy.json "$pkgdir"/etc/chromium/policies/managed/$targetname
+}
+
+sha512sums="48cc93843afed5dc0df4f72956df44b325d6c772ad232d8faa2315c4d8b716eda9b030601e011b280445e0b01528f413e5b19818ad89cd7bcb06f4a1f4f9d3f4  browserpass-3.0.6.tar.gz"
```
maximbaz: Thank you very much for packaging this for Alpine, really appreciate!

testing/browserpass/APKBUILD
Cogitri: Thanks for looking at this! Since we only support FF and Chromium right now (and will never really support Vivaldi/Brave/Chrome due to us using musl, so we can't package proprietary stuff) I think the current approach is fine though?

testing/browserpass/APKBUILD
Cogitri: Fixed now. Didn't look at it to be frank since the Makefile didn't work for me with browserpass 2.x on Exherbo - thanks for the heads up, it works now :)

testing/browserpass/APKBUILD
Cogitri: > Given that extension is handling sensitive data, dedicated packages for browser extensions will help users installing extension from PGP-signed sources, instead of Webstore.

Hm, but don't you release the browser extension and the native host thingie independently? Also, since you're already trusting your browser all your logins when entering them trusting the Webstore isn't unreasonable IMHO.

maximbaz: > (and will never really support Vivaldi/Brave/Chrome due to us using musl, so we can't package proprietary stuff) 

Is there some way for users to "unofficially" install Google Chrome or any other browser? If yes, at least include `Makefile` so they can also configure Browserpass for those browsers 🙂 

> Also, since you're already trusting your browser all your logins when entering them trusting the Webstore isn't unreasonable IMHO.

I was simply thinking that for attackers it is much simpler to break into my Google account and upload a malicious update for Browserpass (which will get installed in your browser automatically thanks to auto-updates) than to steal my PGP key (which is stored on Yubikey) and then additionally break into my Github account to make a release.

One other nice benefit for having extension installed from a package manager is that you can guarantee for your users compatibility between native host and browser extension; although I try to minimize breaking changes, sometimes they are unavoidable, and it will be nice if you can provide for your users simultaneous break-free upgrades.

One other alternative for you to consider is the approach that Debian took:  they distribute native host and browser extensions via a single `browserpass` package (so yeah their package basically follows two releases at the same time).

testing/browserpass/APKBUILD
Ikke: Thanks for the info. I'm trying to keep up with all the go package managers :-)

testing/browserpass/APKBUILD
Cogitri: > I was simply thinking that for attackers it is much simpler to break into my Google account and upload a malicious update for Browserpass (which will get installed in your browser automatically thanks to auto-updates) than to steal my PGP key (which is stored on Yubikey) and then additionally break into my Github account to make a release.

Well, I'd hope you use 2FA on your account too :P

> and it will be nice if you can provide for your users simultaneous break-free upgrades.

Alright, true that. Will add that to the APKBUILD, thanks for the tips :)

testing/browserpass/APKBUILD
Cogitri: Heh, same here :P

maximbaz: Hopefully this is the last package manager to learn, since it's built into Go itself 😄

testing/browserpass/APKBUILD
Ikke: @Cogitri Do you still need to push this?

Cogitri: Oh right! Thanks for noticing? :)

Cogitri: Pushed new version.

testing/browserpass/APKBUILD
maximbaz: > Alright, true that. Will add that to the APKBUILD, thanks for the tips :)

@Cogitri let me know if you still want to implement this, I'll be happy to help - here's how it looks on Arch Linux for [Chromium](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/browserpass-chromium) and [Firefox](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/browserpass-firefox)

Cogitri: Yup, will add seperate pkgs for that though - at least on Chromium it forces the addon to be on because it has been "installed by my organising"? Also, our FF isn't that stable right now so I can't test it right now so I held off of it for now :)

maximbaz: "Installed by an organization" is only when you use "policies", but if you will have a separate package much like `browserpass-chromium` in Arch, it will be like a normal extension. No worries - just ping me when you are done and I'll make sure to add your packages to README on browserpass side 😉

PR #7476, opened 2019-05-01
main/fish: don't install docker completion, -tools needs man-db

Cogitri:
The docker completion is installed by docker-fish-completion already, which
causes conflicts upon install. fish-tools needs man-db for generating completions.

@maxice8

### DIFF
--- a/main/fish/APKBUILD
+++ b/main/fish/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Leo <thinkabit.ukim@gmail.com>
 pkgname=fish
 pkgver=3.0.2
-pkgrel=1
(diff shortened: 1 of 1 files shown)


maxice8:
https://github.com/alpinelinux/aports/pull/7353

Cogitri:
Ah, thanks.

Cogitri:
Can we have "or" dependencies? It works with mdocml too, which is installed by default when once installs "docs"

Cogitri:
Sorry, should've marked this as WIP I guess. What about:

> Can we have "or" dependencies? It works with mdocml too, which is installed by default when once installs "docs"

@ncopa?

PR #7477, opened 2019-05-01
main/pinentry: add pinentry-gnome, use libsecret

Cogitri:
pinentry-gnome is useful when used in GNOME environments for better integration.
Use libsecret to save secrets in the keychain.


ncopa:
maybe we should disable gtk2 and only do gtk3

Cogitri:
I think the gtk2 one integrates better into GTK2 desktop than the gnome3 one.

PR #7313, opened 2019-04-25
[WIP] testing/falkon: new aport

maxice8:

PR #7478, opened 2019-05-01
community/qt5-qtwebengine: fix with musl-1.22

Cogitri:
musl-1.22 introduced SYS_membarrier, which breaks the sandbox of chromium
(and as such also qt5-qtwebengine)


maxice8:
Will help with #7313

Cogitri:
Hum, it still segfaults sometimes for me. APKs are on https://exp.exqa.de/alpine, would be nice if someone else tested it too.

PR #7485, opened 2019-05-02
testing/py3-wifi: upgrade to 0.3.8

maxice8:
- Add missing dependencies
- Fix license
- Rename binary to avoid conflicts
- Enable bash completion
- Use modern style


maxice8:
@fcolista for wireless-tools
@fabaff for py3-wifi

testing/py3-wifi/APKBUILD
```diff
@@ -0,0 +1,43 @@
+# Contributor: Leo <thinkabit.ukim@gmail.com>
+# Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
+# Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
+pkgname=py3-wifi
+_pkgname=wifi
+pkgver=0.3.8
+pkgrel=0
+pkgdesc="Command line tool and library wrappers for network interfaces"
+options="!check" # No testsuite
```
fabaff: Was included in 0.3.5 but not 0.3.8.

maxice8: There are also no github tarballs (which generally include tests/) for 0.3.8, it stops at 0.3.5

PR #7486, opened 2019-05-02
testing/py3-turq: remove

maxice8:
Unmaintained upstream, mitmproxy recommended instead.


maxice8:
@fabaff

maxice8:
0.3.1 requires unpackaged 'dominate' 'falcon' and 'h11'

PR #7487, opened 2019-05-02
community/ceph: upgrade to 14.2.1

iggy:
Bump version to latest and update checksum. No other functional changes
for now.

### DIFF
--- a/community/ceph/APKBUILD
+++ b/community/ceph/APKBUILD
@@ -1,14 +1,15 @@
 # Contributor: John Coyle <dx9err@gmail.com>
 # Maintainer: John Coyle <dx9err@gmail.com>
 pkgname=ceph
-pkgver=14.2.0
-pkgrel=3
+pkgver=14.2.1
+pkgrel=0
 pkgdesc="Ceph is a distributed object store and file system"
 pkgusers="ceph"
 pkggroups="ceph"
 url="http://ceph.com"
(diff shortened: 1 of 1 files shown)


community/ceph/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: John Coyle <dx9err@gmail.com>
 # Maintainer: John Coyle <dx9err@gmail.com>
 pkgname=ceph
-pkgver=14.2.0
-pkgrel=3
+pkgver=14.2.1
+pkgrel=0
 pkgdesc="Ceph is a distributed object store and file system"
 pkgusers="ceph"
 pkggroups="ceph"
```
maxice8: LGPL-2.1
GPL-2.0

are missing the `-or-later` or `-only` suffix for SPDX compliance

use `AND` instead of `and`

use `WITH` keyword with the name exception to denote an exception to the license

e.g: `GPL-2.0-or-later WITH Autoconf-exception-2.0`

iggy: Should I do this in this PR or a follow up PR?

ncopa: in this PR. squash it and git push --force

iggy: Done. Sorry it took so long.

iggy:
Travis times out because C++

PR #7501, opened 2019-05-02
testing/linux-tools: streamline bash-completion package

maxice8:
- Use modern style
- Create perf-bash-completion and replace perf-bash-completions
- Install completions to /usr/share/bash-completion/completions
- perf-bash-completion noarch
- Install completions when bash-completion is installed not when bash is


maxice8:
@ncopa

PR #7503, opened 2019-05-02
main/subunit: modernize

maxice8:
- Fix license
- Fix indent on line 26
- Add check support but keep it disabled due to dependencies being in
  testing/

### DIFF
--- a/main/subunit/APKBUILD
+++ b/main/subunit/APKBUILD
@@ -1,14 +1,16 @@
+# Contributor: Leo <thinkabit.ukim@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=subunit
 pkgver=1.2.0
 pkgrel=0
 pkgdesc="A streaming protocol for test results"
+options="!check" # Dependencies for checking stuff are in testing/
 url="https://launchpad.net/subunit"
 arch="all"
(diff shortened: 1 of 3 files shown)


maxice8:
@fcolista for py-testtools
@ncopa for subunit

main/subunit/APKBUILD
```diff
@@ -1,14 +1,16 @@
...
 pkgdesc="A streaming protocol for test results"
+options="!check" # Dependencies for checking stuff are in testing/
 url="https://launchpad.net/subunit"
 arch="all"
-license="Apache BSD"
+license="Apache-2.0 BSD-3-Clause"
 depends="python2"
 makedepends="check-dev cppunit-dev perl-dev python2-dev"
-options="!check"
+checkdepends="py-testtools py-fixtures py-hypothesis py-testscenarios bash"
```
ncopa: packages in main cannot depend on packages in testing. We need to either move py-testtools to main or move them all to community. Otherwise good.

maxice8: Not even when they are guarded under !check and only in checkdepends ?

ncopa: yes. unfortunately we don't build our world twice, one without tests and one with. We always run the tests on our builders.

PR #7504, opened 2019-05-02
main/bind: split tools

tcely:
- improve descriptions
- added -server-tools subpackage
- added -static subpackage
- provide py3-isc

Resolves alpinelinux/aports#7504

### DIFF
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -282,7 +282,9 @@ _append_deps() {
 		eval "makedepends=\"\$makedepends \${depends_${split}}\""
 	done
 }
-case "$(type 2>/dev/null subpkg_set)" in (subpkg_set\ is\ a\ *function*) _append_deps ;; esac
+is_function subpkg_set && _append_deps || \
+	makedepends="${makedepends} ${depends_dev} ${depends_py3_isc} ${depends_python_tools} ${depends_root_keys} ${depends_server_tools}"
+
 
 gpg_signature_extensions="sha512.asc"
 gpgfingerprints="good:AE3F AC79 6711 EC59 FC00  7AA4 74BB 6B9A 4CBB 3D38, BE0E 9748 B718 253A 28BB  89FF F1B1 1BF0 5CF0 2E57"
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -285,7 +285,7 @@ _append_deps() {
 case "$(type 2>/dev/null subpkg_set)" in (subpkg_set\ is\ a\ *function*) _append_deps ;; esac
 
 gpg_signature_extensions="sha512.asc"
-gpgfingerprints="good:AE3F AC79 6711 EC59 FC00  7AA4 74BB 6B9A 4CBB 3D38"
+gpgfingerprints="good:AE3F AC79 6711 EC59 FC00  7AA4 74BB 6B9A 4CBB 3D38, BE0E 9748 B718 253A 28BB  89FF F1B1 1BF0 5CF0 2E57"
 
 sha512sums="560944951b6c6a2801967813597440f7ffb3687adf455ea27751d5976954643b9ea86d0da4661caf4fa9673c1d5c1ce7f34cce94a8edfa95eb60dd1cf7839c56  bind-9.14.1.tar.gz
 2b32d1e7f62cd1e01bb4fdd92d15460bc14761b933d5acc463a91f5ecd4773d7477c757c5dd2738e8e433693592cf3f623ffc142241861c91848f01aa84640d6  bind.plugindir.patch
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -11,10 +11,10 @@ _isc_rel=0
 _ver=${pkgver%_p*}
 _p=${pkgver#*_p}
 _major=${pkgver%%.*}
-let "_isc_rel+=10*${_ver#${_ver%.*}.}"
+_isc_rel="$(( ${_isc_rel} + ( 10 * ${_ver#${_ver%.*}.} ) ))"
 if [ "$_p" != "$pkgver" ]; then
 	_ver="${_ver}-P$_p"
-	let "_isc_rel+=$_p"
+	_isc_rel="$(( ${_isc_rel} + ${_p} ))"
 fi
 pkgdesc="The ISC DNS server"
 url="https://www.isc.org"
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -4,12 +4,18 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: tcely <bind+aports@tcely.33mail.com>
 pkgname=bind
-pkgver=9.14.1
+pkgver="9.14.1"
+pkgrel=2
+_isc_ver="2.0"
+_isc_rel=0
 _ver=${pkgver%_p*}
 _p=${pkgver#*_p}
 _major=${pkgver%%.*}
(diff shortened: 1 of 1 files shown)


main/bind/APKBUILD
```diff
@@ -174,30 +182,58 @@ dev() {
...
+	depends="$depends_plugins"
+
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/bind "$subpkgdir"/usr/lib/
+}
+
+py3_isc() {
+	pkgdesc="Python functions to support BIND utilities"
+	depends="$depends_py3_isc"
+	provides="py3-isc=${_isc_ver}-r${_isc_rel} py3-$pkgname=$pkgver-r$pkgrel"
```
ncopa: If I understand this correct: `py3-isc-9.14.1-r2` provides `py3-isc-2.0-r0`. In what situation will that be useful?

How is apk supposed to deal with 2 different versions of the same package without messing things up?

I can already now see problems with it. Lets say that we find a bug in `py3-isc-9.14.1-r2` so we add a  patch to the `bind` package and release `bind-9.14.1-r3`. A user upgrade one of his systems, but realizes that the patched `-r3` introduces a bad regression that will literally blow up his system. To save his other servers and prevent that someone by accident upgrade to a newer version of `py3-isc` he runs `apk add 'py3-isc<9.14.1-r3'` to make sure that he never get anything newer that `-r2`.

Now `apk` tries to resolve that, and find this `py3-isc-2.0-r1`, which is way lower than `9.14.1-r3` and installs the package providing that and *kaboom* he blows up his others systems too.

Or other potential situation: `apk add 'py3-isc>9.14'` Maybe apk will consider the provider `py3-isc-2.0` and mark that package as "does not satisfy this dependency", and thus eliminate the the `py3-isc-9.14.1-r2` from further consideration in the dependency graph.

There are just so many things that can go wrong here, and to be honest, I have no clue what apk will do in those situations, nor am I sure what I think it should do.

main/bind/APKBUILD
```diff
@@ -174,30 +182,58 @@ dev() {
...
+	replaces="py3-$pkgname"
+
 	mkdir -p "$subpkgdir"/usr/lib
 	mv "$pkgdir"/usr/lib/python3* "$subpkgdir"/usr/lib/
 }
 
-_dnssec_tools() {
-	pkgdesc="Utilities for DNSSEC keys and DNS zone files management"
-	depends="py3-$pkgname=$pkgver-r$pkgrel"
+pytnon_tools() {
```
ncopa: I wonder how the CI managed to return successful build. `pytnon_tools` is misspelled.

main/bind/APKBUILD
```diff
@@ -174,30 +182,58 @@ dev() {
...
+	provides="$pkgname-dnssec-tools=$pkgver-r$pkgrel"
+	replaces="$pkgname-dnssec-tools"
 
-plugins() {
-	pkgdesc="The ISC DNS server plugins"
-	depends="$depends_plugins"
+	mkdir -p "$subpkgdir"/usr/bin
+	for i in "$pkgdir"/usr/bin/*; do
+		case "${i##*/}" in
+			dig|host|mdig|nslookup) ;;
```
ncopa: `dig` `host` and `nslookup` are provided in a package named `bind-server-tools`? I have always seen those as clients or client tools.

main/bind/APKBUILD
```diff
@@ -240,9 +275,17 @@ libs() {
 	depends="$depends_libs"
 	default_libs
 }
+_append_deps() {
+	local s split
+	for s in $subpackages; do
+		split="$(subpkg_set "$s"; printf -- '%s' "$subpkgsplit")"
+		eval "makedepends=\"\$makedepends \${depends_${split}}\""
```
ncopa: is there some way we can avoid use `eval`? eval is evil.

main/bind/APKBUILD
```diff
@@ -240,9 +275,17 @@ libs() {
...
+	done
+}
+case "$(type 2>/dev/null subpkg_set)" in (subpkg_set\ is\ a\ *function*) _append_deps ;; esac
```
ncopa: I dont like use `$()` in global scope because it forks. It is not problem of a single APKBUILD but things would go horribly slow if every APKBUILD forked in global scope.

main/bind/APKBUILD
```diff
@@ -174,30 +182,58 @@ dev() {
...
+plugins() {
+	pkgdesc="The ISC DNS server plugins"
+	depends="$depends_plugins"
+
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/bind "$subpkgdir"/usr/lib/
+}
+
+py3_isc() {
+	pkgdesc="Python functions to support BIND utilities"
```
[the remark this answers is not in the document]
tcely: The python module isn't called `bind` it's `isc` but I didn't name the thing.

main/bind/APKBUILD
tcely: > `subpackages="$pkgname-doc $pkgname-openrc py3-${pkgname}-isc:py3_isc:noarch`

This is `py3-bind-isc-9.14.1-r2` providing `py3-isc-2.0-r10` so that you can maintain the py3-isc version and ensure that its release is incremented when the source version of bind that bundles it changes.

main/bind/APKBUILD
tcely: They are matched up with the same misspelling.

main/bind/APKBUILD
tcely: Those matches don't have a related command, so we skip over them.

We want `dig`, `host`, `mdig`, and `nslookup` to be in `bind-tools` as before.

main/bind/APKBUILD
tcely: I don't know of a safer way than to use `eval` here. If there is one, I am happy to learn about it.

main/bind/APKBUILD
tcely: This is just a work around for `abuild` not having this sort of thing built-in yet.

Everything is a built-in but I haven't come up with a way to avoid the fork entirely here.

main/bind/APKBUILD
```diff
@@ -4,26 +4,34 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: tcely <bind+aports@tcely.33mail.com>
 pkgname=bind
-pkgver=9.14.1
+pkgver="9.14.1"
+pkgrel=2
+_isc_ver="2.0"
+_isc_rel=0
```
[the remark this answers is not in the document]
tcely: I wanted to provide a way to use `py3-isc=2.0` to avoid upgrading that module if the next version of `bind` bundles `2.1`.

main/bind/APKBUILD
```diff
@@ -4,26 +4,34 @@
...
 _ver=${pkgver%_p*}
 _p=${pkgver#*_p}
 _major=${pkgver%%.*}
-[ "$_p" != "$pkgver" ] && _ver="${_ver}-P$_p"
-pkgrel=1
+let "_isc_rel+=10*${_ver#${_ver%.*}.}"
```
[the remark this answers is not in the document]
tcely: done

main/bind/APKBUILD
tcely: I've switched this to a missing function (when sourcing APKBUILD outside abuild) that should let us avoid the sub-shell here.

tcely:
Superseded by https://github.com/alpinelinux/aports/pull/8932

PR #7507, opened 2019-05-03
testing/openimageio: modernize

maxice8:
- Fix license
- Provide comment on why check is disabled
- Switch to py3-openimageio

### DIFF
--- a/testing/openimageio/APKBUILD
+++ b/testing/openimageio/APKBUILD
@@ -2,36 +2,33 @@
 # Maintainer: Mark Riedesel <mark@klowner.com>
 pkgname=openimageio
 pkgver=1.8.15
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


Ikke:
opencolorio is not built for armv7 so this cannot either.

testing/openimageio/APKBUILD
```diff
@@ -2,36 +2,33 @@
 # Maintainer: Mark Riedesel <mark@klowner.com>
 pkgname=openimageio
 pkgver=1.8.15
-pkgrel=2
+pkgrel=3
 pkgdesc="Image I/O library supporting a multitude of image formats"
+options="!check" # 100 out of 120 tests fail
 url="https://sites.google.com/site/openimageio/"
 arch="all"
```
Ikke:
```suggestion
arch="all !armv7"
```

maxice8: applied

PR #7508, opened 2019-05-03
testing/py2-traceback2: Switch to purely python2

maxice8:
- This is a backport of a python3 module which can be imported as
  traceback
- Enabled tests
- Add missing deps


maxice8:
@itoffshore

Ikke:
Is it still worth the effort considering python2 will be eol soon? Only `py-testtools` depends on it atm.

maxice8:
Do we plan on removing py-testtools soon ?

maxice8:
well i already did the transition process so there isn't much effort to be done

PR #7509, opened 2019-05-03
testing/py2-linecache2: modernize and fix build

maxice8:
- lighten dependency load
- Add missing dependencies
- Work around pbr git problems
- Explain why check is disabled

### DIFF
--- a/testing/py2-linecache2/APKBUILD
+++ b/testing/py2-linecache2/APKBUILD
@@ -3,30 +3,23 @@
 pkgname=py2-linecache2
 _pkgname=${pkgname#py2-*}
 pkgver=1.0.0
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


Ikke:
@itoffshore Can you review?

itoffshore:
@Ikke - yes `PR` is ok

Ikke:
PSF is not a valid SPDX identifier. I've checked the source, but could not find any indication of a license either.

I see testing-cabal/linecache2#10 is still open. The author specifies it's in the setup.py metadata. The only relevant license I could find in the SPDX list is [Python License 2.0][Python-2.0], in which case the license identifier is: 'Python-2.0'.

If that's correct, could that be updated in the same PR?

[Python-2.0]:https://spdx.org/licenses/Python-2.0.html

maxice8:
Thanks for noticing, somehow i messed up my abuild

PR #7510, opened 2019-05-03
testing/pass: modernize

maxice8:
- Add missing dependency on cmd:gpg
- Enable tests, add missing dep on git
- Use modern style


maxice8:
@jomat

PR #7511, opened 2019-05-03
community/py3-rtslib: modernize

maxice8:
- Fix license
- Expand on why check is disabled
- Add missing dependencies
- Use modern style


maxice8:
@ncopa

PR #7514, opened 2019-05-03
testing/py3-compdb: upgrade to 0.2.0, drop py2

maxice8:
- Switch to tagged releases
- Enable tests
- Drop py2- and py- versions
- Add replaces for py- version
- Add missing dependency on py3-setuptools
- Lighten load on makedepends


maxice8:
@orsonteodoro

testing/py3-compdb/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+# Maintainer: Orson Teodoro <orsonteodoro@hotmail.com>
+pkgname=py3-compdb
+_pkgname=${pkgname#py3-*}
+pkgver=0.2.0
+pkgrel=0
+pkgdesc="The compilation database Swiss army knife"
+url="https://github.com/Sarcasm/compdb"
+arch="noarch"
+license="MIT"
+depends="py3-setuptools"
```
andypost: I bet it should be `makedepends`

maxice8: no, it has `/usr/bin/compdb` that has

```
#!/usr/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'compdb==0.2.0','console_scripts','compdb'
__requires__ = 'compdb==0.2.0'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('compdb==0.2.0', 'console_scripts', 'compdb')()
    )
```

so `py3-setuptools` which provides `pkg-resources` is required.

PR #7515, opened 2019-05-03
testing/py-distro: modernize

maxice8:
- Fix license
- Disable check as some tests fail
- Add infrastructure for running tests


maxice8:
@fabaff

PR #7519, opened 2019-05-03
community/networkmanager: update to 1.18.1

Cogitri:


community/networkmanager/APKBUILD
```diff
@@ -1,15 +1,14 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
-# Maintainer: 
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=networkmanager
-pkgver=1.16.0
-pkgrel=0
-pkgdesc="network management tool"
-url="http://projects.gnome.org/NetworkManager/"
+pkgver=1.18.1
+pkgrel=1
```
maxice8: pkgrel should be 0

community/networkmanager/APKBUILD
```diff
@@ -31,9 +30,10 @@ makedepends="$depends_dev
 	util-linux-dev
 	jansson-dev
 	vala
-	autoconf
-	automake
-	libtool
+	meson
+	py-gobject3
+	libpsl-dev
+	python
```
andypost: Better to specify exact py3 version here

Cogitri: Uh? It uses py2.

Ikke: Then specify `python2`

Cogitri: Ah, thanks, didn't see that!

PR #7521, opened 2019-05-03
testing/bcc: new aport

acj:


Ikke:
This fails on `x86`, `armhf` and `armv7` with the error:

> /home/buildozer/aports/testing/bcc/src/bcc-0.8.0/tests/cc/test_usdt_args.cc:65:14: error: 'parser' was not declared in this scope
> REQUIRE(!parser.parse(&arg));
> ^~~~~~

acj:
@Ikke Thanks. Fixed.

testing/bcc/APKBUILD
```diff
@@ -0,0 +1,28 @@
+# Maintainer: Adam Jensen <acjensen@gmail.com>
+pkgname=bcc
+pkgver=0.8.0
+pkgrel=1
```
Ikke: I'll change it when applying, but `pkgrel` should start at 0

PR #7528, opened 2019-05-03
community/py-msgpack: use C extension and modernize

maxice8:
- Use C extension with Cython
- Fix check
- separate makedepends and checkdepends


maxice8:
@fabaff 

Required for https://bugs.alpinelinux.org/issues/10401

fabaff:
If we can improve the performance of `borgbackup` then I'm all for changing it.

PR #6071, opened 2019-01-18
community/rust: upgrade to 1.32.0

jeizsm:
@jirutka review pls

PR #7538, opened 2019-05-04
community/rust: update to 1.32.0

Cogitri:


Cogitri:
This is just so we can bootstrap 1.34 afterwards and isn't actually meant to compile other packages with it.

Ikke:
CI Failed because it took >1h

J0WI:
I was able to compile `1.32.0` and `1.33.0` with these patches, but not `1.34.0`

Cogitri:
What do you mean?

andypost:
@Cogitri I bet @J0WI mean that 1.34.1 released

Cogitri:
Ah. We can't use that immediately though, we have to cycle 1.32->1.34.1 (hopefully, maybe we have to hop onto 1.33 before) due to Rust not really minding backwards compatibility with old compilers for rustc.

Ikke:
This is a duplicate of #6071

PR #7379, opened 2019-04-27
testing/gnome-shell: upgrade to 3.32.1

Cogitri:
~~Depends on #7241~~
~~Depends on #7258~~
~~Depends on #7261~~
~~Depends on #7262~~
~~Depends on #7374~~
~~Depends on #7375~~
~~Depends on #7376~~
~~Depends on #7377~~
~~Depends on #7417~~

PR #7547, opened 2019-05-04
testing/gnome-tweaks: new aport

Cogitri:
Needs #7379


Cogitri:
Should be good to go now

PR #7548, opened 2019-05-04
testing/gucharmap: new aport

Cogitri:


Ikke:
Ci doesn't work because the package in testing does not find the just built package in community.

Cogitri:
Yup.

PR #7549, opened 2019-05-04
testing/bolt: new aport

Cogitri:


Ikke:
It fails on travis with:

> ERROR:../tests/test-common.c:954:test_io_copy_bytes: assertion failed (error == NULL): error while copying data: Function not implemented (g-io-error-quark, 0)

Not sure if this is a issue related to travis or something else.

Cogitri:
Well, since it works on Drone I'm not quite sure what the difference between the runners is, do both run the same Alpine version (edge?)?

Ikke:
Just tested in a 32-bits docker container, and there it also builds fine.

PR #7552, opened 2019-05-04
testing/i3status: add maintainer to i3status and set !check

yjftsjthsd-g:
Add myself as maintainer on i3status, and explicitly specify that it has
no test suite (when testing to make sure I could build locally, it
complained, so I confirmed that upstream doesn't ship any tests in the
current stable version).


yjftsjthsd-g:
That's reasonable, given exactly what pkgrel means.

PR #7553, opened 2019-05-04
community/gnome-bluetooth: upgrade to 3.32.1

Cogitri:
Also fix the homepage&license, streamline the usage of meson/ninja
and adopt the package


Ikke:
Build [fails on armv7][armv7] because polkit is missing.

[armv7]:https://build.alpinelinux.org/buildlogs/build-edge-armv7/community/gnome-bluetooth/gnome-bluetooth-3.32.1-r0.log

PR #7560, opened 2019-05-05
community/flatpak-builder: build docs, enable YAML support

Cogitri:
Most of the new Flatpak templates are written in YAML now, so it makes
sense to enable support for it.

### DIFF
--- a/community/flatpak-builder/APKBUILD
+++ b/community/flatpak-builder/APKBUILD
@@ -2,21 +2,18 @@
 # Maintainer: André Klitzing <aklitzing@gmail.com>
 pkgname=flatpak-builder
 pkgver=1.0.6
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


community/flatpak-builder/APKBUILD
```diff
@@ -26,22 +23,18 @@ build() {
...
-		--disable-documentation \
 		--with-dwarf-header=/usr/include/libdwarf
 
 	make
 }
 
 check() {
-	cd "$builddir"
 	#make -k check
 	./flatpak-builder --version
```
maxice8: @clandmeter asked me to not do such simplistic checks to not give a false sense of proper testing being done.

Cogitri: Sure, but wasn't sure if I should change existing behavior here.

Ikke:
armv7 fails because polkit is missing.

PR #7567, opened 2019-05-05
testing/minio: upgrade to 0.20190502

5paceToast:


maxice8:
`cd "$builddir"`

Isn't necessary on build() and check() phases, abuild does that automatically now.

andypost:
IIRC `builddir` required only for main

PR #7569, opened 2019-05-05
testing/py3-sstash: modernize

maxice8:
- Use modern style
- Explain why check is disabled

### DIFF
--- a/testing/py3-sstash/APKBUILD
+++ b/testing/py3-sstash/APKBUILD
@@ -9,20 +9,16 @@ url="https://github.com/realcr/sstash"
 arch="noarch"
 license="MIT"
 depends="python3 py3-pynacl py3-click"
-makedepends="python3-dev py3-pytest"
-install=""
-options="!check"
-subpackages=""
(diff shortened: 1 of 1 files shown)


maxice8:
@fabaff

Ikke:
@maxice8 This needs a `pkgrel` dump (due to makedepends change).

maxice8:
But makedepends don't affect the final product, do we need to pkgrel if it is going to build the same package again ?

Ikke:
I suppose it wouldn't in this case. There are cases where `makedepends` determine whether features are enabled or not.

maxice8:
On Sun, May 05, 2019 at 09:29:01PM -0700, Kevin Daudt wrote:
> I suppose it wouldn't in this case. There are cases where `makedepends` determine whether features are enabled or not.
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/alpinelinux/aports/pull/7569#issuecomment-489497504

This isn't this case, also changing makedepends to enable/disable a feature warrants a message.

PR #7570, opened 2019-05-05
community/mutagen: modernize

maxice8:
- Enable check
- Split dependencies and lighten them

### DIFF
--- a/community/mutagen/APKBUILD
+++ b/community/mutagen/APKBUILD
@@ -7,29 +7,26 @@ url="https://github.com/quodlibet/mutagen"
 arch="noarch"
 license="GPL-2.0-or-later"
 depends="python3 py-mutagen"
-makedepends="python3-dev"
-checkdepends="py3-pytest py3-setuptools py3-hypothesis"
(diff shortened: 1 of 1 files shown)


community/mutagen/APKBUILD
```diff
@@ -7,31 +7,28 @@ url="https://github.com/quodlibet/mutagen"
...
-options="!check"
-
 builddir="$srcdir/$pkgname-release-$pkgver"
 
 build() {
-	cd "$builddir"
-	python3 setup.py build
-}
+	# Disable failing pep8 test
+	rm -f tests/quality/test_pep8.py
```
Ikke: Is this reported upstream?

maxice8: no

Ikke: Upstream [fixed][] this already in master.

[fixed]:https://github.com/quodlibet/mutagen/commit/0ee86ef9d7e06639a388d0638732810b79998608#diff-543448b48321b1e3f4c34888542f2cf7

PR #7375, opened 2019-04-27
folks: new aport

Cogitri:

PR #7573, opened 2019-05-05
testing/vinagre: new aport

Cogitri:


Cogitri:
Needs #7375

Ikke:
#7375 has been merged.

Cogitri:
As CI showed this will fail on armv* without #7581 I think, should've marked that, sorry!

PR #7581, opened 2019-05-05
main/polkit: enable on armhf&armv7

Cogitri:


Cogitri:
I guess the builder need more time to build mozjs60 for armhf

Cogitri:
Seems to work now, @ncopa.

PR #7586, opened 2019-05-06
community/py-xlsxwriter: upgrade to 1.1.8

maxice8:
- Fix license
- Switch to GitHub tarball so we actually get a testsuite
- Lighten load on makedepends
- Use modern style


maxice8:
@fcolista

PR #7587, opened 2019-05-06
[WiP] testing/spacy: new aport

OTLabs:
https://spacy.io/
Industrial-strength Natural Language Processing (NLP)

Closes https://bugs.alpinelinux.org/issues/8020

### DIFF
--- a/testing/py3-blis/APKBUILD
+++ b/testing/py3-blis/APKBUILD
@@ -1,0 +1,43 @@
+# Contributor: Oleg Titov <oleg.titov@gmail.com>
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+pkgname=py3-blis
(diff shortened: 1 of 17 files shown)


Ikke:
It looks like this only works on x86_64. It uses some inline ASM which is not adopted for different architectures.

OTLabs:
@Ikke I have limited the arch to x86_64 only. Thank you for pointing this out!

Ikke:
I just noticed that you are using a virtualenv to build it in the build phase. Apart from whether this is acceptable or not:

1. It looks like the virtualenv is never packaged, so I wonder if it would work on the target host
2. It seems to build the `spacy._align` extension twice, once during build, and once during package.

It might be better to separately package the dependencies.

PR #7588, opened 2019-05-06
main/libpng: upgrade to 1.6.37

maxice8:
- Add secfixes
- Remove pkg-config detected depends_dev
- Split $pkgname-static


maxice8:
@ncopa

PR #7597, opened 2019-05-06
testing/gnome-online-accounts: upgrade to 3.32.0

Cogitri:


testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
 pkgname=gnome-online-accounts
-pkgver=3.26.0
-pkgrel=1
-pkgdesc="client for GNOME web services"
-url="https://www.gnome.org/"
+pkgver=3.32.0
+pkgrel=0
+pkgdesc="Single sign-on framework for GNOME"
+url="https://wiki.gnome.org/Projects/GnomeOnlineAccounts"
 # limited by webkit2gtk
```
maxwell-k: It looks like [webkit2gtk](https://github.com/alpinelinux/aports/blob/master/community/webkit2gtk/APKBUILD#L9) is now supported on all architectures.

```suggestion
```

testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
 arch="x86 x86_64 ppc64le aarch64"
```
maxwell-k: Can this package be built on all architectures?

```suggestion
arch="all"
```

testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
 		--localstatedir=/var \
-		--disable-telepathy
+		--enable-media-server \
+		--enable-lastfm
 	make
 }
 
 check() {
```
maxwell-k: Same as above.

testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
+	webkit2gtk-dev json-glib-dev rest-dev libsecret-dev
 	docbook-xsl docbook-xml"
-subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-source="https://download.gnome.org/sources/$pkgname/${pkgver%.*}/$pkgname-$pkgver.tar.xz"
-builddir="$srcdir/$pkgname-$pkgver"
+subpackages="$pkgname-dev $pkgname-lang"
+source="https://download.gnome.org/sources/gnome-online-accounts/${pkgver%.*}/gnome-online-accounts-$pkgver.tar.xz"
 
 build() {
-	cd "$builddir"
```
maxwell-k: I think at present it is better to leave any `cd "$builddir"` lines, these make back ports and security fixes easier. Perhaps not as relevant in `testing/` but this is my understanding of the current approach.

testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
+		--enable-lastfm
 	make
 }
 
 check() {
-	cd "$builddir"
 	make check
 }
 
 package() {
```
maxwell-k: Same as above.

testing/gnome-online-accounts/APKBUILD
```diff
@@ -1,40 +1,37 @@
...
-makedepends="gnome-desktop-dev libxslt itstool libxml2-utils 
-	webkit2gtk-dev json-glib-dev rest-dev libsecret-dev 
+license="GPL-2.0-or-later"
+makedepends="gnome-desktop-dev libxslt itstool libxml2-utils
+	webkit2gtk-dev json-glib-dev rest-dev libsecret-dev
 	docbook-xsl docbook-xml"
-subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-source="https://download.gnome.org/sources/$pkgname/${pkgver%.*}/$pkgname-$pkgver.tar.xz"
-builddir="$srcdir/$pkgname-$pkgver"
+subpackages="$pkgname-dev $pkgname-lang"
```
maxwell-k: What happened to the doc sub-package?

testing/gnome-online-accounts/APKBUILD
Cogitri: I've been told that these can be removed for all ports in testing& community because abuild has supported this behaviour for over 6 months now.

testing/gnome-online-accounts/APKBUILD
Cogitri: Will try, thanks for the heads-up

testing/gnome-online-accounts/APKBUILD
Cogitri: Seems like g-o-a doesn't build any docs without gtk-doc anymore.

Cogitri:
Applied feedback, thanks.

Cogitri:
Added libgdata&gvfs to this, it needs the new goa

Cogitri:
I suppose this should be good to go now.

community/libimobiledevice/APKBUILD
```diff
@@ -6,32 +6,29 @@ pkgrel=3
 pkgdesc="Library that talks the protocols to support iPhone and iPod Touch devices on Linux"
 url="http://libimobiledevice.org/"
 arch="all"
-license="GPL-2.0 LGPL-2.1"
-makedepends="gnutls-dev openssl-dev libgcrypt-dev libusbmuxd-dev 
-	libtasn1-dev libplist-dev libtool automake autoconf"
+license="GPL-2.0-or-later LGPL-2.1-or-later"
```
tcely: Using a boolean here would clarify the situation.

Cogitri: Thanks, fixed. Turns out it's just LGPL-2.1-or-later, sorry.

community/gvfs/APKBUILD
```diff
@@ -1,66 +1,57 @@
...
+pkgver=1.40.1
+pkgrel=0
 pkgdesc="Backends for the gio framework in GLib"
-url="http://ftp.gnome.org/pub/gnome/sources/gvfs/${pkgver%.*}/"
+url="https://wiki.gnome.org/Projects/gvfs"
 arch="all"
-license="GPL"
-depends=""
-options="!check"  # no tests available
-triggers="$pkgname.trigger=/usr/lib/gvfs"
```
tcely: Why is this trigger no longer useful?

Cogitri: From the commit msg:

* Remove trigger which is likely to kill gvfs in mid-operation.

Without the trigger newly added gvfs-* packages won't be registered before a restart, but IMHO that's way better than just killing gvfs, which could happen while its in use.

community/libgdata/APKBUILD
```diff
@@ -1,35 +1,34 @@
...
 arch="all"
-license="LGPL"
-makedepends="intltool uhttpmock-dev glib-dev 
-	libsoup-dev json-glib-dev liboauth-dev"
+license="LGPL-2.1-or-later"
+depends="gsettings-desktop-schemas"
+makedepends="intltool uhttpmock-dev glib-dev vala gobject-introspection-dev
+	libsoup-dev json-glib-dev liboauth-dev gnome-online-accounts-dev gcr-dev"
 checkdepends="gsettings-desktop-schemas"
+options="!check" #Needs to bind to ports for testing
```
tcely: We have a lot of tests that bind to ports. I don't understand what goes wrong based on that comment.

Cogitri: Hm, at least in my docker setup that caused it to fail, I assumed the same would happen in other environments. I'll dig out a log

community/libplist/APKBUILD
```diff
@@ -1,35 +1,34 @@
...
-		--mandir=/usr/share/man \
-		|| return 1
-	make || return 1
+		--mandir=/usr/share/man
+	make
+}
+
+check() {
+	# https://github.com/libimobiledevice/libplist/issues/118
+	if ! [ "$CARCH" = armv7 ]; then
```
tcely:
```suggestion
	if [ "$CARCH" != armv7 ]; then
```

Excluding architectures from running check can also be done with:
```
case "${CARCH}" in (armv7) return 0 ;; esac
```

community/libgdata/APKBUILD
Cogitri: Yup, seems like it's just on docker, thanks for noticing.

community/libplist/hack-autotools.patch
```diff
@@ -0,0 +1,25 @@
+For _some_ reason autotools doesn't want to run this check, even if run with --with-cython,
+so let's just remove it...
```
tcely: Did you try `--with-cython=cython2` before patching?

community/libplist/test.patch
```diff
@@ -0,0 +1,9 @@
+--- libplist-2.0.0/test/emptycmp.test.orig	2019-05-11 14:12:09.323356191 +0200
++++ libplist-2.0.0/test/emptycmp.test	2019-05-11 14:12:22.723356503 +0200
+@@ -4,4 +4,6 @@
+ DATAOUT=$top_builddir/test/data
+ TESTFILE=1.plist
+ 
++echo "$(pwd): $top_builddir/test/plist_cmp $DATASRC/$TESTFILE $DATAOUT/$TESTFILE.out"
++
+ $top_builddir/test/plist_cmp $DATASRC/$TESTFILE $DATAOUT/$TESTFILE.out
```
tcely: I don't understand what the purpose of this patch is.

community/libplist/hack-autotools.patch
Cogitri: Ah, already did that but forgot to remove the files. Thanks

community/libplist/test.patch
Cogitri: Sorry, needed the output in CI and removed the file from the APKBUILD but forgot delete the actual file.

community/gvfs/APKBUILD
tcely: Presumably `gvfs` has a signal handler for `SIGUSR1` that does something intelligent about loading the files without just falling over in the middle of doing something.

Cogitri: Alrighty then

Cogitri:
Hm, the same tests fail on the builder for libgdata, I guess I'll disable them for now.

tcely:
docker does not enable `ipv6` by default. Patching those tests to bind to `127.0.0.1` instead will likely fix the issue.

Cogitri:
Yup, meant that. Working on it rn.

Cogitri:
Please take a look at 9fdd23ec5426321674636012c30ec70bd60e715d which fixes it.

tcely:
> Please take a look at [9fdd23e](https://github.com/alpinelinux/aports/commit/9fdd23ec5426321674636012c30ec70bd60e715d) which fixes it.

That's not a solution I would have chosen, but if it only affects `localhost` then it is entirely reasonable.

Cogitri:
Hm, what would've been your way to tackle this? And yes, it only affects `localhost` and only during tests.

tcely:
I was expecting something along the lines of using a `-listenaddr` option.

Cogitri:
Ah, was searching for something like that too, but couldn't find it :c 

Since we don't support ipv6 on CI the uhttpmock change should fix tests for other apps too.

tcely:
> Since we don't support ipv6 on CI the uhttpmock change should fix tests for other apps too.

Yep. Your solution is entirely reasonable. I was just a bit concerned it would prevent all IPv6 binds, but that concern was addressed.

Cogitri:
Good, thanks for taking a look at this! :)

tcely:
A test is failing for `armhf`:

```
ERROR: perf - Bail out! libgdata:ERROR:perf.c:107:test_perf_parsing: assertion failed (per_iteration_time < 2000): (3851 < 2000)
--
10960 | ============================================================================
10961 | Testsuite summary for libgdata 0.17.9
10962 | ============================================================================
10963 | # TOTAL: 345
10964 | # PASS: 344
10965 | # SKIP: 0
10966 | # XFAIL: 0
10967 | # FAIL: 0
10968 | # XPASS: 0
10969 | # ERROR: 1
10970 | ============================================================================
10971 | See gdata/tests/test-suite.log
10972 | Please report to https://bugzilla.gnome.org/enter_bug.cgi?product=libgdata
10973 | ============================================================================
10974 | make[5]: *** [Makefile:1604: test-suite.log] Error 1
10975 | make[5]: Leaving directory '/home/buildozer/aports/community/libgdata/src/libgdata-0.17.9/gdata/tests'
10976 | make[4]: *** [Makefile:1712: check-TESTS] Error 2
10977 | make[4]: Leaving directory '/home/buildozer/aports/community/libgdata/src/libgdata-0.17.9/gdata/tests'
10978 | make[3]: *** [Makefile:1853: check-am] Error 2
10979 | make[3]: Leaving directory '/home/buildozer/aports/community/libgdata/src/libgdata-0.17.9/gdata/tests'
10980 | make[2]: *** [Makefile:1856: check] Error 2
10981 | make[2]: Leaving directory '/home/buildozer/aports/community/libgdata/src/libgdata-0.17.9/gdata/tests'
10982 | make[1]: *** [Makefile:3423: check-recursive] Error 1
10983 | make[1]: Leaving directory '/home/buildozer/aports/community/libgdata/src/libgdata-0.17.9'
10984 | make: *** [Makefile:3721: check] Error 2
10985 | >>> ERROR: libgdata: check failed
```

community/gvfs/APKBUILD
tcely: - [ ] add `triggers=` line again

community/gvfs/APKBUILD
```diff
@@ -1,66 +1,58 @@
...
+license="LGPL-2.0-or-later"
+makedepends="fuse-dev libgudev-dev expat-dev samba-dev libxml2-dev
 	libsoup-dev avahi-dev libarchive-dev udisks2-dev libgphoto2-dev
 	libcdio-paranoia-dev libgcrypt-dev libxslt-dev docbook-xsl
-	libmtp-dev gcr-dev libcap-dev libsecret-dev libnfs-dev meson ninja
-	openssh-client"
+	libmtp-dev gcr-dev libcap-dev libsecret-dev libnfs-dev meson
+	openssh-client libusb-dev gnome-online-accounts-dev libimobiledevice-dev
+	libbluray-dev libgdata-dev"
+triggers="$pkgname.trigger=/usr/libexec/gvfs"
```
Cogitri: The triggers line is here.

Cogitri:
I guess it times out on armhf, are the normal runners faster than the CI ones?

tcely:
> I guess it times out on armhf, are the normal runners faster than the CI ones?

The builder could be faster, but I don't know that it's timing out. Just that the actual value is higher than the asserted limit. You may want to gather the `gdata/tests/test-suite.log` file. Building the changes for this package in isolation (if possible) might be good too.

Cogitri:
`per_iteration_time` very much sounds like a benchmark that fails when it's too slow to me. I don't think splitting this up will change something about that. I don't think I can gather the log from CI, can I?

Anyway, I guess I'll disable the test for armhf since it's so slow.

tcely:
> `per_iteration_time` very much sounds like a benchmark that fails when it's too slow to me. I don't think splitting this up will change something about that. I don't think I can gather the log from CI, can I?
> 

Here you go: https://dpaste.de/PL56

> Anyway, I guess I'll disable the test for armhf since it's so slow.

It's a SIGABRT and a core dump according to the output I gathered. So, not just slowness leading to a time out after all.

Cogitri:
The reason for it coredumping is that the following assertion fails:
(3826 < 2000), because each iteration takes 0.38 seconds, but the test only allows up to 0.2 seconds per iteration. Because 0.38 is above 0.2 the assertion fails, calls SIGABRT and and as such the test coredumps.

tcely:
In the worst case I'd say disabling check when you are running on `armhf` under DroneCI should be used. Here is `libuv` doing something similar.

https://github.com/alpinelinux/aports/blob/49f7b7bbff3084ac325fe4613c14a991f94293d0/main/libuv/APKBUILD#L35-L47

Cogitri:
Thanks, will do that :)

Cogitri:
Nice, it passes now.

community/libgdata/APKBUILD
```diff
@@ -1,36 +1,53 @@
...
+		alpinelinux-linux/arm-armhf)
+			sed -i -e '/perf/d' gdata/tests/Makefile.am
+		;;
+		alpinelinux-*-*)
+			echo "DroneCI: ${DRONE_REPO_OWNER}-${DRONE_RUNNER_PLATFORM}-${CARCH}"
+		;;
+	esac
+
+	intltoolize --copy --force
+	autoreconf -if
```
tcely:
```suggestion
	autoreconf --install --force
```

Cogitri: ...but why? Both are perfectly valid.

tcely: The long options are more explicit. It saves reviewers having to reference the man page.

Cogitri: Or a `autoreconf -h`. Anyway, I get your point, I'll do that for future PRs. (Am not at PC right now)

community/libplist/APKBUILD
```diff
@@ -1,35 +1,34 @@
...
-	make || return 1
+		--mandir=/usr/share/man
+	make
+}
+
+check() {
+	# https://github.com/libimobiledevice/libplist/issues/118
+	if [ "$CARCH" != armv7 ]; then
+		make check -j1 # https://github.com/libimobiledevice/libimobiledevice/issues/777
+	fi
```
tcely: This could switch to using `case` too.

Cogitri:
Is there a rationale behind using case switches for a single match instead of if?

community/libplist/APKBUILD
tcely: The rational is to make modifications easier.

```
case "${CARCH}" in (armv7) return 0 ;; esac # skip matches
```

Adding another architecture to skip is easier and doesn't require thinking through the boolean logic.

Cogitri:
Pushed again.

PR #7598, opened 2019-05-06
community/ostree: enable check

maxice8:


maxice8:
@misery

misery:
Thanks. But checks are still failing. Did you check why?

maxice8:
the tests are so flaky

maxice8:
Tests disabled because they are super flaky

PR #7600, opened 2019-05-06
community/webkit2gtk: security upgrade to 2.24.2

TBK:


Cogitri:
What's the status on this? Can I help?

TBK:
> What's the status on this? Can I help?

All help is appreciated.

I got lost in tracking down the causes.

**x86** fails because of this:
```
CMake Error at CMakeLists.txt:115 (message):
SSE2 support is required to compile WebKit 
```
Should be fixed with https://bugs.webkit.org/show_bug.cgi?id=196488

**armv7** fails because of:
```
make[2]:  ***  [Source/JavaScriptCore/CMakeFiles/JavaScriptCore.dir/build.make:124:  DerivedSources/JavaScriptCore/LLIntAssembly.h] Error 1
```

armhf & aarch64 hit the timecap.

I just discovered that 2.24.2 has been released so will bump and see if that fixes x86 and armv7.

Cogitri:
For x86 we'll have to compile with CFLAGS+="-msse2"

I can take a look at the other build failures tomorrow.

PR #7602, opened 2019-05-06
community/eog: upgrade to 3.32.1

Cogitri:
* streamline APKBUILD
* adopt maintainership

### DIFF
--- a/community/eog/APKBUILD
+++ b/community/eog/APKBUILD
@@ -1,39 +1,34 @@
-# Maintainer: 
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=eog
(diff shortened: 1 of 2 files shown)


prspkt:
@Cogitri `eog` will require `depends=gsettings-desktop-schemas`. The `gsettings-desktop-schemas-dev` package (which is pulled in by `gnome-desktop-dev`) does not explicitly depend on `gsettings-desktop-schemas`.

Would it be a better idea to resolve it [this way](https://github.com/alpinelinux/aports/pull/7647) instead?

Cogitri:
Yes, this will happen a lot otherwise. Thanks for the heads-up.

Cogitri:
Added the dep

PR #7610, opened 2019-05-06
testing/py3-graphviz: rename from py-graphviz

TBK:


maxice8:
Better do this in multiple commits, one to rename then to upgrade, producing a diff that is actually usable

PR #7618, opened 2019-05-06
main/mozjs60: upgrade to 60.6.2, fix s390x

tuan-hoang1:


tuan-hoang1:
Note to myself : reenable mozjs60 reverse dependency packages.

PR #7621, opened 2019-05-06
main/polkit: reenable s390x

tuan-hoang1:


tuan-hoang1:
Requires : https://github.com/alpinelinux/aports/pull/7618

PR #7623, opened 2019-05-06
community/networkmanager: reenable all arch

tuan-hoang1:


tuan-hoang1:
Requires https://github.com/alpinelinux/aports/pull/7621

Ikke:
@tuan-hoang1 The preference would be to have these in a single PR so that they can be applied in one go (unless the PR would become unwieldy to handle)

community/networkmanager/APKBUILD
```diff
@@ -2,10 +2,10 @@
 # Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=networkmanager
 pkgver=1.18.1
-pkgrel=1
+pkgrel=2
```
Ikke: Enabling arches does not require a `pkgrel` bump (which would trigger a rebuild of all packages).

PR #7628, opened 2019-05-07
testing/bcc: upgrade to 0.9.0

acj:


testing/bcc/APKBUILD
```diff
@@ -1,17 +1,34 @@
...
+
+prepare() {
+	cd "$builddir"
+	rm -rf tests/* && touch tests/CMakeLists.txt
+
+	# libbpf does not currently have tagged releases, so we fetch them here
+	# before the patches are applied.
+	mkdir -p "$builddir/build/src/cc"
+	cd "$builddir/src/cc"
+	git clone https://github.com/libbpf/libbpf.git
```
andypost: it should be separate package based on tags or date

acj: Done. Thanks!

acj:
@andypost This is ready for another look.

acj:
@maxice8 I was keeping them separate so that the change was clear for the first reviewer. I just merged the commits and pushed.

testing/bcc/APKBUILD
```diff
@@ -1,17 +1,28 @@
...
-pkgver=0.8.0
+pkgver=0.9.0
 pkgrel=0
 pkgdesc="A toolkit for creating efficient kernel tracing and manipulation programs"
 url="https://github.com/iovisor/bcc/"
-arch="aarch64 x86_64"
+arch="aarch64 x86 x86_64"
 license="Apache-2.0"
 # bcc's test suite requires privileged access to run BPF programs
 options="!check"
```
TBK: Did you try `options="checkroot"` ? -https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#options

acj: The test suite runs BPF programs in the kernel, so I don't think fakeroot will help here.

Ikke:
0.10.0 is out (see #8411)

kaey:
I don't think it's worth making separate package for libbpf until they make a stable release (bcc doesn't build against master branch of libbpf for example)

acj:
@Ikke @maxice8 This PR was approved two weeks ago. Can we merge this first and then deal with 0.10?

@kaey At the beginning my branch was similar to yours, but the reviewers asked for a separate package for libbpf. That's already done. The libbpf maintainers cut their first release last week, and we're dealing with the version change in #8225. The 0.0.3 release builds fine for me, as does master.

PR #7630, opened 2019-05-07
testing/py-fs: move to py3, upgrade, new dependencies (new aports)

TBK:
@fabaff could you review?


testing/py3-fs/APKBUILD
```diff
@@ -0,0 +1,31 @@
+# Contributor: Fabian Affolter <fabian@affolter-engineering.ch>
```
fabaff: Looks like that git/GitHub is not able to detect that the file was moved.

TBK: Yeah, I have experienced that before. Seems to be a bit flaky.

maxice8:
needs to be rebased

PR #7638, opened 2019-05-07
community/runc: upgrade to 1.0.0_rc8

tomalok:
Fixes regression in 1.0.0_rc7, see https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc8

### DIFF
--- a/community/runc/APKBUILD
+++ b/community/runc/APKBUILD
@@ -34,11 +34,6 @@ build() {
 	make man
 }
 
-check() {
-	cd "$builddir"
-	./runc --version
-}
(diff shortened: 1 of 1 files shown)


maxice8:
@clandmeter asked to not use `./binary --version` checks because they give a false sense of security that the package is actually tested.

Please refer to [this](https://github.com/alpinelinux/aports/pull/7326#discussion_r278797457)

tomalok:
@maxice8 updated this PR (and #7640) with regards to trivial `check()`s.

PR #7644, opened 2019-05-07
testing/libpipeline: upgrade to 1.5.1

maxice8:


maxice8:
@awilfox

awilfox:
As I've said multiple times before, I'm no longer maintaining any packages in Alpine Linux.  I had requested the current maintainers to handle this but I suppose I must do it myself.

PR #7647, opened 2019-05-07
community/gsettings-desktop-schemas: fix depends

prspkt:


maxice8:
builddir is default value and can be removed.

PR #7649, opened 2019-05-07
testing/bfs: upgrade to 1.4.1

maxice8:


testing/bfs/APKBUILD
```diff
@@ -1,32 +1,28 @@
...
 pkgdesc="Breadth-first variant of the UNIX find command"
+options="checkroot"
 url="https://github.com/tavianator/bfs"
 arch="all"
 license="0BSD"
-makedepends="linux-headers"
+makedepends="linux-headers libcap-dev acl-dev"
 checkdepends="bash"
 subpackages="$pkgname-doc"
-options="!checkroot"
```
tavianator: Getting rid of this results in
```
error: These tests expect filesystem permissions to be enforced, and therefore
will not work when run as root.
```

maxice8: Solved

tavianator: Doesn't seem to be solved

PR #7653, opened 2019-05-07
community/cmark: upgrade to 0.29.0

maxice8:


maxice8:
@maribu

maribu:
@maxice8: Sorry, I do not have any permission to merge or review to the alpine repos. However, I can confirm that `APKBUILD` works and it looks good to me (whatever this is worth).

I'm no longer using `cmark` personally and would have to rely on users flagging it out of date to update the package. So if you would like to adopt the package, this would be wonderful.

maxice8:
>  Sorry, I do not have any permission to merge or review to the alpine repos. However, I can confirm that APKBUILD works and it looks good to me (whatever this is worth).

No worries, i just tag the maintainer whether i update.

> I'm no longer using cmark personally and would have to rely on users flagging it out of date to update the package. So if you would like to adopt the package, this would be wonderful.

Sure i'll push with adoption changes.

maxice8:
re-pushed with change of maintainership

PR #7675, opened 2019-05-08
remove awilfox as maintainer per request

maxice8:
Sorry, this will be the last tag @awilfox just to confirm.


ncopa:
merged. thanks!

PR #7676, opened 2019-05-08
community/eom: upgrade to 1.22.1

maxice8:


community/eom/install-sh.patch
```diff
@@ -1,10 +0,0 @@
-diff --git a/install-sh b/install-sh
-index defb86a..8175c64 100755
---- a/install-sh
-+++ b/install-sh
```
andypost: Somehow there's no such fine anymore https://github.com/mate-desktop/eom

PR #7679, opened 2019-05-08
unmaintained/opencv: move from testing

maxice8:
Nobody cares about it since it still needs a rebuild for ffmpeg soname
bump, has no maintainer and the only package that consumes it also has
no maintainer.


maxice8:
@ncopa

PR #7682, opened 2019-05-08
unmaintained/liblastfm: move from main

maxice8:
- Only consumer is in unmaintained
- Uses Qt4 which is a security risk


maxice8:
@ncopa

PR #7689, opened 2019-05-08
community/arandr: upgrade to 0.1.10

maxice8:


andypost:
I guess @ScrumpyJack is ok to move to py3

PR #7690, opened 2019-05-08
main/attr: upgrade to 2.4.48

maxice8:


ncopa:
> I am rolling a dice to pick whether to enable $pkgname-static or disabling it until a policy is written down somewhere accessible for it.

I tend to `--disable-static` unless someone specifically asks for static.

maxice8:
On Mon, May 27, 2019 at 01:07:47PM -0700, Natanael Copa wrote:
> > I am rolling a dice to pick whether to enable $pkgname-static or disabling it until a policy is written down somewhere accessible for it.
> 
> I tend to `--disable-static` unless someone specifically asks for static.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7690#issuecomment-496296884

I'll do it then as well.

PR #7707, opened 2019-05-08
main/imagemagick: Add build for PerlMagick packages

timlegge:
Build the PerlMagick perl modules of imagemagick

### DIFF
--- a/main/imagemagick/APKBUILD
+++ b/main/imagemagick/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=imagemagick
-pkgver=7.0.8.44
+pkgver=7.0.8.49
 pkgrel=0
 _pkgver=${pkgver%.*}-${pkgver##*.}
 _abiver=7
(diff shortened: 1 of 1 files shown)


timlegge:
@maxice8, @ncopa, @tcely any possibility of movement on this one?  The trouble with imagemagick as you may have noticed is the frequent releases remove the previous release archive and the tests fail every other release so if this is eventually accepted it needs to get in shortly after a release is announced by Imagemagick on a release that will actually build completely. 

I can build permagick as a separate package but as its a sub package of imagemagick this makes sense to me.

The other option is to point to the github releases as those archives stay around.

Tim

ncopa:
Can we simply move the `usr/lib/perl5` instead of installing it twice?

Something like:
````sh
_perlmagick() {
	pkgdesc="PerlMagick Perl Modules for ImageMagick"
	mkdir -p "$subpkgdir"/usr/lib
	mv "$pkgdir"/usr/lib/perl5 "$subpkgdir"/usr/lib/
}
````
￼

timlegge:
Probably.  I will give it a shot and test it in my use case.

Tim

timlegge:
> Can we simply move the `usr/lib/perl5` instead of installing it twice?
> 
> Something like:
> 
> ```shell
> _perlmagick() {
> 	pkgdesc="PerlMagick Perl Modules for ImageMagick"
> 	mkdir -p "$subpkgdir"/usr/lib
> 	mv "$pkgdir"/usr/lib/perl5 "$subpkgdir"/usr/lib/
> }
> ```

@ncopa I made those changes but had to increment the imagemagick version because they don't keep previous imagemagick  releases at https://www.imagemagick.org/download/releases/ according to the site github is the official mirror which does keep past releases around.


Tim

PR #7746, opened 2019-05-09
testing/urlscan: new aport

maxwell-k:


maxice8:
there is already a py-urwid package

maxwell-k:
@maxice8 thanks, I don't know how I missed it, I amended the PR

PR #7747, opened 2019-05-09
community/qca: upgrade to 2.2.1

maxice8:


maxice8:
@ncopa

community/qca/APKBUILD
```diff
@@ -1,37 +1,34 @@
...
-makedepends="$depends_dev cmake cyrus-sasl-dev"
-install=""
-subpackages="$pkgname-dev $pkgname-doc"
-source="http://download.kde.org/stable/qca/$pkgver/src/qca-${pkgver}.tar.xz
-	qca-openssl-1.1.patch
-	"
-
-builddir="$srcdir"/qca-$pkgver
+license="LGPL-2.1-or-later"
+makedepends="qt5-qtbase-dev cmake cyrus-sasl-dev"
```
tcely: Why did you remove `qt5-qtbase-dev` from `-dev` subpackage?

maxice8: it wasn't removed, it now tracks via pc: provider. It actually tracked before but nobody cared to fix it.

```
qca-dev-2.2.1-r0 depends on:
libqca=2.2.1-r0
pc:Qt5Core
pkgconfig

qca-dev-2.1.3-r6 depends on:
qt5-qtbase-dev
pc:Qt5Core
pkgconfig
qca=2.1.3-r6
```

community/qca/APKBUILD
```diff
@@ -1,37 +1,34 @@
...
 check() {
-	cd "$builddir"
 	make test
 }
 
-sha512sums="0aec277e0695da2e45298f0a9006213829fe4c449a79969e472947db54f45000ba6e22361b782465bdc03f269b7301d318c843f5a83db459a118e58a03f3116a  qca-2.1.3.tar.xz
-13eff6727ddcb1fdfd8d57175bcbee27dcbd60823d91b7b16224067908638ee8ae4b2a141e18b6038d4c0347ddc2011e04c0956309c15032b64ea0056fd7372a  qca-openssl-1.1.patch"
+libs() {
+	default_libs
+	mv "$pkgdir"/usr/lib/qca-qt5 "$subpkgdir"/usr/lib
```
tcely: I don't like depending on how `default_libs` currently operates for directory creation.

```suggestion
	mkdir -p "$subpkgdir"/usr/lib
	mv "$pkgdir"/usr/lib/qca-qt5 "$subpkgdir"/usr/lib
```

maxice8: if there is no "$pkgdir"/usr/lib then the package is broken.

Consider it an early warning system.

tcely: The concern here is `"$subpkgdir"/usr/lib` may not be created by `default_libs` for some reason then this split breaks because it didn't do the creation itself.

maxice8: That is the early warning system part.

tcely: I don't see why not having any shared objects in "$pkgdir/usr/lib" should cause this move to fail.

Perhaps they were moved in a previous split or just install in a subdirectory now? This looks like a totally unrelated late failure for no good reason to me.

PR #7749, opened 2019-05-09
testing/beancount: new aport

maxwell-k:


testing/beancount/APKBUILD
```diff
@@ -0,0 +1,38 @@
+# Contributor: Keith Maxwell <keith.maxwell@gmail.com>
+# Maintainer: Keith Maxwell <keith.maxwell@gmail.com>
+pkgname=beancount
+_pyname=beancount
+pkgver=2.2.1
+pkgrel=0
+pkgdesc="Double-Entry Accounting from Text Files"
+url="http://furius.ca/beancount/"
+arch="x86_64"
```
maxice8: shouldn't it be all ?

maxwell-k: Yes it should thank you!

PR #7750, opened 2019-05-09
testing/rstcheck: new aport

maxwell-k:


testing/rstcheck/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Keith Maxwell <keith.maxwell@gmail.com>
+# Maintainer: Keith Maxwell <keith.maxwell@gmail.com>
+pkgname=rstcheck
+_pyname=rstcheck
```
TBK:
```suggestion

```

testing/rstcheck/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+pkgver=3.3.1
+pkgrel=0
+pkgdesc="Checks syntax of reStructuredText and code blocks nested within it"
+url="https://github.com/myint/rstcheck"
+arch="noarch"
+license="MIT"
+depends="python3 py3-docutils py3-setuptools"
+checkdepends="bash"
+# the test suite isn't part of the pypi package:
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/myint/$pkgname/archive/v$pkgver.tar.gz
```
TBK:
```suggestion
source="$pkgname-$pkgver.tar.gz::https://github.com/myint/$pkgname/archive/v$pkgver.tar.gz
```

testing/rstcheck/APKBUILD
maxwell-k: This variable is in line with the draft policy at <https://wiki.alpinelinux.org/wiki/Python_package_policies>. It was deliberately added to facilitate cross referencing against <https://pypi.org>. Perhaps if it should be removed the linked policy on the wiki could be clarified first.

testing/rstcheck/APKBUILD
maxwell-k: Can you point to a rule or documentation that encourages this change please?

This is just a style preference. Unless there's an explicit policy, I am going to ignore the suggestion in order to avoid extra work that is unnecessary in my view.

I can find five examples of exactly the code before the suggested change, with {}s, in main and thirteen in community: 

```
$ git grep 'source="${pkgname}-${pkgver}' main | wc -l
5
$ git grep 'source="${pkgname}-${pkgver}' community | wc -l
13
```

maxwell-k:
After a conversation on IRC I understand that the use of braces or not is down to the maintainer.

testing/rstcheck/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+# Maintainer: Keith Maxwell <keith.maxwell@gmail.com>
+pkgname=rstcheck
+_pyname=rstcheck
+pkgver=3.3.1
+pkgrel=0
+pkgdesc="Checks syntax of reStructuredText and code blocks nested within it"
+url="https://github.com/myint/rstcheck"
+arch="noarch"
+license="MIT"
+depends="python3 py3-docutils py3-setuptools"
```
TBK:
```suggestion
depends="python3 py3-docutils"
makedepends="py3-setuptools"
```

maxwell-k: setuptools is required at runtime, so `py3-setuptools` is deliberately set as a
runtime dependency.

PEP453 explains that it is just an implementation detail that setuptools is
installed by default on Python 3 so the dependency is required.

> other projects which explicitly require setuptools must still provide an
> appropriate dependency declaration

https://www.python.org/dev/peps/pep-0453/#automatic-installation-of-setuptools

`/usr/bin/rstcheck` contains the line `from pkg_resources import
load_entry_point`. `pkg_resources` comes from
`/usr/lib/python3.7/site-packages/pkg_resources/__init__.py` a part of the
py3-setuptools package.

Installing and then removing `py3-setuptools` breaks the version of setuptools
that is bundled with `python3` and if this runtime dependency is not explicit,
breaks this package.

mps mentioned on IRC that this maybe be discouraged in packaging, but a search
of the documentation yields no results.

TBK: Thanks for taking the time explaining it.

maxwell-k:
@TBK thank you for merging! I am sorry that we had so much back and forward on this one; I learnt a little about Alpine Linux through this.

I appreciate your time.

TBK:
No problem. Have a nice day.

PR #7758, opened 2019-05-09
testing/minio: modernize init.d

5paceToast:
Now that we have openrc 0.41.2, supervise-daemon is actually usable.
Minio also behaves much nicer with it.


maxice8:
pkgrel needs to be bumped for changes to take effect.

5paceToast:
I also forgot to update the checksums... quite silly of me!

PR #7761, opened 2019-05-10
testing/i2pd: upgrade to 2.25.0

r4sas:


testing/i2pd/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: l-n-s <supervillain@riseup.net>
 # Maintainer: l-n-s <supervillain@riseup.net>
 pkgname=i2pd
-pkgver=2.24.0
-pkgrel=1
+pkgver=2.25.0
+pkgrel=0
 pkgdesc="I2P Router written in C++"
 url="https://github.com/PurpleI2P/i2pd"
 arch="all"
```
maxice8: license is not SPDX

maxice8: if you're adding openssl-dev then it brings openssl, so the openssl dependency is redundant

maxice8: has conf.d and init.d files but no $pkgname-openrc

maxice8: would be nice to have a comment on why check is disabled

maxice8: bulddir= is using default value so it can be removed

maxice8: no need for `cd "$builddir"`

r4sas:
Oops, wrong button. 

About license - change to identifier from SPDX list?

maxice8:
https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license

r4sas:
> has conf.d and init.d files but no $pkgname-openrc

Still dont understand about that, because here already present `subpackages="$pkgname-openrc"`.

maxice8:
yes, i missed it.

andypost:
please squash commits into single one

PR #7762, opened 2019-05-10
testing/assimp: new aport

russkel:
http://www.assimp.org/
Open Asset Import Library imports and exports 3D model formats.

I need some guidance on what to do with this package where it builds a static library called irrXML from external source they seem to have included in their tree. Do I create an official package for the irrXML package (https://www.ambiera.com/irrxml/downloads.html) and then force assimp to use this version of irrXML?

### DIFF
--- a/testing/assimp/APKBUILD
+++ b/testing/assimp/APKBUILD
@@ -1,0 +1,42 @@
+# Contributor: Russ Webber <russ@rw.id.au>
+# Maintainer: Russ Webber <russ@rw.id.au>
+pkgname=assimp
+pkgver=4.1.0
+pkgrel=0
+pkgdesc="Open Asset Import Library imports and exports 3D model formats."
+url="http://www.assimp.org/"
+arch="all"
+license="BSD-3-Clause"
+makedepends="cmake zlib-dev minizip-dev"
+subpackages="$pkgname-dev"
+source="assimp-$pkgver.tar.gz::https://github.com/assimp/assimp/archive/v$pkgver.tar.gz
+        01-rm-revision-test.patch"
+
+build() {
+  if [ "$CBUILD" != "$CHOST" ]; then
+    CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
+  fi
+
+  cmake \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DBUILD_SHARED_LIBS=True \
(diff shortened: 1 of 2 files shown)


testing/assimp/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+pkgrel=0
+pkgdesc="Open Asset Import Library imports and exports 3D model formats."
+url="http://www.assimp.org/"
+arch="all"
+license="BSD-3-Clause"
+makedepends="cmake zlib-dev minizip-dev"
+subpackages="$pkgname-dev"
+source="assimp-$pkgver.tar.gz::https://github.com/assimp/assimp/archive/v$pkgver.tar.gz
+        01-rm-revision-test.patch"
+builddir="$srcdir/assimp-$pkgver"
```
maxice8: default value can be removed

testing/assimp/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+
+build() {
+  if [ "$CBUILD" != "$CHOST" ]; then
+  CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
```
maxice8: need indent here

testing/assimp/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+  fi
+  cmake \
+  -DCMAKE_INSTALL_PREFIX=/usr \
```
maxice8: need indent from here

testing/assimp/APKBUILD
```diff
@@ -0,0 +1,42 @@
...
+  -DCMAKE_INSTALL_LIBDIR=lib \
+  -DBUILD_SHARED_LIBS=True \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
+  -DCMAKE_C_FLAGS="$CFLAGS" \
+  -DASSIMP_BUILD_TESTS=True \
+  ${CMAKE_CROSSOPTS} .
```
maxice8: to here

testing/assimp/APKBUILD
russkel: Oops sorry, I think this also needs to be fixed in newapkbuild script as this was how I created it.

Ikke:
I've confirmed with ncopa, and he preferred not to distribute that library. We confirmed that the library does not seem to have any references to that static lib, so it is probably not even needed. I just removed the .a file in the package function.

Can you verify that everything works when it's built?

russkel:
@Ikke thanks for that. I will try and confirm everything works. I will also raise this issue upstream.

PR #7767, opened 2019-05-10
community/qt5-*: upgrade to 5.12.3

fcolista:
Upgrade all the qt5 packages but not qt5-qtwebengine.
The latter fails to build on my buildenv due to memory error, so I'm not 100% sure it would work with all archs.
Moreover, Cogitiri mentioned on IRC that this package (even if builts for him on x86_64, it does not work): 

> Cogitri:  fcolista: http://ix.io/1IjX
> Cogitri:  It doesn't throw any errors anymore, but it still goes to a blank window once it has finished loading the webpage :c

But this is broken also on 5.12.1.

.: Francesco Colista


fcolista:
Committed in 7f9789919af11d5fbb73e1e25596746dbfb3f159

Cogitri:
Could you try launching `nextcloudclient`? It SIGABRTs now and I'm not exactly sure how to debug it :(

PR #7768, opened 2019-05-10
main/postgresql: update secfixes

andypost:
CVE-2019-10127, CVE-2019-10128, CVE-2019-10129, CVE-2019-10130

Ref https://www.postgresql.org/docs/current/release-11-3.html


main/postgresql/APKBUILD
```diff
@@ -36,6 +36,11 @@ builddir="$srcdir/$pkgname-$pkgver"
 options="!checkroot"
 
 # secfixes:
+#   11.3-r0:
+#   - CVE-2019-10127
+#   - CVE-2019-10128
```
jirutka: CVE-2019-10127 and CVE-2019-10128 are Windows-only, so this package has never been affected by those.

https://www.postgresql.org/about/news/1939/

andypost: It was commited as 7cf139bac41c8f2e1885d5f99334daeaeb059ac3 so I updated commit to remove wrong secfixes

andypost:
Tests fails on armv7 with new perl

tcely:
This looks to be merged already. Maybe a new PR for any tweaks would be good?

PR #7775, opened 2019-05-10
main/py-tz: upgrade to 2019.1

OTLabs:
https://github.com/stub42/pytz/releases 2019.1

### DIFF
--- a/main/py-tz/APKBUILD
+++ b/main/py-tz/APKBUILD
@@ -2,8 +2,8 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=py-tz
 _pkgname=pytz
-pkgver=2018.9
-pkgrel=1
+pkgver=2019.1
+pkgrel=0
 pkgdesc="A Python definitions of world timezone"
 url="http://pytz.sourceforge.net/"
 arch="noarch"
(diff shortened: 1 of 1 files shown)


main/py-tz/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=py-tz
 _pkgname=pytz
-pkgver=2018.9
+pkgver=2019.1
 pkgrel=1
```
maxice8: pkgrel should be 0

TBK:
```suggestion
pkgrel=0
```

OTLabs: Fixed. Thank you!

OTLabs: Fixed. Thank you!

OTLabs:
@fabaff could you please take a look at error I get during build: https://cloud.drone.io/alpinelinux/aports/3118

main/py-tz/APKBUILD
```diff
@@ -2,16 +2,16 @@
...
+pkgrel=0
 pkgdesc="A Python definitions of world timezone"
 url="http://pytz.sourceforge.net/"
 arch="noarch"
 license="MIT"
 makedepends="py-setuptools py3-setuptools"
 subpackages="py2-${pkgname#py-}:_py2 py3-${pkgname#py-}:_py3"
 source="https://pypi.io/packages/source/${_pkgname:0:1}/$_pkgname/$_pkgname-$pkgver.tar.gz"
-builddir="$srcdir/$_pkgname-$pkgver"
+builddir="$srcdir/$_pkgname-release_$pkgver/src"
```
tcely: What made you change this? It is breaking your build.

```suggestion
builddir="$srcdir/$_pkgname-$pkgver/src"
```

OTLabs: I had a very strange issue with source file's checksum. Old archive was using different directory so I had to adjust it this way. For some reason it was partially working with new file. Implementing your kind suggestion makes this package build just fine. Thank you!

OTLabs:
@fabaff The package builds fine now. Could you please check the proposed changes and emit your kind opinion? Thank you in advance!

OTLabs:
To use this module we use `import pytz`, that make me think that this module's name is `pytz`, so the package file for Alpine Linux could be named `py-pytz`. What do you think?

tcely:
> To use this module we use `import pytz`, that make me think that this module's name is `pytz`, so the package file for Alpine Linux could be named `py-pytz`. What do you think?

I don't think it is worth changing the name of this package now. It has been this way since at least 2015.

PR #7779, opened 2019-05-10
testing/fzy acceptance tests

3ju:
This brings in a new aport: `ruby-ttycheck`.

I intend to open a new PR to move `fzy`  (and `ruby-ttycheck`) to community once the latter hits testing.

### DIFF
--- a/testing/fzy/APKBUILD
+++ b/testing/fzy/APKBUILD
@@ -1,32 +1,29 @@
 # Contributor: Jonathan Neuschäfer <j.neuschaefer@gmx.net>
+# Maintainer: Eivind Uggedal <eu@eju.no>
 pkgname=fzy
 pkgver=1.0
-pkgrel=0
+pkgrel=1
 pkgdesc="A better fuzzy finder"
 url="https://github.com/jhawthorn/$pkgname"
 arch="all"
 license="MIT"
-depends=""
-makedepends=""
-install=""
+checkdepends="ruby-minitest ruby-ttytest"
 subpackages="$pkgname-doc"
 source="https://github.com/jhawthorn/fzy/releases/download/$pkgver/$pkgname-$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$pkgver"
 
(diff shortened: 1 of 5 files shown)


testing/fzy/APKBUILD
```diff
@@ -1,14 +1,13 @@
 # Contributor: Jonathan Neuschäfer <j.neuschaefer@gmx.net>
+# Maintainer: Eivind Uggedal <eu@eju.no>
 pkgname=fzy
 pkgver=1.0
-pkgrel=0
+pkgrel=1
```
maxice8: don't think pkgrel needs to be bumped.

testing/ruby-ttytest/APKBUILD
```diff
@@ -0,0 +1,48 @@
...
+depends="ruby tmux"
+checkdepends="ruby-minitest ruby-rake ruby-bundler"
+source="$pkgname-$pkgver.tar.gz::$url/archive/v$pkgver.tar.gz
+	skip-yard.patch
+	fix-flaky-test.patch
+	gemspec.patch"
+builddir="$srcdir/$_gemname-$pkgver"
+
+build() {
+	cd "$builddir"
```
maxice8: no need to `cd "$builddir"` this is done implicitly

testing/fzy/APKBUILD
3ju: Will it be built (and tested with the new acceptance tests) on the official edge builders without?

testing/ruby-ttytest/APKBUILD
3ju: Thanks. The perils of not contributing for 4 years...

testing/fzy/APKBUILD
maxice8: no

3ju: I think a revbump therefore is warranted so that it could potentially fail on the official builders in testing before being moved to community.

PR #7785, opened 2019-05-10
main/py-larch: remove, no dependants

3ju:
I packaged and maintained this long ago as a dependeny of
main/obnam. Since obnam is removed from the tree due to
the project being retired, py-larch should also be removed.


3ju:
Closing this, in favor of a new PR which removes more obselete `obnam` deps.

PR #7787, opened 2019-05-10
main/zstd: test

maxice8:


main/zstd/APKBUILD
```diff
@@ -5,15 +5,14 @@ pkgver=1.4.0
 pkgrel=0
```
TBK:
```suggestion
pkgrel=1
```

TBK: PR/commit title could be more descriptive.

> Fix armhf, Enable armhf...

PR #7788, opened 2019-05-10
testing/chrome-gnome-shell: new aport

Cogitri:


testing/chrome-gnome-shell/APKBUILD
```diff
@@ -0,0 +1,39 @@
...
+		-DBUILD_SHARED_LIBS=True \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_CXX_FLAGS="$CXXFLAGS" \
+		-DCMAKE_C_FLAGS="$CFLAGS" \
+		-DBUILD_EXTENSION=FALSE \
+		${CMAKE_CROSSOPTS}
+	make
+}
+
+check() {
```
TBK:
```suggestion

```

If there is no test suite then please remove the check().

Cogitri: Hm, but why? Once they've added tests we can just flick the switch (`options="!check"`) and it'll work.

TBK: Will have to ask higher authority on why...

I have had PRs before with the same reasoning to keep it, but I was asked to remove it.

One reason could be that keeping could cause confusion.

Cogitri: Ah, alright. I mean I don't mind removing it, but I like to know the reason for my doing :)

PR #7812, opened 2019-05-11
main/powertop: update to 2.10, adopt (again)

3ju:
Adopt the unmaintained aport I previously used to maintain.


3ju:
Sorry, did not see your PR @maxice8. I'm closing this in favor of #7326 so that you can maintain the package.

The part of the `musl-fix-headers.patch` patch that touches `src/devices/devfreq.h` can be dropped from your PR since it's been upstreamed.

PR #7814, opened 2019-05-12
community/mpv: Enable lua explicitly and fix makedepends

jantatje:
Enable lua support explicitly, because it got disabled without
anyone noticing instead of throwing an error during build.
Commit cdf8852d1e5ef6c766b94eec7f31fb38410dad8f changed lua5.2-dev
in makedepends to lua5.3-dev, but mpv needs either lua5.2 or luajit.


jantatje:
@maxice8 I removed the `cd "$builddir"` commands.

PR #7818, opened 2019-05-12
testing/plzip: upgrade to 1.8

maxice8:


testing/plzip/APKBUILD
```diff
@@ -1,39 +1,29 @@
...
-install=""
 subpackages="${pkgname}-doc"
 source="https://download.savannah.gnu.org/releases/lzip/$pkgname/$pkgname-$pkgver.tar.gz"
 
-builddir="$srcdir/$pkgname-$pkgver"
 build() {
-	cd "$builddir"
 	./configure \
-		--build=$CBUILD \
-		--host=$CHOST \
```
bratkartoffel: why removing these two?

maxice8: The script of plzip is not a GNU Autoconf configure script, it is handmade and thus doesn't recognize those variables.

same for the others removed.

PR #7820, opened 2019-05-12
community/py-configparser: upgrade to 3.7.4

maxice8:
https://github.com/jaraco/configparser/blob/master/CHANGES.rst#374


maxice8:
@stwa

PR #7821, opened 2019-05-12
main/gdb: upgrade to 8.3

J0WI:


main/gdb/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=gdb
-pkgver=8.2.1
-pkgrel=2
+pkgver=8.3
+pkgrel=0
 pkgdesc="The GNU Debugger"
 url="https://www.sourceware.org/gdb/"
 arch="all"
```
maxice8: empty variables can be removed for brevity

maxice8: GPL-3.0 is not a valid SPDX identifier

main/gdb/APKBUILD
```diff
@@ -15,7 +15,6 @@ source="https://ftp.gnu.org/gnu/$pkgname/$pkgname-$pkgver.tar.xz
 	s390x-use-elf-gdb_fpregset_t.patch
 	ppc-musl.patch
 	ppc-ptregs.patch
-	aarch64-headers.patch
 	"
 
 builddir="$srcdir"/$pkgname-$pkgver
```
maxice8: default value for builddir, can be removed.

main/gdb/APKBUILD
J0WI: Changed to `GPL-3.0-or-later LGPL-3.0-or-later`

J0WI:
Fixed in ef7c6ebe6a056a2777405f4c0749d318eda28e6b

PR #7823, opened 2019-05-12
testing/libuninameslist: upgrade to 20190305

maxice8:


maxice8:
@errm

testing/libuninameslist/APKBUILD
```diff
@@ -1,38 +1,38 @@
...
 
 package() {
-        cd "$builddir"
-        make DESTDIR="$pkgdir" install
+	make DESTDIR="$pkgdir" install
+}
+
+static() {
+	depends=""
+	pkgdesc="$pkgdesc (static library)"
```
andypost: maybe just call `default_static`?

maxice8: Didn't get greenlight from abuild devs to use the newly introduced default_static.

PR #7826, opened 2019-05-12
unmaintained/lxappearance-obconf: move from testing

maxice8:


Ikke:
Can you add to the commit message why you move it to unmaintained in this case?

maxice8:
Is it really necessary for the generic reason 66 'no maintainer' ?

Ikke:
Usually something triggers the move to unmaintained. Most of the time if a package is still building fine and there are no new updates or serious problems, we might just keep it where it is.

maxice8:
On Sun, May 12, 2019 at 01:24:10PM -0700, Kevin Daudt wrote:
> Usually something triggers the move to unmaintained. Most of the time if a package is still building fine and there are no new updates or serious problems, we might just keep it where it is.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7826#issuecomment-491626369

It has pending updates and can't be updated until lxqt-build-tools is updated to 0.6.0

if we keep it as is outdated then there should be no problems.

PR #7606, opened 2019-05-06
 testing/py3-pytest-cov: rename & upgrade to 2.7.1 

TBK:

PR #7827, opened 2019-05-12
community/goobook: move from testing, upgrade and add -doc

maxwell-k:


community/goobook/APKBUILD
```diff
@@ -0,0 +1,29 @@
+# Contributor: Ivan Tham <pickfire@riseup.net>
+# Maintainer: Ivan Tham <pickfire@riseup.net>
+pkgname=goobook
+pkgver=3.3
+pkgrel=0
+pkgdesc="Access Google contacts from the command line"
+url="https://gitlab.com/goobook/goobook"
+arch="noarch"
+license="GPL-3.0"
```
maxice8: GPL-3.0 is not a valid spdx identifier

maxwell-k: @maxice8 thank you! I have fixed this. It was sneaking through the sanity check so I have also filed https://github.com/alpinelinux/abuild/pull/84

maxwell-k:
Amended in line with comment from @maxice8 

The line `cd "$builddir"` is required in a split function

TBK:
My PR should be merged first https://github.com/alpinelinux/aports/pull/7606 and then this should be rebased.

PR #7831, opened 2019-05-12
main/groff: upgrade to 1.22.4

maxice8:


nmeum:
I don't speak groff, what's the purpose of `site.tmac`?

PR #7835, opened 2019-05-12
main/hiredis: add check support, revbump

3ju:
Also make license SPDX compliant.


3ju:
Patch for upstream here: https://github.com/redis/hiredis/pull/665

PR #7839, opened 2019-05-12
main/py-hiredis: update to 1.0.0, adopt

3ju:
* Adopt package (again)
* Modernize
* Add python3 support
* Run tests
* Patch to work with system hiredis 0.14


TBK:
Is py2 support strictly needed or can we move this to being py3 exclusive?

3ju:
> Is py2 support strictly needed or can we move this to being py3 exclusive?

There are no revdeps for either py2 or py3. I can drop the py2 version if Alpine has started on py2 deprecation.

TBK:
We have started the process.

You can add the following for backwards compatibility:
```
replaces="py-hiredis" # for backwards compatibility
provides="py-hiredis=$pkgver-r$pkgrel" # for backwards compatibility
```

3ju:
Py2 version now dropped.

PR #7841, opened 2019-05-12
community/abcde: upgrade to 2.9.3 and fix dependency on py3-eyed3

andypost:
Refs
- release https://git.einval.com/cgi-bin/gitweb.cgi?p=abcde.git;a=tag;h=e6400b13b5c4b8505eddd1e36e133757eddbd447
- changelog https://git.einval.com/cgi-bin/gitweb.cgi?p=abcde.git;a=blob;f=changelog;h=8f99618215bf7666911c3a291e52a58a54a76508;hb=5c6f9e4116d19e6ebabd543a2de0d87eb2f8cb58

PS Also it fix it for https://build.alpinelinux.org/buildlogs/build-3-10-armv7/community/abcde/abcde-2.9.2-r0.log

### DIFF
--- a/community/abcde/APKBUILD
+++ b/community/abcde/APKBUILD
@@ -1,32 +1,27 @@
 # Maintainer:
 pkgname=abcde
-pkgver=2.9.2
+pkgver=2.9.3
 pkgrel=0
 pkgdesc="A command line CD encoder that reads your CD, downloads the track information from a CDDB provider, and rips your CD"
 url="http://abcde.einval.com/"
 arch="noarch"
-license="GPL-2.0"
-depends="bash cd-discid py-eyed3"
-depends_dev=""
-makedepends="$depends_dev"
-install=""
+license="GPL-2.0-or-later"
+depends="bash cd-discid py3-eyed3"
+options="!check"  # No test suite.
 subpackages="$pkgname-doc"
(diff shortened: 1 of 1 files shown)


community/abcde/APKBUILD
```diff
@@ -1,12 +1,12 @@
 # Maintainer:
 pkgname=abcde
-pkgver=2.9.2
+pkgver=2.9.3
 pkgrel=0
 pkgdesc="A command line CD encoder that reads your CD, downloads the track information from a CDDB provider, and rips your CD"
 url="http://abcde.einval.com/"
 arch="noarch"
 license="GPL-2.0"
```
maxice8: GPL-2.0 isn't an SPDX license

community/abcde/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
-depends="bash cd-discid py-eyed3"
+depends="bash cd-discid py3-eyed3"
 depends_dev=""
```
maxice8: empty variables can be removed for brevity

community/abcde/APKBUILD
```diff
@@ -1,12 +1,12 @@
...
 makedepends="$depends_dev"
```
maxice8: can be removed as the variable it uses is empty

andypost: @maxice8 all fixed

PR #7848, opened 2019-05-12
testing/openresty: upgrade to 1.15.8.2

tcely:


Ikke:
@tcely Can you clean these commits up a bit? There are several upgrades and one upgrade revert in this single PR.

tcely:
@Ikke I squashed the RC upgrades into a single commit.

Ikke:
And why is that revert there? (Would be nice to have it included in the commit message)

tcely:
The revert was there to preserve clean history and diffs, but since the commits are squashed now those are being list anyway.

main/nginx/APKBUILD
```diff
@@ -304,7 +306,7 @@ _module() {
 	sonames="${sonames:-"ngx_${name}_module.so"}"
 
 	pkgdesc="$pkgdesc (module $name)"
-	depends="$pkgname $(eval "echo \$_${name}_depends")"
+	depends="$_provide $(eval "echo \$_${name}_depends")"
```
ncopa: @jirutka what do you think about this? both `openresty` and `nginx` will provide `cmd:nginx` but there is no guarantee which implementation the module will pull in.

testing/openresty/APKBUILD
```diff
@@ -139,7 +140,7 @@ _module() {
 	local _deps="$(eval echo "\$_module_${name//-/_}_deps")"
 
 	pkgdesc="$pkgdesc (module $soname)"
-	depends="cmd:nginx${_deps:+ }${_deps}"
+	depends="${_provide}${_deps:+ }${_deps}"
```
ncopa: will `openresty` (1.15) modules work with `nginx` (1.16) and vice versa?

Wouldn't it be better to depend on the exact nginx implementation?

testing/openresty/APKBUILD
```diff
@@ -1,9 +1,10 @@
 # Contributor: tcely <openresty+aports@tcely.33mail.com>
 # Maintainer: tcely <openresty+aports@tcely.33mail.com>
-pkgname=openresty
-pkgver="1.15.8.1"
+_upstream="openresty"
```
ncopa: IMHO, the `_upstream` variable here is meaningless and should be avoided.

testing/openresty/GPGKEYS
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd" >
```
ncopa: Do we really need to store this in html format?

testing/openresty/APKBUILD
```diff
@@ -32,17 +34,33 @@ _add_module() {
...
+_add_module xslt-filter
 
 prepare() {
 	default_prepare
 	cd "$builddir"
 
 	sed -i.orig -e '/--add-module=$dir/s/module/dynamic-&/' configure
+
+	local tag='LuaJIT-2.1'
+	cd "bundle/${tag}"-*
```
ncopa: Could we use system `luajit` instead of the bundled?

PR #7850, opened 2019-05-12
testing/biber: move from community

maxice8:
- Has dependencies in testing:
  - perl-data-dump
  - perl-data-compare


maxice8:
@fcolista

tcely:
Shouldn't `perl-data-dump` and `perl-data-compare` move to `community` instead?

maxice8:
If that is more expedient

PR #7491, opened 2019-05-02
testing/gnome-disk-utility: remove from unmaintained and upgrade to 3.32.1

Cogitri:
Also adopt maintainership

PR #7851, opened 2019-05-12
unmaintained/gnome-disk-utility: remove, is in testing/

maxice8:


andypost:
Maybe better to upgrade package in community? it using **3.12.1** but the one you removing is 3.20.0 so need @ncopa to decide

maxice8:
i think @Cogitri can do it, he is a big gnome fan.

ncopa is out for 2 weeks as far as i remember.

I'd say let's let @Cogitri bump it to the latest gnome version, he is working on bringing up-to-date gnome into Alpine and gnome-disk-utility is part of gnome.

Cogitri:
@andypost See #7491

PR #7865, opened 2019-05-13
community/vdr: use libdvbcsa on all arches except x86

tuan-hoang1:
FFdecsa uses x86-specific flags. We use libdvbcsa for other arches.

See vdr-plugin-dvbapi/INSTALL.


Ikke:
Does this warant a `pkgrel` bump because of the changed build flags?

tuan-hoang1:
@Ikke : very good point as usual :+1: Thanks. Updated.

PR #7880, opened 2019-05-14
community/caddy: upgrade to 1.0.0

andypost:
https://caddyserver.com/blog/announcing-caddy-1_0-caddy-2-caddy-enterprise


5paceToast:
Sorry @andypost, I was going to just suggest the changes in my other PR, but the review system in github (and in emails, actually) is restricted to contextual information in the diff (which doesn't include the actual parts at fault), so I submitted a new one instead.
Hope this is ok.

PR #7888, opened 2019-05-15
testing/minio-client: new aport

5paceToast:
A client for MinIO usage and administration.

### DIFF
--- a/testing/minio-client/APKBUILD
+++ b/testing/minio-client/APKBUILD
@@ -1,0 +1,46 @@
+# Maintainer: Chloe Kudryavtsev <toast@toastin.space>
+pkgname=minio-client
+_pkgname=mc
+_pkgver='RELEASE.2019-05-01T23-27-44Z'
+pkgver=${_pkgver#*.}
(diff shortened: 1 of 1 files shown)


testing/minio-client/APKBUILD
```diff
@@ -0,0 +1,46 @@
+# Maintainer: Chloe Kudryavtsev <toast@toastin.space>
+pkgname=minio-client
+_pkgname=mc
+_pkgver='RELEASE.2019-05-01T23-27-44Z'
+pkgver=${_pkgver#*.}
+pkgver=${pkgver%T*}
+pkgver=0.${pkgver//-}
```
Ikke: I noticed this with minio already but there it's probably too late, but you may want to use something like `0_git<date>` here. Then when upstream starts making releases, you know it will always be selected above this version.

With `0.2019...`, upstream version `0.x` is always going to be considered older than this version.

Ikke: n/m, these are upstream releases.

PR #7889, opened 2019-05-15
testing/asciinema-rs: new aport

TBK:


TBK:
I can not reproduce the failing tests locally and it seems to be specific to Drone since Travis runs the tests successfully.

PR #7891, opened 2019-05-15
community/libsoup: upgrade to 2.66.2

Cogitri:


andypost:
>>> No soname differences for ...
So depends on @ncopa

PR #7892, opened 2019-05-15
[3.7] main/xen: XSA-297

HRio:


HRio:
@rnalrd Please note that its only tested on an AMD box as that's was the only dom0 still on Alpine 3.7

PR #7893, opened 2019-05-15
main/mariadb: after entropy

HRio:
If mariadb is not started after an entropy provider it will take too long to start resulting in:
```
 * Starting mariadb ...190515 22:15:31 mysqld_safe Logging to syslog.
190515 22:15:31 mysqld_safe Starting mysqld daemon with databases from /var/lib/mysql
 [ ok ]
 * ERROR: mariadb failed to start
```


HRio:
The Travis error is `The job exceeded the maximum time limit for jobs, and has been terminated.`

PR #7896, opened 2019-05-16
testing/uacme: new aport

3ju:


testing/uacme/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+		--sysconfdir=/etc \
+		--mandir=/usr/share/man \
+		--localstatedir=/var \
+		--disable-maintainer-mode \
+		--with-openssl
+	make
+}
+
+check() {
+	2>&1 ./uacme -V | grep "$pkgver"
```
Ikke: Running just `$program -v` is not sufficient for a check. It usually only just a tiny portion of the code, giving basically no coverage. In that case, we prefer the package to be marked as not having a test suite (`options="!check"`).

3ju: While having an actual test suite is best, I don’t agree that it’s equivalent to !check for a compiled program with dynamic libraries, but I will remove it.

PR #7897, opened 2019-05-16
[3.9] main/xen: XSA-297

HRio:
Also includes needed patches from the xen stable-4.11 branch


HRio:
@rnalrd tested on a Intel CPU

PR #7899, opened 2019-05-16
main/u-boot: add pine64-lts support

PureTryOut:
`pine64-lts` includes Pine64 A64-LTS, Pine64 Don't be evil (PinePhone devkit) and eventually the actual Pine64 PinePhone. These are Allwinner devices, which require `arm-trusted-firmware` to be built into the image. This firmware is free software (BSD-3 licensed).

I've tested this on my Pine64 Don't be evil device where it boots successfully using this setup.


ncopa:
care to rebase this?

PureTryOut:
@ncopa, rebased.

u-boot for Pine64 requires arm-trusted-firmware in makedepends, but arm-trusted-firmware is in testing at the moment and has to be moved to main.

ncopa:
I will have to revert this. it was so broken that not even shell could parse it.

PureTryOut:
Huh, broken how?

PR #7901, opened 2019-05-16
community/mpv: Use lua5.2

5paceToast:
mpv does not currently support detecting lua5.3, which results in lua support being
disabled.
fixes [bugs.a.o#10461](https://bugs.alpinelinux.org/issues/10461)


jantatje:
#7814 also passes `--enable-lua` to waf, to prevent lua support from unintentionally getting disabled again.

5paceToast:
Closing in favor of #7814.

PR #7903, opened 2019-05-16
main/grub: handle intel-ucode

HRio:


HRio:
About the travis failure `The job exceeded the maximum log length, and has been terminated.`

PR #7906, opened 2019-05-16
[3.8] main/xen: XSA-297

HRio:
* Upgrade to XEN 4.10.3 (from 4.10.2)
* Add fixes for XSA-297


HRio:
@rnalrd unfortunately only build tested, as I currently have no system available to test this on.

PR #7907, opened 2019-05-16
community/caddy: upgrade to 1.0.0

5paceToast:
Superseeds #7880 (there was too much to change for a review).

1. Caddy now uses the Go Modules system:
    -> builds and tests require a new environment variable
    -> it now requires a custom cleanup step
2. Tests now pass (on my builder)

Note: I would be willing to adopt this package if it's too much of a bother to upkeep otherwise.


5paceToast:
Test failures seem to be drone-specific (one related to CPU manipulation, another uses 127.0.0.11 (?)).

clandmeter:
LGTM

5paceToast:
Changes: also adopted package (comment was in response to me asking if that was ok in irc, or at least I assume so).

clandmeter:
Yes sorry i wasnt clear.

PR #7917, opened 2019-05-17
community/rethinkdb: fix tribool errs with explicit cast and re=enable ppc64le

mksully22:
build breaking across all architectures with errors like:
````
src/clustering/administration/auth/user.cc: In member function 'bool auth::user_t::has_read_permission(const database_id_t&, const namespace_id_t&) const':
src/clustering/administration/auth/user.cc:199:20: error: cannot convert 'boost::logic::tribool' to 'bool' in return
             return table_permission;
                    ^~~~~~~~~~~~~~~~
```
add compiler flag -DBOOST_NO_CXX11_EXPLICIT_CONVERSION_OPERATORS to force compiler compatible behavior for tribool

### DIFF
--- a/community/rethinkdb/APKBUILD
+++ b/community/rethinkdb/APKBUILD
@@ -2,10 +2,10 @@
 # Maintainer: Daniel Treadwell <daniel@djt.id.au>
 pkgname=rethinkdb
 pkgver=2.3.6
-pkgrel=9
+pkgrel=10
 pkgdesc="Distributed powerful and scalable NoSQL database"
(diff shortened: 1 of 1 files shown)


community/rethinkdb/APKBUILD
```diff
@@ -5,7 +5,7 @@ pkgver=2.3.6
 pkgrel=9
```
tcely: Do we want to bump `pkgrel` for this change to compiled sources?

mksully22: Yep, I'll update

PR #7927, opened 2019-05-18
community/rspamd: upgrade to 1.9.3

J0WI:


community/rspamd/APKBUILD
```diff
@@ -69,9 +66,11 @@ build() {
 	make
 }
 
-package() {
-	cd "$builddir"
+check() {
+	make test
+}
```
TBK: `make test` does nothing. The make command is: `make rspamd-test` - https://github.com/rspamd/rspamd/blob/master/.circleci/config.yml#L67-L93 & luarocks would be needed as a checkdepends.

J0WI: Thanks for the hint!

TBK:
Was merged with ecba385cdafca2fd88f1c63ac71ea5eabb011e03 by @fcolista

PR #7928, opened 2019-05-18
community/openjdk7: security upgrade to 7.221.2.6.18

J0WI:


fabled:
For some reason this fails on my box with:
```
configure: error: Could not find required system headers; install the appropriate files from glibc-headers, libc6-dev and/or libattr-devel, libattr1-dev or use --disable-compile-against-syscalls if necessary.
>>> ERROR: openjdk7: build failed
```
Any ideas?

J0WI:
`attr/xattr.h` is not found but both, `musl-dev` and `linux-headers`, are present.

fabled:
Seems that header is removed in `attr-dev` in commit http://git.savannah.nongnu.org/cgit/attr.git/commit/?id=7921157890d07858d092f4003ca4c6bae9fd2c38 and users are expected tu include `sys/xattr.h` instead. I wonder if we should provide compatibility header (that just includes the correct header) or compat. symlink in `attr-dev` or perhaps even `libc-dev` or patch openjdk?

community/openjdk7/APKBUILD
```diff
@@ -191,6 +195,7 @@ build() {
 	msg "Icedtea boot"
 	make -j1 icedtea-boot SHELL=/bin/bash
 	msg "Icedtea boot done"
+	make clean-patch
```
fabled: This seems to be breaking build from clean slate.

J0WI:
Rebased.

fabled:
Committed in b548bfb613d67dfd04ddc56b7cc7c628325259da.

PR #7929, opened 2019-05-18
main/samba: replace -initscript with -openrc

tcely:


ncopa:
i squashed the commits. thanks!

PR #7930, opened 2019-05-18
community/py-setuptools_scm: upgrade to 3.3.3

dimon222:
https://github.com/pypa/setuptools_scm/blob/master/CHANGELOG.rst

v3.3.2
======

* fix 335 - fix python3.8 support and add builds for up to python3.8

v3.3.1
======

* fix 333 (regression from 198) - use a specific fallback root when calling fallbacks. Remove old
  hack that resets the root when fallback entrypoints are present.

v3.3.0
======

* fix 198 by adding the ``fallback_version`` option, which sets the version to be used when everything else fails.


dimon222:
Got merged, closing

PR #7944, opened 2019-05-19
main/polkit: upgrade to 0.116

Cogitri:


Cogitri:
It would be appreciated if someone could test main/polkit before merging this. polkit-elogind works as expected, but I can't test polkit non-elogind with my setup.

Cogitri:
fixed.

PR #7946, opened 2019-05-19
community/uacme: moved from testing, up to 1.0.12

3ju:
I have used it successfully in prod from testing


3ju:
> Please add a comment on why check is disabled, it serves as a good reference for the future

Thanks, done.

PR #7952, opened 2019-05-19
main/collectd: upgrade to collectd 5.8.1

ad-m:


main/collectd/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=collectd
-pkgver=5.8.0
-pkgrel=4
+pkgver=5.8.1
+pkgrel=0
 pkgdesc="The system statistics collection daemon"
 url="http://collectd.org"
 arch="all"
```
maxice8: GPL-2.0 is not a valid SPDX identifier, aports will detect that when spdx-licenses-list

main/collectd/APKBUILD
```diff
@@ -1,8 +1,8 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=collectd
-pkgver=5.8.0
-pkgrel=4
+pkgver=5.8.1
+pkgrel=0
 pkgdesc="The system statistics collection daemon"
 url="http://collectd.org"
```
maxice8: url= is now https://

ad-m:
> Empty variables can be removed for brevity (depends=).

Can we do it globally?

maxice8:
They are being done as the APKBUILD is touched

ad-m:
Fixed.

PR #7957, opened 2019-05-19
community/davfs2: upgrade to 1.5.5

ad-m:


TBK:
This aport needs to modernized.

- License needs to be clarified.
- depends="" can be removed
- depends_dev should either be merged with makedepends or a $pkgname-dev subpkg should be added.
- All _builddir can be removed.
- prepare() can either be removed outright or at least reduced to autoconf
- All || return 1 can be removed.
- Add check() or options="!check" # no test suite
- 010-main_code_fix.patch should be updated:
```
010-main_code_fix.patch
patching file src/cache.c
patching file src/webdav.c
Hunk #1 succeeded at 2037 (offset 4 lines). 
```

PR #7963, opened 2019-05-19
main/file: upgrade to 5.37

ad-m:


maxice8:
Please keep it all in one commit

ad-m:
@maxice8, changed to offical ( see https://web.archive.org/web/20190512033023/http://darwinsys.com/file/ ) mirror. This is the CVS archive mirror, and the release names are not optimal.

PR #7965, opened 2019-05-19
community/gegl: upgrade to 0.4.16

ad-m:


Ikke:
The upgrade has been pushed with #8408

PR #7938, opened 2019-05-19
testing/proj4: upgrade to 6.1.0

hjaekel:
testing/proj4: upgrade to 6.1.0
* rename proj4 to proj
* add check()
* add subpackage for java bindings
* add subpackage for utilities
* add subpackage for datum grids
* fixes bug #10434

testing/libgeotiff: upgrade to 1.5.1 

testing/postgis: upgrade to 2.5.2

PR #7972, opened 2019-05-20
testing/proj4: upgrade to 6.1.0

kelvinn:


maxice8:
Duplicate of #7938

kelvinn:
Thank you @maxcnunes for the review! I tried to specify just sqlite-dev, but abuild (with a -r) failed specifying that sqlite was needed, so I explicitly added it. The remaining APKBUILD is unchanged.

maxice8:
Please ignore that then, sqlite-dev is correctly depending on sqlite-libs while sqlite just holds the binary

kelvinn:
@maxice8 - squashed commits. Thanks!

andypost:
It is ok to commit it before #7938 just require to bump related packages because of soname change
```diff
-usr/lib/libproj.so.13
-usr/lib/libproj.so.13.1.1 
+usr/lib/libproj.so.15 
+usr/lib/libproj.so.15.1.0 
```
So `libgeotiff` needs bump

PR #7980, opened 2019-05-21
Wireguard move to community

clandmeter:
This PR moves wireguard to community and also merges aports which are build from same source.

- wireguard-vanilla: vanilla and virt
- wireguard-rpi: rpi and rpi2


misery:
Are you going to update this in stabel release, too? Wireguard should be used with newest snapshot only as the author tells this: https://www.wireguard.com/#work-in-progress

>  If you are packaging WireGuard, you must keep up to date with the snapshots.

PR #7984, opened 2019-05-21
testing/cockroach: bump to 19.1.1

joeholden:


TBK:
Could you add a comment on why the check() is disabled?

joeholden:
> 
> 
> Could you add a comment on why the check() is disabled?

Tests aren't run for a few reasons, mainly: they're broken anyway with release sources (they fail on glibc too), plus there are some musl related test failures but the resulting binaries work just fine.  Unfortunately they're no better now...

TBK:
Please add a short comment after options="!check" # xxxxxx

TBK:
Forgot to ask. Would it be possible to enable other arch now?

> Currently x86_64 only, requires further fixes for other architectures however a new release is due relatively soon.

https://github.com/alpinelinux/aports/pull/6564

joeholden:
Builds correctly on aarch64 also now but not 32bit (looks like its probably still rocksdb that is broken on 32bit architectures)

TBK:
Ok, I will merge it now.

PR #7992, opened 2019-05-21
testing/kde-frameworks-tier1: new aports

PureTryOut:
This PR introduces [KDE Frameworks tier 1](https://api.kde.org/frameworks/index.html).

This is basically a retry of #2495 :wink: 

Note that KDE Frameworks has no stable or unstable releases, but new releases are done every month. Once this is merged I'll introduce the other tiers.

The goal is to eventually fully package Plasma Desktop.

### DIFF
--- a/community/extra-cmake-modules/APKBUILD
+++ b/community/extra-cmake-modules/APKBUILD
@@ -1,5 +1,6 @@
 # Contributor: k0r10n <k0r10n.dev@gmail.com>
 # Contributor: Ivan Tham <pickfire@riseup.net>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
 pkgname=extra-cmake-modules
 pkgver=5.58.0
 pkgrel=1
@@ -11,29 +12,24 @@ depends="cmake"
 makedepends="py-sphinx"
 checkdepends="qt5-qtbase-dev qt5-qtquickcontrols2-dev"
 subpackages="$pkgname-doc"
-source="http://download.kde.org/stable/frameworks/${pkgver%.*}/$pkgname-$pkgver.tar.xz
-	disable-KDEFetchTranslations.patch
-	"
-
-prepare() {
-	default_prepare
-	mkdir -p build
-}
+source="http://download.kde.org/stable/frameworks/${pkgver%.*}/$pkgname-$pkgver.tar.xz"
 
 build() {
-	cd "$builddir"/build
-	cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_TESTING=ON
+	cmake \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DBUILD_TESTING=OFF
 	make
 }
 
 check() {
-	cd "$builddir"/build
 	make test
 }
 
 package() {
-	cd "$builddir"/build
-	make DESTDIR="$pkgdir" install/fast
+	DESTDIR="$pkgdir" make install/fast
+
+	install -Dm644 COPYING-CMAKE-SCRIPTS \
+		"$pkgdir"/usr/share/licenses/$pkgname/COPYING
 }
 
 sha512sums="edff7bcef7effad524743cf234ff6bb41ab59b7d8e4629ee11be9faea97985f32fff0d96cfaf25277acce9806abea3abe475c9dfe2f8c2dc1b61b7f52d2396bb  extra-cmake-modules-5.58.0.tar.xz
--- a/testing/bluez-qt/APKBUILD
+++ b/testing/bluez-qt/APKBUILD
@@ -1,0 +1,28 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=bluez-qt
+pkgver=5.58.0
+pkgrel=0
+arch="all"
+pkgdesc="Qt wrapper for Bluez 5 DBus API"
+url="https://community.kde.org/Frameworks"
+license="LGPL-2.1-or-later"
+depends_dev="qt5-qtbase-dev qt5-qtdeclarative-dev"
+makedepends="$depends_dev extra-cmake-modules doxygen graphviz qt5-qttools-dev"
+source="https://download.kde.org/stable/frameworks/${pkgver%.*}/$pkgname-$pkgver.tar.xz"
+subpackages="$pkgname-dev $pkgname-doc"
+options="!check" # Multiple tests either hang or fail completely
+
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=RelWithDebugInfo \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=lib \
+		-DBUILD_QCH=ON
+	make
+}
+
+package() {
+	DESTDIR="$pkgdir" make install
+}
+sha512sums="c6447bc8464876f91d8a9e2a0bd545791fa627ae94c47990ae58df7ac29ea9bf932482ca0ff94e84946950ddd95c9784d53ac6fd25e98901267d5267251d0d6f  bluez-qt-5.58.0.tar.xz"
--- a/testing/breeze-icons/APKBUILD
+++ b/testing/breeze-icons/APKBUILD
@@ -1,0 +1,28 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=breeze-icons
+pkgver=5.58.0
+pkgrel=0
+pkgdesc="Breeze icon themes"
+arch="noarch"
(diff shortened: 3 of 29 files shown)


maxice8:
There is community/attica already

PureTryOut:
You want _30 individual commits_? Uhm, ok I guess. Not my preference but sure.

maxice8:
> You want 30 individual commits? 

Easier to review 30 small commits than one massive one.

> Uhm, ok I guess. Not my preference but sure.

Do what you prefer i'm not a developer just an outside-the-project contributor

yeah yeah i read it wrong, ignore that part.

PureTryOut:
There you go, 30 individual commits.

Although I personally prefer `cd`'ing to `$builddir` in each function for clarity, even if not needed, I removed them. Now only some packages that require out of source builds have a `cd` to it in the `prepare()` function.

maxice8:
> There you go, 30 individual commits.

Thank you

> Although I personally prefer cd'ing to $builddir in each function for clarity [...]

Thank you

maxice8:
There are still SPDX violations in lots of packages (see first review)

PureTryOut:
> There are still SPDX violations in lots of packages (see first review)

Not true. Like I linked, LGPL-2.1 and LGPL-3.0 are valid SPDX.

maxice8:
```
$ grep GPL-3.0 /usr/share/spdx/license.lst
AGPL-3.0-only
AGPL-3.0-or-later
GPL-3.0-only
GPL-3.0-or-later
LGPL-3.0-only
LGPL-3.0-or-later
deprecated_AGPL-3.0
deprecated_GPL-3.0+
deprecated_GPL-3.0-with-GCC-exception
deprecated_GPL-3.0-with-autoconf-exception
deprecated_GPL-3.0
deprecated_LGPL-3.0+
deprecated_LGPL-3.0
```

Please isntall the spdx-licenses-list package so abuild will warn when the package is built that you're using a non-spdx compliant license identifier in your license= field.

PureTryOut:
Well that's annoying, that means the website is out of date :unamused: 

Updated the licenses, it should be correct now.

maxice8:
It isn't, there is a big red deprecated warning on the links you sent.

PureTryOut:
You're right, I misunderstood that message. I thought it meant that the LGPL-2.1 license was deprecated for the LGPL-3.0 license, and for the latter I just assumed it was a mistake. It would've been nice if they linked the new version from there.

Anyway, everything should be correct now.

maxice8:
On Tue, May 21, 2019 at 04:36:24PM -0700, Bart Ribbers wrote:
> You're right, I misunderstood that message. I thought it meant that the LGPL-2.1 license was deprecated for the LGPL-3.0 license, and for the latter I just assumed it was a mistake. It would've been nice if they linked the new version from there.
> 
> Anyway, everything should be correct now.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/7992#issuecomment-494597667

Looking at the source code header declarations most either declare

LGPL-2.0-or-later

or they declare

LGPL-2.1-only OR LGPL-3.0-or-later

**OR is the SPDX identifier for an optional license**

regardless the COPYING.LIB of all of the ones i checked (most of them) has LGPL-2.1 (Lesser).

So the license field on the packages that define can be be changed to 'LGPL-2.1-only OR LGPL-3.0-only' but it is your call to keep them or change them i didn't dig enough into the complex conditional licensing of Qt to make a definitive call.

PureTryOut:
From all the things this PR could improve on, I never expected it to be licenses... Fixed the BSD one.

All the licenses listed are correct now. This licensing is getting on my nerves, so let's hope this is the final thing.

maxice8:
It is just KDE with its complex licensing

PureTryOut:
For some reason the CI fails on a hanging test in bluez-qt, but I can not reproduce this locally so I'm not sure what to do about this.

Also, @awilfox could you review this too like last time?

awilfox:
The first one I looked at, kcoreaddons, was already incorrect:

```sh
license="LGPL-2.0+ AND (LGPL-2.1-only OR LGPL-3.0-only)"
```

is what we determined in Adélie.  But I suppose that doesn't matter so much.  It also should `depends_dev` on `qt5-qtbase-dev`.  Otherwise it seemed ok.  I really don't have time to review all 30 frameworks here, sorry.

awilfox:
And to save you some time on tier2 licensing and dependencies,

```sh
# Maintainer: A. Wilcox <awilfox@adelielinux.org>
pkgname=kdelibs4support
pkgver=5.54.0
pkgrel=0
pkgdesc="Support library for KDE 4 software"
url="https://www.kde.org/"
arch="all"
options="!check suid"  # Test requires running X11 session.
license="LGPL-2.1+ AND MIT AND LGPL-2.1-only AND LGPL-2.0-only AND (LGPL-2.1-only OR LGPL-3.0-only) AND (LGPL-2.0-only OR LGPL-3.0-only)"
depends=""
depends_dev="qt5-qtbase-dev qt5-qtsvg-dev qt5-qttools-dev kcompletion-dev
        kconfigwidgets-dev kcrash-dev kdesignerplugin kdesignerplugin-dev
        kemoticons-dev kparts-dev kunitconversion-dev kinit kded kded-dev
        kitemmodels-dev knotifications-dev sonnet-dev threadweaver-dev
        kinit-dev"
makedepends="$depends_dev cmake extra-cmake-modules"
```

PureTryOut:
I'm interested to know how you determine those licenses then. Looking at the source code for `kcoreaddons`, there are 3 license files in there. There is a LGPL-2.0 one, which doesn't specify that any later version could be used. Then there is a LGPL-2.1 license which doesn't seem to specify that either, and lastly there is the BSD-3-Clause license for CMake scripts. According to you it uses the LGPL-2.0 license or any later version, and either only LGPL-2.1 or only LGPL-3.0? I'm not sure where you got the "or later" one from or LGPL-3.0?

awilfox:
You can't trust the `LICENSE` files in KDE packages.  You have to audit the source files themselves.

In Adélie we've packaged `licensecheck` which can help a lot, but for full disclosure, the entirety of KDE was packaged by me and for every package I did `vim src/**/*.c{,pp,xx}` and looked at the headers myself.

PureTryOut:
I just asked about this on [#kde:kde.org](https://matrix.to/#/#kde:kde.org), and that statement seems wrong. As I was explained, the license headers in each source file applies when somebody copies the code into a project of theirs. If however the library is just linked too, then the "overall" (the LICENSE and COPYING files in the root of the source) applies.

As a distro you don't just copy bits of code, but you use the entire library for other packages to link to. In this case it means the overall license applies.

Aerdan:
> I just asked about this on [#kde:kde.org](https://matrix.to/#/#kde:kde.org), and that statement seems wrong. As I was explained, the license headers in each source file applies when somebody copies the code into a project of theirs. If however the library is just linked too, then the "overall" (the LICENSE and COPYING files in the root of the source) applies.

This is incorrect. Libraries and programs are licensed under the terms of all of the licenses present in the codebase, not the 'overall' license, and the GPL is explicitly designed around this—otherwise it would be possible to copy GPL code into a BSD-licensed project and lose the GPL licensing on that code.

awilfox:
What thiago is saying in the attached chat log is that you *are* able to "reduce" the set of (L)GPL versions down to a common one shared by all files.  So, in the `ki18n` example, that means the entire distribution can be said as *either* LGPL-2.1-only *or* LGPL-3.0+, because that is the [intersection](https://en.wikipedia.org/wiki/Intersection_(set_theory)) of {LGPL-2.0+} ∩ {LGPL-2.1-only, LGPL-3.0+}.

In some KDE packages, there *are* no common members between sets, and you need to have more complex definitions.  In Adélie we allow the users to know exactly the license they are able to use the software under (more precise and verbose).  But this is why I said above:

> I suppose that doesn't matter so much.

In the case I outlined (`kcoreaddons`), they're all just LGPL versions, and they all match with 2.1.  It's just inexact.  And that's fine for some, maybe even most KDE packages.  But definitely not all.

awilfox:
This also means that, for instance, you could simplify the `kdelibs4support` example above from:

```
license="LGPL-2.1+ AND MIT AND LGPL-2.1-only AND LGPL-2.0-only AND (LGPL-2.1-only OR LGPL-3.0-only) AND (LGPL-2.0-only OR LGPL-3.0-only)"
```

to

```
license="LGPL-2.0-only AND LGPL-2.1-only AND LGPL-3.0-only AND MIT"
```

(because LGPL-2.1+ can be used as LGPL-2.1-only)

PureTryOut:
Well, I hate licensing now... This stuff is confusing and hard to determine.

I made a pass at licensing of all the packages, checked it with the source and LICENSE files. I'm sure there is still stuff wrong, but at least it should be _better_ now. I also added `qt5-qtbase-dev` to `$depends_dev` for `kcoreaddons`.

PureTryOut:
I've just cleaned-up `extra-cmake-modules` a bit (out of source build was not required) and disabled tests for `bluez-qt` completely as they either hang or fail completely.

PR #8000, opened 2019-05-22
community/fzf: add vim help files and create tags

maxwell-k:
install -D creates the directories so removed mkdir:

$ install --help 2>&1 | grep -- -D
        -D      Create leading target directories


Ikke:
Congrats with PR 8000 ;-)

PR #8003, opened 2019-05-22
 main/daq: split static and sfbpf libs

kazkansouh:
`snort` is the only consumer of this package. By default it uses a mix of static and dynamic libraries provided by this package. This PR reduces the install size of `daq` by ensuring that unused dynamic libs are not installed for run time.

Requires a tweak to `snort`'s make dependencies to ensure it continues to compile as before. I included in this PR instead of its own to ensure that nothing is broken.

### DIFF
--- a/main/daq/APKBUILD
+++ b/main/daq/APKBUILD
@@ -1,21 +1,17 @@
-# Contributor: Natanael Copa <ncopa@alpinelinux.org>
+# Contributor: Karim Kanso <kaz.kanso@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=daq
 pkgver=2.0.6
-pkgrel=2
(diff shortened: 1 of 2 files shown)


kazkansouh:
Thanks, I've cleaned up those issues. Also main/snort/APKBUILD has builddir issue to that has been removed.

kazkansouh:
Oh, my bad. I remember doing this on a new port in testing. Just assumed it was same here too :)

maxice8:
On Wed, May 22, 2019 at 11:20:59AM -0700, Karim Kanso wrote:
> Oh, my bad. I remember doing this on a new port in testing. Just assumed it was same here too :)
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8003#issuecomment-494910998

It is done for packages in testing/ and community/

PR #8010, opened 2019-05-23
testing/perl-locale-codes: new aport

timlegge:
Locale::Codes will be removed from the Perl core distribution in the next major release. Please install it from CPAN.
Locale::Country will be removed from the Perl core distribution in the next major release. Please install it from CPAN
Locale::Language will be removed from the Perl core distribution in the next major release. Please install it from CPAN.


timlegge:
Any update.  As I think of this because of the changes coming in perl at some point this should eventually make its was to atleast community and maybe main

PR #8014, opened 2019-05-23
community/rcon: move from testing

TBK:


TBK:
As I recall ncopa mentioned at one point that changes to the license does not warrant a pkgrel bump.

hmm we need docs on when to bump pkgrel.

maxice8:
On Thu, May 23, 2019 at 05:40:54AM -0700, TBK wrote:
> As I recall ncopa mentioned at one point that changes to the license does not warrant a pkgrel bump.
> 
> hmm we need docs on when to bump pkgrel.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8014#issuecomment-495201740

Won't that prevent the users from recieving an updated package that contains the correct license ?

Ikke:
The revbump is required to update the metadata in the package.

TBK:
> The revbump is required to update the metadata in the package.

Fixed.

PR #8020, opened 2019-05-23
community/php7-pecl-vips & community/vips: move from testing 

TBK:
@wjordan Is it alright with you to move vips to community?


wjordan:
@TBK sounds good to me :+1:

PR #8023, opened 2019-05-23
community/rust: upgrade to 1.33.0

Cogitri:
@Ikke

### DIFF
--- a/community/rust/APKBUILD
+++ b/community/rust/APKBUILD
@@ -1,11 +1,12 @@
+# Contributor: Rasmus Thomsen <oss@cogitri.dev>
 # Contributor: Jakub Jirutka <jakub@jirutka.cz>
 # Contributor: Shiz <hi@shiz.me>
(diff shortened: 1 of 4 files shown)


Cogitri:
I'd very much like to take maintainership on this, but I think that can wait until I do the 1.35 bump, let's get this tested&merged and sort of sprint to 1.35 if possible :)

Cogitri:
Currently building Rust 1.34.2 with this.

Cogitri:
FWIW, I've added Rust 1.34.2 to this PR too.

Cogitri:
Both Rusts build fine for me, pass the tests and are able to build my fairly complex app (gxi).

Ikke:
Sadly, this cannot go in a single PR. Only the end result is built, not individual commits, so it would try to build 1.34 with 1.32, which apparently would not work. So we have to split this into two separate pull requests.

Cogitri:
Fixed, thanks for the info :)

PR #8027, opened 2019-05-23
testing/libfort new aport

pdietl:


pdietl:
mkdir build in each function or in prepare? and then just cd to "$builddir/build" in each function?

testing/libfort/APKBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Pete Dietl <petedietl@gmail.com>
+pkgname=libfort
+pkgver=0.1.3
+pkgrel=0
+pkgdesc="A simple crossplatform library to create formatted text tables."
+url="https://github.com/seleznevae/libfort"
+arch="all"
+license="MIT"
+makedepends="cmake binutils-gold"
```
TBK: Any particular reason for requiring binutils-gold?

testing/libfort/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+source="$pkgname-$pkgver.tar.gz::https://codeload.github.com/seleznevae/libfort/tar.gz/v$pkgver"
+builddir="$srcdir/$pkgname-$pkgver"
+
+prepare() {
+	default_prepare
+	mkdir -p "$builddir/build"
+}
+
+build() {
+	cd "$builddir/build"
```
TBK: This is currently the de facto way of making and cd'ing into build dir.
```suggestion
	mkdir build && cd build
```

pdietl: hmm should I assume that `$builddir` is already created and that I am in that directory?
I was trying to have a build directory in a build directory to do an out-of-source build. I'm not sure what the best way to do that is.

testing/libfort/APKBUILD
pdietl: I will get rid of this dependency. The upstream author said:

> Hello, I noticed in https://github.com/alpinelinux/aports/pull/8027/files that you have binutils-gold. I then noticed that for all gcc builds I have option -fuse-ld=gold. Is that the reason why you have to add it or something else?
Anyway, as far as I remember I think that I added it because there were some bugs with sanitizers with some gcc versions. Therefore in develop (after the last release) I made some changes - now I add -fuse-gold only for special builds (sanitizers ...).
Not sure this is important. But in case guys from alpine linux say something about binutils-gold maybe it can be avoided for ordinary builds.

So I think he will give me a version 0.1.4.

testing/libfort/APKBUILD
TBK: builddir is the directory where source is extracted. So build is at the same level as the src files.
It is the common way to doing and "out-of-source build".

TBK:
@pdietl If I can get you to squash the commits into one then this should be good to go.

testing/libfort/APKBUILD
```diff
@@ -1,28 +1,22 @@
...
 
 build() {
-	cd "$builddir/build"
+	mkdir build && cd build
 	if [ "$CBUILD" != "$CHOST" ]; then
 		CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
 	fi
 
-	cmake .. \
+	cmake "$builddir" \
```
TBK: `cmake "$builddir" \`

is the same as
`cmake .. \`

pdietl: TRU

pdietl:
@TBK Ok I've done the squash. :)

pdietl:
@TBK What's the time frame / mechanism for getting this package moved out of testing? Is it automatic or do I have to make another PR after some time?

TBK:
There isn't a defined procedure atm.

Testing is intended to provide a opportunity to battle test the aport so possible bugs/packaging flaws can be caught, maturing the aport and then moving it to community.

It is up to the maintainer to push for the move. Unless it is a dependency for another aport in which "it might get dragged along". So you have to made another PR after "some later time".

PR #8035, opened 2019-05-24
testing/weston: enable logind support via elogind

PureTryOut:
This allows launching Weston by any user when elogind is installed. Old methods of launching Weston still work as before.

### DIFF
--- a/testing/weston/APKBUILD
+++ b/testing/weston/APKBUILD
@@ -2,19 +2,19 @@
 # Maintainer: Valery Kartel <valery.kartel@gmail.com>
 pkgname=weston
 pkgver=6.0.0
-pkgrel=0
+pkgrel=1
 _libname=lib$pkgname
 _libdir=$_libname-${pkgver%%.*}
 pkgdesc="The reference Wayland server"
 url="https://wayland.freedesktop.org/"
 arch="all !s390x !ppc64le"
 license="MIT"
-depends=""
 makedepends="wayland-protocols libxkbcommon-dev xkeyboard-config
 	libinput-dev libunwind-dev mtdev-dev libxcursor-dev glu-dev
 	pango-dev colord-dev libwebp-dev libva-dev dbus-dev
 	linux-pam-dev wayland-dev libevdev-dev libjpeg-turbo-dev
 	freerdp-dev lcms2-dev gstreamer-dev gst-plugins-base-dev meson
+	elogind-dev
 	"
 _cms="cms-colord cms-static"
 _shell="shell-desktop shell-fullscreen shell-ivi"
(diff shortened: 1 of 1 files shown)


PureTryOut:
I mean, really? I didn't touch anything of that, isn't it better for someone to go over all the existing APKBUILDs with some script and fix all of that automatically?

maxice8:
Yes, i review both commits and the files that were A(dded) and M(odified) according to git diff-tree, which is why it says For testing/weston/APKBUILD and not for <commit-hash>

PureTryOut:
Well I made the changes for now, but the point of reviewing a PR is to review _the changes made_, not the entire file. I didn't add those things you mentioned so it doesn't belong in this PR.

testing/weston/APKBUILD
```diff
@@ -2,20 +2,19 @@
...
 arch="all !s390x !ppc64le"
 license="MIT"
-depends=""
 makedepends="wayland-protocols libxkbcommon-dev xkeyboard-config
 	libinput-dev libunwind-dev mtdev-dev libxcursor-dev glu-dev
 	pango-dev colord-dev libwebp-dev libva-dev dbus-dev
 	linux-pam-dev wayland-dev libevdev-dev libjpeg-turbo-dev
 	freerdp-dev lcms2-dev gstreamer-dev gst-plugins-base-dev meson
-	"
+	elogind-dev"
```
tcely:
```suggestion
	elogind-dev
	"
```

maxice8:
Failed to build here

PureTryOut:
Uhm, ok... With what error? It built fine in CI and still builds fine locally for me using `docker-abuild`.

maxice8:
hmm, i sent an image, GitHub decided to do away with it. I'll go build it again

maxice8:
http://ix.io/1L7M

PureTryOut:
Interesting, I can't reproduce that error at all. Quick ducking suggests a wrong order of modules while linking, but that doesn't seem to be the case here.

I guess we could let CI run through it again to make sure it's a problem specific to you.

maxice8:
On Fri, Jun 07, 2019 at 03:29:19AM -0700, Bart Ribbers wrote:
> Interesting, I can't reproduce that error at all. Quick ducking suggests a wrong order of modules while linking, but that doesn't seem to be the case here.
> 
> I guess we could let CI run through it again to make sure it's a problem specific to you.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8035#issuecomment-499836718

I'll disable parallel building and see if it fixes

maxice8:
nope, still fails with same error even with JOBS=1

PR #8038, opened 2019-05-24
main/fish: modernize

maxice8:


Cogitri:
While you're at it: Maybe you could fix the broken deps on fish-tools I've introduced? One can use mdoc instead of man-db

maxice8:
On Sat, May 25, 2019 at 04:49:33AM -0700, Rasmus Thomsen wrote:
> While you're at it: Maybe you could fix the broken deps on fish-tools I've introduced? One can use mdoc instead of man-db
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8038#issuecomment-495911847

Done

Cogitri:
Ah, but that will break it for people who use man-db instead of mdoc. Maybe we can depend on `cmd:man`? Not quite sure if that's really a good option though.

tcely:
Be sure to show results from `apk -U add fish-tools`

maxice8:
WFM, was able to run their generate_manpages python script

PR #8039, opened 2019-05-24
testing/hyperfine: new aport

maxice8:


testing/hyperfine/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+build() {
+	cargo build --release --verbose
+}
+
+check() {
+	cargo test all --release --verbose
+}
+
+package() {
+	cargo install --path . --root="$pkgdir"/usr
```
tcely: Using root in /usr seems odd here. I am not aware of a better way just now, but would welcome one.

PR #8041, opened 2019-05-25
main/: modernize

maxice8:


ncopa:
`main/partimage` needs to bump pkgrel due to license string changed. I would also prefer to bump pkgrel when `makedepends` changes, just to be sure

maxice8:
will bump main/partimage but there is ABSOLUTELY no reason to bump when makedepends change in these circustances since i am only removing DUPLICATE definitions. See: https://github.com/maxice8/atools/blob/master/alint.5.scd#L85 and all the duplicate-* tags as well.

ncopa:
> will bump main/partimage but there is ABSOLUTELY no reason to bump when makedepends change in these circustances since i am only removing DUPLICATE definitions.

It would be useful to have that mentioned in commit message.

PR #8048, opened 2019-05-25
community/network-manager-applet: modernize

maxice8:
@Cogitri want to adopt ? it is needed by gnome-control-center


Cogitri:
These changes and more are already in #7379 :/

PR #8056, opened 2019-05-25
unmaintained/compton-conf: move from testing

maxice8:


ncopa:
this needs to be rebased

PR #8064, opened 2019-05-25
main/memcached: upgrade to 1.5.16

maxice8:
Fixes critical potential segfault/memory corruption bug in 1.5.15 when storing items with client flags of "0". Bug only exists in 1.5.15, and the fix is the only change in this release.

### DIFF
--- a/main/memcached/APKBUILD
+++ b/main/memcached/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor: Jeff Bilyk <jbilyk@alpinelinux.org>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=memcached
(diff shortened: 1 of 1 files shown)


maxice8:
@ncopa

tcely:
@maxice8 The message is unclear. Which version had the bug and which fixes it?

maxice8:
On Sun, May 26, 2019 at 05:59:00AM -0700, tcely wrote:
> @maxice8 The message is unclear. Which version had the bug and which fixes it?
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/alpinelinux/aports/pull/8064#issuecomment-495997948

This was meant as an update to 1.15.16, sorry

PR #8068, opened 2019-05-25
gawk5 fixes for libgpg-error 1.36

djazo:
Taken from 
https://github.com/alpinelinux/aports/commit/f19e074bc6ed6e3be8a13869d0e94dce1f4cad56


tcely:
A better commit message would be:

`main/libgpg-error: gawk5 fixes for 1.36`

ncopa:
`pkgrel` needs to be bumped or builder will not pick it up. I can fix that for you. Thanks!

tcely:
@djazo what is the reason this patch is needed.

It only renames a variable and unescapes '#' from what I can tell.

djazo:
> @djazo what is the reason this patch is needed?
> 
> It only renames a variable and unescapes '#' from what I can tell.

Without unescaping, build fails using gawk5, which is now at least on edge.

tcely:
I can't find a reason for gawk to be installed while this package is building.

See https://build.alpinelinux.org/buildlogs/build-edge-x86_64/main/libgpg-error/libgpg-error-1.36-r0.log

maxice8:
Your PR seems to be missing the patch and has merge commits

tcely:
This was merged by @ncopa in https://github.com/alpinelinux/aports/commit/3497c40f75c25eabecb8252764a4ce1fe3fd9494#diff-4ddc72d2e139bc373dd40f47d7f53011.

PR #8072, opened 2019-05-25
community/sassc: upgrade to 3.6.0

maxice8:


TBK:
https://github.com/sass/sassc/blob/master/docs/testing/unix-instructions.md

maxice8:
That only works if you have a local copy of the libsass repo

PR #8097, opened 2019-05-26
testing/tsung: modernize

maxice8:


maxice8:
@okeuday

okeuday:
@maxice8 Hi, thanks for updating the package.  I haven't had a chance to update the tsung source code and use a newer Erlang/OTP version.  There was a problem when I tried using the package last time, issues were reported though.  Please feel free to commit the change.

PR #8100, opened 2019-05-26
testing/py3-us: modernize

maxice8:


maxice8:
@fabaff

testing/py3-us/APKBUILD
```diff
@@ -11,18 +11,15 @@ license="BSD-3-Clause"
 replaces="py-us" # for backwards compatibility
 provides="py-us=$pkgver-r$pkgrel" # for backwards compatibility
 depends="py3-jellyfish py3-setuptools python3"
```
TBK: Looks incorrect to me. The makedepends should remain, the overlap should be removed from depends.

Would also be great if check was added - https://github.com/unitedstates/python-us/tree/master/tests

maxice8: No, assuming dependency distribution was correct before then you should ever remove from makedepends when against depends.

TBK: Forgot to paste link for reqs - https://github.com/unitedstates/python-us/blob/master/requirements.txt

maxice8:
On Tue, May 28, 2019 at 01:42:27PM -0700, TBK wrote:
> TBK commented on this pull request.
> 
> 
> 
> > @@ -11,18 +11,15 @@ license="BSD-3-Clause"
>  replaces="py-us" # for backwards compatibility
>  provides="py-us=$pkgver-r$pkgrel" # for backwards compatibility
>  depends="py3-jellyfish py3-setuptools python3"
> 
> Forgot to paste link for reqs - https://github.com/unitedstates/python-us/blob/master/requirements.txt
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8100#discussion_r288293435

py3-setuptools is required because /usr/bin/states provided by the package imports pkg_resources and python3 is required because there is a script (/usr/bin/states) that has a shebang on python3

PR #8104, opened 2019-05-26
testing/perl-www-mechanize: modernize

maxice8:


maxice8:
@timlegge

maxice8:
Is it really necessary ? it is just the removal of duplicate dependencies so there is no way the build changes

timlegge:
true I guess but the package is not built with the latest APKBUILD version (obviously no code changes).  You know the rules of the project far better that I so if you believe its fine without incrementing the pkgrel am fine with it.

PR #8115, opened 2019-05-26
testing/kubernetes: modernize

maxice8:


maxice8:
@fcolista

fcolista:
What's the point of adding check() if the check is ignored due to  options="!check"  ?

maxice8:
So in a future version you don't need to add it once someone fixes the tests, it is already there.

fcolista:
It might happen, or not. Better to add the check when is actually used.
Avoids confusion.
Thanks.

maxice8:
On Sun, May 26, 2019 at 11:25:07PM -0700, Francesco Colista wrote:
> It might happen, or not. Better to add the check when is actually used.
> Avoids confusion.
> Thanks.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8115#issuecomment-496093558
Removed

PR #8156, opened 2019-05-26
unmaintained/poco: move from testing

maxice8:


rnalrd:
poco is a build dependency of https://www.collaboraoffice.com/collabora-online/

It's still a WIP which I had no time to dedicate to lately. If none minds it, I'd like to keep it. I've merged the upgrade.

Thanks

PR #8166, opened 2019-05-26
Work on removing gstreamer-1.0

maxice8:


ncopa:
is this ready to merge?

maxice8:
This is as far as i can go at the moment, some packages still need wxgtk2.8 which depends on gst0.10. But this can be done at a later date, due to my notebook being in repairs i am completely immobilized from contributing except in reviews

maxice8:
So yes

PR #8171, opened 2019-05-27
community/lld: disable test eh-frame-hdr-augmentation on armhf & armv7

TBK:


TBK:
pfffffffffffffssssssssssssssssssssssssssssssssssssss llvm issue tracking what a mess :sob:

Ikke:
@TBK Did you verify that it's the tests that are broken? (just to be sure)

TBK:
ELF/eh-frame-hdr-augmentation.s is broken on armv7 (https://build.alpinelinux.org/buildlogs/build-3-10-armv7/community/lld/lld-7.0.1-r0.log). I tried to patch the issue but I could not figure out LLVM svn/git reference system so I gave up in the end.

armhf I do not know. It was disabled in 2017 https://github.com/alpinelinux/aports/commit/88b05a1a15de851dd517e7f5014ecad4514665c1#diff-458e5545a50ba561f5a3275d395e8dd9.

I will enable and see if it passes.

TBK:
Both armhf & armv7 fails due to the fact that ELF/eh-frame-hdr-augmentation.s is broken in 7.x releases - https://cloud.drone.io/alpinelinux/aports/5532

TBK:
After some digging. We can apparently just delete the test file and that disables it. So I have done that for the two arm arches.

PR #8195, opened 2019-05-28
testing/log4cplus: Enable implicit initialization to preserve backwards compatibility

valentinesd:
Such as https://gitlab.isc.org/isc-projects/kea/issues/625


tcely:
See #8011

tcely:
We want to merge the patched kea rather than breaking this package too.

PR #8197, opened 2019-05-28
community & testing: remove extranious $builddir

5paceToast:
This was performed in several steps:
1. An awk script was written such that it removed lines that:
1.1 Come after `build()`, `prepare()`, `check()` or `package()`.
1.2 Match any of the 4 variations (with/without `""` and `{}`) of `cd
  $builddir`.
1.3 The line ends with that.
2. The git diff was examined - no false positives were found.
3. Some functions became invalid, since exhibit A is an error. To work
   around this, a dummy `:` (no-op) was added.
4. One package was so old, it triggered the pre-commit hook (it was
   still using `md5sums`). It was modified to use `sha512sums`.

4 is a bit unfortunate, since it isn't exactly a related change.

Exhibit A:
```shell
function() {
}
```

Note on exhibit A: if there is a comment, it is ignored, and remains
invalid.


5paceToast:
Note: CI will obviously fail (this modifies *many* packages; drone will likely fail due to the sheer amount of difficulty to find ancestry; if travis manages to wade through (uncertain), he will *definitely* run out of build time (2237 packages have been modified!))

5paceToast:
The goal here is to severely reduce the amount of work there is to do in the modernization effort.
Since it's a huge amount of changes, I wanted to be 100% sure there would be no false positives - missing *some* is fine.
Clearly, I've hit quite a few anyways.

Regarding the empty lines, I don't think it's a very big deal - someone will get annoyed eventually and remove them, but there's no outstanding policy saying extraneous whitespace need be removed.

maxice8:
On Mon, May 27, 2019 at 08:57:13PM -0700, Chloe Kudryavtsev wrote:
> The goal here is to severely reduce the amount of work there is to do in the modernization effort.
> Since it's a huge amount of changes, I wanted to be 100% sure there would be no false positives - missing *some* is fine.
> Clearly, I've hit quite a few anyways.

Why not just use apkbuild-lint that also catches the other ones ? false-positives can be checked manually like it must be done anyways

> 
> Regarding the empty lines, I don't think it's a very big deal - someone will get annoyed eventually and remove them, but there's no outstanding policy saying extraneous whitespace need be removed.
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8197#issuecomment-496356271

Shouldn't be too hard to remove the empty lines after removing the cd "$builddir"

5paceToast:
> Why not just use apkbuild-lint that also catches the other ones ? false-positives can be checked manually like it must be done anyways

apkbuild-lint cannot automatically fixup failures.
Editing 2237 files by hand is wholly impractical.

> Shouldn't be too hard to remove the empty lines after removing the cd "$builddir"

I just don't really care.
If it is indeed not very hard, you are welcome to submit an additional patch.

maxice8:
On Mon, May 27, 2019 at 10:11:06PM -0700, Chloe Kudryavtsev wrote:
> > Why not just use apkbuild-lint that also catches the other ones ? false-positives can be checked manually like it must be done anyways
> 
> apkbuild-lint cannot automatically fixup failures.
> Editing 2237 files by hand is wholly impractical.
> 

it gives you specific lines on every match.

ncopa:
I think it would be good to wait with this til after v3.10

5paceToast:
Now that 3.10 is out, do we want to take this approach?
Alternatively, should I (re)publish the awk script so someone with push access can edit (and apply it) to their level of satisfaction?

maxice8:
Can just use apkbuild-fixer from atools to do it mostly properly

5paceToast:
Well the same question remains.
Want to open a PR with a non-main/ run of that?
I'd consider this superseded in that case, and check for any breakages.

maxice8:
> Want to open a PR with a non-main/ run of that?

Yeah sure

5paceToast:
Superseded by #8913.

PR #8198, opened 2019-05-28
community/spice-gtk: upgrade to 0.37

maxice8:
@ncopa


PureTryOut:
Since spice-dev can now not be installed at all, could this PR get some more priority?

PR #8199, opened 2019-05-28
main/celt051: remove

maxice8:


fcolista:
Why this needs to be removed?

maxice8:
On Mon, May 27, 2019 at 11:49:19PM -0700, Francesco Colista wrote:
> Why this needs to be removed?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8199#issuecomment-496387207

Not need but it is a completed experiemnt with no more upstream support with only spice-gtk as its user (which has a PR for switching it over to opus). Void removed in 2015 and Debian in 2012

ncopa:
It would be nice to have a short note in the commit message about why it is removed.

PR #8201, opened 2019-05-28
main/lighttpd: upgrade to 1.4.54

maxice8:
@ncopa


fcolista:
Why kerberos5 has been disabled here?

maxice8:
On Mon, May 27, 2019 at 11:49:44PM -0700, Francesco Colista wrote:
> Why kerberos5 has been disabled here?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8201#issuecomment-496387309

It wasn't, the switch is not recognized so there is no point in having it there since it does nothing.

PR #8203, opened 2019-05-28
main/zfs: clarify license

TBK:


fcolista:
There's no need to bump pkgrel for this modify.
Thanks!

ncopa:
> There's no need to bump pkgrel for this modify.
> Thanks!

It actually is neccessary to bump `pkgrel`. Otherwise will not package get rebuilt and `apk info --license zfs` will show the old license.

TBK:
@ncopa Could you check if the following is correct? and if there are other other conditions could you mention them?

I would like to improve the docs on pkgrel since there is quite a bit of confusion going around when to bump or not.

Increment pkgrel when:
- Changing meta data for a aport which should be propagated to the distributed apk. Such as license...
- Applying patches.
- Changing the build flags/conditions which warrens a rebuild.

When not to increment pkgrel:
- Moving an aport from one repo to another.
- When enabling previously deactivated architectures (arch).

Decrement pkgrel when:
- Upgrading (pkgver increment) an aport. In this case the pkgrel should be reset to 0.

PR #8209, opened 2019-05-28
testing/efl: Update to EFL-1.22.2

billiob:


Ikke:
Build fails with:

>../src/bin/eolian/eolian_gen: exec: line 117: /home/buildozer/aports/testing/efl/src/efl-1.22.2/src/bin/eolian/.libs/lt-eolian_gen: Invalid argument

asriel-danctnix:
How about enable support for NEON, and if the architecture is armhf then disable it? (Line 51)
That can be done by doing something like this:
```
case "${CARCH}" in
armhf*)	DISABLE_NEON="--disable-neon" ;;
esac
```

Also the configuration should be changed if you want to add more support to Enlightenment and other EFL programs.

PR #8213, opened 2019-05-28
main/openssl: upgrade to 1.1.1c

J0WI:


fabled:
Applied in bbbb00832cd1c236a513152b25b725eee42b9b10 and added missing `pkgrel` reset to zero.

PR #8225, opened 2019-05-29
testing/libbpf: use new version scheme

acj:
libbpf now has tagged releases and a version that's independent from the Linux kernel, so I'm updating this package to use them.

### DIFF
--- a/testing/libbpf/APKBUILD
+++ b/testing/libbpf/APKBUILD
@@ -1,7 +1,6 @@
 # Maintainer: Adam Jensen <acjensen@gmail.com>
 pkgname=libbpf
-pkgver=5.0.8
-_versionref="816253000e918f22e3b395d1bf743503d3508b41"
+pkgver=0.0.3
 pkgrel=0
 pkgdesc="A library for interacting with the Linux kernel's Berkeley Packet Filter (BPF) facility from user space"
 url="https://github.com/libbpf/libbpf"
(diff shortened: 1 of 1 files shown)


testing/libbpf/APKBUILD
```diff
@@ -1,7 +1,6 @@
 # Maintainer: Adam Jensen <acjensen@gmail.com>
 pkgname=libbpf
-pkgver=5.0.8
-_versionref="816253000e918f22e3b395d1bf743503d3508b41"
+pkgver=0.0.3
```
maxice8: Won't that fail as 5.0.8 is bigger than 0.0.3 ?

acj: My bcc package is the only one that depends on libbpf, so I think we're safe. What would fail?

maxice8: People using libbpf-dev won't get an update, builders will use libbpf-5.0.8 instead of libbpf-0.0.3 to compile stuff up until the new version scheme reaches a version bigger than 5.0.8.

Trying to install libbpf after compiling your version gives me 

```
(1/2) Installing elfutils (0.168-r2)
(2/2) Installing libbpf (5.0.8-r0)
```

So it would fail as it would make a package with a version number lower than the current one making it never a viable candidate to be used.

acj: I understand, but I think it's acceptable to make the breaking change. This package is new, and probably no one is using it (or libbpf-dev) yet since it's in the testing repo. Is it possible to purge the existing 5.0.8-r0 version from the index?

maxice8:
On Tue, May 28, 2019 at 07:21:00PM -0700, Adam Jensen wrote:
> acj commented on this pull request.
> 
> 
> 
> > @@ -1,7 +1,6 @@
>  # Maintainer: Adam Jensen <acjensen@gmail.com>
>  pkgname=libbpf
> -pkgver=5.0.8
> -_versionref="816253000e918f22e3b395d1bf743503d3508b41"
> +pkgver=0.0.3
> 
> I understand, but I think it's acceptable to make the breaking change. This package is new, and probably no one is using it (or libbpf-dev) yet since it's in the testing repo. Is it possible to purge the existing 5.0.8-r0 version from the index?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8225#discussion_r288370115

Will have to ask @ncopa for that

Cogitri:
5.0.8 will be removed from the repos as soon as this has been merged and builds successfully, Alpine doesn't keep old versions around.

acj:
@maxice8 @ncopa What's the next step here? Is there anything I can do to help resolve the version issue?

PR #8245, opened 2019-05-29
main/perl-carp-clan: upgrade to 6.07

maxice8:


fcolista:
There's an error here:
```
 package() {
-	cd "$_builddir"
+	cd "$_uilddir"
```
BTW: why in some patches you add the builddir, and in some other you remove it?
Now that "cd" is not needed.
Thanks

maxice8:
On Wed, May 29, 2019 at 07:00:20AM +0000, Francesco Colista wrote:
> There's an error here:
> ```
>  package() {
> -	cd "$_builddir"
> +	cd "$_uilddir"
> ```
> BTW: why in some patches you add the builddir, and in some other you remove it?
> Now that "cd" is not needed.
> Thanks
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8245#issuecomment-496811126

cd "$builddir" is sstill needed for packages on main/ that need to support versions of Alpine Linux old enough to still have an abuild that does not automatically change the working directory to "$builddir" on the prepare() build() check() and package() functions.

PR #8248, opened 2019-05-29
testing/rclone: new aport

tcely:
https://rclone.org/
rsync for cloud storage


tcely:
https://forum.rclone.org/t/rclone-1-48-release/10486/2

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+install=""
+subpackages=""
+
+gpg_signature_extensions="."
+gpgfingerprints="good:FBF7 37EC E9F8 AB18 604B  D2AC 9393 5E02 FF3B 54FA"
+gpgsource="${pkgname}-${pkgver}-sha256sums::https://downloads.rclone.org/v${pkgver}/SHA256SUMS"
+source="https://downloads.rclone.org/v${pkgver}/${_pkgname}-v${pkgver}.tar.gz $gpgsource"
+
+builddir="${srcdir}/${_pkgname}-v${pkgver}"
+export GOPATH="${srcdir}/go"
```
5paceToast: The `{}`s are useless.
There is no styleguide, but I'd recommend just having the GOPATH be `$srcdir`, for consistency with other go packages.
```suggestion
export GOPATH="$srcdir"
```

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+
```
5paceToast: It's usually preferred to set `CGO_ENABLED`, we don't support it, and likely wouldn't even if it started working on musl. This is also more consistent with other go packages (some of which require this to be set to function correctly, iirc).
```suggestion
export CGO_ENABLED=0
```

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+sanitycheck() {
```
5paceToast: This is very strange.
Perhaps this functionality should wait until after abuild has these features? (I assume this is done to allow for gpg checks)

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+	default_sanitycheck
+}
+
+cleanup_srcdir() {
+	go clean -modcache
+	default_cleanup_srcdir
+}
+
+build() {
+	make rclone
```
5paceToast: The Makefile for rclone is really weird, and might become incompatible with GO111MODULES later, which are used.
If we want the version to be built-in, feel free to add the relevant ldflags in.
```suggestion
	GO111MODULE=on go build -o bin/"$pkgname"
```

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+}
+
+check() {
+	make quicktest
```
5paceToast: The Makefile for rclone is really weird, and might become incompatible with GO111MODULES later, which are used.
We also want to run more tests, imo.
```suggestion
	GO111MODULE=on go test ./...
```

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+}
+
+package() {
+	make DESTDIR="${pkgdir}" install
```
5paceToast: The `{}`s are useless.
Further, the Makefile for rclone is really weird.
For example, it doesn't even install the docs!
```suggestion
	install -Dm755 "$builddir"/bin/"$pkgname" \
		"$pkgdir"/usr/bin/"$pkgname"

	install -Dm644 "$builddir"/"$pkgname".1 \
		"$pkgdir"/usr/share/man/man1/"$pkgname".1
```

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+}
+
+verify() {
```
5paceToast: This is very strange.
Perhaps this functionality should wait until after abuild has these features? (I assume this is done to allow for gpg checks)

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
+# Contributor: tcely <rclone+aports@tcely.33mail.com>
+# Maintainer: tcely <rclone+aports@tcely.33mail.com>
+_pkgname="rclone"
+pkgname="${_pkgname}"
```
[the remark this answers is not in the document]
tcely: This approach was used to address a concern from @Ikke about upstream pkgname not matching the sport name.

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+makedepends="bash git go perl"
+install=""
+subpackages=""
+
+gpg_signature_extensions="."
+gpgfingerprints="good:FBF7 37EC E9F8 AB18 604B  D2AC 9393 5E02 FF3B 54FA"
+gpgsource="${pkgname}-${pkgver}-sha256sums::https://downloads.rclone.org/v${pkgver}/SHA256SUMS"
+source="https://downloads.rclone.org/v${pkgver}/${_pkgname}-v${pkgver}.tar.gz $gpgsource"
+
+builddir="${srcdir}/${_pkgname}-v${pkgver}"
```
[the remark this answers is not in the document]
tcely: The default does not match the extracted directory.

testing/rclone/APKBUILD
tcely: Why should we use a path that allows conflicts instead of one that does not?

testing/rclone/APKBUILD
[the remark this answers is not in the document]
5paceToast: Is upstream pkgname not "rclone"?

testing/rclone/APKBUILD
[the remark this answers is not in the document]
5paceToast: Ok, fair enough; though I'd still prefer `$srcdir/$pkgname-v$pkgver` rather than `$_pkgname` (since I still don't really understand how having two variables referring to the same constant helps anything)

testing/rclone/APKBUILD
tcely: This is done to allow for verifying before any sums are calculated.

testing/rclone/APKBUILD
5paceToast: Under normal (abuild) circumstances, `$GOPATH` only contains `src/$host/[...]`.
I don't see how this allows for any further conflicts than having a `go/src/$host` - both have exactly one conflict point, both of which are unlikely.
However, not having an extra path means less paths to traverse if debugging ends up required.

As such, can you elaborate on the conflicts part?

testing/rclone/APKBUILD
5paceToast: You're disabling native sanity checks in order to allow for something custom built into the file that's meant to describe how to build the package.
Saying that you're doing it to allow said custom thing isn't really a resolution of any kind.

testing/rclone/APKBUILD
tcely: This verifies the sources from the signed sha256sums provided by upstream. This allows for checking the full chain of trust before calculating sha512 sums.

5paceToast: Again, I understand what this does.
I'm questioning why it's done here, rather than being added to abuild.
APKBUILDs should generally describe how to build and package package-specific things.
I'm not sure they're the correct place to prototype new abuild functionality.

testing/rclone/APKBUILD
tcely: $srcdir has all sources and extracted archives.

It is less likely to have go directory already extracted than src. Ideally this should be distinct like pkgbasedir is.

5paceToast: I'd argue that an archive with a root `src` directory is just as broken as one with a root `go` directory, but fair enough.

5paceToast:
Please don't mark unresolved conversations as resolved.
That makes it very hard to actually follow the ones that aren't resolved, since they look the same externally.

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+pkgrel=0
+pkgdesc="rsync for cloud storage"
+url="https://rclone.org/"
+arch="all"
+license="MIT"
+options=""
+depends=""
+makedepends="bash git go perl"
+install=""
+subpackages=""
```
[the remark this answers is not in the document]
5paceToast: So... man page not desired?
I could fix all of this up after it gets merged but I'd imagine it'd be nicer for it to be working to begin with...
It's not very hard to install a man page either, so I'm not sure why this is just resolved.

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
...
+pkgname="${_pkgname}"
+pkgver="1.48.0"
+pkgrel=0
+pkgdesc="rsync for cloud storage"
+url="https://rclone.org/"
+arch="all"
+license="MIT"
+options=""
+depends=""
+makedepends="bash git go perl"
```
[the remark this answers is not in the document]
tcely: They are used in the Makefile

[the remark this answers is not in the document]
5paceToast: I realize that.
Why are you using the Makefile?
That was another part of my review that was um... marked resolved without any interaction.
The Makefile is extraordinarily strange (read it!) and is wholly unnecessary.
I even gave specific actionable push-single-button-to-apply replacement snippets.
If you had any specific reason to use the Makefile over standard golang build procedures, I'd imagine you would have responded to that part of the review with it...

As-is, you're introducing a potential failure condition and requesting additional makedepends for no noticeable reason.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: It is, I understand that the concern is that URL and extracted directory name's not break when the sport is renamed.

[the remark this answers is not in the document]
5paceToast: I don't understand - are there plans to rename it from the upstream name (which is the policy)?
How does this help in that scenario?
How does this not apply to literally every aport in the tree?

testing/rclone/APKBUILD
tcely: I've disabled selected sanity checks for source and sums counting because they blocked the verification step from running.

5paceToast: So... back to the original question.
Perhaps this functionality should wait until after abuild has these features, and to focus efforts on giving abuild native support for gpg?

You're actively *DISABLING* security features to add in unproven and untested code.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: I'm preferring to use the upstream build process over replacing it.

testing/rclone/APKBUILD
tcely: What security features are you referring to?

testing/rclone/APKBUILD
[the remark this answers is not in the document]
5paceToast: The upstream build process is documented [here](https://rclone.org/install/).
Quotes for your convenience:

> Make sure you have at least Go 1.7 installed. Download go if necessary. The latest release is recommended. Then
> ```
> git clone https://github.com/ncw/rclone.git
> cd rclone
> go build
> ./rclone version
> ```
> You can also build and install rclone in the GOPATH (which defaults to \~/go) with:
> `go get -u -v github.com/ncw/rclone`
> and this will build the binary in $GOPATH/bin (\~/go/bin/rclone by default) after downloading the source to $GOPATH/src/github.com/ncw/rclone (~/go/src/github.com/ncw/rclone by default).

Note that the `Makefile` is never mentioned.

Further, do consider the standard build process for a given language.
What would you do if an upstream had a standard CMake setup, but also a `./build` script that pulled in python, ruby, perl, go and rust, with no additional functionality?

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: I have no plans to rename it.
It helps by splitting aport name from upstream name and using the appropriate one.
This is the only aport in scope.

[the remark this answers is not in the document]
5paceToast: So you're not going to rename it...
Upstream isn't going to rename it...
The two are currently the same...

What is so special about this aport that it needs this special treatment?

testing/rclone/APKBUILD
tcely: I'm classifying this as package specific because published sums vary in the way they are handled quite a bit.

testing/rclone/APKBUILD
5paceToast: By setting source to null, you are disabling [this entire block](https://github.com/alpinelinux/abuild/blob/master/abuild.in#L202).
It does the following things:
1. Makes sure `$install` files are not in `$source`
2. Checks if wget is in makedepends for https sources.
3. Checks to make sure that all source entries have SOME hash attached to them (!).
4. Checks for potentially invalid tarball/source names.

It also disables [this part](https://github.com/alpinelinux/abuild/blob/master/abuild.in#L230) which makes sure that there are no outdated checksums (it's common to manually add a new checksum without removing old ones, while source is emptied, which could introduce problems).

These are not replaced at any point.
Further, even if they were, why would this be done in the APKBUILD rather than as a part of the abuild? (again, the source question, that remains unanswered).

testing/rclone/APKBUILD
5paceToast: So should we go around every aport and add custom `verify()` handling to each of them, with no serious advantage to it?
The developer is perfectly capable of checking the validity of the hash on their own, without this large hacked-on package-specific function.

If we want to do it automatically, it should *still* be done (in a sufficiently generic way) within abuild.
Potentially, you could consider asking for a `default_verify` so that you can at least keep the original `verify()` behavior too, instead of hoping every "custom apkbuild verify" keeps up to date with upstream abuild!

Either way, I'm not seeing why this is absolutely necessary, especially considering all of the introduced problems.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: I was made aware of two distinct usages of package name which just happen to be the same in most cases so I chose to use two variables. Please discuss this with @Ikke if you wish to know more.

testing/rclone/APKBUILD
5paceToast: LGTM with new changes. Feel free to resolve 👍

testing/rclone/APKBUILD
tcely: I absolutely want automated checks over manual work. I don't see how `default_verify` would help. I already did keep the original verify behavior.

5paceToast: Consider other packages implementing this behavior.
Each of them will duplicate original `verify` behavior.
Then, imagine abuild's `verify()` is changed/improved/gets-new-functionality.
Then EVERY package with a custom `verify()` has to update the entire function.
As such, 100% of the common code SHOULD be in the common place (in this case, `abuild`).

In software development, this is normally referred to as the "DRY" (Don't Repeat Yourself) rule.

Ideally, however, `abuild` should be able to handle the entire scenario, not requiring any modification on the APKBUILD side (since the APKBUILD *should* just describe how to correctly build the package; verification of sources shouldn't really be a part of that).

5paceToast:
I think this can be rebased on top of the other (now-merged) PR to purely provide the new gpg features.
Potentially also stay open as a reference for the abuild global implementation (until a resolution (even if EWONTFIX) is made in that upstream)?

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,100 @@
...
+	make rclone
+}
+
+check() {
+	make quicktest
+}
+
+package() {
+	make DESTDIR="${pkgdir}" install
+	install -Dm644 "${_pkgname}.1" "${pkgdir}/usr/share/man/man1"
```
5paceToast: This will install the file as /usr/share/man/man1 rather than what you expect.
```suggestion
	install -Dm644 "$_pkgname.1" "$pkgdir/usr/share/man/man1/$_pkgname.1"
```

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: The pre-built manual page was packaged.

testing/rclone/APKBUILD
tcely: You seem to believe I duplicated common code for this function. That is not the case. I'm already using a call to the common code where possible. So, I still don't see how `default_verify` would help.

5paceToast: Comparing the abuild.in version with yours (both cut out manually) shows a diff of 36 lines; 3 of which are context/separators, and one of which is a modified line (-1).
This results in an effective diff of 32 lines.
Your solution is 49 lines.
Therefore, you share a total of 17 lines exactly.
The abuild version is 17 lines in total.

This means that your content *ONLY* adds new code and copies the entire abuild.in function in verbatim before adding in new stuff.
I realize that your stuff is interleaved inside of the actual function, but the reality is that you *ARE* duplicating common code.

The correct solution was, and remains, to modify abuild to have support for this natively (whether fully, if possible, with some changes to how it's handled to be generic; or in a way that allows piggybacking on the main function in an extended way, which is what I meant by `default_verify`).
Even a standard `default_verify` would help, though it would require some changes in how you're currently handling this code.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
Ikke: @tcely I think you missed the reason why I advised that way. The reason is you want to decouple the package name (which potentially can change) from the upstream name.

Say I want to make some alternative version of a package with some options or some modifications, I change the package name to make it unique, but now the package fails to build because it can not longer find the source.

This says to me that `pkgname` and the upstream name are two separate names, which at some point in time might happen to be the same.

By setting `_pkgname` to `$pkgname` and using that, you just couple those two again (though admittedly would only need to fix one variable).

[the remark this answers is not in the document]
tcely: @Ikke I don't believe I missed the reason. I just don't see why I should assign twice from static text when the two variables are (currently) the same value. When renaming, just update the variable. Creating a single point of change for your case of renaming the aport is the whole reason for using two variables instead of just using `${pkgname}` as @5paceToast repeatedly suggested in the review.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: See https://github.com/alpinelinux/aports/pull/8248#discussion_r297272122

testing/rclone/APKBUILD
```diff
@@ -0,0 +1,98 @@
+# Contributor: tcely <rclone+aports@tcely.33mail.com>
+# Maintainer: tcely <rclone+aports@tcely.33mail.com>
+_pkgname="rclone"
```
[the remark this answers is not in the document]
tcely: See https://github.com/alpinelinux/aports/pull/8248#discussion_r297272122

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: If upstream provided documentation that did not mention `build.sh` and listed manual steps to take, I would also choose `build.sh` in that case as it provides the opportunity for upstream to improve their build process without interaction from downstream.

testing/rclone/APKBUILD
tcely: I think we disagree on what we mean by the words we are using. Modification is not the same as duplication. I'm not sure what you envision a useful purpose for `default_verify` would be.

Please provide a patch to `abuild` that handles this setup (and its variations) for verifying signed sums from upstream in a generic way rather than continuing to argue that my implementation is not package specific. If you have an idea for how `default_verify` would be useful, I'd be very happy to see it.

5paceToast: There appears to be a misunderstanding here.
*You* are the one that wants this functionality merged.
I am perfectly ok with verifying my checksums on bumps, as I do for other packages (including ones that do not have any sort of gpg protection).
The onus on making an implementation and getting it merged is on *you* as a result.
As is, I have little interest in maintaining this monstrosity, and am simply proposing ways as to make it less cumbersome.

Your statement is making it sound like you have no intention on even attempting on patching abuild, and just want this one specific package (and/or others like it, which would have the issue mentioned above, whether or not you "agree with the language" used to describe it) to be some individually rolled prototypes.
That approach is inherently unmaintainable, as it scales exponentially in complexity as the number of handled packages increases.

If you're interest in feedback as to possible ways to handle varying upstream packages, do feel free to ask for that, but please do keep in mind that the one that's interested in having the functionality merged is *you*.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
5paceToast: There are two possible states, with two distinct goals here.
Option 1: if the package is renamed, the APKBUILD should work as-is.
Option 2: the APKBUILD should be trivial when possible.

It is possible to choose Option 1, by having static assignments for both variations, allowing for a renaming of the package.
Under this Option, if someone was to change `pkgname` to `rclone-upstream`, the build would still work.
This, of course, is less trivial than Option 2.

It is also possible to choose Option 2, and minimize the complexity of the actual contents (what I was suggesting, since that is inherently more maintainable).
This means that one does not make a `_pkgname`, and would require changes in case of `pkgname` change.

Further, considering [your comment](https://github.com/alpinelinux/aports/pull/8248#discussion_r296954256), and the fact that Go applications are not trivially configurable, the odds of someone actually renaming the package are relatively slim, suggesting that Option 2 is the correct one.

However, your solution is neither.
It is non-trivial, and it does not allow the APKBUILD to function with a simple rename.
It is the worst of both worlds.

testing/rclone/APKBUILD
tcely: > There appears to be a misunderstanding here.
> _You_ are the one that wants this functionality merged.
> I am perfectly ok with verifying my checksums on bumps, as I do for other packages (including ones that do not have any sort of gpg protection).
> The onus on making an implementation and getting it merged is on _you_ as a result.

I am not happy with "best guess" checksums when a better way of verifying the downloads is available. I made an implementation, you are the one complaining it is not good enough. In a fairly combative tone too.

> As is, I have little interest in maintaining this monstrosity, and am simply proposing ways as to make it less cumbersome.

No one has asked you to either participate in name calling or maintenance.

> 
> Your statement is making it sound like you have no intention on even attempting on patching abuild, and just want this one specific package (and/or others like it, which would have the issue mentioned above, whether or not you "agree with the language" used to describe it) to be some individually rolled prototypes.
> That approach is inherently unmaintainable, as it scales exponentially in complexity as the number of handled packages increases.
> 

You can find the patch to abuild here: https://github.com/alpinelinux/abuild/pull/57

I have no intention of trying to merge a package specific process into `abuild` and I don't agree with your assertion that doing so is the best way forward. If you have a better idea, you are welcome to attempt to get it merged into `abuild`.

> If you're interest in feedback as to possible ways to handle varying upstream packages, do feel free to ask for that, but please do keep in mind that the one that's interested in having the functionality merged is _you_.

That is not a true statement. I don't want to have this package specific verify routine merged into `abuild`. It is your previous posts that have left me with the understanding that you want me to attempt to merge all of the possible ways signed sums can be used by various packages into `abuild` for you so that you (and others) can specify a variable or some such in `APKBUILD` to use one of those options.

I'm simply not willing to do this work for you while my previous work sits unreviewed. I'm happily using a version of `abuild` that validates GPG signatures, so my issue is already solved.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: > There are two possible states, with two distinct goals here.
> Option 1: if the package is renamed, the APKBUILD should work as-is.

This is more accurately stated as: if the package is renamed there should be no additional work beyond updating the single `pkgname` variable. This is silly in my opinion. This argument about having two distinct variables for the distinct usages of package name is also silly and a waste of time.

> Option 2: the APKBUILD should be trivial when possible.

You should probably define what you mean by trivial here.

> 
> It is possible to choose Option 1, by having static assignments for both variations, allowing for a renaming of the package.

Nothing about variable assignments allows or forbids package renaming, suggesting otherwise is a specious argument.

It is possible to assign the static text once and limit the work to the `pkgname` variable too.
```
_pkgname="rclone"
pkgname="${_pkgname}"
```

This is what I did when I first created the `APKBUILD` file.

> Under this Option, if someone was to change `pkgname` to `rclone-upstream`, the build would still work.
> This, of course, is less trivial than Option 2.

You will have to explain as to how you believe this is less trivial than option 2. It is not as apparent as you seem to believe.

> 
> It is also possible to choose Option 2, and minimize the complexity of the actual contents (what I was suggesting, since that is inherently more maintainable).
> This means that one does not make a `_pkgname`, and would require changes in case of `pkgname` change.
> 

This appears to suggest that maintenance is easier by foisting the decision of which is an upstream name and which is a downstream name onto the person coming along next. I don't agree with that notion at all.

> Further, considering [your comment](https://github.com/alpinelinux/aports/pull/8248#discussion_r296954256), and the fact that Go applications are not trivially configurable, the odds of someone actually renaming the package are relatively slim, suggesting that Option 2 is the correct one.
> 

I disagree with this too. Making the (unlikely) future change of the name (from either upstream or downstream) easier to address is the more correct choice in my opinion.

> However, your solution is neither.
> It is non-trivial, and it does not allow the APKBUILD to function with a simple rename.
> It is the worst of both worlds.

This is misstating the situation. I don't view changing one or two variables as non-trivial. Reviewing the entire `APKBUILD` for usages of `pkgname` and deciding on which name is appropriate is certainly non-trivial. Front loading that decision by using two variables is an entirely appropriate step to save a future contributor effort.

testing/rclone/APKBUILD
5paceToast: > I am not happy with "best guess" checksums when a better way of verifying the downloads is available.

How do you know I don't cross-check my checksums against upstream gpg?
Further, why do you care about how checksums are done for packages you don't maintain?
Certainly, the onus of making sure the package isn't broken or compromised is on me, and you need not worry about that.
Instead, I am recommending having a generic way to do what you want - have the better way of verifying the downloads available for all packages, including ones with this type of handling.

> I made an implementation, you are the one complaining it is not good enough. In a fairly combative tone too.

My tone is not meant to be combative, but exasperated.
You continue to defend stylistic choices (that go against common practices), even if I am to be the one maintaining the package (which I am perfectly willing to do and will do whether or not this eventually merges (assuming it merges in a way that is actually maintainable)), you try to make semantic arguments (that nothing is repeated from the original verify function, despite every single logic line being present, unmodified), and now appear to be arguing that my tone is wrong!
If I come across as combative, I'm sorry about that, but it is not intended to be that way, and would ask that you adjust accordingly.

> > As is, I have little interest in maintaining this monstrosity, and am simply proposing ways as to make it less cumbersome.
> 
> No one has asked you to either participate in name calling or maintenance.

As of currently, I maintain this package.
If it is merged in its current form, I will be the one that will have to maintain your code.
It is not up to spec (that you are reluctant to discuss), and has multiple other issues.
*I* will be the one that will need to look at every commit made to abuild in case `verify` is ever changed.
I do not wish to do that.
Further, I do also wish to have an up-to-date version of rclone available.
I suspect that this PR has been mostly untouched for a month not because rclone is controversial (see the PR that did get through), but because of the very large function and strange stylistic choices.
This suggests that any future bumps will cause similar delays.

> > Your statement is making it sound like you have no intention on even attempting on patching abuild, and just want this one specific package (and/or others like it, which would have the issue mentioned above, whether or not you "agree with the language" used to describe it) to be some individually rolled prototypes.
> > That approach is inherently unmaintainable, as it scales exponentially in complexity as the number of handled packages increases.
> 
> You can find the patch to abuild here: [alpinelinux/abuild#57](https://github.com/alpinelinux/abuild/pull/57)
> 
> I have no intention of trying to merge a package specific process into `abuild` and I don't agree with your assertion that doing so is the best way forward. If you have a better idea, you are welcome to attempt to get it merged into `abuild`.

This process is package specific because your original patch is not flexible enough.
I assert that it is possible to have a patch to `abuild` that *is* flexible enough, and only requires minimal changes on the APKBUILD side.
However, again, *I* have no interest in having generic (or non-generic) gpg verification processes in abuild.
I just want my packages to be up to date, and maintainable, so that I may keep them up to date, without, for instance, looking at every commit done to abuild since the last bump.

> > If you're interest in feedback as to possible ways to handle varying upstream packages, do feel free to ask for that, but please do keep in mind that the one that's interested in having the functionality merged is _you_.
> 
> That is not a true statement. I don't want to have this package specific verify routine merged into `abuild`. It is your previous posts that have left me with the understanding that you want me to attempt to merge all of the possible ways signed sums can be used by various packages into `abuild` for you so that you (and others) can specify a variable or some such in `APKBUILD` to use one of those options.

Again, I'm not talking about a package-specific verify routine.
I'm talking about a not-currently-written routine that will adequately handle your goal of allowing a full chain of trust within arbitrary packages.

Regarding what I want, it's simple.
I want packages to be up to date.
I want packages to be maintainable.
This changeset, as it is now, ultimately goes against both points (though only somewhat on one of the two counts).
I am *willing* (but not *interested*) in having a high quality generic approach to having chain of trust style verification.

> I'm simply not willing to do this work for you while my previous work sits unreviewed. I'm happily using a version of `abuild` that validates GPG signatures, so my issue is already solved.

Again, this would not be done *for me*.
I am perfectly content with this getting closed and you blaming me if the built version of rclone ends up compromised, when your approach would have caught the issue.

Since you appear to have trouble imagining a sufficiently generic algorithm for handling these cases, here is a potential sample:
1. Have `gpg_sources` - these must end in `.asc` or `.sig` (even if it's renamed using the `name::source` convention.
2. During fetching/verification, for every gpg_source, check if a non-.asc or non-.sig file exists in `sources`.
  If it does, verify that file using the `.asc|.sig` file.
  If it doesn't, consider it to be a clearsign file, and add the output to sources= (during runtime).

Past that, the only thing "special cases" such as this would need is an override of `abuild checksum`, to generate sums from the verified SHA*SUMS.txt file.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
5paceToast: > > There are two possible states, with two distinct goals here.
> > Option 1: if the package is renamed, the APKBUILD should work as-is.
> 
> This is more accurately stated as: if the package is renamed there should be no additional work beyond updating the single `pkgname` variable. This is silly in my opinion. This argument about having two distinct variables for the distinct usages of package name is also silly and a waste of time.

I agree!
So why are you doing it?
It's a waste of time and space, but you insist on continuing.

> > Option 2: the APKBUILD should be trivial when possible.
> 
> You should probably define what you mean by trivial here.

Trivial: Of little significance or value.
There is little less significant than nothingness.
Defining an extra variable is, by definition, less trivial than not defining it.
This is to minimize how many things one needs to have in mind when looking over the source.

> > It is possible to choose Option 1, by having static assignments for both variations, allowing for a renaming of the package.
> 
> Nothing about variable assignments allows or forbids package renaming, suggesting otherwise is a specious argument.

Of course, I don't mean that it is "impossible to rename the package".
I mean that it can be renamed without changing anything further...

> It is possible to assign the static text once and limit the work to the `pkgname` variable too.
> 
> ```
> _pkgname="rclone"
> pkgname="${_pkgname}"
> ```
> This is what I did when I first created the `APKBUILD` file.
> 
> > Under this Option, if someone was to change `pkgname` to `rclone-upstream`, the build would still work.
> > This, of course, is less trivial than Option 2.
> 
> You will have to explain as to how you believe this is less trivial than option 2. It is not as apparent as you seem to believe.

See above explanation of triviality.

> > It is also possible to choose Option 2, and minimize the complexity of the actual contents (what I was suggesting, since that is inherently more maintainable).
> > This means that one does not make a `_pkgname`, and would require changes in case of `pkgname` change.
> 
> This appears to suggest that maintenance is easier by foisting the decision of which is an upstream name and which is a downstream name onto the person coming along next. I don't agree with that notion at all.

There is no person coming along next.
I'm perfectly ok with actively maintaining my package, without someone else walking in - it's *why* I maintain them after all, to have recent versions available.
Further, there is no decision to make - there is the name of the folder inside the tarball, and there is whatever you call your package.
You don't decide which is which, they simply are.
Regarding maintenance, consider a person that gets confused when they see two identical variables, which are used in different scenarios.
They will try to figure out what's going on *every time they touch the package*, instead of maybe once in a very unlikely situation.

> > Further, considering [your comment](https://github.com/alpinelinux/aports/pull/8248#discussion_r296954256), and the fact that Go applications are not trivially configurable, the odds of someone actually renaming the package are relatively slim, suggesting that Option 2 is the correct one.
> 
> I disagree with this too. Making the (unlikely) future change of the name (from either upstream or downstream) easier to address is the more correct choice in my opinion.

You are allowed to have this opinion.
However, if I'm going to be the one handling the name of change (being the current maintainer), please do respect my desires regarding the package that I am going to be the one doing the work for.
See other comments regarding "up-to-dateness" regarding the obvious alternative.

> > However, your solution is neither.
> > It is non-trivial, and it does not allow the APKBUILD to function with a simple rename.
> > It is the worst of both worlds.
> 
> This is misstating the situation. I don't view changing one or two variables as non-trivial. Reviewing the entire `APKBUILD` for usages of `pkgname` and deciding on which name is appropriate is certainly non-trivial. Front loading that decision by using two variables is an entirely appropriate step to save a future contributor effort.

See above statement of triviality.
More trivial = less code.
More code = less trivial.
Therefore, it is less trivial than the alternative.
It not allowing for the process as described by @Ikke is also a matter of fact - one needs to change additional details (as it is now) in order to rename the package appropriately.

testing/rclone/APKBUILD
tcely: > > I am not happy with "best guess" checksums when a better way of verifying the downloads is available.
> 
> How do you know I don't cross-check my checksums against upstream gpg?

I don't know that you do or don't. This is the problem. There really is not a lot that can be asserted by the sums matching. It is entirely possible for example that the original release was replaced with a trojan from the time the sums were checked into `aports.git` to the time the package was built.

> Further, why do you care about how checksums are done for packages you don't maintain?

As a user, I prefer to know (rather than assume) that the software that was built was that released by the authors.

> Certainly, the onus of making sure the package isn't broken or compromised is on me, and you need not worry about that.

I would call that unreasonable for both the user and maintainer roles. It's much better if both could be assured there is no compromise.

> Instead, I am recommending having a generic way to do what you want - have the better way of verifying the downloads available for all packages, including ones with this type of handling.
> 

Good recommendation. I don't have a generic way. When there is one or a solution that covers sufficient variations I'm happy to switch to it.

> > I made an implementation, you are the one complaining it is not good enough. In a fairly combative tone too.
> 
> My tone is not meant to be combative, but exasperated.
> You continue to defend stylistic choices (that go against common practices), even if I am to be the one maintaining the package (which I am perfectly willing to do and will do whether or not this eventually merges (assuming it merges in a way that is actually maintainable)), you try to make semantic arguments (that nothing is repeated from the original verify function, despite every single logic line being present, unmodified), and now appear to be arguing that my tone is wrong!
> If I come across as combative, I'm sorry about that, but it is not intended to be that way, and would ask that you adjust accordingly.
> 
I will adjust if you will. This is not a stylistic choice. I have an implementation that does what I wanted. You continue to say it is not good enough for you. Feel free to make it better. Until then, please stop telling me how you don't like it. Please also note that I did not say nothing is repeated. I said the function is not duplicated. Which means the verify function that exists in this pull request can not be replaced by the one in `abuild` while preserving the same operation.

> > > As is, I have little interest in maintaining this monstrosity, and am simply proposing ways as to make it less cumbersome.
> > 
> > 
> > No one has asked you to either participate in name calling or maintenance.
> 
> As of currently, I maintain this package.
> If it is merged in its current form, I will be the one that will have to maintain your code.

Yes, a situation that arose after your review of this pull request and the subsequent merging of your alternative pull request.

> It is not up to spec (that you are reluctant to discuss), and has multiple other issues.

I don't agree with that assessment.

> _I_ will be the one that will need to look at every commit made to abuild in case `verify` is ever changed.
> I do not wish to do that.

That's perfectly reasonable.

> Further, I do also wish to have an up-to-date version of rclone available.
> I suspect that this PR has been mostly untouched for a month not because rclone is controversial (see the PR that did get through), but because of the very large function and strange stylistic choices.
> This suggests that any future bumps will cause similar delays.

I don't know or care much about why the pull request was not interacted with. As I'm sure you also noticed, this project has not been promptly reviewing / acting on pull requests for quite a while now. The most common excuse in this period of time was the `3.10` release that was happening. Your suspicious or opinions about "strange stylistic choices" hold very little sway with me. Since I don't put much weight in these things I also don't believe your inference about future bumps is accurate.

> 
> > > Your statement is making it sound like you have no intention on even attempting on patching abuild, and just want this one specific package (and/or others like it, which would have the issue mentioned above, whether or not you "agree with the language" used to describe it) to be some individually rolled prototypes.
> > > That approach is inherently unmaintainable, as it scales exponentially in complexity as the number of handled packages increases.
> > 
> > 
> > You can find the patch to abuild here: [alpinelinux/abuild#57](https://github.com/alpinelinux/abuild/pull/57)
> > I have no intention of trying to merge a package specific process into `abuild` and I don't agree with your assertion that doing so is the best way forward. If you have a better idea, you are welcome to attempt to get it merged into `abuild`.
> 
> This process is package specific because your original patch is not flexible enough.
> I assert that it is possible to have a patch to `abuild` that _is_ flexible enough, and only requires minimal changes on the APKBUILD side.
> However, again, _I_ have no interest in having generic (or non-generic) gpg verification processes in abuild.
> I just want my packages to be up to date, and maintainable, so that I may keep them up to date, without, for instance, looking at every commit done to abuild since the last bump.
> 

You are making the assertion that it is possible. So the proof of the possible is left to you, as you also state that you are uninterested I understand that such proof is not forthcoming. However, please stop bemoaning my implementation that did not meet your notion of the perfect solution when you are unwilling to share anything better.

> > > If you're interest in feedback as to possible ways to handle varying upstream packages, do feel free to ask for that, but please do keep in mind that the one that's interested in having the functionality merged is _you_.
> > 
> > 
> > That is not a true statement. I don't want to have this package specific verify routine merged into `abuild`. It is your previous posts that have left me with the understanding that you want me to attempt to merge all of the possible ways signed sums can be used by various packages into `abuild` for you so that you (and others) can specify a variable or some such in `APKBUILD` to use one of those options.
> 
> Again, I'm not talking about a package-specific verify routine.
> I'm talking about a not-currently-written routine that will adequately handle your goal of allowing a full chain of trust within arbitrary packages.
> 

Again, please stop talking about the non-existent when reviewing the existing implementation. If you want to discuss your theoretical routine, you are welcome to propose such a thing in your own pull request.

> Regarding what I want, it's simple.
> I want packages to be up to date.
> I want packages to be maintainable.
> This changeset, as it is now, ultimately goes against both points (though only somewhat on one of the two counts).
> I am _willing_ (but not _interested_) in having a high quality generic approach to having chain of trust style verification.
> 
> > I'm simply not willing to do this work for you while my previous work sits unreviewed. I'm happily using a version of `abuild` that validates GPG signatures, so my issue is already solved.
> 
> Again, this would not be done _for me_.
> I am perfectly content with this getting closed and you blaming me if the built version of rclone ends up compromised, when your approach would have caught the issue.
> 

Then please stop asking for changes they are not coming. The current function met my needs.

> Since you appear to have trouble imagining a sufficiently generic algorithm for handling these cases, here is a potential sample:
> 
> 1. Have `gpg_sources` - these must end in `.asc` or `.sig` (even if it's renamed using the `name::source` convention.
> 2. During fetching/verification, for every gpg_source, check if a non-.asc or non-.sig file exists in `sources`.
>    If it does, verify that file using the `.asc|.sig` file.
>    If it doesn't, consider it to be a clearsign file, and add the output to sources= (during runtime).
> 
> Past that, the only thing "special cases" such as this would need is an override of `abuild checksum`, to generate sums from the verified SHA*SUMS.txt file.

This is too simplistic. You are welcome to implement it and see for yourself the issues you will run into.

testing/rclone/APKBUILD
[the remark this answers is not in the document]
tcely: > > > There are two possible states, with two distinct goals here.
> > > Option 1: if the package is renamed, the APKBUILD should work as-is.
> > 
> > 
> > This is more accurately stated as: if the package is renamed there should be no additional work beyond updating the single `pkgname` variable. This is silly in my opinion. This argument about having two distinct variables for the distinct usages of package name is also silly and a waste of time.
> 
> I agree!
> So why are you doing it?
> It's a waste of time and space, but you insist on continuing.
> 

I'm continuing because I don't agree with your assertions and do believe that they are worthy of a response.

> > > Option 2: the APKBUILD should be trivial when possible.
> > 
> > 
> > You should probably define what you mean by trivial here.
> 
> Trivial: Of little significance or value.
> There is little less significant than nothingness.
> Defining an extra variable is, by definition, less trivial than not defining it.
> This is to minimize how many things one needs to have in mind when looking over the source.
> 
> > > It is possible to choose Option 1, by having static assignments for both variations, allowing for a renaming of the package.
> > 
> > 
> > Nothing about variable assignments allows or forbids package renaming, suggesting otherwise is a specious argument.
> 
> Of course, I don't mean that it is "impossible to rename the package".
> I mean that it can be renamed without changing anything further...
> 

I already gave an example of how the two variable changes can be reduced to a single change and you gave another.

> > It is possible to assign the static text once and limit the work to the `pkgname` variable too.
> > ```
> > _pkgname="rclone"
> > pkgname="${_pkgname}"
> > ```
> > 
> > 
> > This is what I did when I first created the `APKBUILD` file.
> > > Under this Option, if someone was to change `pkgname` to `rclone-upstream`, the build would still work.
> > > This, of course, is less trivial than Option 2.
> > 
> > 
> > You will have to explain as to how you believe this is less trivial than option 2. It is not as apparent as you seem to believe.
> 
> See above explanation of triviality.
> 
> > > It is also possible to choose Option 2, and minimize the complexity of the actual contents (what I was suggesting, since that is inherently more maintainable).
> > > This means that one does not make a `_pkgname`, and would require changes in case of `pkgname` change.
> > 
> > 
> > This appears to suggest that maintenance is easier by foisting the decision of which is an upstream name and which is a downstream name onto the person coming along next. I don't agree with that notion at all.
> 
> There is no person coming along next.
> I'm perfectly ok with actively maintaining my package, without someone else walking in - it's _why_ I maintain them after all, to have recent versions available.
> Further, there is no decision to make - there is the name of the folder inside the tarball, and there is whatever you call your package.
> You don't decide which is which, they simply are.
> Regarding maintenance, consider a person that gets confused when they see two identical variables, which are used in different scenarios.
> They will try to figure out what's going on _every time they touch the package_, instead of maybe once in a very unlikely situation.
> 

I disagree with everything you've stated here.
You are not going to be the only person to ever touch any package you maintain.
The decision is which name should be used. You proved this yourself when you suggested an inappropriate `builddir` value earlier. Your pull request uses the GitHub generated archive which was named to match the default `builddir` value, where as the signed source archive contains a directory named differently.
These are not "two identical" variables. You could call them similar, but they are distinct variables. They currently do hold an identical value, so some confusion can be understood.

> > > Further, considering [your comment](https://github.com/alpinelinux/aports/pull/8248#discussion_r296954256), and the fact that Go applications are not trivially configurable, the odds of someone actually renaming the package are relatively slim, suggesting that Option 2 is the correct one.
> > 
> > 
> > I disagree with this too. Making the (unlikely) future change of the name (from either upstream or downstream) easier to address is the more correct choice in my opinion.
> 
> You are allowed to have this opinion.
> However, if I'm going to be the one handling the name of change (being the current maintainer), please do respect my desires regarding the package that I am going to be the one doing the work for.
> See other comments regarding "up-to-dateness" regarding the obvious alternative.
> 
> > > However, your solution is neither.
> > > It is non-trivial, and it does not allow the APKBUILD to function with a simple rename.
> > > It is the worst of both worlds.
> > 
> > 
> > This is misstating the situation. I don't view changing one or two variables as non-trivial. Reviewing the entire `APKBUILD` for usages of `pkgname` and deciding on which name is appropriate is certainly non-trivial. Front loading that decision by using two variables is an entirely appropriate step to save a future contributor effort.
> 
> See above statement of triviality.
> More trivial = less code.
> More code = less trivial.
> Therefore, it is less trivial than the alternative.
> It not allowing for the process as described by @Ikke is also a matter of fact - one needs to change additional details (as it is now) in order to rename the package appropriately.

It does allow for the changing of the package name (either upstream or downstream) as described, just not to your liking. The alternative is to use the same variable for every usage, which from a number of decisions and number of changes is not less trivial than one or two changed values in variables.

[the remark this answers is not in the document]
5paceToast: > I'm continuing because I don't agree with your assertions and do believe that they are worthy of a response.

I can respect that.
I'm interested in figuring out the closest opinion to the truth™️.

> I already gave an example of how the two variable changes can be reduced to a single change and you gave another.

That's fair, so it ultimately fits the pre-emptive "what if mv happens" scenario.
However, in my opinion, "looking over the entirety of the APKBUILD" should be a requirement for modifying them in non-trivial ways (in this case, a "trivial" modification is a version bump and nothing else) - bad things can happen.
I would rather things within be simpler, rather than pre-emptively optimizing against unlikely scenarios.
However, the simple variant shouldn't make it impossible to later fix things (see: switch from libressl to openssl, proj4 renaming to proj).

> I disagree with everything you've stated here.
> You are not going to be the only person to ever touch any package you maintain.

My intention, when I start to maintain a package, is for no one to need to touch it except me.
I fully intend to update 100% of my maintained packages in a timely manner, and to support them in case bad things happen.
Sure, someone else could touch the packages I maintain, but my intention is to make that event wholly unnecessary.

> The decision is which name should be used. You proved this yourself when you suggested an inappropriate `builddir` value earlier. Your pull request uses the GitHub generated archive which was named to match the default `builddir` value, where as the signed source archive contains a directory named differently.
> These are not "two identical" variables. You could call them similar, but they are distinct variables. They currently do hold an identical value, so some confusion can be understood.

That's fair.
However, in the vast majority of cases, they do coincidentally match.
I don't think we need to pre-emptively optimize for an unlikely scenario at the expense of the contents (even if the expense is small), because one will accumulate over time (every time someone looks at/edits the build), and one is a simple one-time cost in an unlikely scenario.

> It does allow for the changing of the package name (either upstream or downstream) as described, just not to your liking. The alternative is to use the same variable for every usage, which from a number of decisions and number of changes is not less trivial than one or two changed values in variables.

The changes required to rename the packages are less trivial, I agree.
But if something should be optimized for, it's the general case, imo.

testing/rclone/APKBUILD
5paceToast: > I don't know that you do or don't. This is the problem. There really is not a lot that can be asserted by the sums matching. It is entirely possible for example that the original release was replaced with a trojan from the time the sums were checked into `aports.git` to the time the package was built.

If that happens, verification will fail.
Am I missing something?
Unless we assume that sha512 is broken, but if we assume that sha512 is broken, a trojan pretending to be rclone is the least of our concerns.

> > Further, why do you care about how checksums are done for packages you don't maintain?
> 
> As a user, I prefer to know (rather than assume) that the software that was built was that released by the authors.

Distributions add patches all the time.
If you want that (a legitimate desire, mind you!) I would recommend using something like flatpak when available.
Point taken, though.

> > Certainly, the onus of making sure the package isn't broken or compromised is on me, and you need not worry about that.
> 
> I would call that unreasonable for both the user and maintainer roles. It's much better if both could be assured there is no compromise.

I agree, I just don't want it to be done at the price that it is now.
As mentioned, I would strongly support a real generic way to do it in abuild.

> Good recommendation. I don't have a generic way. When there is one or a solution that covers sufficient variations I'm happy to switch to it.

> I will adjust if you will. This is not a stylistic choice. I have an implementation that does what I wanted. You continue to say it is not good enough for you. Feel free to make it better. Until then, please stop telling me how you don't like it. Please also note that I did not say nothing is repeated. I said the function is not duplicated. Which means the verify function that exists in this pull request can not be replaced by the one in `abuild` while preserving the same operation.

My concern is about needing to regularly check abuild commits in every custom-handled package to keep up with upstream.
What term is used (duplication, repetition, whatever) is irrelevant, imo.

> > It is not up to spec (that you are reluctant to discuss), and has multiple other issues.
> 
> I don't agree with that assessment.

- It actively avoids style guidelines (e.g empty variables)
- It does the _pkgname + pkgname thing to protect against an unlikely hypothetical
- It uses `{}`s in ways that will confuse both new people (since most packages don't use them) and experienced shell writers (since they will wonder why they are there)
- It actively introduces unneeded dependencies
- It actively deviates from standard practice regarding go (build system, GOPATH)
- It shuffles things around post-rebase (look at the current resulting diff)

A conversation can be had on whether or not everything should switch to a new spec (e.g GOPATH=$srcdir/go).
However, I don't think it can be argued that everything is "as it should be".

> > _I_ will be the one that will need to look at every commit made to abuild in case `verify` is ever changed.
> > I do not wish to do that.
> 
> That's perfectly reasonable.

Thank you.
I do believe that even if abuild isn't deeply patched, it should be possible to massage the function to a state where I wouldn't mind maintaining it.
However, based on further comments, I assume there is no intent on doing this.
Once your patch to abuild is merged, do feel free to poke me for brainstorming re: how to do it properly in a generic way.

> > Further, I do also wish to have an up-to-date version of rclone available.
> > I suspect that this PR has been mostly untouched for a month not because rclone is controversial (see the PR that did get through), but because of the very large function and strange stylistic choices.
> > This suggests that any future bumps will cause similar delays.
> 
> I don't know or care much about why the pull request was not interacted with. As I'm sure you also noticed, this project has not been promptly reviewing / acting on pull requests for quite a while now. The most common excuse in this period of time was the `3.10` release that was happening. Your suspicious or opinions about "strange stylistic choices" hold very little sway with me. Since I don't put much weight in these things I also don't believe your inference about future bumps is accurate.

I suppose we can disagree then.

> > > > Your statement is making it sound like you have no intention on even attempting on patching abuild, and just want this one specific package (and/or others like it, which would have the issue mentioned above, whether or not you "agree with the language" used to describe it) to be some individually rolled prototypes.
> > > > That approach is inherently unmaintainable, as it scales exponentially in complexity as the number of handled packages increases.
> > > 
> > > 
> > > You can find the patch to abuild here: [alpinelinux/abuild#57](https://github.com/alpinelinux/abuild/pull/57)
> > > I have no intention of trying to merge a package specific process into `abuild` and I don't agree with your assertion that doing so is the best way forward. If you have a better idea, you are welcome to attempt to get it merged into `abuild`.
> > 
> > 
> > This process is package specific because your original patch is not flexible enough.
> > I assert that it is possible to have a patch to `abuild` that _is_ flexible enough, and only requires minimal changes on the APKBUILD side.
> > However, again, _I_ have no interest in having generic (or non-generic) gpg verification processes in abuild.
> > I just want my packages to be up to date, and maintainable, so that I may keep them up to date, without, for instance, looking at every commit done to abuild since the last bump.
> 
> You are making the assertion that it is possible. So the proof of the possible is left to you, as you also state that you are uninterested I understand that such proof is not forthcoming. However, please stop bemoaning my implementation that did not meet your notion of the perfect solution when you are unwilling to share anything better.

I bemoan your solution because I see no way out for myself.
I just want my packages to be up to date, and usually that means doing it myself (my current mean time to update is under 24h, though I expect it could increase slightly as the number of packages increases).

> Again, please stop talking about the non-existent when reviewing the existing implementation. If you want to discuss your theoretical routine, you are welcome to propose such a thing in your own pull request.

Ok fair.
Regarding the current implementation - I do not want to maintain it as-is.
I do want to maintain the package to make sure that it is consistently up to date, and will put in effort in that direction.
I also don't necessarily trust other people to keep things fully up to date with the same level of zest as me.

re: theoretical proposal, I will open an issue after the move to gitlab, and potentially take a closer look.

> > 1. Have `gpg_sources` - these must end in `.asc` or `.sig` (even if it's renamed using the `name::source` convention.
> > 2. During fetching/verification, for every gpg_source, check if a non-.asc or non-.sig file exists in `sources`.
> >    If it does, verify that file using the `.asc|.sig` file.
> >    If it doesn't, consider it to be a clearsign file, and add the output to sources= (during runtime).
> > 
> > Past that, the only thing "special cases" such as this would need is an override of `abuild checksum`, to generate sums from the verified SHA*SUMS.txt file.
> 
> This is too simplistic. You are welcome to implement it and see for yourself the issues you will run into.

Can you elaborate?
If this does have fundamental problems, I'd rather not spend the time potentially implementing it.

tcely: > > I don't know that you do or don't. This is the problem. There really is not a lot that can be asserted by the sums matching. It is entirely possible for example that the original release was replaced with a trojan from the time the sums were committed into `aports.git` to the time the package was built.
> 
> If that happens, verification will fail.
> Am I missing something?
> Unless we assume that sha512 is broken, but if we assume that sha512 is broken, a trojan pretending to be rclone is the least of our concerns.
> 

Why would verification fail? When the sums that were created by `abuild` were those of the trojan, they will match just fine. You have nothing to assert that the upstream release is the same as what the authors intended after all.

> > > It is not up to spec (that you are reluctant to discuss), and has multiple other issues.
> > 
> > 
> > I don't agree with that assessment.
> 
> * It actively avoids style guidelines (e.g empty variables)

Removing empty variables or in other words removing the template is not helpful for new users and serves no good purpose. If this is out of spec, I'd rather see the spec change. I haven't seen any spec around this yet at all.

> * It does the _pkgname + pkgname thing to protect against an unlikely hypothetical

It does indeed do this, I don't consider that an issue or out of spec.

> * It uses `{}`s in ways that will confuse both new people (since most packages don't use them) and experienced shell writers (since they will wonder why they are there)

What most packages do is immaterial. You can find the explanation for why always quoting and bracing variable uses is better at: https://gist.github.com/tcely/8ef31809f04a494a27ad79d49afdf167.

> * It actively introduces unneeded dependencies

This statement is only true if you ignore the upstream `Makefile`.

> * It actively deviates from standard practice regarding go (build system, GOPATH)

Certainly the `rclone` `Makefile` is not standard. However, as we already covered, using the build system that upstream provided is hardly against standard practice. Please address the `rclone` project to help them improve the `Makefile` instead of ignoring it.

> * It shuffles things around post-rebase (look at the current resulting diff)
> 

The only change in ordering I found was `options` which is not included in the [template](https://github.com/alpinelinux/abuild/blob/297de93aef3110f6ced8926e64e4204692b3b74e/newapkbuild.in#L241-L257) and is found all over the place as a result. In fact, I took specific  care to move my functions around to match the already committed order.

> A conversation can be had on whether or not everything should switch to a new spec (e.g GOPATH=$srcdir/go).
> However, I don't think it can be argued that everything is "as it should be".
> 

I certainly agree with that. `GOPATH` should really be treated much more like `pkgbasedir` in my opinion. That likely requires both `setup_go` and `cleanup_go` functions.

> > > 1. Have `gpg_sources` - these must end in `.asc` or `.sig` (even if it's renamed using the `name::source` convention.
> > > 2. During fetching/verification, for every gpg_source, check if a non-.asc or non-.sig file exists in `sources`.
> > >    If it does, verify that file using the `.asc|.sig` file.
> > >    If it doesn't, consider it to be a clearsign file, and add the output to sources= (during runtime).
> > > 
> > > Past that, the only thing "special cases" such as this would need is an override of `abuild checksum`, to generate sums from the verified SHA*SUMS.txt file.
> > 
> > 
> > This is too simplistic. You are welcome to implement it and see for yourself the issues you will run into.
> 
> Can you elaborate?
> If this does have fundamental problems, I'd rather not spend the time potentially implementing it.

Handling clear signatures, encrypted signatures, and detached signatures is a good first step. After that there's not a good standard about what the sums file will contain. Some considerations are input format (there are at least two common ones), which files are summed, and which algorithm is used.

As for `abuild` modifications, you need to pass the sanity checks, then you need fetch to run to gather all the remote files, then you need verify to pass, then you can think about generating the new sums list for everything that's in sources. I don't see a need to use only the sums that were signed, but that is certainly a valid option.

PR #8249, opened 2019-05-29
community/gpgme: move from main and enable qgpgme subpackage

PureTryOut:
This PR introduces the qgpgme subpackage to have Qt bindings available.

I've also moved `libgpg-error-dev` and `libassuan-dev` to `$makedepends`, as abuild was complaining about them being in `$depends_dev`.

Lastly I removed some unnecessary `cd $builddir`'s.

/cc @Cogitri  @fabaff @ncopa


PureTryOut:
Oh, this built locally in docker-abuild without any problems, but it makes sense to fail as `qt5-qtbase` is in community while `gpgme` is in main...

What could I do about this? Would it be ok to move `gpgme` to community? @ncopa

PureTryOut:
There is a discussion for this going on in #alpine-devel on IRC. The main question is if an "important package" like udisks2 should be moved from main.

PR #8256, opened 2019-05-29
community/wlroots: add elogind support

PureTryOut:
CC @ddevault, could you please review this?

### DIFF
--- a/community/wlroots/APKBUILD
+++ b/community/wlroots/APKBUILD
@@ -2,13 +2,15 @@
 # Maintainer: Drew DeVault <sir@cmpwn.com>
 pkgname=wlroots
 pkgver=0.5.0
-pkgrel=0
+pkgrel=1
 pkgdesc="A modular Wayland compositor library"
 url="https://github.com/swaywm/wlroots"
 arch="all"
 license="MIT"
 options="!check" # contains no test suite
-makedepends="eudev-dev
+makedepends="
+	elogind-dev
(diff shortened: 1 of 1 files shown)


Ikke:
@maxice8 That can also be done when applied.

maxice8:
**I** **will** **do** **it** **later**

ddevault:
LGTM but I would explicitly enable elogind support with meson options to catch issues earlier

Aside: I think that elogind design-wise is pretty bad and I would sooner await a good replacement than ship it with Alpine. YMMV

PureTryOut:
@ddevault thanks for the review! I explicitely enabled logind support and set the logind provider to elogind. About elogind, it is already shipped with Alpine and at the moment it's the only real option aside from the dead-upstream ConsoleKit2. Do you know of any replacement in the works?

@maxice8 and @Ikke: since I had to make a change anyway, I took the liberty to modernize the APKBUILD as well.

ddevault:
The other option is the status quo: setuid. But I'm not torn up about it

community/wlroots/APKBUILD
```diff
@@ -21,22 +21,21 @@ makedepends="eudev-dev
 	wayland-protocols
 	xcb-util-image-dev
 	xcb-util-wm-dev
-	"
+	elogind-dev"
```
tcely:
```suggestion
	elogind-dev
	"
```

PR #8259, opened 2019-05-29
main/acf-chrony: modernize

tcely:


maxice8:
If you plan on modernizing stuff i recommend using apkbuild-lint and aports-lint from github.com/maxice8/atools , have fun

PR #8260, opened 2019-05-29
main/acf-dhcp: modernize

tcely:


maxice8:
Also extra line after the make thing

PR #8272, opened 2019-05-29
testing/ripgrep: upgrade to 11.0.0

5paceToast:


testing/ripgrep/APKBUILD
```diff
@@ -16,17 +16,14 @@ source="$pkgname-$pkgver.tar.gz::https://github.com/BurntSushi/$pkgname/archive/
 builddir="$srcdir/"$pkgname-$pkgver
```
maxice8: This is the default value and thus can be removed

Ikke:
@misery

5paceToast:
I'd be ok with adopting this, if @misery has no interest in actively keeping it up to date. (I was planning to maintain it anyway, for a while now, as can be seen in my user-aports)

misery:
I'm still interested! But very busy at the moment.

misery:
By the way! 11.0.1 is available.

5paceToast:
Good catch! Just compiled it on my builder, new PR over at #8373

PR #8274, opened 2019-05-29
testing/py3-columnize: modernize, enable tests.

maxice8:
- Switch to GitHub tarball since that one provides the tests required
- Bump pkgrel even though it isn't strictly necessary


maxice8:
@fabaff

PR #8282, opened 2019-05-30
testing/py-webob: upgrade to 1.8.5

maxice8:
@fabaff


maxice8:
With this only py-pyramid remains that needs py-webob, after that is converted we can slash down py-webob into only py3-webob.

PR #8288, opened 2019-05-30
testing/traefik: adjust architectures

tcely:


Ikke:
Could you add more details about what and why to the commit message? Thanks.

PR #8291, opened 2019-05-30
main/util-linux: more robust pkgver reduction

tcely:


ncopa:
NACK. this provides little value if any, and makes things more complicated.

PR #8292, opened 2019-05-30
main/sqsh: more robust pkgver reduction

tcely:


ncopa:
NACK. This makes things needlessly complicated and gives little value if any.

PR #8294, opened 2019-05-30
main/shorewall: improve

tcely:
No changes to the built package should result from this commit.


ncopa:
I can add the `main/shorewall: add -openrc subpackage` though.

PR #8296, opened 2019-05-30
testing/xmrig: upgrade to 2.14.4

OTLabs:
- https://github.com/xmrig/xmrig/releases 2.14.4
- License clarification
- No upstream testsuite, removing check()
- License file removed
- builddir has default value and was removed


OTLabs:
@maxice8 Attended all your recommendation in full. Thank you!

PR #8298, opened 2019-05-30
community/isync: upgrade to 1.3.1

itoffshore:
Source: `README`:

> This is a long overdue maintenance release:
> 
> * SSL now uses SNI, which for example GMail requires
> * fixed fallbacks for missing UIDPLUS extension (with e.g. DavMail)
> * fixed UIDVALIDITY recovery with really long Message-id headers
> * fixed GSSAPI authentication with Kerberos
> * fixed support for IMAP servers which do not sort search results (e.g., poczta.o2.pl)
> * fixed CopyArrivalDate on platforms without glibc
> * fixed useless SASL warnings with certain plugins
> * the perl 5.14 requirement is now made explicit
> * improved OpenBSD support
> * fixed a bunch of compiler warnings

### DIFF
--- a/community/isync/APKBUILD
+++ b/community/isync/APKBUILD
@@ -1,20 +1,17 @@
 # Contributor: Stuart Cardall <developer@it-offshore.co.uk>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=isync
-pkgver=1.3.0
-pkgrel=5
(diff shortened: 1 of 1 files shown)


itoffshore:
@maxice8 - rebased with changes

itoffshore:
the `CHANGES` file is wrong then - rebased again with `GPL-3.0-or-later`

PR #8304, opened 2019-05-30
main/shorewall-core: improve

tcely:
No changes to the built package should result from this commit.


ncopa:
see comments from #8294 about the `_ver_maj_min` function. I think something like 49097d22295a841ad109865d72c6fee1cf133dde should make things clearer without breaking anything.

PR #8318, opened 2019-05-30
community/php7: security upgrade to 7.2.19

andypost:
Ref https://www.php.net/archive/2019.php#id2019-05-30-2

CVEs not posted yet


andypost:
Other branches needs backport

PR #8335, opened 2019-05-31
testing/proj4: add java-proj4 subpackage

hjaekel:
* add subpackage for java bindings
* add check()
* remove cd "$builddir"
* remove empty variables
* remove || return 1

no abi changes, so no rebuild of dependent packages

limiting arch since openjdk10 is only available on those arches


hjaekel:
> on makedepends= have the ending double quote on a newline for nicer diffs in the future

done

PR #8339, opened 2019-06-01
main/dhcp: fix circular dependencies

tcely:


tcely:
I just forgot in this case. Other times a rebase wipes out that change.

PR #8348, opened 2019-06-01
testing/sequoia-sqv: upgrade to 0.7.0

tcely:


testing/sequoia-sqv/APKBUILD
```diff
@@ -31,7 +31,10 @@ _cargo_env() {
 build() {
 	_cargo_env
 
-	sed -i -e '/^version =/s/=.*$/= "'"$pkgver"'"/' sqv/Cargo.toml
+	sed -i.orig -e '/^version =/s/=.*$/= "'"$pkgver"'"/' sqv/Cargo.toml
```
Ikke: Could you add a comment explaining the goal of this? Thanks.

tcely: Previous source archives did not update the version to match. This change and diff is for detecting when that happened again.

PR #8352, opened 2019-06-01
testing/sigrok-cli: modernize

maxice8:


testing/sigrok-cli/APKBUILD
```diff
@@ -1,20 +1,16 @@
 # Maintainer: Marian Buschsieweke <marian.buschsieweke@ovgu.de>
-
-pkgname="sigrok-cli"
+pkgname=sigrok-cli
```
tcely:
```suggestion
pkgname="sigrok-cli"
```

I'm never going to approve unquoting values for variable assignments. This is a step backward in my opinion.

PR #8028, opened 2019-05-23
testing/gnome-builder: new aport

Cogitri:

PR #8357, opened 2019-06-01
testing/gnome-calendar: new aport

Cogitri:
Needs #8028


Ikke:
This depends on libdazzle, which apparently is missing.

Cogitri:
See #8028 which includes that (as noted).

maxice8:
ping

Cogitri:
Pushed new version.

PR #8358, opened 2019-06-01
testing/gnome-characters: new aport

Cogitri:


Ikke:
Disabled on s390x because gjs is missing.

Cogitri:
Ah, thanks for the heads-up. Really unfortunate that wr don't have CI for S390x and PPC64 :c

PR #8368, opened 2019-06-01
main/hiredis: bump to build on armv7

Ikke:
hiredis has not been built on armv7, while there should be nothing
preventing it from being built. greenbone-security-assistant
depends on hiredis and is now failing to build on armv7.

Bump pkgrel to force a rebuild.


tcely:
Duplicates alpinelinux/aports#8365

Ikke:
Thanks, closing this one.

PR #8371, opened 2019-06-01
testing/mame: upgrade to 0.210, move to community

tanertas:
This port is in testing for a quite long time. Give a way to the community repo.


tanertas:
> cd "$builddir" on default functions can be removed

Did it for portmidi. Will do for mame for next update because I don't want to mess up with rebase.

community/mame/APKBUILD
```diff
@@ -48,11 +48,10 @@ subpackages="
 _groovymame_patch=groovymame-$pkgver.diff
 source="
 	https://github.com/mamedev/mame/archive/$pkgname$_pkgver.tar.gz
-	$_groovymame_patch::https://drive.google.com/uc?export=download&id=1_h5lcMQ3xMJlKcrh_u1CtCnQhrQgry7W
+	$_groovymame_patch::https://drive.google.com/uc?export=download&id=1K_JmYATSEwcqQx1JpD4ah42Txlg0H7bc
```
tcely: Why is this patch not committed to aports.gir instead of downloaded?

tanertas: it is actually a distribution file which gets updated so often.
and is also too big.

tanertas:
I'm closing this.

PR #8372, opened 2019-06-01
testing/gnome-maps: new aport

Cogitri:


Cogitri:
Needs #7597 for armv7/armhf

PR #8374, opened 2019-06-02
[WIP] testing/qutebrowser: upgrade to 1.6.2

maxice8:
Requires Qt5WebEngine to be fixed with musl like chromium/firefox so in the meantime you need to run with --qt-arg disable-seccomp-filter

Fails to load a page in quickstart, if someone can test would be grateful

### DIFF
--- a/testing/qutebrowser/APKBUILD
+++ b/testing/qutebrowser/APKBUILD
@@ -1,41 +1,29 @@
 # Maintainer: Drew DeVault <sir@cmpwn.com>
 pkgname=qutebrowser
-pkgver=1.5.2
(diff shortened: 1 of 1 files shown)


iggy:
I'm guessing the explicit `py3-qtwebengine` dep was never added because qutebrowser can also use qtwebkit, but since we don't package the python bindings for that, this is probably fine.

I am curious why `py3-setuptools` moved from build dep to runtime dep.

maxice8:
`/usr/bin/qutebrowser` makes uses of `from pkg_resources import load_entry_point`

PR #8388, opened 2019-06-02
testing/kconfigwidgets: use xvfb-run to run testsuite

maxice8:
@PureTryOut


testing/kconfigwidgets/APKBUILD
```diff
@@ -7,11 +7,12 @@ pkgdesc="Widgets for KConfig"
 arch="all"
 url="https://community.kde.org/Frameworks"
 license="LGPL-2.1-only AND LGPL-2.1-or-later"
-depends_dev="kauth-dev kcoreaddons-dev kcodecs-dev kconfig-dev kguiaddons-dev ki18n-dev kwidgetsaddons-dev"
+depends_dev="kauth-dev kcoreaddons-dev kcodecs-dev kconfig-dev kguiaddons-dev
+	ki18n-dev kwidgetsaddons-dev"
```
PureTryOut: I personally prefer all dependencies on a single line, my editor will break the line internally.

maxice8: That is weird, i cut at 80 columns.

Ikke: codestyle is to hard wrap at ~80 characters (if reasonable)

PR #8400, opened 2019-06-02
community/py3-python-gssapi: new aport

dimon222:
This package offers quite comprehensive interface for RFC 2743/2744. It means most of functional wrappers around C-binding of Kerberos. Tests are disabled because dependent test packages are missing from Alpine + they're not even confirmed working for latest versions of pythons 2.7/3.6+ (check forbiddenfruit). They might be added later in the future if I get them to compile and run tests.

### DIFF
--- a/community/py3-python-gssapi/APKBUILD
+++ b/community/py3-python-gssapi/APKBUILD
@@ -1,0 +1,30 @@
+# Contributor: Dmitry Romanenko <dmitry@romanenko.in>
+# Maintainer: Dmitry Romanenko <dmitry@romanenko.in>
+pkgname=py3-python-gssapi
+_pkgname=gssapi
+pkgver=1.5.1
+pkgrel=0
+pkgdesc="A Python interface to RFC 2743/2744 (plus common extensions)"
+url="https://github.com/pythongssapi/python-gssapi"
+arch="all"
+license="ISC"
+depends="python3"
+#checkdepends="py-nose py-shouldbe py-k5test" # Tests are missing alpine packages
+makedepends="python3-dev krb5-dev py-setuptools"
+subpackages=""
+source="https://files.pythonhosted.org/packages/source/g/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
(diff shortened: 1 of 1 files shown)


community/python-gssapi/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+depends=""
+#checkdepends="py-nose py-shouldbe py-k5test" # Tests are missing alpine packages
+makedepends="python2-dev python3-dev krb5-dev py-setuptools"
+subpackages="py3-$_pkgname:_py3 py2-$_pkgname:_py2"
+source="https://files.pythonhosted.org/packages/source/g/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
+	cd "$builddir"
+	python2 setup.py build
```
Ikke: Do you require python2 support for some reason? Python2 is soon eol, so we discourage new packages being built for python2.

community/python-gssapi/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+license="ISC"
+depends=""
+#checkdepends="py-nose py-shouldbe py-k5test" # Tests are missing alpine packages
+makedepends="python2-dev python3-dev krb5-dev py-setuptools"
+subpackages="py3-$_pkgname:_py3 py2-$_pkgname:_py2"
+source="https://files.pythonhosted.org/packages/source/g/$_pkgname/$_pkgname-$pkgver.tar.gz"
+builddir="$srcdir"/$_pkgname-$pkgver
+
+build() {
+	cd "$builddir"
```
Ikke: This can be removed, as this is being done automatically.

community/python-gssapi/APKBUILD
dimon222: Honestly, I just used the typical metapackage template. If this is requirement from now on to only publish python3 packages, no problem, I'll adjust

Ikke: `newapkbuild -y` is already a python3 only template.

dimon222:
@Ikke I applied suggested template, check now.

PR #8412, opened 2019-06-02
community/openjdk9-11: Move from testing

bratkartoffel:
OpenJDK 9 and 10 may be removed once 11 is built.


J0WI:
Can OpenJDK 11 build itself or do we need to keep 10 for updates?

bratkartoffel:
Every JDK can be built either with itself or the previous major version.
So once OpenJDK11 is built and available in community, I'll:
* update the APKBUILD and replace makedepends "openjdk10" with "openjdk11"
* remove openjdk9 and openjdk10 (EOL already)

bratkartoffel:
depends on #8448

ncopa:
> OpenJDK 9 and 10 may be removed once 11 is built.

How do we rebuild the repo from scratch when we build `3.11`?

bratkartoffel:
> 
> 
> > OpenJDK 9 and 10 may be removed once 11 is built.
> 
> How do we rebuild the repo from scratch when we build `3.11`?

This is actually a really interesting question. I have no experience with this kind of issue, how can we deal with it?

PR #8414, opened 2019-06-02
testing/geos: upgrade to 3.7.2

hjaekel:
please do not merge, I want to see on which platforms geos fails


maxice8:
There are 3 commits with the same name

hjaekel:
I think I'm finished now. You can merge, if there are no further review comments.

PR #8425, opened 2019-06-02
testing/gnome-shell: use xvfb-run to run testsuite

maxice8:
Sadly the tests require on gdm which depends on gnome-shell so we can't
enable them, but still, individual maintainers when updating gnome-shell
can make use of an already compiled gdm to run the testsuite.

@Cogitri


Cogitri:
Thanks!

PR #8437, opened 2019-06-03
community/flatpak: enable on aarch64

tcely:
Addresses the build failure for alpinelinux/aports#8028


Ikke:
@Cogitri Is there something changed that flatpak is now working on ppc64le? I see no explanation in the commit message. 

It was disabled in d7ecd27f9f without any explanation either.

Cogitri:
I think you're talking about aarch64?

Ikke:
@Cogitri Yes, sorry, not sure how I got to ppcl64le :confused:. I wonder why it was disabled and now seems to work again.

Cogitri:
Hm, Sadly the commit msg doesn't tell us why it was broken :c

PR #8451, opened 2019-06-04
testing/py3-forbiddenfruit: new aport

dimon222:
This is one of dependencies for [should_be](https://github.com/DirectXMan12/should_be/blob/master/setup.py) which is often used for tests.


dimon222:
@maxice8 applied sugggested corrections

PR #8459, opened 2019-06-04
 Added aarch64 support

ossdev07:
Architecture specific condition has been added for aarch64 in the packages -

community/cassandra-cpp-driver
community/ddrescue
community/elasticsearch
community/john


community/ddrescue/APKBUILD
```diff
@@ -4,7 +4,7 @@ pkgver=1.24
 pkgrel=0
 pkgdesc="Data recovery tool for block devices with errors"
 url="https://www.gnu.org/s/ddrescue/ddrescue.html"
-arch="all !aarch64"
+arch="all aarch64"
```
PureTryOut: "All" already includes aarch64 when it's not explicitely excluded like before.

ossdev07: We have done the required changes in the latest commit. Please have a look at it.
Will take care of the PR title from now onwards, thanks for letting us know. :smiley:

PureTryOut: Seems good, please squash your commits into one though.

I wonder, who is "we"?

bratkartoffel:
I'm unsure of this PR. Why was aarch64 disabled before? Can you ensure that it works now?

Commits responsible disabling aarch64 for these packages:
* ddrescue: https://github.com/alpinelinux/aports/commit/c6e3fd1dd700ae56ad5dbb9ffefb2df3a761dde6
* cassandra-cpp-driver: https://github.com/alpinelinux/aports/commit/b6b191cb348faa02bf6dee681863420febea0483
* elasticsearch: was never enabled
* john: was never enabled

tcely:
@ossdev07 Thank you for contributing!

I have some thoughts on both this PR and the responses to it.

1. The title is not as helpful as I would like it to be. This is not a big deal, I figured out what was going on from the changes.
1. I would have made a commit per APKBUILD change, squashing to one commit makes no sense to me.
1. It does not matter why `aarch64` was disabled. If it builds and checks pass now, we should enable it.
1. We don't ensure packages are bug-free, we build them, then deal with issues that are filed when those come in.
1. Please be sure to check DroneCI results as those cover more architectures.
1. If it passes CI, I think there should be a good reason given not to enable the architecture rather than a reason to enable it.

ossdev07:
Hi, When can we expect the PR to get merge?

ncopa:
we will not merge a PR with a commit that enables aarch64 and then disables it again in the next commit. Instead, squash those two so we have on singe commit that does the right thing on first try.

The commit message for the first commit is also misleading. It enables aarch64 for more packages.

I will clean it up for you this time. Thanks!

PR #8460, opened 2019-06-04
community/flatpak-builder: fix url, disable tests

maxice8:


maxice8:
@Cogitri

Cogitri:
fwiw the tests failed for me with the same error as Travis

community/flatpak-builder/APKBUILD
```diff
@@ -24,15 +26,9 @@ build() {
 		--libexecdir=/usr/lib/$pkgname \
 		--disable-static \
 		--with-dwarf-header=/usr/include/libdwarf
-
 	make
 }
 
-check() {
-	#make -k check
```
misery: Why remove it? Should be fixed someday instead of "ignore it"

maxice8: @fcolista (if i remember correctly) said it causes confusion to have check() defined with check disabled in options.

Cogitri: Yup, @TBK also mentioned that

PR #8462, opened 2019-06-04
testing/shunit2: new aport

oliv3r:
shUnit2 is a xUnit based unit test framework for Bourne based shell scripts.

### DIFF
--- a/testing/shunit2/APKBUILD
+++ b/testing/shunit2/APKBUILD
@@ -1,0 +1,27 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname="shunit2"
+pkgver="2.1.7"
+pkgrel=0
+pkgdesc="A xUnit based unit test framework for Bourne based shell scripts."
+url="https://github.com/kward/shunit2/"
+arch="noarch"
+license="Apache-2.0"
+makedepends="bash dash loksh zsh"
+subpackages="${pkgname}-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/kward/shunit2/archive/v${pkgver}.tar.gz"
+
+check() {
+	_tmpdir="${TMPDIR:-/tmp/}"
+	TMPDIR="$(mktemp -d -p "${_tmpdir}" shunit2.XXXXXXXX)" ./gen_test_report.sh
+	rm -rf "${_tmpdir}/shunit2.*"
+}
+
+package() {
+	install -D -m 644 -t "${pkgdir}/usr/bin/" \
+		"shunit2"
+	install -D -m 644 -t "${pkgdir}/usr/share/doc/${pkgname}/" \
+		"README.md"
+}
+
+sha512sums="35453cf2c469eb1bab2fff0624b8037a09d405c377fec0938875a6339749d5934fe246eec5691a121290ae9e4ee5cbf7768b740c5f0010a5345c4e111c50fbbd  shunit2-2.1.7.tar.gz"


testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname=shunit2
+pkgver="2.1.7"
+pkgrel=0
+pkgdesc="shUnit2 is a xUnit based unit test framework for Bourne based shell scripts."
```
tcely:
```suggestion
pkgdesc="A xUnit based unit test framework for Bourne based shell scripts."
```

tcely:
@oliv3r Thanks for your contribution!

FYI, the customary commit message is:

```
testing/<NAME>: new aport

<URL>
<pkgdesc>
```

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname=shunit2
```
tcely:
```suggestion
pkgname="shunit2"
```

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+pkgrel=0
+pkgdesc="shUnit2 is a xUnit based unit test framework for Bourne based shell scripts."
+url="https://github.com/kward/shunit2/"
+arch="noarch"
+license="Apache-2.0"
+subpackages="${pkgname}-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/kward/${pkgname}/archive/v${pkgver}.tar.gz"
+
+check()
+{
```
tcely: Please put the opening brace on the same line as the function name.

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+	TMPDIR="$(mktemp -d -p "/tmp/" shunit2.XXXXXXXX)" ./gen_test_report.sh
+	rm -rf "/tmp/shunit2.*"
+}
+
+package()
+{
```
tcely: `package() {`

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc="shUnit2 is a xUnit based unit test framework for Bourne based shell scripts."
+url="https://github.com/kward/shunit2/"
+arch="noarch"
+license="Apache-2.0"
+subpackages="${pkgname}-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/kward/${pkgname}/archive/v${pkgver}.tar.gz"
+
+check()
+{
+	TMPDIR="$(mktemp -d -p "/tmp/" shunit2.XXXXXXXX)" ./gen_test_report.sh
```
tcely:
```suggestion
	local tempdir
	tempdir="$(mkdir -p "$TMPDIR"; mktemp -d -t shunit2.XXXXXXXX)"
	TMPDIR="$tempdir" ./gen_test_report.sh
```

TMPDIR could already be set (even to a non-existent path), so we shouldn't hard-code `/tmp` here.

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+	rm -rf "/tmp/shunit2.*"
```
tcely:
```suggestion
	rm -rf "${tempdir}"
```

testing/shunit2/APKBUILD
oliv3r: I just copy pasted the description from the website, but I do agree that this reads nicer.

testing/shunit2/APKBUILD
oliv3r: The irony, I'm usually the one that falls over this :) Sorry, I missed that.

testing/shunit2/APKBUILD
oliv3r: I'd rather not use local, as that is a bashism specific thing (future proofing). However I fully agree that TMPDIR could be set, and we should only set it if it is not set.

testing/shunit2/APKBUILD
oliv3r: I see both styles used in the code base, and have always used this style on my previous contributions. As a C programmer, it obviously makes more sense :)

I tend to see this style used more in POSIX compliant/portable non-bash shell scripts; but true, it's a mixed bag.

So since this is more a reviewer's choice, and don't see it implemented as a linting rule checker of the test suite/rule written down, I'll leave it as is for now

maxice8: **use** the style as pointed by @tcely

[Thank you](https://github.com/maxice8/atools/blob/master/alint.5.scd#L247)

oliv3r:
> For [0876061](https://github.com/alpinelinux/aports/commit/0876061ac7ff8ab0fec9d78439130674557e93ea)
> 
> pkgver doesn't need to be quoted
Yeah but it's very poor/sloppy style to not.
> functions have opening brace in the same line
as per reply
> don't need to install LICENSE
Not sure if there is a strict 'need', but since its there, a common decency. I understand from the alpine 'as small as possible' point of view you want this to be at least part of the -doc package, and not the main package, Debian however for example requires it to be installed at all times, due to legal reasons. So I can imagine, that in the future Alpine would even require it also (due to legal reasons). So at least be prepared and have it in the -doc ready (as do others).

oliv3r:
Many thanks for the quick review guys, and sorry for missing a few things! Much appreciated!

maxice8:
> > For [0876061](https://github.com/alpinelinux/aports/commit/0876061ac7ff8ab0fec9d78439130674557e93ea)
> > pkgver doesn't need to be quoted
> 
> Yeah but it's very poor/sloppy style to not.
> 

It isn't.

> > functions have opening brace in the same line
> 
> as per reply
> 

**Please** apply the suggestions

testing/shunit2/APKBUILD
oliv3r: Thank you for your link, but this points to your personal repository. Can you point me to authoritative unbiased information? E.g. meeting minutes or a quote from @ncopa? So far I still see both styles in the repo, and this linter doesn't seem to run during PR's?

maxice8: [Here](https://github.com/alpinelinux/aports/blob/master/CODINGSTYLE.md)

testing/shunit2/APKBUILD
```diff
@@ -0,0 +1,29 @@
+# Contributor: Olliver Schinagl <oliver@schinagl.nl>
+# Maintainer: Olliver Schinagl <oliver@schinagl.nl>
+pkgname="shunit2"
+pkgver="2.1.7"
+pkgrel="0"
```
maxice8: don't quote literal integers

testing/shunit2/APKBUILD
oliv3r: you do realize that I instantiated that right? :) So that was kind of a low-blow :p

maxice8: I know that very well, feel free to ping when the changes are done

maxice8:
ping

maxice8:
ping

oliv3r:
>     * [x]  adjust commit message
> 
>     * [x]  quote pkgname
> 
>     * [x]  adjust description
> 
>     * [ ]  function style changes
> 
>     * [ ]  save temporary directory in a variable then remove it later
@tcely done

tcely:
@oliv3r you are still using a glob match to remove the temporary directory.

oliv3r:
@tcely and a bad one  at that (it's inside the quote). Sorry for missing that. I missed that. Looks like it has been merged. I think it should fail, as it will try to remove the string 'shunit.*' and so leaves the cruft behind. I'll send a hotfix now; as we can just nuke the entire _tmpdir due to it being generated as a subdir anyway. Again, sorry for missing it, i'll open the new PR on gitlab though.

PR #8467, opened 2019-06-04
testing/qtchooser: new aport

PureTryOut:
QtChooser is normally meant to allow the user to switch between several installations of Qt easily (including Qt4), by symlinking all Qt binaries from the actually installed location (`/usr/lib/qt5/bin` for the distro packages) to `/usr/bin`. Currently all Qt binaries are available with a `-qt5` suffix, but some software (most notably KDE Plasma) expects it without the suffix. Rather than patching all the individual applications, this is the proper solution.

Note that this can work with Qt4 quite easily, but my understanding is that we should try to get rid of Qt4 instead and not make any effort in improving it.

### DIFF
--- a/community/qt5-qtbase/APKBUILD
+++ b/community/qt5-qtbase/APKBUILD
@@ -7,11 +7,11 @@ _ver=${_ver/_/-}
 _ver=${_ver/beta0/beta}
 _ver=${_ver/rc0/rc}
 _V=${_ver/rc/RC}
-pkgrel=0
+pkgrel=1
 pkgdesc="Qt5 - QtBase components"
-url="http://qt-project.org/"
+url="https://qt.io/developers/"
 arch="all"
-license="LGPL-2.0 with exceptions or GPL-3.0 with exceptions"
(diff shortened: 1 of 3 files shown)


community/qt5-qtbase/APKBUILD
```diff
@@ -7,13 +7,14 @@ _ver=${_ver/_/-}
 _ver=${_ver/beta0/beta}
 _ver=${_ver/rc0/rc}
 _V=${_ver/rc/RC}
-pkgrel=0
+pkgrel=1
 pkgdesc="Qt5 - QtBase components"
 url="http://qt-project.org/"
```
maxice8: While you're at it fix the url https://www.qt.io/developers/

community/qt5-qtbase/APKBUILD
```diff
@@ -7,13 +7,14 @@ _ver=${_ver/_/-}
...
 arch="all"
 license="LGPL-2.0 with exceptions or GPL-3.0 with exceptions"
```
maxice8: those are not SPDX identifiers but whatever

PureTryOut: I _think_ I corrected them properly, but someone with more knowledge of this will have to verify.

PR #8472, opened 2019-06-04
community/mumble: update to 1.3.0_rc1 and move to Qt5

PureTryOut:
The update to 1.3.0_rc1 is required to get it building with the latest libraries in Alpine and Qt5.

CC @jomat


community/mumble/APKBUILD
```diff
@@ -2,66 +2,51 @@
 # Contributor: Johannes Matheis <jomat+alpinebuild@jmt.gr>
 # Maintainer: Johannes Matheis <jomat+alpinebuild@jmt.gr>
 pkgname=mumble
-pkgver=1.2.19
-pkgrel=9
+pkgver=1.3.0_rc1
+_pkgver=${pkgver/_/-}
+__pkgver=${pkgver/_rc1/}
```
5paceToast: This will change whenever rc2 comes out.
Perhaps use wildcards instead?
```suggestion
__pkgver=${pkgver/_rc*/}
```

community/mumble/APKBUILD
```diff
@@ -2,66 +2,51 @@
...
-	CVE-2018-20743.2.patch
-	boost-1.66.patch
+source="https://github.com/mumble-voip/$pkgname/releases/download/$_pkgver/$pkgname-$_pkgver.tar.gz
 	murmur.initd"
-builddir="$srcdir"/$pkgname-$pkgver
-
-# secfixes:
-#   1.2.19-r9:
-#     - CVE-2018-20743
+builddir="$srcdir/$pkgname-${pkgver/_rc1/}"
```
5paceToast: Shouldn't this be `$__pkgver`?
```suggestion
builddir="$srcdir/$pkgname-$__pkgver"
```

PureTryOut: No. The `_rc1` tarball contains the directory without the `_rc1` suffix. Not sure why, but that is how it is

PR #8475, opened 2019-06-04
community/automoc4: remove

PureTryOut:
CC @fcolista, do you mind me taking over maintainership of attica? It's part of KDE Frameworks and I'm maintaining the rest of it since last and this month.


maxice8:
Might as well remove unmaintained/tomahawk since it depends on it.

PR #8481, opened 2019-06-05
testing/py3-shouldbe: new aport

dimon222:
Depends on #8451


dimon222:
@maxice8 adjusted according to suggestions

PR #8514, opened 2019-06-05
testing/ostinato: update to latest Git and move to Qt5

PureTryOut:
This should be the last of the Qt4 dependent packages :tada:

Thanks to SpaceToast on IRC on helping it get to compile.

The update to a Git version is required for Qt5 support, as no new release with support for it has been made yet.

CC @little-dude

### DIFF
--- a/testing/ostinato/ModelTest.patch
+++ b/testing/ostinato/ModelTest.patch
@@ -1,0 +1,18 @@
+Undefined reference to ModelTest::ModelTest under defaults.
+Clearly, we're not in a debug mode, but the sources think we are.
+
+Patch is not upstream, made by SpaceToast on #alpine-devel
+diff -ur a/client/ostinato.pro b/client/ostinato.pro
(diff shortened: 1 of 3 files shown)


testing/ostinato/ModelTest.patch
```diff
@@ -0,0 +1,16 @@
+Undefined reference to ModelTest::ModelTest under defaults.
+Clearly, we're not in a debug mode, but the sources think we are.
+diff -ur a/client/ostinato.pro b/client/ostinato.pro
```
Cogitri: Is this upstream?

PureTryOut: No, the patch is made by SpaceToast

Cogitri: Maybe it'd be a good idea mentioning that in the patch header then, makes it easier for other people to touch this :)

PR #8529, opened 2019-06-05
testing/wgetpaste: new aport

Cogitri:


testing/wgetpaste/APKBUILD
```diff
@@ -0,0 +1,30 @@
+# Contributor: Rasmus Thomsen <oss@cogitri.dev>
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
+pkgname=wgetpaste
+pkgver=2.28
+pkgrel=0
+pkgdesc="Command-line interface to various pastebins"
+url="http://wgetpaste.zlin.dk/"
+arch="noarch"
+license="None"
```
maxice8: Public Domain

Cogitri: But that's not listed in SPDX either and there's no license file attached :c

maxice8: i think it should be fine, public domain isn't a license anyways

Cogitri: I'm currently working with upstream to get this fixed, I'll push again then.

PR #8532, opened 2019-06-05
main/binutils: security upgrade to 2.32

J0WI:


Ikke:
These security fixes might need to be backported as well? 3.9 has binutils-2.31, 3.8 has binutils-2.30.

J0WI:
Only 2.31 is affected.

PR #8537, opened 2019-06-06
community/py-hypothesis: upgrade to 4.24.2

maxice8:


maxice8:
@ddevault

ddevault:
Do you want to adopt this package?

maxice8:
I wouldn't mind

maxice8:
@ddevault updated to include maintainership change.

maxice8:
Thanks for the quick response and taking care of py-hypothesis.

PR #8538, opened 2019-06-06
Correct PowerDNS licenses

tcely:
Fixes https://bugs.alpinelinux.org/issues/10536

/cc @mattx86


Ikke:
@prspkt already changed this, but without bumping the pkgrel:

* 956c729
* beaddd0
* 69c9165

tcely:
Yes, are you or @prspkt going to correct that oversight?

prspkt:
I intentionally did not increment _pkgrel_ to avoid rebuilding and upgrading client systems with the (3) affected packages for a license change.

PR #8550, opened 2019-06-06
main/syslog-ng: Provided support for aarch64

ossdev07:
Architecture specific condition has been added for aarch64.


ossdev07:
Hi, When can we expect the PR to get merge?

PR #8555, opened 2019-06-06
testing/bpftrace: new aport

kaey:


testing/bpftrace/APKBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: Konstantin Kulikov <k.kulikov2@gmail.com>
+pkgname=bpftrace
+pkgver=0.9
+pkgrel=0
+pkgdesc="High-level tracing language for Linux eBPF"
+url="https://github.com/iovisor/bpftrace"
+arch="x86_64"
+license="Apache-2.0"
+makedepends="build-base cmake llvm7-dev llvm7-static clang-dev clang-static flex-dev bison elfutils-dev linux-headers bcc"
```
maxice8: Please cut this line at 80 columns

testing/bpftrace/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+source="$pkgname-$pkgver.tar.gz::https://github.com/iovisor/bpftrace/archive/v$pkgver.tar.gz"
+# Tests require root.
+options="!check"
```
maxice8: can't the tests be ran with the `checkroot` option ?

testing/bpftrace/APKBUILD
maxice8: and build-base isn't required

kaey: done

testing/bpftrace/APKBUILD
kaey: I've updated comment.
It's possible, but requires more work - download gmock, patch some tests and disable others (which depend on kernel config for ex).
I'll do all this, but a bit later.

maxice8: understandable

PR #8559, opened 2019-06-07
main/nfs-utils: fix nfsidmap nss module

jantatje:


ncopa:
LGTM. Thanks!

is this reported upstream?

PR #8566, opened 2019-06-07
testing/libtorrent-rasterbar: use py3 for python bindings

maxice8:


maxice8:
@AmatCoder

maxice8:
This will temporarily break deluge until deluge-2.0.0 is updated on #8567

PR #8571, opened 2019-06-07
testing/py3-automat: modernize, switch to py3

maxice8:
Needs:

- [x] #8570


maxice8:
@myhro

myhro:
Hi @maxice8 ,

I'm not sure I understood the purpose of this change. Are Python 2 packages being dropped from Alpine? Are there any functional differences between the current [py3-automat][1] and the proposed one other that?

Regards,
Tiago.

[1]: https://pkgs.alpinelinux.org/package/edge/testing/x86_64/py3-automat

maxice8:
Yes, python2 will become EOL soon and Python Policy is to remove python2 as soon as convenient. https://wiki.alpinelinux.org/wiki/Python_package_policies#Python_3_only

No there are no functional differences from it and the previous py3 except now there is a py3-twisted available for it to depend on.

If you wish i can keep as is

myhro:
Oh, that's great. Thanks for the explanation. I have no merge powers, but the PR looks fine to me.

Also, feel free to add yourself to the `Contributor` list if you like.

maxice8:
> Oh, that's great. Thanks for the explanation. I have no merge powers, but the PR looks fine to me.

No worries, i'll merge it once its last consumer, deluge, switches over to 2.0.0 which is python3

PR #8574, opened 2019-06-07
testing/gxi: new aport

Cogitri:
Needs #8573


Cogitri:
Seems like gtk-rs is especially good at triggering SIGSEVs in Rust.

PR #8580, opened 2019-06-07
testing/mutter: fix cogl path build order and re-enable ppc64le

mksully22:
on ppc64le mutter build fails with errors of type:
```
In file included from ../cogl/cogl/cogl.h:153,
                 from ../clutter/clutter/clutter-types.h:32,
                 from ../clutter/clutter/clutter-test-utils.h:29,
                 from ../clutter/clutter/clutter-test-utils.c:3:
../cogl/cogl-path/cogl-path.h:55:10: fatal error: cogl-path/cogl-path-enum-types.h: No such file or directory
 #include <cogl-path/cogl-path-enum-types.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The error occurs because of subcomponent build ordering issues where the cogl-path component must be built before the clutter components so that the generated cogl-path-enum-types.h file is available. This PR  modifies the clutter meson.build to enforce this order.

### DIFF
--- a/testing/mutter/APKBUILD
+++ b/testing/mutter/APKBUILD
@@ -2,10 +2,10 @@
 # Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=mutter
 pkgver=3.32.2
-pkgrel=0
+pkgrel=1
 pkgdesc="clutter-based window manager and compositor"
 url="https://wiki.gnome.org/Projects/Mutter/"
-arch="all !s390x !ppc64le" # limited by gnome-settings-daemon
+arch="all !s390x" # limited by gnome-settings-daemon
 license="GPL-2.0-or-later"
 depends="xkeyboard-config zenity gsettings-desktop-schemas xorg-server-xwayland"
 makedepends="gnome-desktop-dev libcanberra-dev upower-dev json-glib-dev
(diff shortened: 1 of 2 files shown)


testing/mutter/APKBUILD
```diff
@@ -2,10 +2,10 @@
 # Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=mutter
 pkgver=3.32.2
-pkgrel=0
+pkgrel=1
 pkgdesc="clutter-based window manager and compositor"
 url="https://wiki.gnome.org/Projects/Mutter/"
-arch="all !s390x !ppc64le" # limited by gnome-settings-daemon
+arch="all !s390x !armv7" # limited by gnome-settings-daemon
```
Cogitri: Why have you added armv7 to the list here?

mksully22: drone exposed that armv7 build of mutter will fail because of a failing dependency on the libgweather apk package. Looking at the edge repository community/libgweather isn't available for armv7.

Cogitri: That's due to a temporary failure in the armv7 builder, which should he fixed "soon"™. Please leave armv7 enabled and ignore the CI failure.

mksully22: Thanks Cogitri. Sounds good. I'll update the PR and re-enable armv7.

Cogitri: Thank you :)

PR #8584, opened 2019-06-07
WIP: main/llvm-friends: bump to 8.0.0

Cogitri:
This depends on LLVM8 being moved to main/


J0WI:
There's also `community/llvm-libunwind`

Cogitri:
Oh right, missed that one. Thanks!

community/lld/APKBUILD
```diff
@@ -45,11 +45,12 @@ check() {
 		armhf|armv7) rm ../test/ELF/eh-frame-hdr-augmentation.s;;
 	esac
 
-	#make check-lld
+	# make check-lld doesn't like using system lit
```
J0WI: I was trying with `LLVM_EXTERNAL_LIT` in #8533

Cogitri: I think calling lit manually should do the same thing, but thanks for the pointer!

PR #8599, opened 2019-06-07
main/minicom: fix url

maxice8:


maxice8:
@ms13sp

ncopa:
i bumped `pkgrel`

PR #8602, opened 2019-06-07
main/curl: upgrade to 7.65.1

prspkt:
[Changelog](https://curl.haxx.se/changes.html)

### DIFF
--- a/main/curl/APKBUILD
+++ b/main/curl/APKBUILD
@@ -3,7 +3,7 @@
 # Contributor: Łukasz Jendrysik <scadu@yandex.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=curl
-pkgver=7.65.0
+pkgver=7.65.1
 pkgrel=0
 pkgdesc="URL retrival utility and library"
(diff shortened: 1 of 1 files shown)


main/curl/APKBUILD
```diff
@@ -101,12 +104,11 @@ build() {
...
 
 check() {
-	cd "$builddir"
-	make -C tests nonflaky-test
+	cd "$builddir"/tests
+	make nonflaky-test
 }
 
 package() {
-	cd "$builddir"
```
Ikke: This is a package in main, ncopa asked to leave these in main. Especially for backporting, you want to introduce as less noise as possible.

main/curl/APKBUILD
```diff
@@ -101,12 +104,11 @@ build() {
 }
 
 check() {
-	cd "$builddir"
-	make -C tests nonflaky-test
+	cd "$builddir"/tests
```
Ikke: This change does not improve a lot and only introduces noise. I would leave it out.

prspkt:
@Ikke Proposed changes applied. Thanks for reviewing.

PR #8606, opened 2019-06-07
main/mkinitfs: modernize

maxice8:


main/mkinitfs/APKBUILD
```diff
@@ -9,18 +9,15 @@ arch="all"
 license="GPL-2.0"
 # currently we do not ship any testsuite
 options="!check"
-makedepends_build=""
 makedepends_host="busybox kmod-dev util-linux-dev cryptsetup-dev linux-headers"
-makedepends="$makedepends_build $makedepends_host"
+makedepends="$makedepends_host"
```
5paceToast: If we're removing `$makedepends_build`, can't we just throw the contents of `$makedepends_host` into `$makedepends` and call it a day?

ncopa: i think those are used by `scripts/bootstrap.sh`.

PR #8611, opened 2019-06-07
testing/ttf-hack: fix url

maxice8:


maxice8:
@pickfire

maxice8:
It needs to be bumped so the package is rebuilt and users are able to query for the new url.

All changes of user-facing metadata (url, license) require a pkgrel bump to take effect

pickfire:
Then that's fine I guess.

PR #8635, opened 2019-06-08
community/slibtool: upgrade to 0.5.28, move to community

maxice8:


ncopa:
I think i want keep `slibtool` in main because I'd like to completely replace `libtool` with `slibtool` at some point. There is just not any easy way to do it currently.

PR #8659, opened 2019-06-08
community/sg3_utils: upgrade to 1.44, move from main

maxice8:


maxice8:
@clandmeter added missing email on your Contributor: tag

@ncopa

ncopa:
i would guess sg3_utils is a tool you typically would want long time support for on server installs? Its not causing much maintenance work in stable branch either, so I'm curious why we want move it to community?

maxice8:
On Fri, Jun 28, 2019 at 08:03:35AM -0700, Natanael Copa wrote:
> i would guess sg3_utils is a tool you typically would want long time support for on server installs? Its not causing much maintenance work in stable branch either, so I'm curious why we want move it to community?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/8659#issuecomment-506767000

If that is the case we can keep it in main/

PR #8666, opened 2019-06-08
testing/gnome-maps: upgrade to 3.32.2.1

Cogitri:
* run tests


Cogitri:
armv7 fails due to libgweather not being present on the builders

PR #8628, opened 2019-06-07
community/rust: upgrade to 1.35.0

Cogitri:
* use llvm8

Needs #8620

PR #8678, opened 2019-06-08
testing/bat: new aport

5paceToast:
bat is a cat(1) clone that acts like standard cat when the output is not
a tty.
If the output *is* a tty, it provides syntax highlighting and integrated
paging on top of the typical cat(1) features.

### DIFF
--- a/testing/bat/APKBUILD
+++ b/testing/bat/APKBUILD
@@ -1,0 +1,33 @@
+# Contributor: Chloe Kudryavtsev <toast@toastin.space>
+# Maintainer: Chloe Kudryavtsev <toast@toastin.space>
+pkgname=bat
+pkgver=0.11.0
+pkgrel=0
+pkgdesc="A cat(1) clone with wings"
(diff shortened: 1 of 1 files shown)


testing/bat/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+pkgdesc="A cat(1) clone with wings"
+url="https://github.com/sharkdp/bat"
+arch="x86_64" # limited by rust/cargo
+license="Apache-2.0"
+makedepends="cargo"
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/sharkdp/bat/archive/v$pkgver.tar.gz"
+
+export CARGO_HOME="$srcdir"/cargo
+export RUSTFLAGS="-C target-feature=-crt-static"
```
Cogitri: Does our Rust not do this by default? I mean that shouldn't block this PR, but means I'd have to adjust #8628

PR #8699, opened 2019-06-09
unmaintained/llvm7: move from community

J0WI:
https://lists.llvm.org/pipermail/llvm-announce/2019-May/000083.html


Ikke:
Fails with:

> \>\>\> llvm7-libs*: Running split function libs...
> mv: can't rename '/home/buildozer/aports/main/llvm7/pkg/llvm7/usr/lib/llvm7/lib/libLLVM-7.so': No such file or directory

Ikke:
According to ncopa, this is an ABI breaking change, so all dependencies need to be rebuilt.

ncopa:
we upgraded to llvm8 instead. I suppose we could upgrade llvm7 and remove the other commits?

7.1 breaks ABI so we'd need to rebuild anything linked to libLLVM

ncopa:
I get this build failure:
````
>>> llvm7-libs*: Running split function libs...
mv: can't rename '/home/ncopa/aports/community/llvm7/pkg/llvm7/usr/lib/llvm7/lib/libLLVM-7.so': No such file or directory
````

J0WI:
Are there any packages left depending on LLVM7?

Ikke:
@J0WI atm, only bpftrace

> testing/bpftrace/APKBUILD
> 9:makedepends="cmake llvm7-dev llvm7-static clang-dev clang-static

J0WI:
`bpftrace` have been updated to llvm8, so llvm7 can be removed.

PR #8700, opened 2019-06-09
testing/sd: new aport

5paceToast:
Fast and intuitive sed-like.
Useful in most cases, when sed/awk are overkill.


5paceToast:
Updates in force push:
1. A new major version has come out (0.6.4, from 0.5.x)
2. It now generates shell completions and a man page
3. Add support for rootbld (rust packages require option_has net)
4. They seem to have fixed the strangeness with their release process/tarball naming

PR #8705, opened 2019-06-09
main/git: upgrade to 2.22.0

J0WI:


main/git/APKBUILD
```diff
@@ -10,17 +10,16 @@
...
 pkgdesc="Distributed version control system"
-url="https://www.git-scm.com"
+url="https://www.git-scm.com/"
 arch="all"
 license="GPL-2.0-or-later"
 depends=""
 # we need tcl and tk to be built before git due to git-gui and gitk
 makedepends="zlib-dev openssl-dev curl-dev expat-dev perl-dev python2-dev
 	pcre2-dev asciidoc xmlto perl-error tcl tk"
-options="!check" # Check is currently failing on t3900-i18n-commit.sh ISO-2022-JP.txt
```
Ikke: Any reason you removed this?

andypost: > >>> WARNING: git: APKBUILD does not run any tests!

Means it needs `check()`

Ikke: Like the comment that got removed says, some checks are failing. We could perhaps see if we can exclude those specific checks though.

I've been working on getting all tests green for Alpine, but not there yet.

J0WI: I forgot to readd this line. UTF-16 checks are not passing.
Changes of this PR have been applied in b7386174f378d261c2503b73aabebf079b27238a and 8778e17f073a8ba0c242d15f9c50c69918a66fc3.

PR #8712, opened 2019-06-10
main/openrc: Fix respawn_max=0

5paceToast:
This is an off-by-one error fixed in April.
Latest release is in February.
Ths adds a patch for the interim with a note on when to remove it.


maxice8:
pkgrel needs to be bumped for changes to take effect

PR #8717, opened 2019-06-10
testing/kde-frameworks: update to 5.59.0

PureTryOut:
I very much doubt this will succeed in the CI within the output or 1 hour time limit.

The previous `-DCMAKE_BUILD_TYPE=RelWithDebugInfo` was incorrect and an invalid value, it should be `-DCMAKE_BUILD_TYPE=RelWithDebInfo`.

### DIFF
--- a/community/attica/APKBUILD
+++ b/community/attica/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor:
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=attica
-pkgver=5.58.0
-pkgrel=1
+pkgver=5.59.0
+pkgrel=0
 pkgdesc="Freedesktop OCS binding for Qt"
 url="http://www.kde.org/"
 arch="all"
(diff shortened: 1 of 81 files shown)


Ikke:
Current test failures don't seem to be due to time. It might be that the community package is causing trouble (while the rest is in testing).

PureTryOut:
```
CMake Warning at CMakeLists.txt:7 (find_package):
  Could not find a configuration file for package "ECM" that is compatible
  with requested version "5.59.0".

  The following configuration files were considered but not accepted:

    /usr/share/ECM/cmake/ECMConfig.cmake, version: 5.58.0



-- 
 * ECM (required version >= 5.59.0), Extra CMake Modules., <https://projects.kde.org/projects/kdesupport/extra-cmake-modules>

CMake Error at /usr/share/cmake/Modules/FeatureSummary.cmake:457 (message):
  feature_summary() Error: REQUIRED package(s) are missing, aborting CMake
  run.
  CMakeLists.txt:9 (feature_summary)
```

You're right, for some reason it doesn't pick up `extra-cmake-modules`. It did build that one successfully though, why would it matter being in a different repository? As long as it's not community depending on something from testing, this should work should it not?

maxice8:
is it OK to have everything in a single commit ?

PureTryOut:
I mean, we've done that with Qt updates and other stuff before. I'm not really willing to do 81 separate commits every month KDE Frameworks updates.

maxice8:
> I mean, we've done that with Qt updates and other stuff before.

Sounds fine by me then.

PR #8719, opened 2019-06-10
testing/proj4: add -static and -utils subpackages

hjaekel:


kelvinn:
Quick note if anybody upgrades from 6.1.0-r0 to 6.1.0-r2, and immediately see this:

`Proj executable not found. Please set PROJ_DIR variable.`

The `proj` CLI can be installed by running

`apk add proj4-util`

PR #8724, opened 2019-06-10
testing/gdm: with deps fixed re-enable for ppc64le

mksully22:
with mutter, gnome-shell, and gnome-shell deps fixed, re-enable ppc64le and bump pkgrel to force rebuild.


Ikke:
@mksully22 Any particular reason for the bump? Just enabling an arch where a specific package was not built yet does not need a pkgrel bump.

mksully22:
I thought the bump was necessary to have the builders kicked off? If not necessary I will remove it. Thanks Ikke...

PR #8739, opened 2019-06-11
community/micro-tetris: fix Maintainer email

maxice8:
@ncopa


community/micro-tetris/APKBUILD
```diff
@@ -1,34 +1,25 @@
-# Maintainer: Natanael Copa <ncopa@alpinepinux.org>
+# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
```
ncopa: :rofl:

PR #8754, opened 2019-06-12
testing/firefox: upgrade to 67.0.4

timbru31:
This updates Firefox to 67.0.4

### DIFF
--- a/testing/firefox/APKBUILD
+++ b/testing/firefox/APKBUILD
@@ -1,16 +1,13 @@
 # Contributor:
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
-pkgver=66.0.5
-_pkgver=$pkgver
-_xulver=$pkgver
-pkgrel=1
+pkgver=67.0.4
+pkgrel=0
 pkgdesc="Firefox web browser"
 url="https://www.firefox.com/"
 # limited by rust and cargo
 arch="x86_64"
-license="GPL LGPL MPL"
-depends=""
+license="GPL-3.0-only AND LGPL-2.1-only AND LGPL-3.0-only AND MPL-2.0"
 makedepends="
 	alsa-lib-dev
 	autoconf2.13
@@ -37,7 +34,7 @@ makedepends="
 	libvorbis-dev
 	libxt-dev
 	libxcomposite-dev
-	llvm5-dev
+	llvm8
 	mesa-dev
 	nasm
 	nodejs
@@ -54,7 +51,6 @@ makedepends="
 	zip
 	"
 
-install=""
 source="https://ftp.mozilla.org/pub/firefox/releases/$pkgver/source/firefox-$pkgver.source.tar.xz
 	stab.h
 
@@ -83,11 +79,7 @@ ldpath="$_mozappdir"
 
 prepare() {
 	default_prepare
-
 	cp "$srcdir"/stab.h toolkit/crashreporter/google-breakpad/src/
-	# https://bugzilla.mozilla.org/show_bug.cgi?id=1341234
-	mkdir -p "$builddir"/objdir
-	echo "ac_add_options BINDGEN_CFLAGS='-I/usr/include/nspr -I/usr/include/pixman-1'" >>objdir/.mozconfig
 }
 
 build() {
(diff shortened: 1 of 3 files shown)


Cogitri:
I realize this is still a WIP; but please don't merge this before #8628, the build can SIGSEV randomly with LLVM7.
Also, please enable rust-simd once that PR is merged since that significantly speeds up some ops in FF. Thanks!

mi2iec:
> Also, please enable rust-simd once that PR is merged since that significantly speeds up some ops in FF. Thanks!
 
Firefox 67.0.2 won't build if the `--enable-rust-simd` flag is enabled:

    ERROR: --enable-rust-simd does not work with Rust 1.33 or later. See https://bugzilla.mozilla.org/show_bug.cgi?id=1521249

Cogitri:
Oh, oops, read that as Rust 1.33 or earlier somehow :)

I guess we'll have to wait until FF 68.x for this then.

timbru31:
Strange. PR fails because of llvm-objdump:

```
checking for llvm-objdump... not found
--
588 | DEBUG: llvm_objdump: Trying llvm-objdump
589 | ERROR: Cannot find llvm-objdump
```

Buuut, it should be available?!

Cogitri:
It works for you locally?

mi2iec:
> Strange. PR fails because of llvm-objdump:
> 
> ```
> checking for llvm-objdump... not found
> --
> 588 | DEBUG: llvm_objdump: Trying llvm-objdump
> 589 | ERROR: Cannot find llvm-objdump
> ```
> 
> Buuut, it should be available?! https://pkgs.alpinelinux.org/contents?file=llvm-objdump&path=&name=llvm*&branch=edge&repo=community&arch=x86_64

In makedepends llvm5-dev should be changed to llvm5 (or llvm8) as llvm-objdump is not in the -dev packages.

timbru31:
```
checking for NSPR - version >= 4.21... no
--
691 | configure: error: you do not have NSPR installed or your version is older than 4.21.
```

We need a nspr package update first: https://pkgs.alpinelinux.org/package/edge/main/x86_64/nspr

testing/firefox/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor:
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
-pkgver=66.0.5
+pkgver=67.0.3
```
J0WI:
```suggestion
pkgver=67.0.4
```

testing/firefox/APKBUILD
```diff
@@ -37,7 +37,7 @@ makedepends="
 	libvorbis-dev
 	libxt-dev
 	libxcomposite-dev
-	llvm5-dev
+	llvm8
```
Cogitri: I think we still need -dev for llvm-config

Cogitri:
Please change the license to `GPL-3.0-only AND MPL-2.0 AND  LGPL-2.1-only AND LGPL-3.0-only` (see toolkit/content/license.html`.

testing/firefox/APKBUILD
timbru31: Passed without it :)

Cogitri:
Alright, could you make this a non-draft? I think this should be OK and having the security fixes in the repos would be nice.

liqa:
community/firefox-esr has cleanups that should be applied here.
https://github.com/alpinelinux/aports/blob/master/community/firefox-esr/APKBUILD
Both packages should be synced with differences only where really necessary.

liqa:
You can delete _pkgver=$pkgver and _xulver=$pkgver.
You can delete empty depends="" and empty install=""
You can drop the custom prepare (the bug mentioned is resolved upstream).
You can delete the "$CARCH" = "x86" part as firefox is not built for that arch.

There should be more cleanups as firefox is for a long time in the repos. I suspect a few other commands in the package are cruft from the past.

timbru31:
Thanks @liqa, I will apply those cleanup advises in this PR!

Cogitri:
Mind taking a look at the CI failure?

timbru31:
Done!

Cogitri:
Nice, mind squashing the commits together then? Then it looks good to me.

PR #8760, opened 2019-06-12
testing/py-zipp: new aport

prspkt:
new depends for:
- [pytest-4.6.3](https://github.com/pytest-dev/pytest/blob/4.6.3/setup.py#L6) (packaging, pluggy>=0.12, importlib-metadata>=0.12)
- [pluggy-0.12.0](https://github.com/pytest-dev/pluggy/blob/0.12.0/setup.py#L40) (importlib-metadata>=0.12)


maxice8:
Assuming there is a valid reason to introduce new py2- stuff i believe we should have the reason in the commit so people can check why it was added and if it ok to remove it.

PR #8764, opened 2019-06-12
main/mesa: enable Lima and Panfrost support

PureTryOut:
Depends on Mesa 19.1.1 (next stable release) which isn't out yet, shouldn't be merged till then.

Note that devices using the Lima or Panfrost driver also need to install `mesa-dri-kmsro`, as the former is just the renderer and the latter the display driver.


PureTryOut:
This now depends on #9153.

ncopa:
you need to bump pkgrel if you want the builders to build it.

PR #8776, opened 2019-06-13
community/openjdk9: Fix and enable build on x86, armv7 and armhf

bratkartoffel:
Built JVM seems to work (at least on my machine), although one gtest fails on x86:
``` 
[ RUN ] UninitializedDoubleElementWorkerDataArrayTest.print_summary_on_test_test_vm 
/home/buildozer/aports/community/openjdk9/src/jdk9u-jdk-9.0.4+12/hotspot/test/native/gc/g1/test_workerDataArray.cpp:277: Failure 
Value of: print_summary() 
Actual: "Test array Min: 5.0, Avg: 6.2, Max: 7.2, Diff: 2.2, Sum: 12.2, Workers: 2\n" 
Expected: print_expected_summary() 
Which is: "Test array Min: 5.0, Avg: 6.0, Max: 7.2, Diff: 2.2, Sum: 12.4, Workers: 2\n" 
[ FAILED ] UninitializedDoubleElementWorkerDataArrayTest.print_summary_on_test_test_vm (0 ms) 
```

### DIFF
--- a/community/openjdk9/APKBUILD
+++ b/community/openjdk9/APKBUILD
@@ -3,10 +3,10 @@
 pkgname=openjdk9
 pkgver=9.0.4_p12
 _pkgver=${pkgver/_p/+}
-pkgrel=1
+pkgrel=2
 pkgdesc="Oracle OpenJDK 9"
 url="https://hg.openjdk.java.net/jdk-updates/jdk9u"
-arch="all !x86 !armhf !armv7" # build fails on 32bit arches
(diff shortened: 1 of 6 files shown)


bratkartoffel:
Build fails due to timeouts. Jtreg results are (as far as i can see) similar on x86 and x86_64.

bratkartoffel:
i'll open one PR for the whole chain (9, 10 and 11)

PR #8780, opened 2019-06-13
testing/watchexec: new aport

5paceToast:
An inotify-related tool to run commands in response to filesystem changes.

### DIFF
--- a/testing/watchexec/APKBUILD
+++ b/testing/watchexec/APKBUILD
@@ -1,0 +1,32 @@
+# Contributor: Chloe Kudryavtsev <toast@toastin.space>
+# Maintainer: Chloe Kudryavtsev <toast@toastin.space>
+pkgname=watchexec
+pkgver=1.10.2
+pkgrel=0
+pkgdesc='Executes commands in response to file modifications'
+url='https://github.com/watchexec/watchexec'
+arch='x86_64' # limited by rust/cargo
+license='Apache-2.0'
+makedepends='cargo'
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/watchexec/watchexec/archive/$pkgver.tar.gz"
+
+export CARGO_HOME="$srcdir"/cargo
+
+build() {
+	cargo build \
+		--release \
+		--verbose
+}
+
+check() {
+	cargo test --all \
+		--release \
+		--verbose
+}
+
+package() {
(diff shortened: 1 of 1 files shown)


testing/watchexec/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Chloe Kudryavtsev <toast@toastin.space>
+# Maintainer: Chloe Kudryavtsev <toast@toastin.space>
+pkgname=watchexec
+pkgver=1.10.2
+pkgrel=0
+pkgdesc='Executes commands in response to file modifications'
```
maxice8: should use " " not ' ' for quoting variables

5paceToast: Is there a standard saying this?
Generally, in shell, `''` quotes mean "no substitution", while `""` mean that substitution is allowed.
As such, if anything, constants should prefer using `''` (unless there's a spec/standard/policy saying otherwise, which is why I ask).

maxice8: No, it is just what newapkbuild does by default and the overwhelming majority does as well.

5paceToast: Then I'm not sure why this is considered a blocking issue.
"should" generally implies a stronger reason than "this is common".

maxice8: It isn't blocking but is a deviation from a **super heavily** established norm

5paceToast: I don't see how a change from `'` to `"` or vice versa is a significant deviation, or a significant norm.
On the other hand, `'` is less error prone (since it explicitly outlines constants, and prevents accidental escapes).
This honestly looks like bikeshedding over ultimately inconsequential details.
Since the practice is bound to cause significant time wastage in the future, I would rather settle this definitively in a preemptive manner.

Should we perhaps establish a policy regarding quoting rules?
(as another possible target, `"$srcdir"/something` vs `"$srcdir/something"` vs `$srcdir/something` and so on)

maxice8: > Should we perhaps establish a policy regarding quoting rules?

Do what you will just write it somewhere official so i can implement on `atools`.

I personally prefer keeping double quotes because they (subjectively) are easier to distinguish at than single quote and is used by default and nearly all the time. Overall i don't care about what is decided i just want it done consistently **everywhere**.

PR #8784, opened 2019-06-13
main/linux-vanilla: add patch to export now GPL-only fpu functions

Cogitri:
This significantly speeds up ZFS' due to allowing it to use SIMD. Other distros
such as NixOS use the same patch.

Also do some minor cleanups while we're at it.

### DIFF
--- a/main/linux-vanilla/APKBUILD
+++ b/main/linux-vanilla/APKBUILD
@@ -7,7 +7,7 @@ case $pkgver in
 	*.*.*)	_kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=0
+pkgrel=1
 pkgdesc="Linux vanilla kernel"
 url="http://kernel.org"
 depends="mkinitfs"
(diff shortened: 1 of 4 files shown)


nmeum:
`kernel_fpu_begin` is `EXPORT_SYMBOL_GPL` but the patch exports `__kernel_fpu_begin` instead of changing the export of `kernel_fpu_begin` to `EXPORT_SYMBOL` does this work around the license compatibility issues outlined in https://lkml.org/lkml/2019/5/2/527?

Though this doesn't apply directly to the patch proposed here but I am wondering in general if we can just change an `EXPORT_SYMBOL_GPL` to `EXPORT_SYMBOL` downstream or would that have any legal implications?

Cogitri:
I'm not a legal expert, but I doubt it'd have legal implications. I just went with this pwtch because other distros (like NixOS) also use it.

ncopa:
Has this been reported and submitted upstream?

What is the upstream's response?

Cogitri:
See: https://github.com/zfsonlinux/zfs/issues/8259

TL;DR upstream "broke" this themselves on purpose and doesn't want to export the symbols for non-GPL compliant stuff anymore.

ncopa:
> Has this been reported and submitted upstream?
> 
> What is the upstream's response?

I read the thread referenced in https://github.com/alpinelinux/aports/pull/8784#issuecomment-501872454

I think we should follow upstream.

If we at some point changes from a `linux-vanilla` to a `linux-alpine` for alpine specific kernel, then we can reconsider this again. But I think we should try avoid adding patches as long as we call the kernel `vanilla`.

Cogitri:
Alrighty

PR #8786, opened 2019-06-13
community/py-sqlparse: upgrade to 0.3.0

andypost:
Ref https://github.com/andialbrecht/sqlparse/blob/master/CHANGELOG#L16

PS: required for pgcli https://github.com/dbcli/pgcli/blob/master/setup.py#L21

### DIFF
--- a/community/py-sqlparse/APKBUILD
+++ b/community/py-sqlparse/APKBUILD
@@ -2,8 +2,8 @@
 # Maintainer: Thomas Boerger <thomas@webhippie.de>
 pkgname=py-sqlparse
 _pkgname=sqlparse
-pkgver=0.2.4
-pkgrel=1
+pkgver=0.3.0
+pkgrel=0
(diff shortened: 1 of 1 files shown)


andypost:
@tboerger Looks other dependencies are not requiring strictly 0.2 version

maxice8:
`mycli` has 'sqlparse>=0.2.2,<0.3.0',

tboerger:
I will try to get back to the packages I have built asap to update them, then it should be fine.

andypost:
@maxice8 good catch, I think to add patch with 3 last commits from https://github.com/dbcli/mycli/commits/master/setup.py

andypost:
I manually checked #8798  and it works perfect with 0.3 version so they should be commited altogether

andypost:
pgcli also works for me after upgrade, @tboerger I think all 3 PRs are ready

PR #8798, opened 2019-06-14
testing/mycli: use upstream patch

andypost:
it blocks upgrade of `community/py-sqlparse`

Ref https://github.com/alpinelinux/aports/pull/8786#issuecomment-502015864


maxice8:
Can the commit title be more descriptive ? upstream patch (for what?)

PR #8802, opened 2019-06-14
main/llvm8: use default LLVM_TARGETS_TO_BUILD

ids1024:
This enables a few additional targets, such as the new WebAssembly target.

### DIFF
--- a/main/llvm8/APKBUILD
+++ b/main/llvm8/APKBUILD
@@ -8,7 +8,7 @@ _pkgname=llvm
 pkgver=8.0.0
 _majorver=${pkgver%${pkgver##[0-9]}}
 pkgname=$_pkgname$_majorver
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


ids1024:
For reference, as can been seen from `CMakeLists.txt` or the build log, this includes all the targets that were already enabled, as well as Hexagon, Lanai, MSP430, Sparc, WebAssembly, and XCore.

`WebAssembly` is what I'm interested in, but it seems it's best to just use the defaults if there isn't a compelling reason to do otherwise (I suppose it increases the size slightly?). This is what Arch Linux does for the llvm package, for instance.

Some of these targets seem a bit obscure to me, but then so do some of the ones that were already enabled.

Cogitri:
You'll need to bump the pkgrel to make this actually do anything on the builders though.

ids1024:
Ah yes, of course. I've amended the commit to bump the pkgrel.

PR #8806, opened 2019-06-15
main/wxgtk2.8: remove

maxice8:


maxice8:
This will also free me to remove even more stuff from the deprecated gstreamer0.10

PR #8814, opened 2019-06-16
 testing/proj4: rename proj4 to proj

hjaekel:
With the release of version 5.0 in February 2018, the software was renamed to PROJ, removing version four (".4") from the name. With the accusation of the proj.org domain name in June 2019, the project was also renamed as PROJ.


testing/libgeotiff/APKBUILD
```diff
@@ -1,24 +1,26 @@
...
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
 pkgname=libgeotiff
-pkgver=1.4.2
-pkgrel=2
+pkgver=1.5.1
+pkgrel=0
 pkgdesc="TIFF based interchange format for georeferenced raster imagery."
 url="https://trac.osgeo.org/geotiff"
-arch="all"
+arch="s390x aarch64 x86_64 ppc64le"
```
maxice8: Are they exclusive to these arches or are the not available on any 32bit arch ?

maxice8: if the latter i'd prefer to use !arch to exclude 32bit ones.

testing/libgeotiff/APKBUILD
```diff
@@ -31,4 +33,15 @@ package() {
 	make DESTDIR="$pkgdir" install
 }
 
-sha512sums="059c6e05eb0c47f17b102c7217a2e1636e76d622c4d1bdcf0bd89fb3505f3130bffa881e21c73cfd2ca0d6863b81322f85784658ba3539b53b63c3a8f38d1deb  libgeotiff-1.4.2.tar.gz"
+check() {
+	bin/makegeo
+	bin/listgeo newgeo.tif > metadata.txt
+	grep "Ellipsoid: 7016/Everest 1830 (1967 Definition) (6377298.56,6356097.55)" metadata.txt
```
maxice8: Can we get a comment that explains what this is about ? i can't make out what this is all supposed to mean.

testing/mapserver/APKBUILD
```diff
@@ -1,18 +1,17 @@
...
-pkgver=7.2.1
+pkgver=7.4.0
 pkgrel=0
 pkgdesc="MapServer is CGI-based platform for publishing spatial data and interactive mapping applications to the web."
 url="http://mapserver.org/"
 arch="x86_64"
 license="MIT"
-depends="geos gdal proj4 cairo curl fcgi freetype libjpeg libpng libxml2 libpq giflib protobuf-c"
-depends_dev="cmake postgresql-dev geos-dev gdal-dev jpeg-dev giflib-dev libxml2-dev proj4-dev fcgi-dev curl-dev freetype-dev cairo-dev protobuf-c-dev"
+depends="geos gdal proj cairo curl fcgi freetype libjpeg libpng libxml2 libpq giflib protobuf-c"
```
maxice8: shouldn't depends be removed since depends_dev will find it automatically ?

testing/libgeotiff/APKBUILD
hjaekel: changed to `arch="all !x86 !armhf !armv7"`

testing/libgeotiff/APKBUILD
hjaekel: added some comments where I try to explain the test, which is from https://github.com/OSGeo/libgeotiff#building-libgeotiff

testing/mapserver/APKBUILD
hjaekel: Yes, depends are automatically detected. I removed them.

PR #8819, opened 2019-06-16
main/syslinux: upgrade to 6.04-pre3

liqa:
The change to zytor.com is because kernel.org didn't get the new release tarball.
But they are all official repositories. Confirmation from syslinux website: https://wiki.syslinux.org/wiki/index.php?title=Development#Browse_the_SYSLINUX_source_code

"These repositories are official, but they are sometimes not fully synchronized during development periods between stable releases."

I'd also like to thank the Alpine developers for keeping syslinux around and rocking for x86_64 and x86.
I understand some architectures are doomed to use other (bigger, more complex) bootloaders, but it's great we can keep using syslinux where possible.

Regards,
Liq

### DIFF
--- a/main/syslinux/APKBUILD
+++ b/main/syslinux/APKBUILD
@@ -1,12 +1,12 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=syslinux
-pkgver=6.04_pre1
-pkgrel=3
+pkgver=6.04_pre3
+pkgrel=0
 _ver=${pkgver/_/-}
 pkgdesc="Boot loader for the Linux operating system"
 url="http://syslinux.org"
 arch="x86 x86_64"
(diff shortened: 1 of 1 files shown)


ncopa:
It fails to boot in qemu:

````
ISOLINUX 6.04 6.04-pre3 ETCD Copyright (C) 1994-2015 H. Peter Anvin et al       
                                                                                
Failed to load ldlinux.c32                                                      
Boot failed: press a key to retry...                                            
````

liqa:
I found this problem listed on syslinux's common problems page.
https://wiki.syslinux.org/wiki/index.php?title=Common_Problems#Failed_to_load_ldlinux

It's a bit generic, so I'm not sure it helps.
The issue could indicate a mismatch between syslinux modules and the syslinux bootloader.

PR #8823, opened 2019-06-16
main/py-twisted: add missing dependencies

prspkt:


main/py-twisted/APKBUILD
```diff
@@ -2,7 +2,7 @@
 pkgname=py-twisted
 _pkgname=Twisted
 pkgver=19.2.1
-pkgrel=0
+pkgrel=1
 pkgdesc="Asynchronous networking framework written in Python."
 url="http://twistedmatrix.com/"
```
maxice8: While we're at it `https://twistedmatrix.com/trac/`

ncopa:
The introduced deps are all in testing. they need to move to main (including their deps).

We also need to have the py3 deps in makedepends so the builder can build them in correct order.

PR #8831, opened 2019-06-17
main/vala: upgrade to 0.44.5

Cogitri:
* Adopt maintership
* Enable tests

### DIFF
--- a/main/vala/APKBUILD
+++ b/main/vala/APKBUILD
@@ -1,20 +1,35 @@
-# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
+# Contributor: Rasmus Thomsen <oss@cogitri.dev>
+# Contributor: Natanael Copa <ncopa@alpinelinux.org>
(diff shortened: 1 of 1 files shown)


Cogitri:
Times out on armhf on CI :/

main/vala/APKBUILD
```diff
@@ -1,31 +1,50 @@
...
+
+	# https://gitlab.gnome.org/GNOME/vala/issues/809
+	sed '/array-uint8-uchar-compat.vala/d' -i tests/Makefile.am
+
+	autoreconf --install --force
+}
+
 build() {
 	cd "$builddir"
+	export CFLAGS="-O2 -pipe"
```
ncopa: I wonder why this is needed?

Cogitri: Oh my, good catch. Tried this to debug the one failing test. Will remove :)

Cogitri: Fixed.

PR #8834, opened 2019-06-17
testing/py3-avro: Adding check()

gena01:
* Fixing type of package we download (and download URL).
* Download the full avro package (w/ tests)
* Add check() to run python3 setup.py test
NOTE: Tests use python 3 unittest (not pytest)

### DIFF
--- a/testing/py3-avro/APKBUILD
+++ b/testing/py3-avro/APKBUILD
@@ -1,24 +1,28 @@
 # Maintainer: Gennady Feldman <gena01@gmail.com>
 # Contributor: Gennady Feldman <gena01@gmail.com>
 pkgname=py3-avro
-_pkgname=avro-python3
+_pkgname=avro
 pkgver=1.9.0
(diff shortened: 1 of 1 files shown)


testing/py3-avro/APKBUILD
```diff
@@ -1,24 +1,28 @@
...
+_pkgname=avro
 pkgver=1.9.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Avro is a serialization and RPC framework."
 url="https://avro.apache.org/"
 arch="noarch"
 license="Apache-2.0"
 depends="py3-six"
 makedepends="py3-setuptools python3-dev"
```
maxice8: i don't think python3-dev is necessary

maxice8:
Commit title should tell what it does not what it is doing. `Add check()`

maxice8:
Please comment when the proposed changes are done so i am notified and can review your contribution again.

PR #8835, opened 2019-06-17
community/gitea: upgrade to 1.8.3

TBK:


community/gitea/APKBUILD
```diff
@@ -68,7 +68,7 @@ check() {
 	"$builddir"/$pkgname help > /dev/null
```
maxice8: Please use the upstream testsuite or disable check with a comment on why.

Ikke:
Fails on aarch64:

> [.../runtime/asm_arm64.s:1128 goexit()] [E] updateRepoIndexer: exit status 128 - fatal: Unable to read current working directory: No such file or directory

TBK:
> Fails on aarch64:
> 
> > [.../runtime/asm_arm64.s:1128 goexit()] [E] updateRepoIndexer: exit status 128 - fatal: Unable to read current working directory: No such file or directory

It is apparently an issue with Gitea - https://github.com/go-gitea/gitea/issues/5324 should I just disable aarch64?

PR #8836, opened 2019-06-17
testing/py-pandas: Remove double package() call

sgerrand:
💁 Removes the duplicated function definition in the `APKBUILD` file.


maxice8:
I think it is better to remove the first definition

sgerrand:
> I think it is better to remove the first definition

For what reason?

maxice8:
Looks more correct, doesn't install unnecessary LICENSE file and leaves installing the files in the _py functions

sgerrand:
:bow: @maxice8: Done. ♻️

PR #7990, opened 2019-05-21
main/imagemagick: upgrade to 7.0.8.49

J0WI:
Source url changed, because older releases are no longer available on imagemagick.org

PR #8838, opened 2019-06-17
main/imagemagick: security upgrade to 7.0.8-53

J0WI:
Remaining change from #7990:
Source url changed, because older releases are no longer available on imagemagick.org

cc @tcely

### DIFF
--- a/main/imagemagick/APKBUILD
+++ b/main/imagemagick/APKBUILD
@@ -2,6 +2,7 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=imagemagick
+_pkgname=ImageMagick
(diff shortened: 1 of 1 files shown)


J0WI:
Depends on #8991

J0WI:
Something is still wrong with perlmagick. @timlegge can you help here?

timlegge:
will take a look

timlegge:
I believe it needs:

chrpath -d "$subpkgdir"/usr/lib/perl5/vendor_perl/auto/Image/Magick/Magick.so 

I am testing a patch for certain.

I am not familiar enough with rpath to know whether there is a way to prevent this from reoccuring or whether stripping it is the only solution.

Tim

timlegge:
See PR https://github.com/alpinelinux/aports/pull/9520 which should finish building cleanly

J0WI:
Closing in favor of #9520

PR #8839, opened 2019-06-18
main/perl: upgrade to 5.30.0

J0WI:
Many packages needs to be rebuild against this new version.


Ikke:
Something looks broken:

> ls: cpan/Time-Piece/t/02core_dst.t: No such file or directory

J0WI:
I was trying to disable this test, but I didn't find a proper way to do so.

J0WI:
b5681c5d3d34d42440228c6041d9318b70989be4

PR #8854, opened 2019-06-18
testing/py-schema: new aport

sgerrand:
💁 https://github.com/keleshev/schema
Schema is a library for validating Python data structures

### DIFF
--- a/testing/py3-schema/APKBUILD
+++ b/testing/py3-schema/APKBUILD
@@ -1,0 +1,32 @@
+# Contributor: Sasha Gerrand <alpine-pkgs@sgerrand.com>
+# Maintainer: Sasha Gerrand <alpine-pkgs@sgerrand.com>
+pkgname=py3-schema
+_pkgname=schema
+pkgver=0.6.8
+pkgrel=0
+pkgdesc="Schema is a library for validating Python data structures"
+url="https://github.com/keleshev/$_pkgname"
+arch="noarch"
+license="MIT"
+depends="python3"
+checkdepends="pytest"
+makedepends="py3-setuptools"
+source="$pkgname-$pkgver.tar.gz::https://github.com/keleshev/$_pkgname/archive/v$pkgver.tar.gz"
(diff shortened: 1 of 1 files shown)


maxice8:
Is there a good reason to add a py2- variant ? we'd rather not add new packages for an soon to be EOL language.

maxice8:
ping

sgerrand:
> Is there a good reason to add a py2- variant ?

None really. I'd created a `py-` variant to support my own local use case but there's no need to have it exist here.

maxice8:
Then please keep it only as py3-schema, py2 is soon to be EOL and we are working hard to remove it from everywhere possible.

sgerrand:
:bow: That's done.

testing/py3-schema/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+pkgname=py3-schema
+_pkgname=schema
+pkgver=0.6.8
+pkgrel=0
+pkgdesc="Schema is a library for validating Python data structures"
+url="https://github.com/keleshev/$_pkgname"
+arch="noarch"
+license="MIT"
+depends="python"
+makedepends="py-setuptools pytest python2-dev python3-dev"
```
maxice8: can be only py3-setuptools and pytest should be in `checkdepends`

testing/py3-schema/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+# Maintainer: Sasha Gerrand <alpine-pkgs@sgerrand.com>
+pkgname=py3-schema
+_pkgname=schema
+pkgver=0.6.8
+pkgrel=0
+pkgdesc="Schema is a library for validating Python data structures"
+url="https://github.com/keleshev/$_pkgname"
+arch="noarch"
+license="MIT"
+depends="python"
```
maxice8: should be python3

sgerrand:
♻️

PR #8856, opened 2019-06-18
community/shotwell: Provided support for aarch64

ossdev07:
AARCH64 architecture support has been provided


maxice8:
Needs a shorter title and use present not past time.

PR #8863, opened 2019-06-18
main/clang: bump pkgrel, to use new llvm

ids1024:
Clang statically links against libraries from LLVM, since LLVM_LINK_LLVM_DYLIB is not used.

LLVM was updated in 84dff3bf837eeb0e47439bbd3dd2b4e6fd53ed56.

### DIFF
--- a/main/clang/APKBUILD
+++ b/main/clang/APKBUILD
@@ -4,7 +4,7 @@
 pkgname=clang
 # Note: Update together with llvm.
 pkgver=8.0.0
-pkgrel=0
+pkgrel=1
 _llvmver=${pkgver%%.*}
 pkgdesc="A C language family front-end for LLVM"
 arch="all"


maxice8:
Please use standard form for commit titles

`<repo>/<package>: <what you're doing in present tense>`

`main/clang: rebuild to use new llvm`

ids1024:
@ncopa

ids1024:
I've built this locally, and it seems to work (I'm now able to compile with the `wasm32` target, which wasn't available).

Travis failure is, of course, a timeout.

ncopa:
Why don't we link dynamically instead?

ncopa:
It seems like clang already is linked dynamically against llvm.

ncopa:
The commit message is wrong. We do link dynamically against llvm:
```
ncopa-desktop:~$ apk info -R clang
clang-8.0.0-r0 depends on:
so:libLLVM-8.so
so:libc.musl-x86_64.so.1
so:libclang.so.8
so:libgcc_s.so.1
so:libstdc++.so.6
so:libxml2.so.2

ncopa-desktop:~$ apk info -R clang-libs
clang-libs-8.0.0-r0 depends on:
so:libLLVM-8.so
so:libc.musl-x86_64.so.1
so:libgcc_s.so.1
so:libstdc++.so.6
```

But we may need to rebuild clang anyway.

ids1024:
I assumed it was static linking given the problem and after seeing commit https://github.com/alpinelinux/aports/commit/3fa7615b906a3d6e3b3c8ac244b54ed3dd27f90c.

> But we may need to rebuild clang anyway.

It seems so. It wasn't able to compile to WebAssembly without rebuilding, but did after I rebuilt.

But if it's not linking statically... I don't know why. I can change the commit message, but ideally it would help to know why the rebuild is needed...

ids1024:
I suppose this should be unnecessary due to 4f34aeb947cdd47eb6a8939860b90855f9305a2b, so I'm closing it.

PR #8878, opened 2019-06-18
community/ceph: Switch to python3

iggy:
Change packages to be built against python3. I tried to build both
python2 and python3 packages, but cmake doesn't like me.


iggy:
On the drone build, I think the compiler ran out of memory. It's c++, so not surprising. It will probably timeout on the Travis build (all of my other PRs have)

PR #8900, opened 2019-06-19
testing/kontact: new aport

PureTryOut:
This application is useless on itself, it needs others like KMail, KOrganizer, KAddressBook, Akregator and KNotes to be useful. However, these are all optional dependencies and the user can choose to use any number of them.


maxice8:
Merged

PR #8903, opened 2019-06-19
main/mariadb: upgrade to 10.4.6

J0WI:
https://mariadb.com/kb/en/library/changes-improvements-in-mariadb-104/


main/mariadb/APKBUILD
```diff
@@ -184,7 +183,6 @@ check() {
 	# exclude test-connect which seems to be buggy. testsuite does not set port env var
 	# ma_test_loghandler_readonly seems to fail if /etc/my.conf.d is missing
 	ctest -E '(test-connect|ma_test_loghandler_readonly)'
```
TBK:
```suggestion
	ctest -E 'test-connect'
```
I also worked on upgrading to 10.4.6 and the test ma_test_loghandler_readonly passes on my machine (x86_64).

J0WI: Thanks, added

andypost:
Meantime it is interesting host many packages can work with newly added features, I seen some issues because of new system tables and other changes
So it needs to check lots of apps

ncopa:
179bcc61994dd6f23dcb44f7f2cdb5a7b0b3bc68  456e3a801ed0b427cd5c2aa73bedb6b944a5d404

PR #8914, opened 2019-06-19
Add Drone CI jsonnet template

TBK:
cloud.drone.io does currently not native support jsonnet - https://discourse.drone.io/t/drone-cloud-and-jsonnet/4621 

Docs https://docs.drone.io/extend/config/jsonnet/

Example https://github.com/drone/drone/blob/master/.drone.jsonnet

@clandmeter is this of interest?


Cogitri:
I guess this isn't relevant anymore now that we use Gitlab CI, sorry.

PR #8933, opened 2019-06-20
community/wine: Provided support for aarch64

ossdev07:
Architecture specific support is provided for aarch64.


maxice8:
Use present tense and a shorter title

`community/wine: enable aarch64`

PR #8936, opened 2019-06-20
testing/korganizer: new aport

PureTryOut:


PureTryOut:
CI fails on multiple arches because of some IO erros when installing openssl. Probably just needs a CI restart.

PR #8937, opened 2019-06-20
testing/latte-dock: new aport

PureTryOut:


testing/latte-dock/APKBUILD
```diff
@@ -0,0 +1,36 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=latte-dock
+pkgver=0.8.8
+pkgrel=0
+arch="all"
```
maxice8: set !ppc64le and !s390x because of libksysguard -> qt5-qtwebengine (not available on those arches)

PR #8939, opened 2019-06-20
testing/kate & testing/kwrite: new aports

PureTryOut:


PureTryOut:
For some reason it fails gzipping `kwrite-doc` on aarch64... Not sure what to do about this.

Cogitri:
Seeing how these are both built from the same source maybe it'd make more sense to make kwrite a subpackage of kate?

PureTryOut:
I thought of doing that, but I'm not sure how to split out `kwrite-doc` in that case. I somehow need to tell `default_doc` for `kwrite-doc` to use the doc files of the `kwrite` subpackage rather than the main package.

Cogitri:
I guess you'd need to make a custom doc function then.

PureTryOut:
Obviously, that I understand. But I'm not sure what to do _inside_ that function, as I don't know what `default_doc` would do in there.

PR #8943, opened 2019-06-20
community/nextcloud-client: re-enable all architectures, modernize

PureTryOut:
CC @rnalrd

### DIFF
--- a/community/nextcloud-client/APKBUILD
+++ b/community/nextcloud-client/APKBUILD
@@ -1,31 +1,31 @@
 # Maintainer: Leonardo Arena <rnalrd@alpinelinux.org>
 pkgname=nextcloud-client
 pkgver=2.5.2
-pkgrel=0
+pkgrel=1
 pkgdesc="Nextcloud Desktop Client"
-url="https://github.com/nextcloud/client_theming"
-arch="all !x86 !s390x !ppc64le" # qt5-qtwebengine not avail
+url="https://github.com/nextcloud/desktop"
+arch="all"
 license="GPL-2.0-only"
-makedepends="$depends_dev cmake doxygen qt5-qtwebengine-dev qt5-qttools-dev
-	qt5-qtwebkit-dev qt5-qtkeychain-dev"
+makedepends="$depends_dev cmake doxygen qt5-qttools-dev qt5-qtkeychain-dev"
 subpackages="$pkgname-dev"
 options="!check" # no testsuite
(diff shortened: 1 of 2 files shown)


Cogitri:
Not really related to this, but does it work for you now?

PureTryOut:
I'll be honest, I haven't tried running it (I don't have a pure x86 installation anywhere). But the comment said that x86 was disabled due to qt5-qtwebengine not being available (which is not the case anymore), so re-enabling it makes sense. I don't see why it wouldn't work on x86, but I guess someone would have to test.

Cogitri:
Ah, sorry that I wasn't clear, I was just curious if it worked on x86_64 for you, QtWebengine still crashes instantly for me :c 

Some user posted a patch to use Nextcloudclient without QtWebengine (requiring you to use an app password for Nextcloud), so maybe we should add that for now?

PureTryOut:
Ah ok. Yeah crashes instantly for me too, it has more problems anyway. If that patch can make it work right now, I'd go for it. But eventually we should move back to QtWebengine.

Cogitri:
Here's the URL to the patch: http://ix.io/1Jld

PureTryOut:
Thanks, I've applied that patch and removed `qt5-qtwebengine` from the `$makedepends` for now. Because we don't use that dependency anymore, I've also re-enabled the package for all architectures.

From the quick testing I did it seems to work!

Cogitri:
Yup, works for me too. Thanks for checking! :)

community/nextcloud-client/APKBUILD
```diff
@@ -1,31 +1,31 @@
...
 builddir="$srcdir"/desktop-$pkgver
 
 build() {
-	cd "$builddir"
 	cmake \
-		-DCMAKE_BUILD_TYPE="Release" \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo \
 		-DCMAKE_INSTALL_LIBDIR=lib \
 		-DCMAKE_INSTALL_PREFIX=/usr \
 		-DCMAKE_INSTALL_SYSCONFDIR=/etc/$pkgname
```
Cogitri: I guess we should add `-DNO_SHIBBOLETH=1` here to get rid of the qt5-qtwebkit dependency, upstream has supported app passwords for some years now (and qt5-qtwebkit is super insecure at this point).

maxice8:
@rnalrd

PureTryOut:
Wait, not the full thing got merged. It still has the webkit dependency and lacks `-DNO_SHIBBOLETH=1`.

PR #8956, opened 2019-06-21
testing/xdg-desktop-portal: new aport

Cogitri:
Needs #8954

### DIFF
--- a/testing/xdg-desktop-portal-kde/APKBUILD
+++ b/testing/xdg-desktop-portal-kde/APKBUILD
@@ -1,0 +1,34 @@
+# Contributor: Rasmus Thomsen <oss@cogitri.dev>
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
+pkgname=xdg-desktop-portal-kde
+pkgver=5.16.1
+pkgrel=0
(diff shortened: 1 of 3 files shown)


testing/xdg-desktop-portal-kde/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+
+build() {
+	if [ "$CBUILD" != "$CHOST" ]; then
+		CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
+	fi
+	cmake \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=lib \
+		-DBUILD_SHARED_LIBS=True \
+		-DCMAKE_BUILD_TYPE=Release \
```
PureTryOut: This should probably have been built with RelWithDebInfo instead

Cogitri: Huh, I just went with whatever newapkbuild gave me, maybe we should change it there then?

PR #8965, opened 2019-06-22
main/nss: upgrade to 3.45

J0WI:


nmeum:
Seems to be [backwards compatible](https://abi-laboratory.pro/index.php?view=objects_report&l=nss&v1=3.44.1&v2=3.45), so no packages should require a rebuild. We could merge this as is and afterwards upgrade firefox (as per #9431). 

@ncopa you maintain both packages (firefox and nss) any objections?

J0WI:
9e7c729c647b61578615346d94b3927608c67755

PR #8976, opened 2019-06-22
testing/xdg-desktop-portal-kde: take over maintainership

PureTryOut:
It's part of the Plasma stack which I already maintain, so I'll maintain this one too. Agreed upon by @Cogitri on IRC.

Also set build type to `RelWithDebInfo` like all the other KDE packages.

### DIFF
--- a/testing/xdg-desktop-portal-kde/APKBUILD
+++ b/testing/xdg-desktop-portal-kde/APKBUILD
@@ -1,14 +1,14 @@
 # Contributor: Rasmus Thomsen <oss@cogitri.dev>
-# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
 pkgname=xdg-desktop-portal-kde
 pkgver=5.16.1
-pkgrel=0
+pkgrel=1
 pkgdesc="Qt/KF5 implementation of xdg-desktop-portal"
 url="https://phabricator.kde.org/source/xdg-desktop-portal-kde"
 arch="all"
 license="GPL-3.0-or-later"
-makedepends="cmake qt5-qtbase-dev glib-dev kio-dev kwayland-dev libepoxy-dev
-	pipewire-dev kcoreaddons xdg-desktop-portal extra-cmake-modules cups-dev"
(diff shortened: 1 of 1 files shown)


Cogitri:
> Also set build type to RelWithDebInfo like all the other KDE packages..

Won't we strip the debug symbols anyway though (unless we have a -dbg package)?

testing/xdg-desktop-portal-kde/APKBUILD
```diff
@@ -1,14 +1,14 @@
...
-pkgrel=0
+pkgrel=1
 pkgdesc="Qt/KF5 implementation of xdg-desktop-portal"
 url="https://phabricator.kde.org/source/xdg-desktop-portal-kde"
 arch="all"
 license="GPL-3.0-or-later"
-makedepends="cmake qt5-qtbase-dev glib-dev kio-dev kwayland-dev libepoxy-dev
-	pipewire-dev kcoreaddons xdg-desktop-portal extra-cmake-modules cups-dev"
+makedepends="extra-cmake-modules qt5-qtbase-dev glib-dev kio-dev kwayland-dev libepoxy-dev
+	pipewire-dev kcoreaddons xdg-desktop-portal cups-dev"
```
Cogitri: I don't really see the reason for this change?

Also, maybe you could take a look at qt5-qtbase? I think that the cups-dev dependency actually belongs to it.

PureTryOut: `extra-cmake-modules` already depends on `cmake`, there isn't really a need to declare it separately. All other KDE packages already do it like this, this is just making it do the same.

PureTryOut:
> > Also set build type to RelWithDebInfo like all the other KDE packages..
> 
> Won't we strip the debug symbols anyway though (unless we have a -dbg package)?

Hmm good call, I'm not actually sure about this. Might be worth adding `-dbg` to all the packages then, but I'd need to know for sure.

Cogitri:
I'm pretty sure we strip our binaries. Try loading any binary of those packages into `gdb` and you'll see that it can't load debug symbols.

testing/xdg-desktop-portal-kde/APKBUILD
Cogitri: Oh, oops, didn't realise you replace `cmake` here, thought you just moved e-c-m for some reason

PR #8985, opened 2019-06-22
testing/{bluedevil,drkonqi,kinfocenter}: remove depends_dev

PureTryOut:
Added them by mistake while not having `$pkgname-dev` subpackages. Build output won't change from this modification, so I decided not to increase the `$pkgrel` for these applications. Is this correct?


maxice8:
yes

PR #8991, opened 2019-06-23
main/imagemagick: Fix install location for PerlMagick modules

timlegge:
Original code puts perl modules in /usr/lib/perl5/site_perl which is not in the @INC for perl.  This puts it in /usr/lib/perl5/vendor_perl which is the location for packaged modules.

Sorry I did not catch this previously.


timlegge:
forced a push to increment the pkgrel...

J0WI:
:ping_pong: @tcely could you have a look at this?

tcely:
>  @tcely could you have a look at this?

Done. @J0WI

timlegge:
@tcely Is there anything else required on this to be merged?

Tim

PR #8993, opened 2019-06-23
testing/plasma-vault: new aport

PureTryOut:


testing/plasma-vault/APKBUILD
```diff
@@ -0,0 +1,39 @@
+# Contributor: Bart Ribbers <bribbers@disroot.org>
+# Maintainer: Bart Ribbers <bribbers@disroot.org>
+pkgname=plasma-vault
+pkgver=5.16.1
+pkgrel=0
+pkgdesc="Plasma applet and services for creating encrypted vaults"
+arch="all"
```
maxice8: this package depends on libksysguard which is restricted on ppc64le and s390x

PR #9003, opened 2019-06-23
testing/qbittorrent-nox: fix license, build against libtorrent-rasterbar

maxice8:
Use a git release to fix building against newer libtorrent-rasterbar


maxice8:
@jantatje

PR #9018, opened 2019-06-24
testing/rclone: new aport

5paceToast:


maxice8:
There is #8248

5paceToast:
Oh, I must have missed it.
Looking at it though, it's a 100 LOC monster, lacks the -doc subpackage and uses the weird makefile.
I think it makes sense to treat it like any other go package (i.e this) and add the gpg and other things once abuild actually supports it.

Where should that conversation take place? (i.e I don't know if I really care for trying to get the gpg situation resolved, especially with the review seemingly being ignored (latest commit still has empty var etc))

maxice8:
> Where should that conversation take place? 

Anywhere where you and the author of the other PR want, i just want a resolution of what i should review and ultimately merge.

tcely:
Feel free to pick this one. It is clear to me that @5paceToast cares about this more than I do. I am happy to maintain one less package.

PR #9022, opened 2019-06-24
community/exo: upgrade to 0.12.6

maxice8:


ncopa:
i want to fix webkit2gtk

maxice8:
On Fri, Jun 28, 2019 at 08:26:47AM -0700, Natanael Copa wrote:
> i want to fix webkit2gtk
> 
> -- 
> https://github.com/alpinelinux/aports/pull/9022#issuecomment-506775273

@Cogitri said there is the possibility of 32bit support on the next version

Cogitri:
I don't have time to implement this rn, but upstream mentioned that it might work with CLang which apparently takes less memory. Please add clang-dev to makedepends, set CC=clang CXX=clang++ and try if it works on 32-bit arches.

ncopa:
I managed to get webkit2gtk built on armhf with gcc. I'm testing it with armv7 and x86 now.

Cogitri:
Huh, doesn't x86 work already? Anyway, thanks for looking into it!

PR #9032, opened 2019-06-24
testing/googler: upgrade to 3.9

maxice8:


maxice8:
@pickfire

pickfire:
LGTM except I don't get why the space changes.

maxice8:
replaces spaces with tabs which is the correct usage.

PR #9051, opened 2019-06-24
testing/sweeper: new aport

PureTryOut:


testing/sweeper/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=lib
+	make
+}
+
+check() {
+	CTEST_OUTPUT_ON_FAILURE=TRUE ctest
```
maxice8: xvfb-run isn't used here (is it used implicitly during tests ?)

PureTryOut: Sorry my mistake, checkdepends is not needed.

PR #9052, opened 2019-06-24
testing/kmouth: new aport

PureTryOut:


testing/kmouth/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=lib
+	make
+}
+
+check() {
+	CTEST_OUTPUT_ON_FAILURE=TRUE ctest
```
maxice8: xvfb-run in checkdepends but not used (used implicitly ?)

PureTryOut: Sorry my mistake, checkdepends is not needed. Fixed it.

PR #9096, opened 2019-06-25
community/dunst: add dbus-dev dependency to fix build

mksully22:
Build breaks with error:
Error: Makefile:25: *** "Failed to query pkg-config for package 'dbus-1'!".  Stop.

The Makefile uses pkg-conf and requires /usr/lib/pkgconfig/dbus-1.pc provided by the dbus-dev package. Add this as a Make dependancy to fix.


maxice8:
If you're fixing the build then i don't think it is necessary to bump pkgrel

maxice8:
Use a shorter title on the commit and depend**a**ncy is misspelled

maxice8:
On the commit, you changed the title on the Pull Request

PR #9103, opened 2019-06-25
community/py3-faker: upgrade to 1.0.7

prspkt:


maxice8:
@ddevault

ddevault:
I know this package was wrong when you got to it, but while you're here can you update the style to match the python guidelines:

https://wiki.alpinelinux.org/wiki/Python_package_policies

Cheers

prspkt:
@ddevault Will do. Thanks.

PR #9105, opened 2019-06-25
testing/knotes: new aport

PureTryOut:


PureTryOut:
The CI failure is an IO error on aarch64. Just needs a restart.

PR #9107, opened 2019-06-25
testing/kdepim-addons: new aport

PureTryOut:


testing/zxing-cpp/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+license="Apache-2.0"
+makedepends="cmake"
+source="$pkgname-$pkgver.tar.gz::https://github.com/nu-book/$pkgname/archive/v$pkgver.tar.gz"
+subpackages="$pkgname-dev"
+
+build() {
+	cmake \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_INSTALL_LIBDIR=lib \
```
maxice8: This is unnecessary according to CMake

PR #9113, opened 2019-06-25
testing/dolphin-plugins: new aport

PureTryOut:


testing/dolphin/APKBUILD
```diff
@@ -15,7 +15,7 @@ makedepends="extra-cmake-modules qt5-qtbase-dev kdoctools-dev kinit-dev
 	kactivities-dev"
 checkdepends="xvfb-run"
 source="https://download.kde.org/stable/applications/$pkgver/src/$pkgname-$pkgver.tar.xz"
-subpackages="$pkgname-doc $pkgname-lang"
+subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
```
maxice8: shouldn't depends_dev also be added since we now have a -dev subpackage ?

PureTryOut: You're completely right, I didn't think of that. Fixed.

PR #9140, opened 2019-06-26
testing/kitinerary: remove libphonenumber from makedepends on x86_64

PureTryOut:
35ee7505c6f44c55d2098b1dcc1712ca7a7c8404 clearly failed, so let's try this approach instead. Since the build output doesn't change for other architectures, a `$pkgrel` bump shouldn't be needed. 35ee7505c6f44c55d2098b1dcc1712ca7a7c8404 should probably be reverted (if possible).


PureTryOut:
This has been fixed by @Ikke, seems a force rebuild of libphonenumber directly in the builder container worked.

PR #9141, opened 2019-06-26
testing/networkmanager-openvpn: create nm-openvpn user/group on pkg install/upgrade

Cogitri:
This is required for operation.

@PureTryOut


PureTryOut:
LGTM

PR #9162, opened 2019-06-27
community/mpv: enable Vulkan support

maxice8:


maxice8:
@ncopa

maxice8:
A sore point to fix is that the user needs to install the proper mesa-vulkan-\<vendor> package, for my machine it is intel and there is AMD

PureTryOut:
> A sore point to fix is that the user needs to install the proper mesa-vulkan-<vendor> package, for my machine it is intel and there is AMD

Does the `setup-xorg-base` script handle this yet?

maxice8:
On Thu, Jun 27, 2019 at 01:28:27PM -0700, Bart Ribbers wrote:
> > A sore point to fix is that the user needs to install the proper mesa-vulkan-<vendor> package, for my machine it is intel and there is AMD
> 
> Does the `setup-xorg-base` script handle this yet?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/9162#issuecomment-506498627

Apparently not

http://ix.io/1MX7

PR #9178, opened 2019-06-28
community/newsboat: upgrade to 2.16.1

maxice8:


maxice8:
@mcrute

mcrute:
Thanks @maxice8! Does this build correctly in the main repo? As I recall Newsboat took a hard dependency on Rust that's missing in this aport.

maxice8:
It has been restricted since 2.15.0 to only x86_64 (only arch where rust is available)

PR #9208, opened 2019-06-30
main/xrdp: upgrade to 0.9.10

maxice8:


maxice8:
@alacerda

main/xrdp/APKBUILD
```diff
@@ -1,15 +1,13 @@
 # Contributor: Alan Lacerda <alacerda@alpinelinux.org>
 # Maintainer: Alan Lacerda <alacerda@alpinelinux.org>
 pkgname=xrdp
-pkgver=0.9.9
+pkgver=0.9.10
 pkgrel=0
 pkgdesc="An open source RDP server"
-url="http://www.xrdp.org/"
+url="http://xrdp.org/"
```
nmeum: `xrdp.org` and `www.xrdp.org` actually have different IP address.  The former doesn't load here, the latter is a github page. Besides: `www.xrdp.org` seems to support HTTPS.

maxice8: Fixed, thanks for catching

PR #9211, opened 2019-06-30
testing/aria2: new aport

Magicloud:
Aria2 is quite a useful downloading tool. Since it supports RPC, it is
popular to be the downloading server as well.


maxice8:
Already present in main/

Magicloud:
My bad. Did search with the "2".

PR #9222, opened 2019-07-01
community/sn0int: move from testing

kpcyrd:
I've tested the package for a while and it works well.


maxice8:
please bump the revision in a commit separate from the commit that moves it from testing/ to community/

kpcyrd:
@maxice8 done. Note that it seems travis died during an apk update and might need to be retriggered.

PR #9234, opened 2019-07-01
main/nginx: fix http-geoip2 bug

aptalca:
Currently, there are ~3~2 bugs in the nginx package (and modules)

1) Nginx is loading `ngx_stream_geoip2_module.so` before `ngx_stream_module.so` and that prevents nginx from starting with error `nginx: [emerg] dlopen() "/var/lib/nginx/modules/ngx_stream_geoip2_module.so" failed (Error relocating /var/lib/nginx/modules/ngx_stream_geoip2_module.so: ngx_stream_add_variable: symbol not found) in /etc/nginx/modules/http_geoip2.conf:1`
2) When there is more than one so defined for a module, the line `echo "load_module \"modules/$soname\";" > ./etc/nginx/modules/$name.conf` only writes the last so listed into the conf file, so `ngx_http_geoip2_module.so` is not being loaded
~3) sha512sum for the package `lua-nginx-module-0.10.15.tar.gz` has changed, so nginx currently doesn't build with abuild~ fixed in #10076 

This PR fixes these issues by
1) Splitting stream-geoip2 from http-geoip2 into its own package, and loading the stream-geoip2 conf after stream (alphabetical)
2) Fixing the echo line to append into conf file
~3) Updating the sha512sum for the lua nginx module~ fixed in #10076 

Note: Splitting stream-geoip2 into a new package is a somewhat hacky method because we are pulling the same github repo twice and building it twice (once for http-geoip2 and once for stream-geoip2). Each build results in two so's one for http and one for stream. After the second build the first pair are renamed ${soname}.old so I added a line `rm -f "$pkgdir"/$_modules_dir/${soname}.old` to remove these duplicate files before packaging.

### DIFF
--- a/main/nginx/APKBUILD
+++ b/main/nginx/APKBUILD
@@ -19,7 +19,7 @@ pkgname=nginx
 # NOTE: Upgrade only to even-numbered versions (e.g. 1.14.z, 1.16.z)!
 # Odd-numbered versions are mainline (development) versions.
 pkgver=1.16.1
-pkgrel=0
+pkgrel=1
 # Revision of nginx-tests to use for check().
 _tests_hgrev=40e5f2a0a238
 _njs_ver=0.3.4
@@ -151,7 +151,9 @@ _rtmp_provides="$pkgname-rtmp"  # for backward compatibility
 _add_module "http-vod" "1.24" "https://github.com/kaltura/nginx-vod-module"
 
 _add_module "http-geoip2" "3.2" "https://github.com/leev/ngx_http_geoip2_module"
-_http_geoip2_so="ngx_http_geoip2_module.so ngx_stream_geoip2_module.so"
+
+_add_module "stream-geoip2" "3.2" "https://github.com/leev/ngx_http_geoip2_module"
+_stream_geoip2_depends="$pkgname-mod-stream"
 
 prepare() {
 	local file; for file in $source; do
@@ -314,7 +316,8 @@ _module() {
 
 	local soname; for soname in $sonames; do
 		mv "$pkgdir"/$_modules_dir/$soname ./$_modules_dir/$soname
-		echo "load_module \"modules/$soname\";" > ./etc/nginx/modules/$name.conf
+		rm -f "$pkgdir"/$_modules_dir/${soname}.old
+		echo "load_module \"modules/$soname\";" >> ./etc/nginx/modules/$name.conf
 	done
 }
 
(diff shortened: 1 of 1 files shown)


aptalca:
Added the dependency of `mod-stream` for `stream-geoip2` that wasn't picked up automatically

aptalca:
Just following up as it's been about a month now. The package remains broken in both edge and 3.10. It breaks nginx once installed.

Any feedback would be appreciated.

Cogitri:
@andypost / @ncopa mind taking a look?

main/nginx/APKBUILD
```diff
@@ -310,7 +312,8 @@ _module() {
 
 	local soname; for soname in $sonames; do
 		mv "$pkgdir"/$_modules_dir/$soname ./$_modules_dir/$soname
-		echo "load_module \"modules/$soname\";" > ./etc/nginx/modules/$name.conf
+		rm -f "$pkgdir"/$_modules_dir/${soname}.old
```
jirutka: What’s the purpose of this? There’s no file with suffix `.old` generated, is it?

aptalca: Because we build the geoip2 module twice (once for http and once for stream) and since the module puts out both `so`s every time, the second build results in the first set of `so`s to be renamed `*.old`.

Building it twice is a bit of a hack, but it ensures that we can have two separate packages, one for the http version and one for the stream version, each package with the respective `so`

aptalca:
To clarify, building the `ngx-http-geoip2` module results in two `so`s: `ngx_http_geoip2_module.so` and `ngx_stream_geoip2_module.so`. The latter depends on `ngx_stream_module.so` and also needs to be loaded into nginx after that.

If both `so`s are provided by the same package named `nginx-mod-http-geoip2`, `ngx_stream_geoip2_module.so` gets attempted to be loaded before `ngx_stream_module.so` (alphabetical). Moving `ngx_stream_geoip2_module.so` into a separate package named `nginx-mod-stream-geoip2` not only fixes the alphabetical loading issue as it is loaded after `ngx_stream_module.so`, but it is also more consistent with the package structures of the v1 geoip modules.

The simplest (least modification to APKBUILD) solution to split the two `so`s into separate packages was to build the module twice (which results in 4 `so`s named `ngx_http_geoip2_module.so`, `ngx_http_geoip2_module.so.old`, `ngx_stream_geoip2_module.so` and `ngx_stream_geoip2_module.so.old`), and define the specific `so` for each package entry. Then we clean up by deleting the duplicate `.old` files that are left behind.

aptalca:
@jirutka just making sure you saw the responses above

aptalca:
@ncopa can you re-tag it with `S-waiting-maintainer-review`?
Thanks

ncopa:
I had a closer look at this and the proposed fix is pretty clever :+1:

But to be honest, it is a bit too hackish for my taste and does not solve the underlying problem: that you cannot specify the dependency order of loading the modules. I think we should use the `_<mod>_depends` to figure out the load order somehow. Not sure how though.

I am not sure how many of the modules that are affected by this but at least those in addition to `stream-geoip2`:
```
_stream_js_depends="$pkgname-mod-stream"
_http_lua_depends="$pkgname-mod-devel-kit"
_http_lua_upstream_depends="$pkgname-mod-http-lua"
_http_set_misc_depends="$pkgname-mod-devel-kit"
```

One way to do it would be to have a two digit number prefix with the load order.

What do you think @jirutka?

PeterDaveHello:
Hi guys, is there any chance that we can push this to get merged as that package is still broken since its first release, many thanks 😄

aptalca:
@PeterDaveHello it seems the maintainer for this repo is AWOL

Just a heads up, in our nginx based letsencrypt image, we work around this issue by creating/appending the following linked module files under `/etc/nginx/modules`: https://github.com/linuxserver/docker-letsencrypt/tree/master/root/etc/nginx/modules

That way nginx starts properly and the modules work

Cogitri:
Well @aptalca could you comment on what has been mentioned in https://github.com/alpinelinux/aports/pull/9234#issuecomment-527929484?

aptalca:
@Cogitri I think https://github.com/alpinelinux/aports/pull/9234#issuecomment-527929484 is an excellent idea and I agree would be the proper way to fix it. However, it is a structural change and is therefore best discussed with the maintainer first.

If there is interest, I can bounce some ideas off of you and @ncopa here and would be happy to work on a PR that is hopefully acceptable to your team.

jirutka:
The primary underlying problem was that `nginx-mod-http-geoip2` provided two separate modules: `http_geoip2` and `stream_geoip2`, and doesn’t declare a dependency on `nginx-mod-stream` required by the latter. This was my fault and I finally fixed it now.

The secondary underlying problem is that the modules are not being loaded in the order reflecting their dependencies, as @ncopa described in https://github.com/alpinelinux/aports/pull/9234#issuecomment-527929484. Actually, they are now, but just by a coincidence. We can fix this in the same/similar way as I did it in the php7 package ([here](https://github.com/alpinelinux/aports/blob/f29c859c687342147661b62edb89d3f890877c3b/community/php7/APKBUILD#L648-L662)).

PR #9243, opened 2019-07-02
testing/wabt: new aport

ids1024:


testing/wabt/APKBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Ian Douglas Scott <ian@iandouglasscott.com>
+# Maintainer: Ian Douglas Scott <ian@iandouglasscott.com>
+pkgname=wabt
+pkgver=1.0.11
+pkgrel=0
+pkgdesc="WebAssembly binary toolkit"
+arch="all"
+url="https://github.com/WebAssembly/wabt"
+license="Apache-2.0"
+makedepends=cmake
```
maxice8: makedepends should be quoted

testing/wabt/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+subpackages="$pkgname-doc"
+source="$pkgname-$pkgver.tar.gz::$url/archive/$pkgver.tar.gz"
+
+prepare() {
+	default_prepare
```
maxice8: prepare() is defined as `default_prepare` by default, so this whole definition can be removed.

ids1024:
The tests inconveniently use vendored dependencies. I've raised an issue about that upstream: https://github.com/WebAssembly/wabt/issues/1103.

Probably including a patch to use the packaged versions would be a good short-term solution, until it is (hopefully) addressed upstream. But I'll see what response that issue gets.

maxice8:
You can set !check in options="" with a comment in the mean time

ids1024:
I've amended the commit to address these concerns.

PR #9244, opened 2019-07-02
add mongo-cxx-driver

russkel:
Requires https://github.com/alpinelinux/aports/pull/9486

Added the CXX driver which is also Apache 2.0 licensed.

### DIFF
--- a/testing/mongo-cxx-driver/APKBUILD
+++ b/testing/mongo-cxx-driver/APKBUILD
@@ -1,0 +1,81 @@
+# Contributor: Russ Webber <russ@rw.id.au>
+# Maintainer: Russ Webber <russ@rw.id.au>
+pkgname="mongo-cxx-driver"
+pkgver="3.4.0"
+_commit="131fa1a67acd45c0eebcbdcfee42b212af8d2e80" # master compiles with GCC8.3/CXX17
+pkgrel=0
+pkgdesc="C++ driver for MongoDB"
+url="http://mongocxx.org/"
+arch="all"
+license="Apache-2.0"
+makedepends="cmake mongo-c-driver-dev libbson-dev openssl-dev python3 snappy-dev"
+subpackages="$pkgname-dev:_dev libbsoncxx libbsoncxx-dev"
+depends_dev="libbson-dev"
+source="mongo-cxx-driver-$pkgver.tar.gz::https://github.com/mongodb/mongo-cxx-driver/archive/$_commit.tar.gz
+	01-dont-build-mongo-tests.patch"
+builddir="$srcdir/mongo-cxx-driver-$_commit"
+
+prepare() {
+	mkdir -p "$builddir/build"
+	default_prepare
+}
+
+build() {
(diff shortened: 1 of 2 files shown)


russkel:
Okay, so the tests are failing because of:

```
-------------------------------------------------------------------------------
Create streams.events and assert we can read a single event
-------------------------------------------------------------------------------
/aports/testing/mongo-cxx-driver/src/mongo-cxx-driver-131fa1a67acd45c0eebcbdcfee42b212af8d2e80/src/mongocxx/test/change_streams.cpp:277
...............................................................................

/aports/testing/mongo-cxx-driver/src/mongo-cxx-driver-131fa1a67acd45c0eebcbdcfee42b212af8d2e80/src/mongocxx/test/change_streams.cpp:277: FAILED:
due to unexpected exception with message:
  No suitable servers found (`serverSelectionTryOnce` set): [connection refused
  calling ismaster on 'localhost:27017']: generic server error
```

Is it worthwhile running the server for testing? *Can* the server be run for testing?

russkel:
```
>>> mongo-cxx-driver-dev*: Package size: 1.2 MB
>>> mongo-cxx-driver-dev*: Compressing data...
>>> mongo-cxx-driver-dev*: Create checksum...
>>> mongo-cxx-driver-dev*: Create mongo-cxx-driver-dev-3.4.0-r0.apk
>>> mongo-cxx-driver*: Tracing dependencies...
>>> ERROR: mongo-cxx-driver*: libbsoncxx.so._noabi: path not found
so:libbson-1.0.so.0
so:libc.musl-x86_64.so.1
so:libgcc_s.so.1
so:libmongoc-1.0.so.0
so:libstdc++.so.6 
```
Is this a bug with abuild not liking non-numeric SOVERSIONS?

ncopa:
> Is it worthwhile running the server for testing? _Can_ the server be run for testing?

Only if the server is built together with this package, but I think its not. As I understad we dont even ship a server, so we cannot have a server as checkdepends either.

I think we can simply disable tests that requires a server.

russkel:
Will rebase when #9486 is merged.

maxice8:
rebase please

russkel:
mongo-c-driver-dev should depend on libbson-dev

testing/mongo-cxx-driver/APKBUILD
```diff
@@ -0,0 +1,60 @@
...
+	make test
+}
+
+package() {
+	cd "$builddir/build"
+	make DESTDIR="$pkgdir" install
+	rm -rf "$pkgdir"/usr/share/mongo-cxx-driver/
+}
+
+libbsoncxx() {
```
maxice8: doesn't it also have its own -dev contents ?

russkel: It does. I see you changed mongo-c-driver. I will modify to suit.

russkel: :tada: working now.

PR #9250, opened 2019-07-02
testing/libglvnd: new aport

russkel:
https://github.com/NVIDIA/libglvnd
libglvnd is a vendor-neutral dispatch layer for arbitrating OpenGL API calls between multiple vendors.

Part of my work to get OpenGL routing through Alpine linux with Docker Nvidia runtime. I haven't been successful yet, but this package seems to compile fine so my issues might lie elsewhere.

### DIFF
--- a/testing/libglvnd/APKBUILD
+++ b/testing/libglvnd/APKBUILD
@@ -1,0 +1,48 @@
+# Contributor: Russ Webber <russ@rw.id.au>
+# Maintainer: Russ Webber <russ@rw.id.au>
+pkgname=libglvnd
+pkgver=1.1.1
+pkgrel=0
+pkgdesc="libglvnd is a vendor-neutral dispatch layer for arbitrating OpenGL API calls between multiple vendors."
+url="https://github.com/NVIDIA/libglvnd"
+arch="all"
+license="custom"
+options="!check"	# test suite requires an Xorg display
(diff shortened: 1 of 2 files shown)


maxice8:
Doesn't NVIDIA not work with musl ?

Also this is definitively not enough since you also need to modify mesa to compile with it and deal with the shlib changes.

russkel:
> Doesn't NVIDIA not work with musl ?

I'll admit I am not sure what's going on with the Nvidia runtime. I don't really know how the internals work.

> Also this is definitively not enough since you also need to modify mesa to compile with it and deal with the shlib changes. (Source: did the libglvnd work on Void Linux that went nowhere due to musl incompatibilities)

I am just trying to replicate that's happening with these ubuntu Dockerfiles: https://gitlab.com/nvidia/opengl/tree/ubuntu18.04 Is this the same as what you are talking about? I can't see any MESA compiling steps.

If you have tried this already then I'll just set fire to it and find something else.

maxice8:
> I'll admit I am not sure what's going on with the Nvidia runtime. I don't really know how the internals work.

the NVIDIA proprietary driver doesn't work with musl

See the fedora spec file for mesa: https://src.fedoraproject.org/rpms/mesa/blob/master/f/mesa.spec

russkel:
You have saved me a bit of time then. I guess I'll have to see if VirtualGL works.

PR #9258, opened 2019-07-02
main/curl: define static() for older stable versions

maxice8:


ncopa:
I doubt we will ever backport the introduction of `-static` subpackage as it can potentially break things if people expect it to be in the `-dev` package.

PR #9262, opened 2019-07-02
testing/vulkan-tools: new aport

PureTryOut:


maxice8:
commit body should have 

$url
$pkgdesc

when adding a new aport

maxice8:
ping

PR #9263, opened 2019-07-02
testing/code-server: new aport

wenerme:


wenerme:
will set arch="x86_64"

PR #9268, opened 2019-07-02
community/php7-pecl-redis: upgrade to 5.0.0

andypost:
Ref https://pecl.php.net/package-changelog.php?package=redis&release=5.0.0

PS It depends on `php7-json` because it supposed to be bundled https://www.php.net/manual/en/extensions.membership.php#extensions.membership.bundled


andypost:
@faabiosr are you ok with major upgrade?

faabiosr:
Yes Mate!

PR #9277, opened 2019-07-03
testing/bfs: upgrade to 1.5

maxice8:


tavianator:
Strange test failures on arm{hf,v7}.  Is it possible to get the output from something like this?

    $ ./tests.sh --bfs="strace /path/to/bfs" test_basic

tavianator:
Wild guess: there's a seccomp profile or something blocking `statx()` for those architectures

tavianator:
Maybe try applying https://github.com/tavianator/bfs/commit/71a1b65ba926ff9c6db3cb671580b5491db0bd97 as a patch

PR #9278, opened 2019-07-03
community/clsync: Enable aarch64 

ossdev07:
Architecture specific support is provided for aarch64.

### DIFF
--- a/community/clsync/APKBUILD
+++ b/community/clsync/APKBUILD
@@ -5,7 +5,7 @@ pkgver=0.4.2
 pkgrel=1
 pkgdesc="File live sync daemon based on inotify"
 url="https://github.com/xaionaro/clsync"
-arch="x86 x86_64"
+arch="x86 x86_64 aarch64"
 license="GPL-3.0-or-later"
 depends=""
 depends_dev="glib-dev fts-dev libcap-dev libexecinfo-dev linux-headers musl-dev"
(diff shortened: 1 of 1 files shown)


community/clsync/APKBUILD
```diff
@@ -26,17 +26,35 @@ prepare() {
...
+		./configure \
+			CFLAGS="$CFLAGS -lfts -lexecinfo" \
+			--build=$CBUILD \
+			--host=$CHOST \
+			--prefix=/usr \
+			--sysconfdir=/etc \
+			--mandir=/usr/share/man \
+			--localstatedir=/var \
+			--enable-capabilities \
+			--enable-seccomp \
```
maxice8: I think it is beter we make a _seccomp variable that holds the value that is required.

ossdev07: As you suggested, I have added a seccomp variable.

Please, kindly review.

community/clsync/APKBUILD
```diff
@@ -15,6 +15,7 @@ source="$pkgname-$pkgver.tar.gz::https://github.com/xaionaro/$pkgname/archive/v$
 	musl-fix.patch"
 builddir="$srcdir/$pkgname-$pkgver"
 options="!check"  # upstream does not provide tests
+seccomp=""
```
maxice8: this is not necessary

PR #9287, opened 2019-07-04
main/mesa: enable etnaviv and tegra support

PureTryOut:
Depends on #9288


PureTryOut:
Now depends on #9288 instead.

PR #9288, opened 2019-07-04
main/libdrm: switch to meson and enable tegra & etnaviv

z3ntu:
Depends & includes the commit from #9266


main/libdrm/APKBUILD
```diff
@@ -19,29 +19,26 @@ source="https://dri.freedesktop.org/$pkgname/$pkgname-$pkgver.tar.bz2
 builddir="$srcdir/$pkgname-$pkgver"
```
PureTryOut: As this is the default value, this can be removed

z3ntu:
Not merged yet... I've re-opened the PR as #9319

PR #9290, opened 2019-07-04
main/avahi: move to Python 3 and integrate avahi-ui

PureTryOut:
I've moved `gdbm` out of the main Python package like we do currently with Python 2.
I'm not sure if there are other packages depending on this functionality that need to have this new `py3-gdbm` dependency.


main/avahi/APKBUILD
```diff
@@ -118,14 +123,51 @@ lidns_sd() {
...
 
 	mkdir -p "$subpkgdir"/usr/lib
-	mv "$pkgdir"/usr/lib/python2.* "$subpkgdir"/usr/lib/
+	mv "$pkgdir"/usr/lib/python3.* "$subpkgdir"/usr/lib/
+}
+
+ui() {
+	pkgdesc="GTK user interface library for Avahi"
+
+
```
maxice8: extra line

ncopa:
NACK. according commit 79df80b23fcb065914d045cc68ebc355c5a7952d:

>     We splitted out the ui libs and tools to avoid the circular deps:
>     avahi ->  gtk -> cups -> avahi

PR #9296, opened 2019-07-04
community/mpv: add PulseAudio support

PureTryOut:


maxice8:
Is Pulseaudio really required ? i get it working just fine with pulseaudio and pulseaudio-alsa/alsa-plugins-pulse

PureTryOut:
It's not required no but it's definitely recommended if you want stuff like surround sound. Yes it can work with Alsa, but [it requires manually changing your Alsa configs](https://github.com/mpv-player/mpv/wiki/ALSA:-Surround-Sound-and-Upmixing) and such. I basically always prefer direct PulseAudio support for ease of use.

Btw, it's still completely optional after this change. When I look back at adding PulseAudio support for audacious-plugins, a new runtime dependency `libpulse` will be added which is only 1.46MB in size. However, it won't be used or even usable without full PulseAudio installed. So if the user wants mpv but not PulseAudio, just don't install the latter.

PR #9297, opened 2019-07-04
testing/firefox: add PulseAudio support

PureTryOut:


Cogitri:
Sooo...sorry for not looking at this sooner, but do you still want to have this merged?

PureTryOut:
Personally, yes. But we should see if the extra PulseAudio bits can be split out of the main package somehow.

maxice8:
Sadly no, the pulseaudio library will be linked to /usr/lib/firefox/libxul.so just like libasound (alsa) is linked for us since we use ALSA

PureTryOut:
Damn. Maybe the application can be built 2 times, once with Pulse and once without?

Honestly, I'm amazed it can even still compile without PulseAudio, as Firefox dropped Alsa support a few versions ago.

maxice8:
have a `firefox` and `firefox-pulse` ?

PureTryOut:
Basically yes, should be easy enough

Cogitri:
> Damn. Maybe the application can be built 2 times, once with Pulse and once without?

That seems kinda hacky though, having flavours by adding two different packages is kind of annoying :/

 > Honestly, I'm amazed it can even still compile without PulseAudio, as Firefox dropped Alsa support a few versions ago.

They just dropped support for it from the official binaries, it's still supported in the source code IIRC. Anyway, I personally wouldn't mind just enabling pulse here.

PureTryOut:
The fact that Mozilla doesn't actively support anything other than Pulse anymore should be enough reason in itself to enable it imo. If you don't want PulseAudio, Firefox is just not the right browser for you anymore although that could of course change again in the future.

So should I rebase this and can we get it merged after all? I'll recreate it as a MR on Gitlab.

Cogitri:
Sounds good

PureTryOut:
Recreated on [Gitlab](https://gitlab.alpinelinux.org/alpine/aports/merge_requests/1051), closing this.

PR #9273, opened 2019-07-03
testing/mongo-c-driver: APKBUILD cleanup

sergey-safarov:

PR #9336, opened 2019-07-06
mongo-php7-library: reverted from non-free to testing

sergey-safarov:
mongo-php7-library: reverted part of 8a901de31fa055ed591d487e12f8bb9ffcc0df21 commit
mongo-php7-library use [Apache 2.0 license](https://github.com/mongodb/mongo-php-library/blob/master/LICENSE)


andypost:
Btw this library depends on https://www.php.net/manual/en/set.mongodb.php

sergey-safarov:
yes, also this depend of [`mongo-c-driver`](https://github.com/alpinelinux/aports/pull/9273)
Need to wait now.

maxice8:
This package needs `php7-pecl-mongodb` to be moved from non-free to testing as well.

PR #9339, opened 2019-07-06
community/php7: upgrade to 7.3.7

andypost:
Ref https://www.php.net/archive/2019.php#id2019-07-04-1


andypost:
aarch64 is very slow on tests `7766/12660` (but takes more then hour)

PR #9353, opened 2019-07-08
community/rust: enable use of our triplets, add support for aarch64

Cogitri:
Once this is merged and the new tarballs have been built it should be OK
building packages against the new arches for rust.


community/rust/alpine-target.patch
```diff
@@ -23,229 +23,99 @@ and distribution-specific quirks instead of polluting the main musl target of
...
-+        linker_flavor: LinkerFlavor::Gcc,
-+        options: TargetOptions {
-+            abi_blacklist: super::arm_base::abi_blacklist(),
-+            target_mcount: "\u{1}mcount".to_string(),
-+            .. base
-+        },
-+    })
++    let mut base = super::i686_unknown_linux_musl::target()?;
++
++       base.otpions.cpu = "pentium4".to_string();
```
tcely:
```suggestion
+       base.options.cpu = "pentium4".to_string();
```

community/rust/alpine-target.patch
```diff
@@ -23,229 +23,99 @@ and distribution-specific quirks instead of polluting the main musl target of
...
-+        target_vendor: "alpine".to_string(),
-+        linker_flavor: LinkerFlavor::Gcc,
-+        options: TargetOptions {
-+            abi_blacklist: super::arm_base::abi_blacklist(),
-+            .. base
-+        },
-+    })
++    let mut base = super::armv7_unknown_linux_musleabihf::target()?;
++
++    base.llvm_target = "armv7-alpine-linux-musleabihf".to_string();
```
tcely: The linker errors suggest the float-abi needs adjusting.

community/rust/alpine-target.patch
Cogitri: Yup, couldn't push while I'm connected to the alpine infra vpn to correct this

tcely: No problem. I pushed that and x86_64 built successfully.

Cogitri: Yup, works for me on x86_64 too and aarch64. Currently compiling on aarch64 with our triplets again, since that didn't work before.

community/rust/alpine-target.patch
Cogitri: Oh, but do we build another armv7 target into LLVM? It was my impression that that's our only armv7 target

Cogitri:
Hey @tcely, although I very much appreciate the help could we maybe first get Rust actually working on the crossarches (pinned it down to the triplets), and then do other changes please? I just don't want to break yet another thing before I'm sure that it works :)

tcely:
@Cogitri That's fine. FYI, the build system for rust worked with python3 without any trouble. lldb may need further attention.

For aarch64, what's the plan for bootstrapping so that a version can get to the mirrors?

tcely:
Is there any reason not to use `--enable-parallel-compiler`? It seems like it could be a big win in these days of huge numbers if cores.

Cogitri:
> aarch64, what's the plan for bootstrapping so that a version can get to the mirrors?

The same as the other arches if it doesn't segfault for some reason on aarch64 CI and builders (but magically works on my box 🤔)

> build system for rust worked with python3 without any trouble

Upstream still hardcodes py2 so I'd prefer to keep it at that

> Is there any reason not to use --enable-parallel-compiler? It seems like it could be a big win in these days of huge numbers if cores.

That it's still experimental and we're already having problems with Rust SIGSEVing on musl so I'd rather not worsen that :/

community/rust/APKBUILD
```diff
@@ -6,8 +6,8 @@
 pkgname=rust
 pkgver=1.36.0
 _llvmver=8
-_bootver=1.35.0
-pkgrel=0
+_bootver=1.36.0
+pkgrel=1
 pkgdesc="The Rust Programming Language"
 url="https://www.rust-lang.org"
 arch="x86_64 armv7 armhf"
```
tcely:
```suggestion
arch="x86_64 aarch64 armv7 armhf"
```
We need the builder to not exit immediately.

tcely:
Removing `noarch` from `-src` looks like a better option than disabling `archcheck` in `options`.

community/rust/APKBUILD
Cogitri: Well, I'm currently working on that but until then we'll need to have aarch64 disabled.

Cogitri:
Will reopen as draft.

PR #9367, opened 2019-07-08
testing/retroarch: fix s390x

tuan-hoang1:
Include directfb header path explicitly


tuan-hoang1:
It seems x86_64 has the same issue.

PR #9371, opened 2019-07-08
main/alpine-baselayout: improve color_prompt

dunielpls:
This commit will alter color_prompt in order to display the default $PS1
prompt in the form of 'user@hostname' instead of 'hostname' alone.

It will also display the full hostname rather than just the hostname up
to the first '.' character.

### DIFF
--- a/main/alpine-baselayout/color_prompt
+++ b/main/alpine-baselayout/color_prompt
@@ -4,7 +4,7 @@ NORMAL="\[\e[0m\]"
 RED="\[\e[1;31m\]"
 GREEN="\[\e[1;32m\]"
 if [ "$USER" = root ]; then
-	PS1="$RED\h [$NORMAL\w$RED]# $NORMAL"
+	PS1="$RED\u$NORMAL@$RED\H [$NORMAL\w$RED]$NORMAL# "
 else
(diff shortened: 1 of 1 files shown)


ncopa:
I guess this is a question of taste. I personally prefer when the prompt is as compact as possible, so there is as much space as possible for the path. I would even like to remove the `[` and `]` chars from prompt. Something like:

````sh
NORMAL='\[\e[0m\]'
case "$USER" in
	root) COLOR='\[\e[1;31m\]';; # red
	*) COLOR='\[\e[1;32m\]';; # green
esac
export PS1="${COLOR}\h${NORMAL}:\w${COLOR}\\\$${NORMAL} "
````

dunielpls:
I decided I'd modify as little as possible and retain the original prompt - only really wanted to add the username and full hostname for security reasons.

What do you want to do about this?

nmeum:
Honestly, I don't see the benefit of shipping this file in the first place. It is not enabled by default and prompt configuration is (as the discussion here shows) a matter of personal preference. Just my two cent.

dunielpls:
This is in essence a stale mate, and no conclusion will be reached.

I am therefore closing this pull request.

PR #9373, opened 2019-07-09
WIP: testing/firefox: upgrade to 68.0

timbru31:
This updates Firefox to 68.0.  
It also enables the `enable-rust-simd` flag as proposed in https://github.com/alpinelinux/aports/pull/8754#issuecomment-501457795

### DIFF
--- a/testing/firefox/APKBUILD
+++ b/testing/firefox/APKBUILD
@@ -1,7 +1,7 @@
 # Contributor:
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
-pkgver=67.0.4
+pkgver=68.0
 pkgrel=0
 pkgdesc="Firefox web browser"
 url="https://www.firefox.com/"
@@ -121,6 +121,7 @@ build() {
 		--enable-system-sqlite \
 		--enable-ffmpeg \
 		--enable-hardening \
+		--enable-rust-simd \
 		\
 		--with-system-bz2 \
 		--with-system-icu \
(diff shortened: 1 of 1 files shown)


PureTryOut:
Do you actually use Firefox on Alpine? It has been suffering from a crash for quite a while if a password field is focused. This basically makes the browser unusable for me, as I can not sign in to _any_ services, not even Firefox Sync.

https://bugs.alpinelinux.org/issues/10451

Could this be fixed together with this upgrade maybe?

timbru31:
Yes we use this package in our CI pipeline for headless testing. I wasn't aware of any crashes so far.

liqa:
You should delete all instances of paxmark and paxflags. Alpine doesn't use grsecurity for a long time.
You can delete the xulrunner symlink as its cruft from a very old release.

You probably can delete the launcher symlink below the xulrunner one. It says it was needed for firefox-7 we are now at firefox-68 (but I haven't tried myself, worth checking)

The less stuff and cruft in the APKBUILD the easier to spot bugs.

Cogitri:
@PureTryOut Huh, FWIW I'm not hit by that bug :o

Cogitri:
Could you change `--enable-default-toolkit=cairo-gtk3` to `--enable-default-toolkit=cairo-gtk3-wayland` please? That makes it possible to use FF in wayland mode (but keeps using Xorg if you're not on wayland).

liqa:
There're similar bugs in other distros.
https://github.com/void-linux/void-packages/issues/12885
https://github.com/gentoo/musl/issues/44#issuecomment-397825350

Anyway this shouldn't prevent this PR from being merged IMO.
That bug is a different discussion...

stefson:
The choice of using either bfd/gold/lld is now done via `--enable-linker=`, where the value is one of those three. With musl you should force bfd via `--enable-linker=bfd`, as using gold results in a broken binary with missing symbols all over the place. Same for spidermonkey by the way. 

`elfhack` should work again, if using gcc and not clang.

stefson:
And some of the patches have to be backported, in case you haven't noticed yet.

Ikke:
#9432 already has upgraded firefox to 68.0, closing.

PR #9378, opened 2019-07-09
testing/gdal: upgrade to 3.0.0

abellgithub:


abellgithub:
The current PDAL package (1.9.1) doesn't work with the new version of GDAL, but it works with the current version.  Is there some way that I can pin it and maintain the older version of PDAL and not make it a dependency of this GDAL 3.0 package?

PR #9393, opened 2019-07-10
main/python2: enable optimizations

J0WI:


main/python2/APKBUILD
```diff
@@ -57,13 +56,14 @@ build() {
...
 		--enable-ipv6 \
-		--with-system-ffi \
+		--enable-optimizations \
+		--enable-shared \
+		--enable-unicode=ucs4 \
 		--with-system-expat \
+		--with-system-ffi \
 		--with-system-zlib \
-		--enable-unicode=ucs4
+		--with-threads
```
Cogitri: Why the move of options here? It makes the diff more confusing than it has to be.

J0WI: I sorted them just like in `main/python3`.

ncopa:
e3c2cc82f61b903619474889096739a168a1a552

PR #9397, opened 2019-07-10
[3.10] main/nsd: upgrade to 4.2.1

iggy:
**Summary**

* Bump version
* Add myself as a contributor

Backporting specifically for `Disable TLS1.0, TLS1.1 and weak ciphers` that was added in 4.2.0.

Not sure if this is considered too risky for `-stable`

@ncopa (hope it's okay to @ you since you are listed as maintainer, if not, feel free to let me know and I won't do it in the future)

### DIFF
--- a/main/nsd/APKBUILD
+++ b/main/nsd/APKBUILD
@@ -1,9 +1,10 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
 # Contributor: Matt Smith <mcs@darkregion.net>
 # Contributor: Leonardo Arena <rnalrd@alpinelinux.org>
+# Contributor: Iggy Jackson <iggy@theiggy.com>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=nsd
-pkgver=4.1.27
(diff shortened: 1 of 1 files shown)


ncopa:
> Backporting specifically for `Disable TLS1.0, TLS1.1 and weak ciphers` that was added in 4.2.0.

I want the above line in commit message, to make it clear for everyone why it is backported.

> Not sure if this is considered too risky for `-stable`

The general rule of thumb is that people using stable branches should be able to do `apk upgrade`, without anything breaking or without need to modify any configuration files. So this is find as long as the 4.1 -> 4.2 upgrade does not require any config changes. (so a link to the release notes would have been helpful in this case)

Can you please also squash the commits into a single? there is no point having them in 2 separate commits.
 
> @ncopa (hope it's okay to @ you since you are listed as maintainer, if not, feel free to let me know and I won't do it in the future)

Its perfectly fine to @ me, but they kind of drown, since I get a notification for every single PR anyway.

Thanks!

iggy:
I'll just let this one go for now. It sounds like it might be too big of a change.

PR #9415, opened 2019-07-12
main/libpciaccess: upgrade to 0.15

maxice8:


main/libpciaccess/APKBUILD
```diff
@@ -29,6 +28,12 @@ package() {
 	install -Dm644 COPYING "$pkgdir"/usr/share/licenses/$pkgname/COPYING
 }
 
-sha512sums="bf40214dfd22f2a223f7c948566e6eaa4bfed60116e71b3eb19f6fe7ab2c4546649eb0416c06d5983e476ea5e27de036936646eb5f800ea97763b6d1f1976f4b  libpciaccess-0.14.tar.bz2
+static() {
```
Ikke: Any reason you manually define this function? There is a default static function implemented that should cover this.

PR #9424, opened 2019-07-12
main/mc: upgrade to 4.8.23

andypost:
Ref https://midnight-commander.org/wiki/NEWS-4.8.23


maxice8:
There is #9041 which you approved

PR #9436, opened 2019-07-13
community/neovim: enable aarch64

z3ntu:
See NixOS/nixpkgs#61588

### DIFF
--- a/community/lua-bitop/APKBUILD
+++ b/community/lua-bitop/APKBUILD
@@ -1,0 +1,48 @@
+# Contributor:
+# Maintainer:
+pkgname="lua-bitop"
+pkgver="1.0.2"
+pkgrel=0
+pkgdesc="Bitwise operations on numbers"
+url="https://bitop.luajit.org/"
+arch="all"
+license="MIT"
+source="https://bitop.luajit.org/download/LuaBitOp-$pkgver.tar.gz"
+builddir="$srcdir/LuaBitOp-$pkgver"
+
+_luaversions="5.1 5.2 5.3"
+for _v in $_luaversions; do
+	makedepends="$makedepends lua$_v-dev"
+	subpackages="$subpackages lua$_v-${pkgname#lua-}:_subpackage"
+done
+
+prepare() {
+	default_prepare
+
+	local lver; for lver in $_luaversions; do
+		cp -r "$builddir" "$builddir-$lver"
+	done
+}
+
+build() {
+	local lver; for lver in $_luaversions; do
+		msg "Building for Lua $lver..."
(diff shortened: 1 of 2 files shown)


community/lua-bitop/APKBUILD
```diff
@@ -0,0 +1,51 @@
+# Contributor:
+# Maintainer:
```
maxice8: Why no maintainer ?

community/lua-bitop/APKBUILD
```diff
@@ -0,0 +1,51 @@
...
+pkgname="lua-bitop"
+pkgver="1.0.2"
+pkgrel=0
+pkgdesc="Bitwise operations on numbers"
+url="https://bitop.luajit.org/"
+arch="all"
+license="MIT"
+makedepends=""
```
maxice8: empty variables can be removed for brevity

maxice8:
1. New package should start in testing then be moved to community/
2. New package has no maintainer

community/lua-bitop/APKBUILD
z3ntu: I'd rather not put myself as maintainer for that package as I have little interest in lua and probably don't know how to fix any problems if they come up...

community/lua-bitop/APKBUILD
z3ntu: Will do

z3ntu:
> New package should start in testing then be moved to community/

As lua-bitop will be a dependency of neovim on aarch64, I don't think it can be in testing?

> New package has no maintainer

@jirutka You seem to be maintaining a lot of lua packages, would you maybe adopt this one?

community/neovim/APKBUILD
```diff
@@ -19,10 +19,15 @@ pkgdesc="Vim-fork focused on extensibility and agility"
 url="https://neovim.io"
 # s390x: needs luajit
 # aarch64: fails to build
-arch="all !s390x !aarch64"
+arch="all !s390x"
 license="Apache-2.0 custom"
 makedepends="cmake gettext-dev gperf libtermkey-dev libuv-dev libvterm-dev
-	lua5.1-lpeg lua5.1-mpack luajit-dev msgpack-c-dev unibilium-dev"
+	lua5.1-lpeg lua5.1-mpack msgpack-c-dev unibilium-dev"
+if [ "$CARCH" = "aarch64" ]; then
```
tcely: Using `case "${CARCH}" in` allows for easier expansion to other architectures.

tcely:
We have LuaJIT for `aarch64` is using that an option?

PR #9446, opened 2019-07-14
community/fossil: enable check

tcely:


tcely:
`Final results: 50 errors out of 36205 tests`

A lot of tests are bogus because of outdated lists or expected output. I'm not going to review them all now.

PR #9450, opened 2019-07-14
community/py-sip: split into -pyqt5 and -sip packages

maxice8:
resolves #10529

@fcolista


fcolista:
Thanks @maxice8 : 
LGTM. Can you please also update the dependency from py-sip?
The pacakges are only the following ones (quick and dirty search):
``
$ grep -Er 'py.?-sip' */APKBUILD   | cut -d/ -f1 | uniq

gns3-gui
py-sip
py3-qt5
py3-qtwebengine
py3-sip
``
py-sip is used only on community repo, no depends to/from main/testing.
Thanks.
.: Francesco

maxice8:
Updated

PR #9451, opened 2019-07-14
main/taskd: fix url add missing bash dep

maxice8:


maxice8:
Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/5026

PR #9452, opened 2019-07-14
main/check: enable fork

maxice8:
Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/8538

@clandemeter

### DIFF
--- a/main/check/APKBUILD
+++ b/main/check/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Fabian Affolter <fabian@affolter-engineering.ch>
 pkgname=check
 pkgver=0.12.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A unit test framework for C"
 url="https://libcheck.github.io/check/"
 arch="all"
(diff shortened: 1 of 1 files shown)


main/check/APKBUILD
```diff
@@ -2,30 +2,25 @@
...
-pkgrel=1
+pkgrel=2
 pkgdesc="A unit test framework for C"
 url="https://libcheck.github.io/check/"
 arch="all"
 license="LGPL-2.0-or-later"
 makedepends="pkgconfig"
 checkdepends="diffutils gawk"
-subpackages="$pkgname-doc $pkgname-dev"
+subpackages="$pkgname-static $pkgname-doc $pkgname-dev"
```
nmeum: I wasn't around for the entire `-static` discussion but does it really make sense to move a `76.0K` ar archive to a separate subpackage?

maxice8: Yes, the point of removing it from -dev is to avoid accidentally linking to it.

But in the case of check there isn't much point since check is compiled statically into tests most of the time

PR #9465, opened 2019-07-15
community/docker: enable seccomp support

maxice8:
@tomalok

Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/9028


tomalok:
I'm all for getting this in, but we should be clear where this needs to be enabled...  here, `containerd`, and/or `runc`?  (I'm not able to dig into this at the moment.

maxice8:
well `runc` and `containerd` already have seccomp enabled, `docker` seems to be the last component.

tomalok:
@maxice8 - excellent, thanks for checking!

PR #9470, opened 2019-07-15
community/arpack: Removed depends

ossdev07:


maxice8:
If you're going to clean up the package might as well fix all violations found by `apkbuild-lint` from the atools package.

PR #9474, opened 2019-07-15
community/ntpsec: upgrade to 1.1.6

tcely:


community/ntpsec/APKBUILD
```diff
@@ -1,18 +1,19 @@
 # Contributor: tcely <ntpsec+aports@tcely.33mail.com>
 # Maintainer: tcely <ntpsec+aports@tcely.33mail.com>
-pkgname=ntpsec
-pkgver=1.1.3
+_pkgname="ntpsec"
+pkgname="${_pkgname}"
```
ncopa: I don't think `_pkgname` makes any sense in this APKBUILD

tcely: It is a way to distinguish which uses of pkgname we control. I prefer to make the decision up front. This reduces cognitive load when renaming this package. Don't let it bother you too much.

PR #9483, opened 2019-07-16
community/pdns-recursor: upgrade to 4.2.0

tcely:


community/pdns-recursor/APKBUILD
```diff
@@ -1,22 +1,24 @@
 # Contributor: Olivier Mauras <olivier@mauras.ch>
 # Maintainer: tcely <pdns-recursor+aports@tcely.33mail.com>
-pkgname=pdns-recursor
-pkgver=4.2.0
+_pkgname="pdns-recursor"
+pkgname="${_pkgname}"
```
ncopa: why is the `_pkgname` variable name introduced?

tcely: It is a convenient way to distinguish between upstream name and the Alpine Linux package name.

PR #9486, opened 2019-07-16
non-free/mongo-c-driver: move to testing, upgrade to 1.14

russkel:
Split https://github.com/alpinelinux/aports/pull/9244 into two.

Is incorrectly located in nonfree so it has been moved to testing.


tcely:
See also #9273

russkel:
"I was first"

PR #9500, opened 2019-07-16
main/strace: upgrade to 5.2

nmeum:
Currently testing this on my machine, also wanted to check if it passes on other architectures.


nmeum:
Been casually using this for the last couple of days and didn't run into any issues.

PR #9503, opened 2019-07-16
testing/py3-typed-ast: fix tests

prspkt:


maxice8:
Should we use `local` ? it is not specified in POSIX

prspkt:
yeah, I know. we are already using it elsewhere to limit the scope so it wouldn't kill us. It is probably OK for now.

maxice8:
Sounds fair, feel free to push if CI is green

PR #9510, opened 2019-07-17
community/knot: upgrade to 2.8.3

tcely:


ncopa:
Please do not add style changes to PRs which are marked as high prio, or don't mark PRs with style changes as high prio. It only slows us down and forces us to either spend time on "meaningless" style changes or makes us ignore the high prio label.

tcely:
You are welcome to spend less time caring about the style changes I prefer to see in the package I maintain. The high priority is getting the upgrade into the mirrors. I already split the commits so making separate pull requests would be very little trouble.

ncopa:
> The high priority is getting the upgrade into the mirrors.

Ok. I will push that separately then and deal with the other stuff another time.

Thanks!

PR #9511, opened 2019-07-17
main/libxkbcommon: provide static library

maxice8:
Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/10663


ncopa:
I believe that commit message `fixes #10663` should work

PR #9516, opened 2019-07-17
main/libcrystalhd: Enable aarch64 Support

ossdev07:
Added support for aarch64 in APKBUILD by providing a patch to enable the support.


ncopa:
This is a rather large/intrusive patch. Is it applied upstream? If not, why?

ossdev07:
I have rebased the PR on the latest master as per your suggestion, Please can you consider this for a merge.

Cogitri:
Well, why isn't this patch upstream? I'd rather not have a ~2K LOC patch downstream :/

PR #9517, opened 2019-07-17
main/gcc: Add libstdc++-dev subpackage

j8r:
When compiling a program statically, installing `libstdc++` isn't enough. `g++` has to be added to to have `libstdc++.a`. I noticed it in https://github.com/j8r/dockerfiles/issues/4.

### DIFF
--- a/main/gcc/APKBUILD
+++ b/main/gcc/APKBUILD
@@ -114,7 +114,7 @@ fi
 
 _languages=c
 if $LANG_CXX; then
-	subpackages="$subpackages libstdc++:libcxx:$CTARGET_ARCH g++$_target:gpp"
+	subpackages="$subpackages libstdc++:libcxx:$CTARGET_ARCH libstdc++-dev:libcxx_dev:$CTARGET_ARCH g++$_target:gpp"
 	_languages="$_languages,c++"
 fi
 if $LANG_OBJC; then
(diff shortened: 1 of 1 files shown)


j8r:
I don't quite understand why it's failing, `gcc-gnat` should be already installed when we want to build ADA.

j8r:
Any idea @ncopa ?

j8r:
In fact adding `gcc-gnat` to `makedepends_build` is ignored.

Cogitri:
It most likely is ignored because pkgs can't depend on themselves (or their subpkgs). See e.g. Rust which needs to provide cargo-bootstrap and rust-bootstrap to hack around this. I think GCC should work around this with the bootstrapping magic that's in the APKBUILD though?

j8r:
Related GitLab issue: [GCC fails to build due to gcc-gnat not being installed](https://gitlab.alpinelinux.org/alpine/aports/issues/10738)

Cogitri:
@ncopa

ncopa:
So, if I understand this correctly, you want be able to link against `libstd++.a` without having `g++` installed?

The price for this is that we get same problem with `libstdc++-dev` as we have with `gcc-gnat`. To build the package, you will need to have `libstdc++-dev` installed, which introduces another circular dependency. I`m not sure we want that?

ncopa:
Also, if we do this, shouldn´t we also include all the c++ headers and symlinks to shared libs?

j8r:
Yes, it was mainly to link against the `.a` without installing the whole `g++` package.
I agree to avoid cycling dependencies, it will complicate bootstrap.

[Fedora is using a macro](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping) to avoid circular dependencies for bootstrapping.
Debian seems to have `udeb` for this purpose.

PR #9521, opened 2019-07-17
testing/stgit: new aport

jbenden:
http://www.procode.org/stgit/
Manage a stack of patches using GIT as a backend

### DIFF
--- a/testing/stgit/APKBUILD
+++ b/testing/stgit/APKBUILD
@@ -1,0 +1,51 @@
+# Contributor: Joseph Benden <joe@benden.us>
+# Maintainer: Joseph Benden <joe@benden.us>
+
+pkgname=stgit
+pkgver=0.19
+pkgrel=0
+pkgdesc="Manage a stack of patches using GIT as a backend"
+arch="noarch"
+url="http://www.procode.org/stgit/"
+license="GPL-2.0-or-later"
+options="!check !strip" # unit-tests fail due to iconv usage
+depends="git python3"
+makedepends="asciidoc xmlto perl"
+subpackages="$pkgname-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/ctmarinas/stgit/archive/v${pkgver}.tar.gz"
+builddir="$srcdir/${pkgname}-$pkgver"
+
+prepare() {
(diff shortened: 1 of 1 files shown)


testing/stgit/APKBUILD
```diff
@@ -0,0 +1,51 @@
+# Contributor: Joseph Benden <joe@benden.us>
+# Maintainer: Joseph Benden <joe@benden.us>
+
+pkgname=stgit
+pkgver=0.19
+pkgrel=0
+pkgdesc="Manage a stack of patches using GIT as a backend"
+arch="noarch"
+url="http://www.procode.org/stgit/"
+license="GPL"
```
maxice8: ambiguous and not SDPX compliant

testing/stgit/APKBUILD
```diff
@@ -0,0 +1,51 @@
...
+options="!check !strip" # unit-tests fail due to iconv usage
+depends="git python3"
+makedepends="python3 asciidoc xmlto perl"
```
maxice8: python3 is in depends=, doesn't need to be in makedepends

testing/stgit/APKBUILD
jbenden: Fixed. Thank you for the review!

testing/stgit/APKBUILD
jbenden: Fixed. Thanks!

testing/stgit/APKBUILD
```diff
@@ -0,0 +1,51 @@
...
+pkgdesc="Manage a stack of patches using GIT as a backend"
+arch="noarch"
+url="http://www.procode.org/stgit/"
+license="GPL-2.0-or-later"
+options="!check !strip" # unit-tests fail due to iconv usage
+depends="git python3"
+makedepends="asciidoc xmlto perl"
+subpackages="$pkgname-doc"
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/ctmarinas/stgit/archive/v${pkgver}.tar.gz"
+builddir="$srcdir/${pkgname}-$pkgver"
```
maxice8: default value, can be removed

PR #9522, opened 2019-07-17
testing/topgit: new aport

jbenden:
https://github.com/mackyle/topgit
A different patch queue manager

### DIFF
--- a/testing/topgit/APKBUILD
+++ b/testing/topgit/APKBUILD
@@ -1,0 +1,27 @@
+# Contributor: Joseph Benden <joe@benden.us>
+# Maintainer: Joseph Benden <joe@benden.us>
+
+pkgname=topgit
+pkgver=0.19.12
+pkgrel=0
+pkgdesc="A different patch queue manager"
+arch="noarch"
+url="https://github.com/mackyle/topgit"
+license="GPL-2.0-or-later"
+options="!check !strip"
+depends="git gawk sed"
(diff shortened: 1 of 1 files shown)


testing/topgit/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Joseph Benden <joe@benden.us>
+# Maintainer: Joseph Benden <joe@benden.us>
+
+pkgname=topgit
+pkgver=0.19.12
+pkgrel=0
+pkgdesc="A different patch queue manager"
+arch="noarch"
+url="https://github.com/mackyle/topgit"
+license="GPL"
```
maxice8: Ambiguous and not SPDX compliant

testing/topgit/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+options="!check !strip"
+depends="git gawk sed"
+makedepends="gawk sed perl"
```
maxice8: gawk and sed are in depends= no need to be in makedepends

testing/topgit/APKBUILD
jbenden: Fixed. Thank you for the review!

testing/topgit/APKBUILD
jbenden: Ok, thanks!

PR #9523, opened 2019-07-17
testing/virtualbox: new aport

jbenden:
https://virtualbox.org/
General-purpose full virtualizer for x86 hardware

This aport deprecates all other virtualbox related aports (and may conflict with); as it provides for both host and/or guest operation.

Given the size of this aport, I am **completely** open to discussion and improvements upon it, prior to full commit to master.

### DIFF
--- a/testing/virtualbox/002-dri-driver-path.patch
+++ b/testing/virtualbox/002-dri-driver-path.patch
@@ -1,0 +1,28 @@
+--- a/src/VBox/Additions/common/crOpenGL/fakedri_drv.c
++++ b/src/VBox/Additions/common/crOpenGL/fakedri_drv.c
+@@ -56,24 +56,8 @@ typedef enum {
+ 
+ //#define DEBUG_DRI_CALLS
+ 
+-/// @todo this could be different...
+-#ifdef RT_ARCH_AMD64
+-# ifdef RT_OS_FREEBSD
+-#  define DRI_DEFAULT_DRIVER_DIR "/usr/local/lib/dri"
+-#  define DRI_XORG_DRV_DIR "/usr/local/lib/xorg/modules/drivers/"
+-# else
+-#  define DRI_DEFAULT_DRIVER_DIR "/usr/lib64/dri:/usr/lib/dri:/usr/lib/x86_64-linux-gnu/dri:/usr/lib/xorg/modules/dri"
++#  define DRI_DEFAULT_DRIVER_DIR "/usr/lib/xorg/modules/dri"
+ #  define DRI_XORG_DRV_DIR "/usr/lib/xorg/modules/drivers/"
+-# endif
+-#else
+-# ifdef RT_OS_FREEBSD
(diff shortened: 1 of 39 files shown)


jbenden:
Why does Drone error due to not being able to build for x86 - shouldn't it skip this?

Also, Travis won't build this package due to log length - should I reduce the build verbosity levels?

jbenden:
I have finished building/testing the package, within the docker-abuild builder and locally.

Thanks,
-Joe

ncopa:
I think we should have the kernel modules in separate APKBUILD to avoid need to rebuild the entire virtualbox package every time there is a new kernel version.

jbenden:
Is there a way to keep duplication of variables and such to a minimum? Would I be able to safely `source` a common file between the multiple packages?

Thanks,
-Joe

jbenden:
Hello,

I have found two ways of keeping duplication to a minimum and wish for feedback to help steer me in the best direction.

Before enumerating the options, I wanted to share that I am able to `source` other `APKBUILD` files using relative paths. This works with `abuild -r` and `abuild -r rootbld`. I am hopeful this will not cause errors elsewhere.

**Option # 1**: Use a shell script in the "main" package directory, to directly copy all files to the other package directories. This shell script is used during package maintenance, to ensure all files are properly distributed outwards towards the other sub-packages.

**Option # 2**:  With two small adjustments to `abuild` software, I was able to support `source` containing relative paths. The small adjustments change abuild's `unpack` and `prepare` functions; so they use `startdir` instead of `srcdir`. This option allows each of the subpackages to refer to files stored in the main primary package.

On an aside: I'd like to recommend a globally included build file, which contain predefined variables of interest to all packages. It would be nice if packages just "knew" the current versioning details of the kernel package. Alternatively, a function to return this information, given a package, would also work.

I look forward to comments, suggestions, and ideas!
-Joe

Cogitri:
Well, we have the virtual-guest* things as seperate packages too, so it'd make sense to split this up too,  I suppose.

jbenden:
Yes, it would.

I am waiting for a plan on de-duplication and reliability against change.

I have found that FreeBSD has handled the same problem by using an include which contains shared variables. [See here](https://github.com/freebsd/freebsd-ports/blob/master/Mk/bsd.default-versions.mk) for their instance of all shared default versions.

I wish for our copy to contain the default kernel version we're all building against... Also, the default version of virtualbox, too.

Thoughts, comments, suggestions are welcomed,
-Joe

Cogitri:
Well, that'd be a rather bit change as to how abuild currently works and would make it a fair bit more complex, so I think we'd rather not do that

jbenden:
FYI: It's also not possible to relatively include other build files into APKBUILD files.

This, plus the lack of shared variables, translates to each and every package of Virtualbox requiring a ton of duplicate data. [So does the firefox and thunderbird APKBUILD files, too]

At this time I'd rather not handle virtualbox - so I am closing this PR.

Best regards,
-Joe

PR #9532, opened 2019-07-18
main/openssh: use ldns

nmeum:
Other distributions, e.g. Arch Linux also built openssh with ldns by default. Doing so fixes a segfault when trying to verify SSH host keys using DNS records.

---

Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/8323

### DIFF
--- a/main/openssh/APKBUILD
+++ b/main/openssh/APKBUILD
@@ -4,7 +4,7 @@
 pkgname=openssh
 pkgver=8.0_p1
 _myver=${pkgver%_*}${pkgver#*_}
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


ncopa:
This adds 360kb in dependencies. Wouldn't it be better to fix the segfault instead?

nmeum:
> Wouldn't it be better to fix the segfault instead?

Definitely, would just propose this as a workaround in the meantime. If I get around to it I will debug this further soonish.

nmeum:
Superseded by ed91ec54fcc777d8b5f9fb32f78bf14c4e8ea343

PR #9534, opened 2019-07-18
testing/shairport-sync: Improve APKBUILD

henriknil:
* Fix comments from apkbuild-lint
* Fix doc build warning


testing/shairport-sync/APKBUILD
```diff
@@ -2,29 +2,25 @@
 # Maintainer:
 pkgname=shairport-sync
 pkgver=3.3.1
-pkgrel=0
+pkgrel=2
```
maxice8: pkgrel should be bumped from 0 to 1

PR #9540, opened 2019-07-19
main/bind: upgrade to 9.14.4

tcely:
~~Depends on #8932~~

### DIFF
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -243,8 +243,12 @@ libs() {
 	default_libs
 }
 
-#gpg_signature_extensions="sha512.asc"
-#gpgfingerprints="good:AE3F AC79 6711 EC59 FC00  7AA4 74BB 6B9A 4CBB 3D38"
+
+gpg_signature_extensions="sha512.asc"
+gpgfingerprints="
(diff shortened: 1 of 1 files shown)


main/bind/APKBUILD
```diff
@@ -174,30 +182,58 @@ dev() {
...
+	depends="$depends_plugins"
+
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/bind "$subpkgdir"/usr/lib/
+}
+
+py3_isc() {
+	pkgdesc="Python functions to support BIND utilities"
+	depends="$depends_py3_isc"
+	provides="py3-isc=${_isc_ver}-r${_isc_rel} py3-$pkgname=$pkgver-r$pkgrel"
```
ncopa: why do you provide different version of py3-isc? This will create trouble and I cannot see the value it brings.

tcely: > why do you provide different version of py3-isc? This will create trouble and I cannot see the value it brings.

It has its own version. Why should it be given the version of `bind`? Just because it is bundled?

Ignoring the version of the python `isc` module is more likely to create trouble in my opinion.

PR #9543, opened 2019-07-20
main/linux-vanilla: Enables SUN4I USB PHY as a module

trixpan:
Fixes #10677


ncopa:
fixed with 37ed4242850483c6cdb6c4f3378050643cc76f16.

PR #9560, opened 2019-07-21
[3.10] main/py-django: fix CVE-2019-12781

maxice8:


ncopa:
Thanks, but in this case I think we should use the `1.11.22` release. (2bf8de409de29218b845158bf2dcfdce10b33611)

PR #9584, opened 2019-07-21
scripts/bootstrap.sh: Changes required to bootstrap ARM on x86

trixpan:
Introduce changes required to perform successful bootstrap of v3.10 ARMv7 environment on x86_64 hosts.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -101,10 +101,10 @@ for PKG in fortify-headers linux-headers musl libc-dev pkgconf zlib \
 	   gmp mpfr3 mpc1 isl cloog gcc \
 	   openrc alpine-conf alpine-baselayout alpine-keys alpine-base build-base \
 	   attr libcap patch sudo acl fakeroot tar \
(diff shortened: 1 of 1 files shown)


trixpan:
@fabled any feedback regarding this PR?

trixpan:
@fabled ?

fabled:
Removing  `community/go` and `community/ghc` seems unacceptable to me. Those need to be cross-compiled as they are "self-hosting" compilers. Do their cross-compile fail? Is it possible to fix that instead? Also, please file the fix against edge first and then request backport. We don't merge directly to stable branches if the issue affects edge too (unless the fix needs to be completely different).

The other hunk looks ok, but seems I bumped into it too, and had fixed it individually in edge. I cherry-picked that commit. Sorry for the overlapping work, and thank you for your effort on this.

fabled:
The sensible hunks were merged in other commit. No response on the bits that we consider not appropriate. Please reopen with alternate plan or additional explanation if problem persists. Closing for now.

PR #9603, opened 2019-07-21
testing/glog: Create libglog pkg-config file

maxice8:
@z3ntu


z3ntu:
@maxice8 As I've written yesterday on IRC, the pkg-config file should be called `libglog.pc` and not `glog.pc`, otherwise it won't be found by pkg-config

PR #9604, opened 2019-07-21
main/alpine-baselayout: create /var/mail

maxice8:
@ncopa

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s11.html

### DIFF
--- a/main/alpine-baselayout/APKBUILD
+++ b/main/alpine-baselayout/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=alpine-baselayout
 pkgver=3.1.2
-pkgrel=0
+pkgrel=1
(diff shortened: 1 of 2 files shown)


ncopa:
> @ncopa
> 
> https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s11.html

I would like an explanation in the commit message on why we do this change. We could cite FHS3:

> The logical location for this directory was changed from /var/spool/mail in order to bring FHS in-line with nearly every UNIX distribution. This change is important for inter-operability since a single /var/mail is often shared between multiple hosts and multiple UNIX distribution (despite NFS locking issues).

And add the link to commit message.

maxice8:
Done

ncopa:
i changed the version to `3.2.0` instead, since i think this is a rather significant change.

PR #9627, opened 2019-07-22
main/alpine-baselayout: remove postgres account

nmeum:
This change was made due to complains that /etc/passwd contains system
    accounts with login shells, postgres being one of them [0].
    
    Setting the shell of the postgres account to /sbin/nologin was not
    deemed feasible as the official document suggests login in as the
    postgres user via su(1) [1]. Other Linux distribution (Debian, Gentoo,
    Arch) also create the account with a shell.
    
    Not creating the postgres account by default (via alpine-baselayout) and
    instead creating it on demand from the package was deemed a reasonable
    compromise.
    
    [0]: https://gitlab.alpinelinux.org/alpine/aports/issues/10579
    [1]: https://wiki.postgresql.org/wiki/First_steps

Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/10579


main/postgresql/APKBUILD
```diff
@@ -3,13 +3,13 @@
...
-pkgrel=0
+pkgrel=1
 pkgdesc="A sophisticated object-relational DBMS"
 url="https://www.postgresql.org/"
 arch="all"
 license="PostgreSQL"
 depends="postgresql-client tzdata"
-install="$pkgname.pre-upgrade"
+install="$pkgname.pre-install $pkgname.pre-upgrade"
 pkgusers="postgres"
```
nmeum: I always get confused about `pkgusers`/`pkggroups` is that still needed when creating the users in a `pre-install` script?

ncopa: The `pre-install` script will not run when you build the package, so if the `postgres` user does not exist on the builder, things like `chown postgres` in APKBUILD will fail. abuild will create the `pkgusers` on the building system. `pre-install` creates it on the runtime system.

PR #9630, opened 2019-07-22
community/extra-cmake-modules: use py3-sphinx to build docs

maxice8:
@PureTryOut


PureTryOut:
Well, LGTM, but it would be nice to wait for a response before merging next time.

PR #9637, opened 2019-07-23
community/mtd-utils: upgrade to 2.1.1

oliv3r:
Version bump of mtd-utils


maxice8:
Commit title is wrong

oliv3r:
> Commit title is wrong

hah, I was only checking to see if the test where actually running on the CI, but it's fixed :) thanks

maxice8:
> hah, I was only checking to see if the test where actually running on the CI

So not supposed to be merged now ?

oliv3r:
> > hah, I was only checking to see if the test where actually running on the CI
> 
> So not supposed to be merged now ?

As all tests have passed, and the previous change about tests not working is resolved, yes, why not!

PR #9647, opened 2019-07-23
[3.9] main/zeromq: security upgrade to 4.3.2

maxice8:


main/zeromq/APKBUILD
```diff
@@ -10,13 +10,14 @@ license="LGPL-3.0-or-later with exceptions"
 makedepends="util-linux-dev libsodium-dev
 	perl python2 asciidoc xmlto autoconf automake
 	"
-subpackages="$pkgname-dev $pkgname-doc libzmq:libs"
+subpackages="libzmq-static $pkgname-dev $pkgname-doc libzmq:libs"
```
ncopa: we should not introduce static subpackage in stable branch unless absolutely necessary. It may break builds for users who only do `apk add zeromq-dev` to do static builds.

maxice8:
Sorry, copied it from 3.10/Edge

PR #9671, opened 2019-07-24
testing/sbcl: Correcting SBCL compilation failure, bumping package to…

durrendal:
… 1.5.4

Removing <signal.h> from musl-fixes.patch allows SBCL to compile correctly. Added myself as the maintainer.


durrendal:
It appears Drone is failing to build the package due to an unexpected error, however travis builds it normally with the exact same output.

I'm a little bit at a loss as to why Drone won't compile it when it works manually on an Alpine box, and Travis can do so as well.

maxice8:
pushed with a corrected commit title, please take note on how write future commit titles.

Also changed pkgrel to 0 as when the version is bumped the pkgrel **MUST** be at 0 **ALWAYS**

durrendal:
Thanks Maxice. I wasn't aware of the commit title. I'll keep both that and the pkgrel in mind next time.

PR #9686, opened 2019-07-25
CODINGSTYLE.md: use 'than' rather than 'then'

maxice8:
@ncopa

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -78,7 +78,7 @@ This ensures code is always neatly aligned and properly indented.
 Space before tab should be avoided.
 
 ## Line length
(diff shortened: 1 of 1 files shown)


CODINGSTYLE.md
```diff
@@ -78,7 +78,7 @@ This ensures code is always neatly aligned and properly indented.
 Space before tab should be avoided.
 
 ## Line length
-A line should not be longer then 80 lines. While this is not a hard limit, it
+A line should not be longer than 80 lines. While this is not a hard limit, it
```
prspkt:
```suggestion
A line should not be longer than 80 characters. While this is not a hard limit, it
```

maxice8: Fixed, thank you

PR #9710, opened 2019-07-27
[3.10] community/atools: upgrade to 18.1

maxice8:
@Ikke imo we should keep atools updated everywhere


Ikke:
It's not really policy to backport major releases especially because this is in community, so I would not merge this without consulting others.

That being said, I wonder how necessary it is to backport this.

PR #9719, opened 2019-07-27
Add support for Beagle Bone Black

vicencb:
Hi,
this patch series adds support for the Beagle Bone Black.
The first patch just enables the `IIO` module.
The second patch just enables support for `SOC_AM33XX`.
This first two changes generate an avalanche of differences, so, they are on standalone patches.
The third patch adds drivers for the BBB.

Regards,
  Vicenç.


vicencb:
There is a PR version 2 at #10043.

PR #9731, opened 2019-07-28
testing/fractal: upgrade to 4.2.0

Cogitri:
* Use musl gettext instead of GNU gettext


Cogitri:
Rust really likes to SIGSEV out of nowhere, but I guess this is good to to anyway.

PR #9748, opened 2019-07-29
mongo-c-driver, mongo-cxx-driver: fix depends_dev

russkel:
Should no longer need to specify libbson[cxx]-dev as an explicit makedep.

EDIT: Actually:

```
/usr/lib # ldd libmongocxx.so.0.0.0 
        /lib/ld-musl-x86_64.so.1 (0x7faddbd73000)
        libmongoc-1.0.so.0 => /usr/lib/libmongoc-1.0.so.0 (0x7faddbc54000)
        libbson-1.0.so.0 => /usr/lib/libbson-1.0.so.0 (0x7faddbc15000)
        libbsoncxx.so._noabi => /usr/lib/libbsoncxx.so._noabi (0x7faddbbf5000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7faddbaa0000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7faddba8c000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7faddbd73000)
        libssl.so.1.1 => /lib/libssl.so.1.1 (0x7faddba0c000)
        libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0x7faddb78e000)
        libsnappy.so.1 => /usr/lib/libsnappy.so.1 (0x7faddb782000)
        libz.so.1 => /lib/libz.so.1 (0x7faddb56b000)
```

but:
```
mongo-cxx-driver-3.4.0-r0 depends on:
so:libbson-1.0.so.0
so:libc.musl-x86_64.so.1
so:libgcc_s.so.1
so:libmongoc-1.0.so.0
so:libstdc++.so.6
```

So, we're not correctly finding the libbsoncxx.so._noabi dependency...

Should we add it explicitly or fix the underlying issue?

### DIFF
--- a/testing/mongo-c-driver/APKBUILD
+++ b/testing/mongo-c-driver/APKBUILD
@@ -1,12 +1,13 @@
 # Maintainer: Leonardo Arena <rnalrd@alpinelinux.org>
 pkgname=mongo-c-driver
 pkgver=1.14.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Client library written in C for MongoDB"
 url="https://github.com/mongodb/mongo-c-driver"
 arch="all"
 license="Apache-2.0"
 makedepends="openssl-dev snappy-dev zlib-dev libtool py3-sphinx cmake"
+depends_dev="libbson-dev"
 subpackages="
 	$pkgname-static:_static
 	libbson-static:_libbson_static
--- a/testing/mongo-cxx-driver/APKBUILD
+++ b/testing/mongo-cxx-driver/APKBUILD
@@ -9,7 +9,7 @@ url="http://mongocxx.org/"
 arch="all"
 license="Apache-2.0"
 makedepends="cmake mongo-c-driver-dev libbson-dev openssl-dev python3 snappy-dev"
-subpackages="$pkgname-dev:_dev libbsoncxx libbsoncxx-dev"
+subpackages="libbsoncxx-dev $pkgname-dev:_dev libbsoncxx"
 depends_dev="libbsoncxx-dev"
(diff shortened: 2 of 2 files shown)


maxice8:
That is weird, the pkgconfig file has Requires: libbson, which should be automatically detected.

Can you flip the order in which the 2 dev subpackages are made and see if it works ?

maxice8:
The function also needs to be adapted

maxice8:
I'll try here for mongo-c-driver

maxice8:
yeah no dice

russkel:
:thinking: is this an underlying issue with abuild?

maxice8:
Yes, it can't find /usr/lib/pkgconfig/libbson-1.0.pc because it only searches on the system.

maxice8:
while that file is actually in one of the packages being made, i filled an issue https://gitlab.alpinelinux.org/alpine/abuild/issues/9975

russkel:
Cool cool. I thought it used ldd to find the linked libs?

maxice8:
The Pkg-config is not about linked libs it is about files that describe
what the library name is, the version, what cflags ldflags you need to use
to link to the library and also a field called Requires: which lists other
Pkg-config files.

Abuild parses Requires and finds other dependencies that way but it only
works on packages installed on the system not other subpackages that are
also being created.

maxice8:
I pushed a modified version of your fix (added comment and moved the depends_dev to inside the function, otherwise libbson and libbsoncxx will depend on itself)

russkel:
Thanks. 

https://github.com/alpinelinux/aports/commit/01694a410638e63683aafe0601e4e8903de9f050#diff-8b52dc7bb0f63a7524005052934a8744R10
Is this right?

PR #9752, opened 2019-07-29
testing/gdm: gdm.initd needs dbus

Cogitri:


testing/gdm/gdm.initd
```diff
@@ -15,4 +15,5 @@ start_pre() {
 
 depend() {
         need net
+        need dbus
```
maxice8: can be on the same line i think

PR #9766, opened 2019-07-29
testing/sbcl: Upgrade to 1.5.5

durrendal:
Upgrade SBCL to 1.5.5, removed the fix-generate-version.patch as it doesn't appear to be necessary any longer.


maxice8:
Pushed with a few changes:

- Commit title was corrected (upgrade should be fully lowercase)
- Fixed typo on Contributor: (was written Contribut**e**r)

PR #9790, opened 2019-07-31
main/linux-vanilla: Enable CONFIG_SND_HDA_RECONFIG

Cogitri:
It's required for alsa-tools-gui's hdajackretask


ncopa:
what about armhf and armv7?

Cogitri:
Those didn't have the `# CONFIG_SND_HDA_RECONFIG not set` entry, so I assumed that they don't support that setting

PR #9801, opened 2019-07-31
testing/emulators, libraries etc...

sodomon2:
Contribution of emulators, libraries, programs, games etc.


maxice8:
The commits with the wrong title are still present, the new commits are missing a space between `:` and `new aport`

the new commits need to have their url= and pkgdesc= written down in the body.

Use this an an example: https://github.com/alpinelinux/aports/commit/6dea2fc511ed6636ef2f67328566e65c73cbc63f

PR #9815, opened 2019-08-01
community/php7: security upgrade 7.3.8

andypost:
- CVE-2019-11041
- CVE-2019-11042

Ref https://www.php.net/archive/2019.php#2019-08-01-1


andypost:
tests are slow on arch64 as always

PR #9823, opened 2019-08-01
testing/synapse: new aport

maxice8:
https://github.com/matrix-org/synapse
Matrix reference homeserver

### DIFF
--- a/testing/synapse/synapse.initd
+++ b/testing/synapse/synapse.initd
@@ -1,0 +1,23 @@
+#!/sbin/openrc-run
+supervisor=supervise-daemon
+
+name="synapse"
+description="Daemon for Synapse, the Matrix reference homeserver"
+
+: ${command_user:="synapse:synapse"}
+: ${config:="/etc/synapse/homesever.yaml"}
+
+command="/usr/bin/python3"
+command_args="-m synapse.app.homeserver --config-path=$config $command_args"
+
+pidfile="/run/${RC_SVCNAME}.pid"
+
+depends() {
+	need net
+}
+
+start_pre() {
+	# /var/lib/synapse needs to be owned by the synapse user and so does the config location
+	checkpath --directory --owner "$command_user" /var/lib/synapse
+	checkpath --file --owner "$command_user" $config
+}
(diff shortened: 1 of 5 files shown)


maxice8:
needs changes to #9829

Cogitri:
Would be nice if it also provided an initd script like https://www.br0tkasten.de/knowledge/matrix-chat-server

Cogitri:
And a synapse user

testing/synapse/synapse.initd
```diff
@@ -0,0 +1,15 @@
...
+command_args="-m synapse.app.homeserver --config=$config $command_args"
+
+depends() {
+	need net
+}
+
+start_pre() {
+	# /var/lib/synapse needs to be owned by the synapse user and so does the config location
+	checkpath --directory --owner synapse:synapse /var/lib/synapse
+	checkpath --directory --owner synapse:synapse $config
```
Cogitri: --file

testing/synapse/synapse.initd
```diff
@@ -0,0 +1,15 @@
+#!/sbin/openrc-run
+
+name="synapse"
+command="/usr/bin/python3"
+command_args="-m synapse.app.homeserver --config=$config $command_args"
```
Cogitri: --config-file

testing/synapse/synapse.initd
Cogitri: Should synapse really own the config and not root?

maxice8: yes

testing/synapse/synapse.post-install
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+if [ ! -e etc/synapse/homeserver.yaml ]; then
+	python3 -m synapse.app.homeserver \
+		--server-name my.domain.name \
+		--config-path /etc/synapse/homeserver.yaml \
+		--generate-config \
+		--report-stats=yes
```
Cogitri: Ah, shouldn't this do `2>/dev/null` too?

PR #9848, opened 2019-08-02
testing/grafana: new aport

kaey:


testing/grafana/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+arch="all"
+license="Apache-2.0"
+makedepends="go nodejs yarn python"
+install="$pkgname.pre-install"
+subpackages="$pkgname-openrc"
+source="$pkgname-$pkgver.tar.gz::https://github.com/grafana/$pkgname/archive/v$pkgver.tar.gz
+	grafana.initd
+	grafana.confd
+	"
+builddir="$srcdir/$pkgname-$pkgver"
```
maxice8: default value, can be removed

testing/grafana/APKBUILD
```diff
@@ -0,0 +1,49 @@
...
+pkgname=grafana
+pkgver=6.2.5
+pkgrel=0
+_commit=6082d1982 # git rev-parse --short HEAD
+_stamp=1561485379 # git --no-pager show -s --format=%ct
+pkgdesc="Open source, feature rich metrics dashboard and graph editor"
+url="https://grafana.com"
+arch="all"
+license="Apache-2.0"
+makedepends="go nodejs yarn python"
```
maxice8: can python3 be used ?

testing/grafana/APKBUILD
kaey: done

testing/grafana/APKBUILD
kaey: python dep was optional, removed

kaey:
Nothing i can do about travis timeout failure. Welcome to the amazing world of webpack.

testing/grafana/APKBUILD
kaey: ok, it seems to be optional for x86_64, but required for x86. I'll remove x86 support then

kaey:
Another way we can go is to use prebuilt frontend assets published by grafana devs (they're just js/css files, arch independent) and avoid running webpack ourselves. This will reduce build times from 20mins to 1-2.

kaey:
I brought python dep back to support x86. python3 can't be used.
I also removed phantomjs blob since it depends on glibc and building it is a pain
arm builds timeout in 1hour unfortunately so they're disabled

testing/grafana/APKBUILD
maxice8: Please remove x86 support, we are also working on removing python2

kaey: done

kaey:
Updated to v6.3.0

PR #9859, opened 2019-08-03
community/influxdb: upgrade to 1.7.7

kaey:
switch from gdm to dep
remove pkgusers/groups - they're unneeded at build time
enable tests
disable overly verbose build logging
add conf.d
move config to /etc/influxdb.conf
simplify init.d script
disable x86 arch - many tests fail, it most likely doesn't work or there's something wrong with builder


maxice8:
@stwa

kaey:
Original maintainer didn't update this for almost a year (https://github.com/influxdata/influxdb/blob/v1.7.7/CHANGELOG.md#v163-2018-09-14)
They might not be coming back.

PR #9869, opened 2019-08-04
community/caddy: add checkconfig and reload commands to init.d

kaey:


maxice8:
Need to bump pkgrel for changes to take affect

maxice8:
@5paceToast

kaey:
>Need to bump pkgrel for changes to take affect

Done

community/caddy/caddy.initd
```diff
@@ -5,9 +5,37 @@ command="/usr/sbin/caddy"
...
+extra_started_commands="reload"
 
 depend() {
 	need net localmount
 	after firewall
 }
+
+checkconfig() {
+	ebegin "Checking configuration for $RC_SVCNAME"
+	start-stop-daemon --start --exec $command \
```
5paceToast: This is a bit unusual.
Further, I'm planning on rewriting this thing to use supervise-daemon later, and would appreciate having an easy transition (not a problem in `reload()` since supervise-daemon is also signal-aware).

Since we're hard-coding the user/group, it may be appropriate to do this using su(1).
Alternatively, since we specify an explicit `-conf` by default, it could be appropriate to just run the validation sequence as root (it's a read-only operation, after all).

kaey:
In case of su(1) i don't know how to correctly handle $command_args if it would contain (double)quotes, openrc does eval here for example, but I didn't investigate how it works exactly yet https://github.com/OpenRC/openrc/blob/master/sh/start-stop-daemon.sh#L48

Validating config as root would miss some checks, for example given this config I see correct error
```
cat /etc/caddy.conf.d/main.conf
:80 {
    root /root/www
    timeouts 24h
}

/etc/init.d/caddy checkconfig
 * Checking configuration for caddy ...
2019/08/05 01:57:42 /etc/caddy.conf.d/main.conf:2 - Error during parsing: Unable to access root path '/root/www': stat /root/www: permission denied
 * start-stop-daemon: failed to start `/usr/sbin/caddy'
```

As for switch to s-d, it has a few problems
1. it doesn't implement $error_logger, thus it's impossible to send logs to syslog, except via named pipe, which is complicated
2. $error_log is pretty much useless (save for named pipes) since there is no way to reopen log file after rotation without restarting child. ssd has the same problem, but it has $error_logger.
3. it doesn't like self-forking children, in case of caddy it forks itself when receiving USR2 allowing for binary upgrades without downtime. While i didn't implement it here, it's possible to do with ssd (eg make caddy write pidfile, instead of letting ssd do it), but not with s-d
4. s-d uses the same variables $command $command_args and $command_user, even if you decide to switch to it, checkconfig can be left as is. Furthermore it will most likely just work if you add `supervisor=supervise-daemon` to `/etc/conf.d/caddy` except for reload func, which will send signal to s-d instead of caddy (i can add check for this you want)

Given all this I'd ask you not to change to s-d by default and instead let users opt-in.

5paceToast:
> In case of su(1) i don't know how to correctly handle $command_args if it would contain (double)quotes, openrc does eval here for example, but I didn't investigate how it works exactly yet https://github.com/OpenRC/openrc/blob/master/sh/start-stop-daemon.sh#L48

`--` is a common (including in `su(1)`) convention.
`su user -- "$command" $command_args` should be sufficient.

> Validating config as root would miss some checks, for example given this config I see correct error
> 
> ```
> cat /etc/caddy.conf.d/main.conf
> :80 {
>     root /root/www
>     timeouts 24h
> }
> 
> /etc/init.d/caddy checkconfig
>  * Checking configuration for caddy ...
> 2019/08/05 01:57:42 /etc/caddy.conf.d/main.conf:2 - Error during parsing: Unable to access root path '/root/www': stat /root/www: permission denied
>  * start-stop-daemon: failed to start `/usr/sbin/caddy'
> ```

Fair enough.

> As for switch to s-d, it has a few problems
> 
>     1. it doesn't implement $error_logger, thus it's impossible to send logs to syslog, except via named pipe, which is complicated

Any serious system log (rsyslog, syslog-ng) has file-sources.
Further, the default (currently) is not to send things to syslog anyways, so this point is moot.

>     2. $error_log is pretty much useless (save for named pipes) since there is no way to reopen log file after rotation without restarting child. ssd has the same problem, but it has $error_logger.
> 
>     3. it doesn't like self-forking children, in case of caddy it forks itself when receiving USR2 allowing for binary upgrades without downtime. While i didn't implement it here, it's possible to do with ssd (eg make caddy write pidfile, instead of letting ssd do it), but not with s-d

s-d is a daemontools-like, sure, which does lose that.
I would argue that anyone with high uptime requirements should be using coordinated clusters.
In the meantime, s-d allows for direct child-parent supervision, which allows for healthchecks and automated service recovery.

>     4. s-d uses the same variables $command $command_args and $command_user, even if you decide to switch to it, checkconfig can be left as is. Furthermore it will most likely just work if you add `supervisor=supervise-daemon` to `/etc/conf.d/caddy` except for reload func, which will send signal to s-d instead of caddy (i can add check for this you want)

`reload()` would not work with s-d with that single change, and I'd consider it bad form to be using s-s-d to do things other than actually manage a daemon (including `-verify`).
Further, `$pidfile` is a noop under supervise-daemon, so no, it'll just trip on its own feet.
Finally, I recall there being some sort of post-processing on conf.d (not allowing the overriding of some variables), but I don't have the time to go fetch it right now.

> Given all this I'd ask you not to change to s-d by default and instead let users opt-in.

In short:
1. su(1) should interpret `--` correctly (same as s-s-d), but you may be interested in using `-l`.
2. The change to s-d will be happening for the sake of proper monitoring, simplified internal data (pidfiles are quite bad, just ask the systemd devs), and automated health recovery.
3. I do not consider the logging redirection a serious issue, given that anyone that seriously cares about logging will be using a competent system logger that has file-input capabilities. For basic access to logs, simple output to files is similarly sufficient.
4. `reload()` will not function as is, and it is bad form to utilize s-s-d in an s-d service (and vice versa).

kaey:
>-- is a common (including in su(1)) convention.
>su user -- "$command" $command_args should be sufficient.

it seems i've been worried for nothing. While your suggestion doesn't work, the simplest `su -s /bin/sh ${command_user%:*} -c "$command $command_args -validate"` does. Commit updated.

>Any serious system log (rsyslog, syslog-ng) has file-sources.

I don't care about file sources. I care about proper log file rotation my daemon writes to and s-d cannot do it. Only option is for a daemon to implement output to syslog or file with rotation (or at least support reopening log file when it's rotated by other tool), which caddy thankfully does. However I'd prefer if all daemons would just send all logs to stderr and allow launcher to choose where those logs go.

>Further, the default (currently) is not to send things to syslog anyways

I override this default with conf.d file.

> reload() would not work with s-d with that single change, and I'd consider it bad form to be using s-s-d to do things other than actually manage a daemon (including -verify).

i was thinking of doing something like that

```
reload() {
        checkconfig || return $?
        ebegin "Reloading $RC_SVCNAME"
        case $supervisor
        start-stop-daemon) start-stop-daemon --signal USR1 --pidfile "$pidfile" ;;
        supervise-daemon) supervise-daemon $RC_SVCNAME --signal USR1 ;;
        *) eend 1 "Unsupported supervisor: $supervisor" ;;
        esac
        eend $?
}
```

>Further, $pidfile is a noop under supervise-daemon.

No.
https://github.com/OpenRC/openrc/blob/master/sh/supervise-daemon.sh#L33

>Finally, I recall there being some sort of post-processing on conf.d (not allowing the overriding of some variables), but I don't have the time to go fetch it right now.

I don't see any.
https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in#L228

In fact I tried creating test service and specify `supervisor=supervise-daemon` in conf.d and it worked like a charm.

5paceToast:
> > Any serious system log (rsyslog, syslog-ng) has file-sources.
> 
> I don't care about file sources. I care about proper log file rotation my daemon writes to and s-d cannot do it. Only option is for a daemon to implement output to syslog or file with rotation (or at least support reopening log file when it's rotated by other tool), which caddy thankfully does. However I'd prefer if all daemons would just send all logs to stderr and allow launcher to choose where those logs go.

You appear to misunderstand my point.
Proposed workflow is as follows:
1. Caddy writes to stdout/err
2. s-d receives it and sends it to /var/log/caddy.whatever
3. rsyslog/syslog-ng/whatever uses /var/log/caddy.whatever as a file-input and places that in the desired location, with all the usual syslog management things.

In other words, you get proper logging with rotation.
If you do not have a real syslog daemon running, chances are you do not actually care all that much.

Anyway, the reason it isn't implemented is because back when it was suggested, s-s-d didn't have any ability like that, and thus s-d was actively stopped from having it to keep parity.
Please do go to the linked bug and get that changed, though!
I'm fully in favor of s-d being able to send stdout/err to syslog.
https://github.com/OpenRC/openrc/issues/118

> > Further, the default (currently) is not to send things to syslog anyways
> 
> I override this default with conf.d file.

That does not change what the default is/does.

> > reload() would not work with s-d with that single change, and I'd consider it bad form to be using s-s-d to do things other than actually manage a daemon (including -verify).
> 
> i was thinking of doing something like that
> 
> ```
> reload() {
>         checkconfig || return $?
>         ebegin "Reloading $RC_SVCNAME"
>         case $supervisor
>         start-stop-daemon) start-stop-daemon --signal USR1 --pidfile "$pidfile" ;;
>         supervise-daemon) supervise-daemon $RC_SVCNAME --signal USR1 ;;
>         *) eend 1 "Unsupported supervisor: $supervisor" ;;
>         esac
>         eend $?
> }
> ```

Works for me.
Please do test that it works first though (I'm not in a position to do that yet (or anytime soon) due to my recent move).

> > Further, $pidfile is a noop under supervise-daemon.
> 
> No.
> https://github.com/OpenRC/openrc/blob/master/sh/supervise-daemon.sh#L33

I meant supervise-daemon.c, the C program (not the shell script).
However, it looks like that decision was reversed about a month after it was made, in order to work on it for the next release and I missed it.
However, we still shouldn't be setting it.
See:
https://github.com/OpenRC/openrc/commit/af70862a7a115d6269affca663423e9340d6e929
https://github.com/OpenRC/openrc/commit/0f704402a236d385e8b30083ccf9aca327c0a57e
https://github.com/OpenRC/openrc/blob/master/man/supervise-daemon.8 (lack of documented option)

> > Finally, I recall there being some sort of post-processing on conf.d (not allowing the overriding of some variables), but I don't have the time to go fetch it right now.
> 
> I don't see any.
> https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in#L228
> 
> In fact I tried creating test service and specify `supervisor=supervise-daemon` in conf.d and it worked like a charm.

It only worked because you did not have `supervisor=` in the actual init.d file.
https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in#L246 - notice that loading the init.d file happens *after* loading the conf.d file (which, as you point out, happens 18 lines earlier).
This means that any declaration within the init.d file will override the conf.d file.

kaey:
>You appear to misunderstand my point.

Yes, I thought you meant write to file then read from it, when you meant fifo. I completely forgot about that and always used `error_logger="logger -t $RC_SVCNAME"`.

>Please do go to the linked bug and get that changed, though!

I REALLY don't want to write any C code

>Please do test that it works first though (I'm not in a position to do that yet (or anytime soon) due to my recent move).

Done. I also added if for pidfile, so switch to s-d via conf.d is now possible.
```
# Pidfile only required for start-stop-daemon.
if [ -z "$supervisor" ]; then
	pidfile="/var/run/caddy.pid"
fi
```

>It only worked because you did not have supervisor= in the actual init.d file.

This is unfortunate. I thought `supervisor=${supervisor:-supervise-daemon}` would work, however `supervisor=start-stop-daemon` is not valid, to use ssd `$supervisor` must be null. That means if you change default to s-d there is no way to opt-out of it via config, opt-in does work though. My only concern was logging, since there is a way to properly do it, I don't particularly care about supervisor choice any more.

One more thing about healthchecks and restarting (in case you intend to implement those). Restart on crash is fine (but still better leave configurable), but restart based on arbitrary healthcheck defined by you is not fine. Partial outages are very common and HC are strongly tied to specific system user is running, so defaults should be conservative and easily overridable.

5paceToast:
> > You appear to misunderstand my point.
> 
> Yes, I thought you meant write to file then read from it, when you meant fifo. I completely forgot about that and always used `error_logger="logger -t $RC_SVCNAME"`.

No worries, misunderstandings happen :)

> > Please do go to the linked bug and get that changed, though!
> 
> I REALLY don't want to write any C code

And that's why it still hasn't been done (I've been spoiled by D and modern C++).

> > Please do test that it works first though (I'm not in a position to do that yet (or anytime soon) due to my recent move).
> 
> Done. I also added if for pidfile, so switch to s-d via conf.d is now possible.
> 
> ```
> # Pidfile only required for start-stop-daemon.
> if [ -z "$supervisor" ]; then
> 	pidfile="/var/run/caddy.pid"
> fi
> ```

Feel free to keep the plain statement - I prefer simplicity.
Removing one line when I do get around to rewriting the thing isn't a big deal.

> > It only worked because you did not have supervisor= in the actual init.d file.
> 
> This is unfortunate. I thought `supervisor=${supervisor:-supervise-daemon}` would work, however `supervisor=start-stop-daemon` is not valid, to use ssd `$supervisor` must be null. That means if you change default to s-d there is no way to opt-out of it via config, opt-in does work though. My only concern was logging, since there is a way to properly do it, I don't particularly care about supervisor choice any more.

The point of supervisor= toggling is that users shouldn't really care about it, and I intend to have s-d be implemented as well as possible for all cases.

> One more thing about healthchecks and restarting (in case you intend to implement those). Restart on crash is fine (but still better leave configurable), but restart based on arbitrary healthcheck defined by you is not fine. Partial outages are very common and HC are strongly tied to specific system user is running, so defaults should be conservative and easily overridable.

My intent currently:
1. Set restart-on-crash to a value such that it does not trigger rate-limits in case something is wrong with DNS.
2. Have a "health_address" variable (empty by default) that can be set in conf.d.
If "health_address" is defined, healthchecks are performed at regular intervals (currently thinking every 30 seconds) using curl.
There would be an example snippet on how to make such an address a reality (it's quite simple, but my current situation makes it difficult to fetch an example and I've little inclination on composing one from scratch quite right now).

Incoming is a (hopefully final) review.

community/caddy/caddy.initd
```diff
@@ -3,11 +3,46 @@
...
 command="/usr/sbin/caddy"
 command_args="$CADDY_OPTS"
-pidfile="/var/run/caddy.pid"
 command_background=yes
-start_stop_daemon_args="--user caddy --group caddy"
+command_user=caddy:caddy
+extra_commands="checkconfig"
+extra_started_commands="reload"
+
+# Pidfile only required for start-stop-daemon.
```
5paceToast: Please keep the singular statement - I will remove it once the move to s-d happens.

community/caddy/APKBUILD
```diff
@@ -11,7 +11,7 @@ license="Apache-2.0"
 options="net !check"
 # tests fail on x86 builders due to containerization edge-cases
 case $CARCH in
-	x86) options="$options !check";;
+x86) options="$options !check" ;;
```
5paceToast: I would ask that you keep the original formatting.

community/caddy/caddy.initd
```diff
@@ -3,11 +3,46 @@
...
+	ebegin "Checking configuration for $RC_SVCNAME"
+	su -s /bin/sh ${command_user%:*} -c "$command $command_args -validate"
+	eend $?
+}
+
+reload() {
+	checkconfig || return $?
+	ebegin "Reloading $RC_SVCNAME"
+	case "$supervisor" in
+	"") start-stop-daemon --signal USR1 --pidfile "$pidfile" ;;
```
5paceToast: Please indent `PATTERN)` sections, and do not leave a space before the `;;`s.
Ditto for the next two lines.

kaey:
All done.
Formatting in this way was done by shfmt (https://github.com/mvdan/sh/), as one Go proverb says:

>Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.

5paceToast:
Thanks for the mention @kaey, I'll look into it, maybe suggest some improvements :)

5paceToast:
ping @maxice8

maxice8:
good to merge ?

5paceToast:
@maxice8 Yeah, good to merge.

PR #9875, opened 2019-08-05
[3.10] community/vault: security upgrade to 1.1.1

mtneug:
This fixes CVE-2019-11075.

The current 1.1.x version is 1.1.5. I wasn't sure if I should use that instead.


mtneug:
Should I also create PRs for older Alpine versions?

ncopa:
> The current 1.1.x version is 1.1.5. I wasn't sure if I should use that instead.

I we could use 1.1.5, as long as it does not break anything or require users to modify their config.

> Should I also create PRs for older Alpine versions?

I don't think thats necessary. We only support latest stable release for the community repo.

PR #9883, opened 2019-08-05
main/nodejs: upgrade to 10.16.1

timbru31:
Changelog: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#10.16.1


andypost:
Follow-up https://github.com/alpinelinux/aports/pull/9943

PR #9892, opened 2019-08-05
community/networkmanager: use supervise-daemon for OpenRC initd

maxice8:


community/networkmanager/APKBUILD
```diff
@@ -101,15 +101,15 @@ package() {
 To modify system network connections without the root password: add your user account to the 'plugdev' group, or use Polkit.
 EOF
 
-	cat > "$pkgdir"/etc/NetworkManager/conf.d/iwd.conf <<EOF
-[device]
-wifi.backend=iwd
+	cat > "$pkgdir"/etc/NetworkManager/conf.d/iwd.conf <<-EOF
+	[device]
+	wifi.backend=iwd
 EOF
```
Cogitri: I guess this should be removed (see #9894)

maxice8: later

PR #9894, opened 2019-08-05
Fix wpa_supplicant and re-enable it in NetworkManager again

MartijnBraam:
This makes wpa_supplicant run in dbus mode if no config file has been created for it, fixing the networkmanager integration. This config has been used for a long time in postmarketOS and works fine.

It also removes the iwd dependency from networkmanager again so wpa_supplicant is used because iwd doesn't support as many chipsets as wpa_supplicant and it doesn't run on older kernels.

Ping @Cogitri

### DIFF
--- a/community/networkmanager/APKBUILD
+++ b/community/networkmanager/APKBUILD
@@ -2,12 +2,12 @@
 # Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=networkmanager
 pkgver=1.18.2
-pkgrel=2
+pkgrel=3
 pkgdesc="Network Management daemon"
 url="https://wiki.gnome.org/Projects/NetworkManager"
 arch="all !s390x" # Limited by iwd
 license="GPL-2.0-or-later"
-depends="dhcpcd iptables dbus iwd"
+depends="dhcpcd iptables dbus"
 install="$pkgname.pre-install $pkgname.pre-upgrade"
 makedepends="$depends_dev
 	curl-dev
@@ -99,11 +99,11 @@ package() {
 	mv "$pkgdir/usr/share/doc/NetworkManager" "$pkgdir/usr/share/doc/$pkgname"
 	cat > "$pkgdir"/usr/share/doc/$pkgname/README.alpine <<EOF
 To modify system network connections without the root password: add your user account to the 'plugdev' group, or use Polkit.
-EOF
 
-	cat > "$pkgdir"/etc/NetworkManager/conf.d/iwd.conf <<-EOF
-	[device]
-	wifi.backend=iwd
+To use iwd instead of the default wpa_supplicant install iwd, start its service and add the following to your /etc/NetworkManager/NetworkManager.conf:
+
+[device]
+wifi.backend=iwd
 EOF
 }
 
--- a/community/networkmanager/networkmanager.pre-install
+++ b/community/networkmanager/networkmanager.pre-install
@@ -2,6 +2,8 @@
 
 addgroup -S plugdev 2>/dev/null
 
-printf "  *\n  * To setup system connections, regular users must either use Polkit for authentication or be a member of the 'plugdev' group."
+printf "  *\n  * To setup system connections, regular users must be member of 'plugdev' group.\n  *\n"
+printf "  *\n  * To control WiFi devices, enable wpa_supplicant service: 'rc-update add wpa_supplicant default'\n"
+printf "  * then reboot the system or restart 'wpa_supplicant' and 'networkmanager' services respectively.\n  *\n"
 
 exit 0
--- a/main/wpa_supplicant/wpa_supplicant.initd
+++ b/main/wpa_supplicant/wpa_supplicant.initd
@@ -5,7 +5,14 @@
 command=/sbin/wpa_supplicant
 : ${wpa_supplicant_conf:=/etc/wpa_supplicant/wpa_supplicant.conf}
 wpa_supplicant_if=${wpa_supplicant_if:+-i}$wpa_supplicant_if
-command_args="$wpa_supplicant_args -B -c$wpa_supplicant_conf $wpa_supplicant_if"
+
+# Run in dbus mode if no config file is found
+if [ -f "${wpa_supplicant_conf}" ]; then
+	command_args="$wpa_supplicant_args -B -c$wpa_supplicant_conf $wpa_supplicant_if"
+else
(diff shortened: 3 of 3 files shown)


community/networkmanager/networkmanager.pre-install
```diff
@@ -2,6 +2,8 @@
 
 addgroup -S plugdev 2>/dev/null
 
-printf "  *\n  * To setup system connections, regular users must either use Polkit for authentication or be a member of the 'plugdev' group."
+printf "  *\n  * To setup system connections, regular users must be member of 'plugdev' group.\n  *\n"
+printf "  *\n  * To control WiFi devices, enable wpa_supplicant service: 'rc-update add wpa_supplicant default'\n"
+printf "  * then reboot the system or restart 'wpa_supplicant' and 'networkmanager' services respectively.\n  *\n"
```
Cogitri: Hm, can't wpa_supplicant be dbus-activated from NM too? Then users wouldn't have to do this and we wouldn't have to mess with the initd service

community/networkmanager/APKBUILD
```diff
@@ -97,11 +97,11 @@ package() {
...
 	# post-install message
 	mv "$pkgdir/usr/share/doc/NetworkManager" "$pkgdir/usr/share/doc/$pkgname"
-	cat > "$pkgdir"/usr/share/doc/$pkgname/README.alpine <<EOF
+	cat > $pkgdir/usr/share/doc/$pkgname/README.alpine <<EOF
 To modify system network connections without the root password: add your user account to the 'plugdev' group, or use Polkit.
-EOF
 
-	cat > "$pkgdir"/etc/NetworkManager/conf.d/iwd.conf <<EOF
+To use iwd instead of the default wpa_supplicant install iwd, start its service and the following to your /etc/NetworkManager/NetworkManager.conf:
+
```
Cogitri: /s/and the following/and add the following/ :)

community/networkmanager/APKBUILD
```diff
@@ -97,11 +97,11 @@ package() {
 	mkdir -p "$pkgdir/usr/share/apk-tools/$pkgname"
 	# post-install message
 	mv "$pkgdir/usr/share/doc/NetworkManager" "$pkgdir/usr/share/doc/$pkgname"
-	cat > "$pkgdir"/usr/share/doc/$pkgname/README.alpine <<EOF
+	cat > $pkgdir/usr/share/doc/$pkgname/README.alpine <<EOF
```
Cogitri: Why unquote this? This could potentially have a space in it

community/networkmanager/networkmanager.pre-install
MartijnBraam: I have no idea if openrc can dbus-activate services, I know that's how it's done on systemd systems.

Cogitri: Ah no, I was taking about NM starting wpa_supplicant directly. But I guess activating the service isn't too bad.

maxice8:
Sorry for the obstruction, but i merged some changes to networkmanager so it works without backgrounding

MartijnBraam:
I rebased it

liqa:
There's a reason wpa_supplicant is able to run without dbus. It's a core package that basically everyone uses. Making it depend by default on dbus would mean users now can't have a dbus-free system.

Please consider a package split.
wpa_supplicant and wpa_supplicant-dbus

I see there's an old issue about the matter.

https://gitlab.alpinelinux.org/alpine/aports/issues/7415

maxice8:
> There's a reason wpa_supplicant is able to run without dbus. 

From what i understand it still does run without it.

> It's a core package that basically everyone uses. Making it depend by default on dbus would mean users now can't have a dbus-free system.

It already depends on libdbus.so.1 for a long time and the usage is only in a specific use-case, one where you don't configure wpa_supplicant because networkmanager (which depends on dbus) will do so.

Cogitri:
@liqa it'll only run in dbus mode if you don't have a config for wpa_supplicant supplied (which you _need_ in non-dbus mode anyway).

Danct12:
Is there anyone taking the job?

main/wpa_supplicant/wpa_supplicant.initd
```diff
@@ -5,7 +5,14 @@
...
 wpa_supplicant_if=${wpa_supplicant_if:+-i}$wpa_supplicant_if
-command_args="$wpa_supplicant_args -B -c$wpa_supplicant_conf $wpa_supplicant_if"
+
+# Run in dbus mode if no config file is found
+if [ -f "${wpa_supplicant_conf}" ]; then
+	command_args="$wpa_supplicant_args -B -c$wpa_supplicant_conf $wpa_supplicant_if"
+else
+	command_args="$wpa_supplicant_args -B -u"
+fi
+
```
ncopa: Is it possible to run wpa_supplicant in dbus mode with a config file? If so, I would recommend to have a conf.d option for dbus mode and make dbus mode default if config is missing.

Something like this:
```sh
if [ -f "${wpa_supplicant_conf}" ]; then
	command_args="${command_args} -c$wpa_supplicant_conf"
	: ${dbus_mode:=no}
else
	: ${dbus_mode:=yes}
fi

start_pre() {
	if yesno $dbus_mode; then
		command_args="${command_args} -u"
	fi
}
```

Then you can explicitly enable/disable dbus mode regardless of the existance of the wpa_supplicant config.

Otherwise you risk that you start with a no-config machine, and when you later need to set some unrelated config option for wpa_supplicant (lets say you want to disable use of some channels), you as a sideffect lose your dbus mode

MartijnBraam: wpa_supplicant ignores the config files you supply to it if you enable dbus mode so that wouldn't be useful. `-u` overrides `-c` and `-i`

ncopa: seems like that is not the case. I think `-u` will only enable the dbus interface. Nothing more.

Danct12: Any updates on this?

ollieparanoid:
For reference, one of the two commits were merged: be27851dffcfb0d3ddf656ff1798f7838bdcbb62

ncopa:
this was fixed in different way in commit 978f22f7b0a0c5fc92eb9f49ecc7c08dd962bf3b. Was trickier than expected to get it right.

PR #9904, opened 2019-08-06
main/rsync: use supervise-daemon for running rsyncd, fix license

maxice8:


main/rsync/rsyncd.confd
```diff
@@ -5,4 +5,4 @@
 
 # See man pages for rsync or run `rsync --daemon --help`
 # for valid cmdline options.
-#command_args=""
+#RSYNC_OPTS
```
tcely:
```suggestion
#RSYNC_OPTS=""
```

maxice8: Good catch, thank you

tcely:
I'm not seeing the supervisor change and checksums need updating.

PR #9906, opened 2019-08-06
community/bcc: move from testing

acj:
This also moves libbpf because it's a dependency.


kaey:
Would be great to split package into subpackages before moving to community.
`bcc-tools`, `bcc-doc`, `bcc-tools-doc`, `py-bcc`, maybe `bcc-dev` but I'm not sure if tools require c headers or not. This way other tools which only depend on libbcc (bpftrace for example) will not install stuff user is not interested in.

acj:
Thanks for the feedback. `bcc-*doc` should be easy to do. The others may be difficult to separate, but I'll give it a try.

acj:
Subpackage changes are in #9937.

acj:
Closing this until the subpackage PR lands.

PR #9909, opened 2019-08-06
main/bluez: package btmgmt tool

marmota:
This is not installed by default because upstream does not consider it a "stable" tool (i.e. its usage can change anytime).

However, it is a very useful tool for debugging Bluetooth problems, or to change the Bluetooth MAC address, since no other tool exists to talk to the Bluetooth Management API.


ncopa:
related issue: https://gitlab.alpinelinux.org/alpine/aports/issues/10036

PR #9923, opened 2019-08-06
community/utmps: move from testing

maxice8:


maxice8:
@skarnet

skarnet:
Thanks. I don't use GitHub in my workflow, and I'm not part of the alpinelinux organisation so I don't have the power to merge here. If it's okay with you, though, I'll apply the changes in my local repository and send a patch to the ML.
Note that explicitly listing skalibs-dev in makedepends is intentional. I find it bad practice to use the value of depends-dev as makedepends; those variables are semantically different and it's not guaranteed that their values are the same, even if it's often the case. Listing the values explicitly forces a contributor to _think_ about what should really go in there.

PureTryOut:
Wasn't this supposed to go to mainline instead?

maxice8:
Feel free to move it to main, i want it at least in community so people can use sddm

skarnet:
I have no objection to your submitting patches yourself if you have push rights and they don't break stuff, but if you're going to do it, I would have appreciated at least a notice, so I understand what's going on when I get to it and don't find the code I expect.

I wonder what the Maintainer line is good for if everyone can push stuff to master no questions asked. It's not the first time it happens with Alpine and it's pretty concerning.

PR #9937, opened 2019-08-07
testing/bcc: split into subpackages

acj:
~Now that libbpf has releases, I'm including it as part of `source` to simplify the build and patches.~ See discussion below.

### DIFF
--- a/testing/bcc/APKBUILD
+++ b/testing/bcc/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Adam Jensen <acjensen@gmail.com>
 pkgname=bcc
 pkgver=0.10.0
-pkgrel=0
+pkgrel=1
 pkgdesc="A toolkit for creating efficient kernel tracing and manipulation programs"
 url="https://github.com/iovisor/bcc/"
 arch="aarch64 x86 x86_64"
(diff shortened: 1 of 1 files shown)


kaey:
pkg-config files should be in -dev.
bcc-tools should also depend on py-bcc.
Everything else looks good.

acj:
Fixed. Thanks.

acj:
@maxice8 @Ikke Sorry for the direct ping, but if you have time for a review, I would love to get this merged. It's blocking work on several downstream projects.

testing/bcc/APKBUILD
```diff
@@ -1,22 +1,27 @@
...
+pkgrel=1
 pkgdesc="A toolkit for creating efficient kernel tracing and manipulation programs"
 url="https://github.com/iovisor/bcc/"
 arch="aarch64 x86 x86_64"
 license="Apache-2.0"
 # bcc's test suite requires privileged access to run BPF programs
 options="!check"
 depends="python"
+subpackages="$pkgname-dev:_dev $pkgname-doc:_doc $pkgname-tools:_tools $pkgname-lua:_lua py-$pkgname:_py"
+_libbpfver=0.0.4
```
maxice8: should libbpf be a separate APKBUILD ?

acj: I had hoped to simplify the build, but I just noticed that there's been a recent effort to make an RPM package for libbpf. We will probably want one as well.

Force-pushed a change to continue using testing/libbpf-dev.

testing/bcc/APKBUILD
```diff
@@ -33,6 +34,28 @@ build() {
 package() {
 	cd "$builddir/build"
 	make install DESTDIR="$pkgdir"
+
+	mv "$pkgdir"/usr/share/bcc/man "$pkgdir"/usr/share/man
```
kaey: Tools which these man pages refer to are not in PATH, so it's better to leave them in `/usr/share/bcc/man`.
Furthermore in a shared space they would conflict with man pages provided by bpftrace

PR #9940, opened 2019-08-07
community/sddm: enable on all arches and use supervise-daemon

maxice8:
@PureTryOut


PureTryOut:
LGTM, although as far as re-enabling on all arches goes, I already have a patch for that on https://lists.alpinelinux.org...

PR #9948, opened 2019-08-08
testing/firefox: enable wayland capabilities

Cogitri:
This doesn't change anything for Xorg sessions, but for Wayland sessions it makes
use of the GDK Wayland backend (if set by the session, e.g. via GDK_BACKEND=wayland),
which enables things like fractional scaling and different scaling factors across
different monitors

### DIFF
--- a/testing/firefox/APKBUILD
+++ b/testing/firefox/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
 pkgver=68.0
-pkgrel=0
+pkgrel=1
 pkgdesc="Firefox web browser"
 url="https://www.firefox.com/"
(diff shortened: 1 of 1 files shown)


liqa:
@Cogitri could you make the same changes to community/firefox-esr?
It's a good practice to keep both firefox in sync as much as possible.
So it's easier for future upgrades.

liqa:
Right now here's the 3 differences between the packages
```
+license="GPL LGPL MPL"
-license="GPL-3.0-only AND LGPL-2.1-only AND LGPL-3.0-only AND MPL-2.0"

+       nss-dev>=3.45
-       nss-dev>=3.44.1

+               --enable-default-toolkit=cairo-gtk3 \
-               --enable-default-toolkit=cairo-gtk3-wayland \
```

License should be double-checked

PR #9959, opened 2019-08-08
community/pdns: upgrade to 4.1.13

tcely:


tcely:
It'd be nice to have a decision about #9866 first, but I don't want to wait on that.

maxice8:
It was decided to include it

PR #9963, opened 2019-08-08
main/bind: split tools

tcely:
- improve descriptions
- added -server-tools subpackage
- added -libs-static subpackage
- provide py3-isc

### DIFF
--- a/main/bind/APKBUILD
+++ b/main/bind/APKBUILD
@@ -3,13 +3,13 @@
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: tcely <bind+aports@tcely.33mail.com>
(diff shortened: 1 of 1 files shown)


main/bind/APKBUILD
```diff
@@ -176,30 +193,57 @@ dev() {
...
+	depends="$depends_plugins"
+
+	mkdir -p "$subpkgdir"/usr/lib
+	mv "$pkgdir"/usr/lib/bind "$subpkgdir"/usr/lib/
+}
+
+py3_isc() {
+	pkgdesc="Python functions to support BIND utilities"
+	depends="$depends_py3_isc"
+	provides="py3-isc=${_isc_ver}-r${_isc_rel} py3-$pkgname=$pkgver-r$pkgrel"
```
ncopa: Why is the versioned provides needed? I think it adds complexity for no good reason.

PR #9966, opened 2019-08-09
community/kodi: enable on all arm architectures

PureTryOut:
Don't mind me, just need ARM devices to test the build on (I don't have a usable ARM device atm and [Java is broken on VM's](https://gitlab.alpinelinux.org/alpine/aports/issues/10711)) so I'm using the Drone CI.


PureTryOut:
Well that I didn't expect, it went through first time. Awesome, ready to merge then.

PR #9974, opened 2019-08-09
[3.10] community/chromium: security upgrade to 76

maxice8:
ref #10596

@clandmeter @ncopa


rnalrd:
Did we make a policy exception for browsers in stable branches? We don't do such version jump in stable branches afaik.

maxice8:
No, but i don't see anyone patching the very big number of CVEs, they are so much we don't even track them.

rnalrd:
I believe that browsers maintenance was discussed in the past and iirc we basically support only firefox-esr. Firefox and Chromium basically die young in every Alpine stable release due to the fast upstream release cycle.

Il giorno 17 ago 2019, 02:26, alle ore 02:26, maxice8 <notifications@github.com> ha scritto:

maxice8:
On Tue Aug 20, 2019 at 11:17 AM Leonardo wrote:
> I believe that browsers maintenance was discussed in the past and iirc we basically support only firefox-esr. Firefox and Chromium basically die young in every Alpine stable release due to the fast upstream release cycle.
> 

Then might as well move to testing instead of exposing users to lots of CVEs

PR #9988, opened 2019-08-10
main/vala: upgrade to 0.44.7

Cogitri:


Cogitri:
Timeouts on armhf, but seems good to go to me

maxice8:
Ah, sorry

PR #10013, opened 2019-08-11
community/kde-frameworks: upgrade to 5.61.0

PureTryOut:
For some reason this patch breaks git-send-email for me, so I'm doing it the bad 'ol Github way.

CVE-2019-14744 has been fixed in kconfig 5.61.0 so the patch isn't needed anymore.


maxice8:
Please do more on the bad way, it has CI :D

PureTryOut:
Which doesn't work often anyway because it has time and output limits which it exceeds lol. Travis is even worse.

maxice8:
Can't the packages be separated in groups of PRs ?

Regardless i was never a fan of single-commit updating a bunch of packages.

PureTryOut:
They could. Is more maintenance and work though, I prefer not to really.

PR #5640, opened 2018-11-13
testing/creduce: upgrade to 2.10.0

J0WI:
Depends on #5639

PR #10038, opened 2019-08-12
unmaintained/clang6: move from testing

J0WI:
Was only used in `testing/creduce` (#5640)


maxice8:
ponyc still needs it

J0WI:
`ponyc` uses only llvm6, doesn't it?

maxice8:
Yes i read it badly.

maxice8:
Merged

PR #10039, opened 2019-08-12
main/dns-root-hints: upgrade to 2019073000

tcely:
See also alpinelinux/aports#5950


main/dns-root-hints/APKBUILD
```diff
@@ -1,39 +1,69 @@
...
+builddir="${srcdir}/internic-domain-data-${pkgver}"
 
-build() {
-	mkdir -p "$builddir"
+prepare() {
+	default_prepare
+
+	builddir="${builddir}/build"
+	mkdir -p "${builddir}"
 }
```
Cogitri: Although it's a nitpick don't we usually set builddir to $srcdir/<name>/build already and just cd to $srcdir/<name> in prepare?

main/dns-root-hints/APKBUILD
```diff
@@ -1,39 +1,69 @@
...
+	_copy_from_source_archive
+	options_has net && _fetch_from_url
+	cd -
+	_check_sig
+	rm -rf "${_tmp}"
+}
+
+build() {
+	cd "${builddir}"
+	test ! -d build || cd build
```
Cogitri: Why would build not exist at this point? It has been created in prepare

main/dns-root-hints/APKBUILD
tcely: Normally, `builddir` is set to the directory that is contained in the source archive. This is usually `$pkgname-$pkgver` that is created by the unpack function under `$srcdir` which is why that is the default.

This package was created with an empty build directory in mind. So when extracting the source archive, the easiest way back to that state is to create an empty directory.

main/dns-root-hints/APKBUILD
tcely: The value of `builddir` depends on which functions have already run. When `/build` was not appended the build subdirectory exists and we cd into it.

tcely: You did point me at a problem I didn't handle. The value of `builddir` was not accurate for other functions when the `cd` case hit. bc4a0cc corrects that problem.

ncopa:
I think this is wrong solution for the problem.

Relying on a second git repository (https://github.com/tcely/internic-domain-data/) is worse than copying the data to our aports repo. How do we update this when you are on vacation or need to be long time away for private reasons?

We should avoid it to bitrot but we should do so by monitoring it with external tools, similar to what we do with other packages. We should have a service that notifies the maintainer when its outdated. Til that is in place you (or other users) can flag the package as outdated on pkgs.alpinelinux.org.

Also, this data is is not really supposed to be updated via `apk upgrade`. You should keep your root hints update by requesting this data via DNS lookups or by fetching it directly from https://www.internic.net/domain/ via cron job or similar. This package is only to facilitate bootstrapping this process. This means that it is not the end of the world if this data gets outdated. I think our time is better spent elsewhere than on debating this.

So I am letting @rnalrd keep this package.

PR #10043, opened 2019-08-12
Add support for Beagle Bone Black [V2]

vicencb:
Hi,
this patch series adds support for the Beagle Bone Black.
The first patch just enables the IIO module.
The second patch just enables support for SOC_AM33XX.
This first two changes generate an avalanche of differences, so, they are on standalone patches.
The third patch adds drivers for the BBB.

V2 removes the touchscreen, backlight and sound drivers which are less probably used in an embedded system.
sha512sums are updated at each commit

Regards,
  Vicenç.


Ikke:
@vicencb Note that it's not necessary to open a new pull request to include new changes. You can either add new commits to the branch you used for the pull request, or amend existing commits and force push that.

That way, we keep all the comments and labels for the pull request.

vicencb:
Hi @Ikke, thanks for the note.
From now on, i know about `git commit --amend`.

rnalrd:
LGTM. Delegating to @ncopa the kernel maintainer.

PR #10055, opened 2019-08-12
main/socat: fix IP-SENDTO on musl

nmeum:
Without this patch IP-SENDTO doesn't work on Alpine (probably due to musl) because socat uses `SOCK_DGRAM (2)` as a socktype even for raw ip sockets. This patchs makes sure that it uses `SOCK_RAW (3)` for raw ip sockets.

To verify that `IP-SENDTO` doesn't work without this patch run the following command:

```
# socat - IP6-SENDTO:localhost:6
2019/08/12 22:48:34 socat[12846] E getaddrinfo("localhost", "NULL", {1,10,2,6}, {}): Unrecognized service
```

The element `2` in the output  `{1,10,2,6}` indicates the incorrect socktype (`SOCK_DGRAM`).

---

*I also submitted this patch upstream, haven't received a response yet though.*

### DIFF
--- a/main/socat/0001-xiogetaddrinfo-support-SOCK_RAW.patch
+++ b/main/socat/0001-xiogetaddrinfo-support-SOCK_RAW.patch
@@ -1,0 +1,43 @@
+From dd103797da3a4bf8c662119b3ff717e0ca072cd7 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?S=C3=B6ren=20Tempel?= <soeren+git@soeren-tempel.net>
+Date: Mon, 12 Aug 2019 22:46:36 +0200
+Subject: [PATCH] xiogetaddrinfo: support SOCK_RAW
+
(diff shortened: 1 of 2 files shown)


ncopa:
Is this patch sent upstream? What does upstream think about it?

nmeum:
I did sent an email to upstream containing the patch but never received a response.

nmeum:
Received a response via email from upstream:

> My fix will probably remove the fallback logic that you fixed;

We could use my fix in the meantime. I certainly find it annoying that raw IP sockets don't work.

ncopa:
can you please verify if musl-1.1.24 fixes it?

nmeum:
I can still reproduce the issue with musl-1.1.24. BTW: The issue [seems to be reproducible on glibc systems](https://github.com/RIOT-OS/RIOT/pull/12001#issuecomment-522534682) as well, it's not musl specific.

PR #10063, opened 2019-08-13
main/dns-root-hints: do not mention tcely

tcely:
This has been a hijacked package from the first commit.
I want nothing to do with this monstrous reinterpretation.
I am against everything about how this works.

We need a source of truth for our DNS software.
This is just more rot in the git repo like we had already.
The only goal that was met was to put the rot into fewer places.

Go read alpinelinux/aports#5950 and alpinelinux/aports#10039 instead.


rnalrd:
Your commits were not merged as-is because they did not meet the requirements despite the explanations.
Closing PR.

tcely:
Rewritten as https://github.com/alpinelinux/aports/commit/0b7e719034b2e852916cc28f145f2b144de37c0e

PR #10068, opened 2019-08-13
main/xen: 4.12.1

HRio:
Fixes XSA-295, CVE not yet assigned


HRio:
@rnalrd @ncopa tested on a small Intel box. Should be cherry-picked to v3.10 as well.

PR #10069, opened 2019-08-13
testing/fd: more arches

J0WI:


maxice8:
Is it really required ? i'd rather wait until rust is on all arches as work is being done by mps and Cogitri

J0WI:
It's not required, just nice to have and a way to test the rust ports in real world examples.

testing/fd/APKBUILD
```diff
@@ -2,10 +2,10 @@
 # Maintainer: Leo <thinkabit.ukim@gmail.com>
 pkgname=fd
 pkgver=7.3.0
-pkgrel=0
+pkgrel=1
```
maxice8: no need to bump pkgrel if you're enabling new archs

maxice8:
I'll wait until full rust is available

J0WI:
https://github.com/alpinelinux/aports/commit/cc4e034e2668c4b6a6150d39f3d44200b1562675

PR #10075, opened 2019-08-13
testing/zimg: upgrade to 2.9.2

prspkt:


prspkt:
For future reference, tests fail across all architectures on CI but succeed on our build servers.

PR #10082, opened 2019-08-13
main/apache2: upgrade to 2.4.41

J0WI:


J0WI:
The announcement is here:
https://lists.apache.org/thread.html/c950015da2f121ccdf9a1256efa9398042e17c6362f58b694afe9735@%3Cdev.httpd.apache.org%3E

PR #10089, opened 2019-08-14
community/suricata: re-enable all arches

J0WI:


rnalrd:
Merged after removing the no longer relevant comment: 
<pre>
# luajit is not available for disabled arches
</pre>

PR #10107, opened 2019-08-15
testing/ldc: new aport

Geod24:
```
This add a package to support the D Programming Language.
Currently, one has to jump through hoops in order to use D on Alpine.
This packages installs the official (and only) LDC build that came
with an Alpine build, LDC-1.8.0.
While it is an old version of the compiler, 2.078.3,
it can be used to bootstrap other compiler versions.
```

Note: D compilers have their frontend written in D since a few years.
This cause obvious problems of self-hosting.
We have 3 compilers:
- DMD, the reference compiler;
- GDC, based on GCC and included with it since GCC9
- LDC, based on LLVM

My plan was to originally use GCC9, but some backports are necessary to support musl, and response time has been slow.
The LDC guys however provided a build for musl last year, but it was never made a package.
Once this is in, I'd like to use it as a base to compile the new compilers (new DMD and new LDC, a new GCC will come as self hosted) and other tools (e.g. our package manager).
[This is the same approach that was used for Rust](https://github.com/alpinelinux/aports/commit/2a8472b921baf8c2ac35688df5188f034b526441#diff-0d0f9a278fdcdf00a9167fd4861cdfe7R16-R25), except it uses an upstream-provided binary instead of one compiled by a contributor.

Things I wasn't sure about:
- The "LICENSE" file was a bit complicated: I included "Some part" but didn't see an example for multi-licensed code.

### DIFF
--- a/testing/ldc/APKBUILD
+++ b/testing/ldc/APKBUILD
@@ -1,0 +1,112 @@
+# -*- shell-script -*-
+# Contributor: Mathias LANG <pro.mathias.lang@gmail.com>
+# Maintainer: Mathias LANG <pro.mathias.lang@gmail.com>
+#
+# This package is divided between the actual package,
+# and the two subpackages, 'static' and 'runtime'.
+# The main package depends on them.
+# A user might want to install 'runtime' when they have
+# a dynamically-linked D program.
+pkgname=ldc
+pkgver=1.8.0
+pkgrel=0
+pkgdesc="The LLVM-based D Compiler"
+url="https://github.com/ldc-developers/ldc"
+# Bootstrapping was done with an upstream-provided package,
+# with only x86_64 available
+# Once we can use GDC for bootstrapping, this can be "all"
+arch="x86_64"
+license="BSD-3-Clause AND BSL-1.0 AND (Artistic-1.0 OR GPL-2.0-or-later) AND NCSA AND NCSA AND MIT"
+depends="libexecinfo"
+makedepends="chrpath cmake llvm5-libs libexecinfo-static"
+subpackages="$pkgname-runtime $pkgname-doc $pkgname-static"
+provides="dlang-compiler"
+source="https://github.com/ldc-developers/ldc/releases/download/v$pkgver/ldc2-$pkgver-alpine-linux-x86_64.tar.xz
+		ldc2.package.conf"
+builddir="$srcdir/"
+
+build() {
+	rootdir="$srcdir/ldc2-$pkgver-alpine-linux-x86_64/"
+
+	# Unfortunately the upstream-provided version has the builder's /home/ in its rpath - strip it
+	chrpath -d "$rootdir/bin/ldc2"
+	chrpath -d "$rootdir/bin/ldmd2"
+	# This exe is not installed, just used in build, but do it for completeness
+	chrpath -d "$rootdir/bin/ldc-build-runtime"
+
+	# Now build the runtime - the compiler itself is already built by upstream
+	"$rootdir"/bin/ldc-build-runtime --buildDir="$builddir/ldc-runtime-build/"
+
+	# CMake added the rpaths to the shared libs - strip them
+	chrpath -d "$builddir"/ldc-runtime-build/lib/*.so
+	:
+}
+
+check() {
+	rootdir="$srcdir/ldc2-$pkgver-alpine-linux-x86_64/"
+	libroot="$builddir/ldc-runtime-build/lib/"
+
+	echo 'import std.stdio; void main () { try throw new Exception(null); catch (Exception) writeln("Hello World!"); }' > hello_world.d
+	if [ "$($rootdir/bin/ldmd2 -conf='' -I$rootdir/import/ldc -I$rootdir/import/ -L-L$libroot -defaultlib=phobos2-ldc,druntime-ldc -L-lexecinfo -L-rpath=$libroot -run hello_world.d)" != "Hello World!" ]; then
+		return 1
+	fi
+	:
+}
+
+package() {
+	depends="$pkgname-runtime $pkgname-static"
+	rootdir="$srcdir/ldc2-$pkgver-alpine-linux-x86_64/"
+
+	install -s -D "$rootdir/bin/ldc2"  "$pkgdir/usr/bin/ldc2"
+	install -s -D "$rootdir/bin/ldmd2" "$pkgdir/usr/bin/ldmd2"
+
+	install -D "$srcdir/ldc2.package.conf" "$pkgdir/etc/ldc2.conf"
+
+	install -D "$rootdir/LICENSE" "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+
+	# Install the import files
+	for impfile in $(find "$rootdir/import" -name "*.d*");
+	do
+		target="${impfile##*import/}"
+		install -D "$impfile" "$pkgdir/usr/include/dlang/$pkgname-$pkgver/$target"
+	done
+
+	# Install the static runtime libraries
+	for libn in "libdruntime-ldc" "libdruntime-ldc-debug" "libphobos2-ldc" "libphobos2-ldc-debug";
+	do
+		install -D "$builddir/ldc-runtime-build/lib/$libn.a" "$pkgdir/usr/lib/$libn.a"
+	done
+
+	# Install the shared runtime libraries
+	for libn in "libdruntime" "libphobos2";
+	do
+		install -D "$builddir/ldc-runtime-build/lib/$libn-ldc-shared.so.2.0.78" "$pkgdir/usr/lib/$libn-ldc-shared.so.2.0.78"
+	done
+}
+
+runtime() {
+	depends="libexecinfo-dev"
+	pkgdesc="Dynamic runtime library for D code compiled with $pkgname-$pkgver"
+
+	mkdir -p "$subpkgdir/usr/lib/"
+	for libn in "libdruntime" "libphobos2";
+	do
+		mv "$pkgdir/usr/lib/$libn-ldc-shared.so.2.0.78" "$subpkgdir/usr/lib/$libn-ldc-shared.so.2.0.78"
+		ln -s "$libn-ldc-shared.so.2.0.78" "$pkgdir/usr/lib/$libn-ldc-shared.so.78"
+		ln -s "$libn-ldc-shared.so.2.0.78" "$pkgdir/usr/lib/$libn-ldc-shared.so"
+	done
+
+	mv "$pkgdir"/usr/lib/*.so* "$subpkgdir/usr/lib/"
+	:
+}
+
+static() {
+	depends="libexecinfo-static"
+	pkgdesc="$pkgdesc (static library)"
+
+	mkdir -p "$subpkgdir/usr/lib/"
+	mv "$pkgdir"/usr/lib/*.a "$subpkgdir/usr/lib/"
+}
+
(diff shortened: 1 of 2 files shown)


Geod24:
Fixed the test failure. I am not quite sure what a proper `check` is, given the testsuite is not available to run for this package, so I went with compiling and running an "Hello World".

Geod24:
So, I gave this extensive testing via docker containers.

Since it's an old version of the compiler, the next thing I did was to build the newest version of the compiler (v1.16.0). When this PR is merged and the package is available, I will PR it.
The new package looks much more like a multi-purpose package than the current version. You can find it [here](https://github.com/bpfkorea/docker-ldc/tree/master/ldc) if you are interested.

The next thing any D developer will want to do is to use `dub`, the package manager.
I have a package for it ready as well, compiled with LDC-1.16.0.
Additionally, there is a package for `rdmd`, a very common build tool, and `ddemangle`, a CLI tool to demangle D symbols.
Those 3 packages are available [here](https://github.com/bpfkorea/docker-dub).

The reason I built all those images was to be able to build and deploy our project on Alpine-based Docker container. Our project is [here](https://github.com/bpfkorea/agora/blob/6c8de94b0bad7ae983ab8aeef0cddad50d37a414/Dockerfile) and although I still have to setup a proper system integration test, it works as expected and is a reasonable test that this work.

Geod24:
Ping @maxice8 , any chance you could take a look at this ?

ZombineDev:
@Cogitri can you please take a look?

Cogitri:
Thanks for the reminder, will do in a bit :)

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+	:
+}
+
+package() {
+	depends="${pkgname}-runtime ${pkgname}-static"
+	rootdir="${srcdir}/ldc2-1.8.0-alpine-linux-x86_64/"
+
+	install -s -D "${rootdir}/bin/ldc2"	 "${pkgdir}/usr/bin/ldc2"
+	install -s -D "${rootdir}/bin/ldmd2" "${pkgdir}/usr/bin/ldmd2"
+
```
JohanEngelen: also install `ldc-profdata` (needed for PGO) and `ldc-prune-cache` (needed for LDC compile cache management)

jacob-carlborg:
There’s an Alpine build available for 1.13.0 as well: https://github.com/ldc-developers/ldc/releases/download/v1.13.0/ldc2-1.13.0-alpine-linux-x86_64.tar.xz.

testing/ldc/APKBUILD
Geod24: Those binaries are not in the tar. If memory serves me well, PGO wasn't well developed at the time ? Or perhaps they were not included in the alpine package.

However I will have a PR for 1.17.0 up after this is merged, and will make sure to ping you. This is merely used for bootstrap.

Geod24:
> There’s an Alpine build available for 1.13.0 as well: https://github.com/ldc-developers/ldc/releases/download/v1.13.0/ldc2-1.13.0-alpine-linux-x86_64.tar.xz.

Thanks, didn't knew about it. However I have a script to build v1.16.0 (will update to v1.17.0 when this is merged), and this package is merely a bootstrap phase.

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+check() {
+	rootdir="${srcdir}/ldc2-1.8.0-alpine-linux-x86_64/"
+	libroot="${builddir}/ldc-runtime-build/lib/"
+
+	echo 'import std.stdio; void main () { try throw new Exception(null); catch (Exception) writeln("Hello World!"); }' > hello_world.d
+	if [ "$(${rootdir}/bin/ldmd2 -conf='' -I${rootdir}/import/ldc -I${rootdir}/import/ -L-L${libroot} -defaultlib=phobos2-ldc,druntime-ldc -L-lexecinfo -L-rpath=${libroot} -run hello_world.d)" != "Hello World!" ]; then
+		return 1
+	fi
+	:
```
Cogitri: I'm pretty sure we have a policy against making trivial test cases like these since they can make a package to appear to be working when it actually doesn't

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+}
+
+package() {
+	depends="${pkgname}-runtime ${pkgname}-static"
+	rootdir="${srcdir}/ldc2-1.8.0-alpine-linux-x86_64/"
+
+	install -s -D "${rootdir}/bin/ldc2"	 "${pkgdir}/usr/bin/ldc2"
+	install -s -D "${rootdir}/bin/ldmd2" "${pkgdir}/usr/bin/ldmd2"
```
Cogitri: The indention seems odd here

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+	echo 'import std.stdio; void main () { try throw new Exception(null); catch (Exception) writeln("Hello World!"); }' > hello_world.d
+	if [ "$(${rootdir}/bin/ldmd2 -conf='' -I${rootdir}/import/ldc -I${rootdir}/import/ -L-L${libroot} -defaultlib=phobos2-ldc,druntime-ldc -L-lexecinfo -L-rpath=${libroot} -run hello_world.d)" != "Hello World!" ]; then
+		return 1
+	fi
+	:
+}
+
+package() {
+	depends="${pkgname}-runtime ${pkgname}-static"
+	rootdir="${srcdir}/ldc2-1.8.0-alpine-linux-x86_64/"
```
Cogitri: Please don't hardcode the version number here (maybe make a `_bootver` variable)

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+	do
+		install -D "${builddir}/ldc-runtime-build/lib/${libn}.a" "${pkgdir}/usr/lib/${libn}.a"
+	done
+
+	# Install the shared runtime libraries
+	for libn in "libdruntime" "libphobos2";
+	do
+		install -D "${builddir}/ldc-runtime-build/lib/${libn}-ldc-shared.so.2.0.78" "${pkgdir}/usr/lib/${libn}-ldc-shared.so.2.0.78"
+	done
+}
```
Cogitri: No need for the curly braces about all of these. Please install `atools` and run apkbuild-lint which will automatically find such errors for you.

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+	depends="libexecinfo-dev"
+	pkgdesc="Dynamic runtime library for D code compiled with ${pkgname}-${pkgver}"
+
+	mkdir -p "${subpkgdir}/usr/lib/"
+	for libn in "libdruntime" "libphobos2";
+	do
+		mv "${pkgdir}/usr/lib/${libn}-ldc-shared.so.2.0.78" "${subpkgdir}/usr/lib/${libn}-ldc-shared.so.2.0.78"
+		ln -s "${libn}-ldc-shared.so.2.0.78" "$pkgdir/usr/lib/${libn}-ldc-shared.so.78"
+		ln -s "${libn}-ldc-shared.so.2.0.78" "$pkgdir/usr/lib/${libn}-ldc-shared.so"
+	done
```
Cogitri: We won't have to do this once we have the bootstrapping done, right?

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+# a dynamically-linked D program.
+pkgname="ldc"
+pkgver="1.8.0"
+pkgrel=0
+pkgdesc="The LLVM-based D Compiler"
+url="https://github.com/ldc-developers/ldc"
+arch="x86_64"
+license="BSD-3-Clause; Some parts BSL-1.0; Some parts Artistic-1.0 or GPL-2.0-or-later; Some parts NCSA; Some parts NCSA and MIT"
+depends="libexecinfo"
+makedepends="chrpath cmake llvm5-libs libexecinfo-static"
```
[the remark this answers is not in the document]
Geod24: This was [released March 2018](https://github.com/ldc-developers/ldc/releases/tag/v1.8.0) and only supports v3.7-5.0 according to the release.
But once bootstrapping in done, we can switch to the latest LLVM.

testing/ldc/APKBUILD
```diff
@@ -0,0 +1,110 @@
...
+# The main package depends on them.
+# A user might want to install `runtime` when they have
+# a dynamically-linked D program.
+pkgname="ldc"
+pkgver="1.8.0"
+pkgrel=0
+pkgdesc="The LLVM-based D Compiler"
+url="https://github.com/ldc-developers/ldc"
+arch="x86_64"
+license="BSD-3-Clause; Some parts BSL-1.0; Some parts Artistic-1.0 or GPL-2.0-or-later; Some parts NCSA; Some parts NCSA and MIT"
```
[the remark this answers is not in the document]
Geod24: Alright, I changed everything to `AND`. The part I'm most concerned about is "Some parts Artistic-1.0 or GPL-2.0-or-later` as I'm not sure if the new version implies that GPL-2.0-or-later is required now.

testing/ldc/APKBUILD
Geod24: Yep, this is what I have after bootstrap: https://github.com/bpfkorea/docker-ldc/blob/530be2999018866e0330ef4c11efd4e90db04d9f/ldc/APKBUILD#L96-L109

Geod24: Or, by "this", do you mean the runtime package ? I assumed your comment apply to creating the symlink.

testing/ldc/APKBUILD
Geod24: Removed curly braces everywhere.
`apkbuild-lint` didn't warn me about it, but did point two other issues (pkgver & pkgname shouldn't be quoted). I also had to change some comments, as it complained about my usage of backtips in comments.

Side note: Shouldn't that be part of the CI ?

testing/ldc/APKBUILD
Geod24: What would be a proper test case ?
This did help me a lot during development, as I could ensure programs are linked correctly (D uses `backtrace` which is in Glibc but not in Musl).
When we're building from source we could run the compiler test suite, although it is quite long. I couldn't find clear guidelines on the wiki on what is a good `check` function, w.r.t time, complexity, space, is there such a description somewhere ?

Geod24:
Updated. Marked resolved comments as such and left open the ones that need additional discussion.

testing/ldc/APKBUILD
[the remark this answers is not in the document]
Cogitri: Ah, good.

testing/ldc/APKBUILD
Cogitri: A good check function is what upstream provides, so once we have the proper compiler we should use the testsuite you've mentioned. The length doesn't matter, we have plenty of build time to ensure our packages work :)

testing/ldc/APKBUILD
Cogitri: Yes, referred to the symlinks

Cogitri:
Thanks. This LGTM now other than the license thingie and the commit message. Please us the standard `<repo>/<package>: new aport`

Geod24:
> Thanks. This LGTM now other than the license thingie and the commit message. Please us the standard `<repo>/<package>: new aport`

Thanks. Done.

Cogitri:
Thank you! Please ping me once you have the actual compiler ready.

PR #10116, opened 2019-08-16
main/ghostscript: remove jasper-dev from makedepends

maxice8:
Upstream already removed support for it in favour of openjpeg

@ncopa @Ikke


Ikke:
Would this warant a pkgrel bump?

maxice8:
No, this was already removed by upstream so it doesn't link to jasper, presumably on the version where it switched over to openjpeg.

So this removal just removes a package that doesn't affect compilation from the build dependencies, call it a cleanup if you wish.

Ikke:
Alright, thanks

PR #10118, opened 2019-08-16
testing/synapse: upgrade to 1.3.0

maxice8:
ref matrix-org/synapse#5732


maxice8:
Tracking issue for systemd: https://github.com/matrix-org/synapse/issues/5865

PR #7986, opened 2019-05-21
community/namecoin: upgrade to 0.18.0

andypost:
Ref https://github.com/namecoin/namecoin-core/releases/tag/nc0.18.0

PR #10143, opened 2019-08-17
community/bitcoin: upgrade to 0.18.1

israelfaria:
Ref https://github.com/bitcoin/bitcoin/releases/tag/v0.18.1

There's another [patch that upgraded to 0.18.0](https://patchwork.alpinelinux.org/patch/4896/) but unfortunately, similarly to PR #7986, simply updating the version causes a test to fail and therefore the build. My approach is identical to the one applied for namecoin.

### DIFF
--- a/community/bitcoin/APKBUILD
+++ b/community/bitcoin/APKBUILD
@@ -1,9 +1,9 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=bitcoin
-pkgver=0.17.1
+pkgver=0.18.1
 _ver=${pkgver/_/}
-pkgrel=2
+pkgrel=0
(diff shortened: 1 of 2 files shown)


community/bitcoin/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Natanael Copa <ncopa@alpinelinux.org>
 # Maintainer: Stuart Cardall <developer@it-offshore.co.uk>
 pkgname=bitcoin
-pkgver=0.17.1
+pkgver=0.18.1
 _ver=${pkgver/_/}
 pkgrel=2
```
maxice8: when upgrading a package reset pkgrel to 0

israelfaria: Thanks for the prompt response, didn't thought it would be as fast!
Fixed!

maxice8:
@itoffshore

PR #10206, opened 2019-08-18
testing/qutebrowser: upgrade to 1.7.0

maxice8:


maxice8:
@ddevault 

Still broken, loads pages but enter renderer crash loop shortly after

ddevault:
I haven't considered this package worth keeping up to date so long as dark magic deep in the cesspool of qtwebengine is broken on musl.

PR #10218, opened 2019-08-19
main/augeas: upgrade to 1.12.0

maxice8:


Ikke:
Test failures on non-x86* arches.

PR #10244, opened 2019-08-20
main/busybox-initscripts: update rpi bluetooth command

joneskoo:
The command hciattach is deprecated in bluez.
Replace the command with btattach.

### DIFF
--- a/main/busybox-initscripts/mdev.conf
+++ b/main/busybox-initscripts/mdev.conf
@@ -1,4 +1,4 @@
-# 
+#
 # This is a sample mdev.conf.
 #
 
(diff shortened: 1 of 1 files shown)


joneskoo:
I tested this (uncommented obviously) on my Raspberry Pi Zero W. Installing bluez and uncommenting this line seem to be sufficient to enable built-in Bluetooth.

btattach does not fork. There three lines lines of output from btattach at start which could be discarded but as this needs to be commented out by user to enable, I don't think that's necessary.

joneskoo:
I updated https://wiki.alpinelinux.org/wiki/Raspberry_Pi_3_-_Setting_Up_Bluetooth too.

ncopa:
we also need to update checksum and bump `pkgrel`.

ncopa:
fixed with 760c4286035013366ef0919e2d4f904f09d572f8

PR #10123, opened 2019-08-16
community/firefox-esr: security upgrade to 68.0.2

J0WI:
https://www.mozilla.org/en-US/security/advisories/mfsa2019-24/

PR #10252, opened 2019-08-20
testing/firefox: security upgrade to 68.0.2

timbru31:
See https://www.mozilla.org/en-US/security/advisories/mfsa2019-24/ and #10123

### DIFF
--- a/testing/firefox/APKBUILD
+++ b/testing/firefox/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
 pkgver=68.0.2
-pkgrel=1
+pkgrel=0
 pkgdesc="Firefox web browser"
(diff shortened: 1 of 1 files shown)


testing/firefox/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Sören Tempel <soeren+alpine@soeren-tempel.net>
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=firefox
-pkgver=68.0.1
+pkgver=68.0.2
 pkgrel=1
```
maxice8: pkgrel should be 0

timbru31:
Oh indeed. I’ll change it :)

testing/firefox/APKBUILD
timbru31: Done

PR #10264, opened 2019-08-20
main/pmacct: upgrade to 1.7.3

maxice8:
- Build uacctd daemon by enabling nflog
- Use supervise-dameon for proper process supervision
- Add name and description to daemons
- Remove examples since they contain compiled code


maxice8:
depends on #10267 and #10268

maxice8:
@rnalrd

PR #10272, opened 2019-08-21
main/ncurses: upgrade to 6.1-20190817

crwilliams:


crwilliams:
As seen previously in https://github.com/alpinelinux/aports/pull/7166, the previous version results in a HTTP 404 response from  https://invisible-mirror.net . In that PR, @ncopa suggested downloading from https://invisible-mirror.net/archives/ncurses/6.1/ rather than https://invisible-mirror.net/archives/ncurses/current/ . Unfortunately in this case, the previous version is missing from both locations so that suggestion would not work this time.

PR #10281, opened 2019-08-21
main/libx11: provide static libraries in libx11-static

maxice8:


ncopa:
i modified the commit message a bit.
f5595d8af20a0f627c505fe7c5c33e94744ef077

PR #10314, opened 2019-08-21
main/hwdata: upgrade to 0.326

maxice8:


ncopa:
i think we should deprecate this package and use hwids instead. See https://gitlab.alpinelinux.org/alpine/aports/issues/10716

PR #10344, opened 2019-08-22
testing/trafficserver: new aport

tcely:
https://trafficserver.apache.org
high-performance building block for cloud services


testing/trafficserver/APKBUILD
```diff
@@ -0,0 +1,162 @@
...
+# Maintainer: tcely <trafficserver+aports@tcely.33mail.com>
+_upstream="trafficserver"
+pkgname="${_upstream}"
+pkgver="8.0.5"
+pkgrel=0
+pkgdesc="high-performance building block for cloud services"
+url="https://trafficserver.apache.org"
+arch="x86_64"
+license="Apache-2.0"
+options=""
```
maxice8: empty variables can be removed

testing/trafficserver/APKBUILD
```diff
@@ -0,0 +1,162 @@
...
+depends_static="${pkgname}-dev"
+checkdepends="findutils"
+makedepends="${depends_dev} autoconf automake perl python3"
+install="${pkgname}.pre-install"
+subpackages="
+	${pkgname}-libs-static
+	${pkgname}-libs
+	${pkgname}-dev
+	${pkgname}-doc
+	${pkgname}-perl
```
maxice8: shouldn't it be ::noarch ?

testing/trafficserver/APKBUILD
```diff
@@ -0,0 +1,162 @@
...
+	trafficserver-musl-errno.patch
+	trafficserver-pthread_getname_np.patch
+	trafficserver-python3.patch
+	trafficserver-sed-regex-fix.patch
+	"
+builddir="${srcdir}/${_upstream}-${pkgver}"
+
+prepare() {
+	default_prepare
+	cd "${builddir}"
```
maxice8: can be removed

testing/trafficserver/APKBUILD
tcely: It could be, however this includes `/usr/bin/tspush` so I chose not to.

Is there any compelling reason to use `noarch` when there is only one supported architecture anyway?

Cogitri:
What's the status of any of these patches? Does upstream know about them?

testing/trafficserver/APKBUILD
```diff
@@ -0,0 +1,162 @@
+# Contributor: tcely <trafficserver+aports@tcely.33mail.com>
+# Maintainer: tcely <trafficserver+aports@tcely.33mail.com>
+_upstream="trafficserver"
+pkgname="${_upstream}"
```
Cogitri: Why not directly define pkgname as trafficserver?

testing/trafficserver/APKBUILD
```diff
@@ -0,0 +1,162 @@
...
+pkgver="8.0.5"
+pkgrel=0
+pkgdesc="high-performance building block for cloud services"
+url="https://trafficserver.apache.org"
+arch="x86_64"
```
Cogitri: Please add an option as to why this is only available on x86_64

testing/trafficserver/APKBUILD
Cogitri: ?

testing/trafficserver/APKBUILD
tcely: 64-bit only and `aarch64` is broken.

testing/trafficserver/APKBUILD
tcely: The package name is more likely to change than the upstream project name.

tcely:
@Cogitri 
> What's the status of any of these patches? Does upstream know about them?

What are you looking for with this comment? Submitting AlpineLinux specific patches upstream does not make sense to me. In any case, I'm not planning to submit these until the new package is merged.

PR #10345, opened 2019-08-22
testing/fceux: new aport

Danct12:
FCEUX is a Nintendo Entertainment System (NES), Famicom, Famicom Disk System (FDS), and Dendy emulator.

Signed-off-by: Danct12 <danct12@disroot.org>

### DIFF
--- a/testing/fceux/APKBUILD
+++ b/testing/fceux/APKBUILD
@@ -1,0 +1,38 @@
+# Contributor: Danct12 <danct12@disroot.org>
+# Maintainer: Danct12 <danct12@disroot.org>
+pkgname=fceux
+pkgver=2.2.3
+pkgrel=0
+pkgdesc="Fast and ultra-compatible NES/Famicom emulator"
+url="http://www.fceux.com"
+arch="all"
+license="GPL-2.0-only"
+depends_dev="sdl-dev gd-dev gtk+2.0-dev mesa-dev glu-dev"
+makedepends="$depends_dev scons"
+subpackages="$pkgname-doc"
+source="https://downloads.sourceforge.net/project/fceultra/Source%20Code/$pkgver%20src/$pkgname-$pkgver.src.tar.gz
+	scons-python3.patch"
+options="!check" # no testsuite
+
+build() {
+	scons -C "$pkgname-$pkgver"
+}
+
+package() {
+	scons install -i --prefix="$pkgdir/usr"
+
+	install -d "$pkgdir/usr/share/doc/$pkgname"
+	cp -r documentation/* "$pkgdir/usr/share/doc/$pkgname/"
+	install -Dm644 changelog.txt "$pkgdir/usr/share/doc/$pkgname/"
+
+	# desktop shortcut
+	install -Dm644 "$pkgname.desktop" \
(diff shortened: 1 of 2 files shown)


testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
+# Contributor: Danct12 <danct12@disroot.org>
+# Maintainer: Danct12 <danct12@disroot.org>
+pkgname=fceux
+pkgver=2.2.3
+pkgrel=0
+pkgdesc="Fast and ultra-compatible NES/Famicom emulator"
+url="http://www.fceux.com"
+arch="all"
+license="GPL"
```
maxice8: needs SPDX identifier

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+depends="sdl gd gtk+2.0"
```
maxice8: there should be automatically found by the package manager by looking at the linked libraries

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+depends_dev="sdl-dev gd-dev gtk+2.0-dev scons mesa-dev glu-dev"
```
maxice8: I think scons isn't meant to be in the developer dependencies but rather in makedepends

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+makedepends="$depends_dev"
+install=""
```
maxice8: empty variable can be removed for brevity

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+subpackages="$pkgname-doc"
+source="https://downloads.sourceforge.net/project/fceultra/Source%20Code/$pkgver%20src/$pkgname-$pkgver.src.tar.gz
+	scons-python3.patch"
+options="!check" # no testsuite
+builddir="$srcdir/$pkgname-$pkgver"
```
maxice8: default value can be removed

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+
+prepare() {
```
maxice8: Why define prepare if you're going to define it as the default definition ?

remove this

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+        default_prepare
+}
+
+build() {
+	cd "$builddir"
```
maxice8: no need for cd "$builddir" statements, it is done implicitly

testing/fceux/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+	scons -C "$pkgname-$pkgver"
+}
+
+package() {
+	cd "$builddir"
```
maxice8: same here

testing/fceux/APKBUILD
Danct12: So.. this depends line could be removed entirely?

PR #10351, opened 2019-08-22
main/tzdata: upgrade to 2019b

andypost:
https://mm.icann.org/pipermail/tz-announce/2019-July/000056.html


andypost:
Btw is it backportable to older releases?

ncopa:
> Btw is it backportable to older releases?

I think this should be backported yes

PR #10380, opened 2019-08-23
community/mplayer:provided support for aarch64

ossdev07:
Architecture specific support for aarch64 is added.

### DIFF
--- a/community/mplayer/APKBUILD
+++ b/community/mplayer/APKBUILD
@@ -4,22 +4,34 @@ pkgver=1.4.0
 pkgrel=1
 pkgdesc="A movie player for linux"
 url="http://www.mplayerhq.hu/design7/news.html"
-arch="x86_64"
+arch="x86_64 aarch64"
 license="GPL-2.0-or-later"
 subpackages="$pkgname-doc"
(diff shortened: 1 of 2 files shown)


community/mplayer/APKBUILD
```diff
@@ -4,22 +4,34 @@ pkgver=1.4.0
...
+			vdpau="disable-vdpau"	;;
+		*)
+			cpudetection="enable-runtime-cpudetection"
+	esac
+
 	./configure \
 		--prefix=/usr \
-		--enable-runtime-cpudetection \
+		--$cpudetection	\
+		--$vdpau	\
```
Cogitri: This doesn't work, if you don't set `vdpau` you get `unknown argument: --` as error.

community/mplayer/APKBUILD
```diff
@@ -47,6 +59,38 @@ build() {
 		--disable-nemesi \
 		--enable-freetype \
 		--enable-xvmc
+	case "$CARCH" in
+		aarch64)
+			cd ffmpeg/
+			./configure	\
```
Cogitri: Why can't we use the system's aarch64?

PR #10382, opened 2019-08-23
main/ansible: upgrade to 2.8.4

maxice8:


tcely:
I typically use "security upgrade" for these commits and list the CVE numbers in the commit message too.

maxice8:
It is not a security upgrade it is a normal upgrade that happens to have positive security implications. 

Security upgrades happen where normal upgrades don't happen, the stable branches.

tcely:
Check the git log for "security upgrade" you will find many commits just like this one.

maxice8:
> Check the git log for "security upgrade" you will find many commits just like this one.

That is very nice but it is still a normal upgrade that ends up having security implications that is properly tagged into a security upgrade on backports.

PR #10391, opened 2019-08-23
main/openssl: add openssl-static subpackage

maxice8:


tcely:
@maxice8 This is one of those cases where `openssl-libs-static` makes more sense to me.

Having a static `openssl` command may be very useful. So, using `openssl-static` for that rather than libraries is better.

ncopa:
In general, we don't do static builds (with a very few exceptions). Lets use `openssl-static` for now and reconsider whenever the need for a static openssl binary shows up.

PR #10398, opened 2019-08-23
[3.8] main/openldap: security upgrade to 2.4.48

maxice8:
ref #10752

fixes:

CVE-2019-13565
CVE-2019-13057


ncopa:
do we also need 900be1f?

PR #10404, opened 2019-08-23
community/libcanberra: add pulseaudio support

maxice8:
closes GH-6936

@Ikke @ncopa


community/libcanberra/APKBUILD
```diff
@@ -71,4 +65,11 @@ gtk3() {
 		"$subpkgdir"/usr/lib/gnome-settings-daemon-3.0/gtk-modules/
 }
 
+pulse() {
```
Ikke: Would it make sense to add an `isntall_if=` here so that it gets automatically installed when you have pulseaudio installed?

maxice8:
On Fri Aug 23, 2019 at 10:55 PM Kevin Daudt wrote:
> Ikke commented on this pull request.
> 
> 
> 
> > @@ -71,4 +65,11 @@ gtk3() {
>  		"$subpkgdir"/usr/lib/gnome-settings-daemon-3.0/gtk-modules/
>  }
>  
> +pulse() {
> 
> Would it make sense to add an `isntall_if=` here so that it gets automatically installed when you have pulseaudio installed?
> 
> -- 
> https://github.com/alpinelinux/aports/pull/10404#pullrequestreview-279277169

Added

community/libcanberra/APKBUILD
maxice8: Added

PR #10415, opened 2019-08-24
main/mpg123: upgrade to 1.25.12

maxice8:


main/mpg123/APKBUILD
```diff
@@ -20,7 +18,6 @@ build() {
 		--host=$CHOST \
 		--prefix=/usr \
 		--disable-dependency-tracking \
-		--with-ipv6 \
```
ncopa: why?

rnalrd:
@maxice8 thank you for the hard work you're doing in keeping the packages updated.

When some changes are coupled with the upgrade I think it would be useful for the reviewers to have some additional comments why the changes are made.

For example in this case ipv6 support was removed. But there other cases with other PRs. I don't mind having them in the same commit if they are small but it would make the review go fast if there's a comment about it.

Thanks!

main/mpg123/APKBUILD
prspkt: it appears to be on by default

ncopa: I'll add that info to the commit message

ncopa:
> When some changes are coupled with the upgrade I think it would be useful for the reviewers to have some additional comments why the changes are made.

And it is helpful to have those comments in the commit message, because people will wonder why was this done when looking at the git log. The commit message should answer "why" if that is not obvious.

PR #10418, opened 2019-08-24
main/fuse3: upgrade to 3.6.2

maxice8:


main/fuse3/APKBUILD
```diff
@@ -11,7 +11,7 @@ depends="fuse-common"
 makedepends="gettext-dev linux-headers meson eudev-dev python3
 	py3-six py3-pluggy py3-attrs py3-py"
 checkdepends="py3-pytest"
-subpackages="$pkgname-dev $pkgname-doc fuse-common:_common fuse-openrc:_openrc"
+subpackages="$pkgname-dev $pkgname-doc fuse-common:_common fuse-openrc:_openrc lib$pkgname:libs"
```
ncopa: I'd like to call it `$pkgname-libs` rather than `lib$pkgname`.

PR #10473, opened 2019-08-25
testing/neard: new aport

Minecrell:


testing/neard/neard.initd
```diff
@@ -0,0 +1,11 @@
+#!/sbin/openrc-run
+supervisor=supervise-daemon
+
+name="NFC"
```
maxice8: a description="" would be nice, but not blocking

PR #10557, opened 2019-08-27
testing/py-case: upgrade to 1.5.3

maxice8:
@ddevault


ddevault:
Drop py2?

ddevault:
Can we drop py2 from those, too?

PR #10563, opened 2019-08-27
testing/py3-case: upgrade to 1.5.3

maxice8:
@ddevault


ddevault:
I meant, can we drop python2 entirely from all of these packages?

ddevault:
https://git.sr.ht/~sircmpwn/aports/tree/testing-drop-py2

This tree should have py3-only versions of all of these packages

ddevault:
Your call, but I'm personally for doing them when we get to them.

PR #10564, opened 2019-08-27
testing/py3-kombu: drop py2

maxice8:
@ddevault


ddevault:
Sorry for being a pedantic dick here, if you want me to step in and do some of this work just lmk

PR #10571, opened 2019-08-27
main/gtk+3.0: support broadway backend

CrazyPandar:
update configure to support broadway backend


Cogitri:
That should be a single commit, simply amend the rev bump to the first commit.

PR #10572, opened 2019-08-27
[WIP] community/rust: upgrade to 1.37.0, enable on aarch64&x86 again

Cogitri:
Use upstream triplets for now, rust doesn't seem to want to work with our
triplets...

See https://github.com/rust-lang/rust/issues/62447

Still needs a bit more testing


Cogitri:
@mksully22 Could you test this on ppc64le again? :)

mksully22:
After applying the patch and re-enabling arch for ppc64le the package built successfully.

Cogitri:
Great, thank you!

PR #10577, opened 2019-08-27
main/unbound: upgrade to 1.9.3

maxice8:


nmeum:
I would personally separate the upgrade from the OpenRC service change. Maybe even merge it separately.

PR #10650, opened 2019-08-29
community/py3-factory-boy: drop py2

maxice8:
@ddevault


maxice8:
This will break py-django-oscar which depends on serveral py2- only packages.

ddevault:
Do you have a patch for py-django-oscar?

maxice8:
Nah, that package needs to convert 4 or 5 py2- only packages to py3- for it to be transfered

maxice8:
all of them py-django-* stuff

maxice8:
Packages for django-oscar

py3-django-extra-views
py3-django-haystack
py3-django-phonenumber-field
py3-django-sorl-thumbnail
py3-django-treebeard
py3-django-widget-tweaks

maxice8:
These need to be merged to add the requires py3- dependencies to django-oscar

#10706
#10707
#10708 
#10709 
#10710 
#10711

PR #10665, opened 2019-08-29
community/py3-ruamel.yaml: upgrade to 0.16.5

maxice8:


maxice8:
@maxwell-k

maxwell-k:
Before this PR a few tests fail and they are disabled with `pytest --deselect`

After this PR a few more tests fail and all of the tests are disabled with `options="!check"`

I would prefer to be consistent; for me the best approach is to keep disabling tests that cause a problem with `--deselect`.

I hope to have more time to look at this tomorrow.

maxice8:
Deselect the offending tests

maxwell-k:
LGTM! 🎉

PR #10668, opened 2019-08-29
main/py-scandir: upgrade to 1.10.0

maxice8:
@clandemeter


maxice8:
@clandmeter

ncopa:
looks like tests fails on 32bit arches

ncopa:
tests still fail on 32 bit

PR #10671, opened 2019-08-29
main/py-sphinx: depend on py-attrs

maxice8:
@ncopa


maxice8:
fixes build of sn0int

PR #10673, opened 2019-08-29
community/py3-pytest-xdist: upgrade to 1.29.0

maxice8:
drop py2 in the process


maxice8:
This will break checkdepends for py2-sqlalchemy

maxice8:
@dimon222

PR #10675, opened 2019-08-29
testing/grafana: security upgrade to 6.3.4 (CVE-2019-15043)

kaey:


maxice8:
need to add secfixes comment

kaey:
>need to add secfixes comment

where is it documented? why doesn't abump do it?

For now made by example from postgresql.

PR #10678, opened 2019-08-29
main/py-pluggy: upgrade to 0.12.0

maxice8:
This will break pytest on python2


maxice8:
@prspkt @itoffshore

prspkt:
to be safe we could add `depends="python3"` to `main/py3-importlib-metadata` and bump `pkgrel`, otherwise LGTM

maxice8:
i added py3-zipp (and python3) dependencies. which was missed when the package was first introduced

PR #10684, opened 2019-08-29
community/webkit2gtk: add secfixes

J0WI:


maxice8:
@Cogitri 

If there are security issues fixed we should upgrade the other branches as well

Cogitri:
Oh, I didn't do the 2.24.3 upgrade so I didn't see this. Will upgrade the other arches tomorrow.

PR #10706, opened 2019-08-30
main/py3-django-haystack: drop py2

maxice8:


main/py3-django-haystack/APKBUILD
```diff
@@ -0,0 +1,29 @@
+# Contributor: Kaarle Ritvanen <kaarle.ritvanen@datakunkku.fi>
+# Maintainer: Kaarle Ritvanen <kaarle.ritvanen@datakunkku.fi>
+pkgname=py3-django-haystack
+pkgver=2.8.1
+pkgrel=0
+pkgdesc="Modular search for Django"
+options="!check" # requires unpackaged pysolr and whoosh
```
ddevault: Should package them, then

PR #10707, opened 2019-08-30
main/py3-django-extra-views: upgrade to 0.12.0

maxice8:


main/py3-django-extra-views/APKBUILD
```diff
@@ -0,0 +1,35 @@
+# Contributor: Kaarle Ritvanen <kaarle.ritvanen@datakunkku.fi>
+# Maintainer: Kaarle Ritvanen <kaarle.ritvanen@datakunkku.fi>
+pkgname=py3-django-extra-views
+pkgver=0.12.0
+pkgrel=0
+pkgdesc="Extra class-based views for Django"
+options="!check" # Requires unpackaged django-nose
```
ddevault: So let's package it

maxice8: Good Luck :D

PR #10740, opened 2019-08-30
main/nss: upgrade to 3.46

prspkt:


J0WI:
I think `export NSS_ENABLE_WERROR=0` can be removed.

ncopa:
> I think `export NSS_ENABLE_WERROR=0` can be removed.

Seems like it adds `-Werror` without it and we don't need/want `-Werror` for releases.

prspkt:
> > I think `export NSS_ENABLE_WERROR=0` can be removed.
> 
> Seems like it adds `-Werror` without it and we don't need/want `-Werror` for releases.

I agree. Also if the next release raises a warning we will have to add it back.

PR #10744, opened 2019-08-30
testing/sniffglue: enable on arches supported by rust >= 1.37.0-r0

Cogitri:
* Remove unneccesary `cd "$builddir"`


maxice8:
x86 needs -fno-stack-protector workaround

aarch64 failed for new reasons (yay ?!)

Cogitri:
> aarch64 failed for new reasons (yay ?!)

Ah, that seems to be on sniffglue using syscalls that apparently aren't supported on aarch64 by the crate it uses

PR #10745, opened 2019-08-30
testing/mdbook: enable on arches supported by rust >= 1.37.0-r0

Cogitri:
* remove unneccesary RUSTFLAGS statements, we default to -crt-static


maxice8:
x86 needs -fno-stack-protector workaroudn

PR #10748, opened 2019-08-30
testing/bat: enable on arches supported by rust >= 1.37.0-r0

Cogitri:


maxice8:
need x86 -fno-stack-protector workaround

maxice8:
@5paceToast

PR #10749, opened 2019-08-30
testing/fractal: enable on arches supported by rust >= 1.37.0-r0

Cogitri:


maxice8:
armhf got the upstream LLVMxRust SIGSEGV

Cogitri:
Let's give it another go

PR #10750, opened 2019-08-30
testing/ffsend: new aport

Cogitri:


testing/ffsend/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Rasmus Thomsen <oss@cogitri.dev>
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
+pkgname="ffsend"
```
maxice8: quoting

testing/ffsend/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+pkgver="0.2.50"
```
maxice8: quoting

PR #10785, opened 2019-08-31
main/libssh: upgrade to 0.9.0

maxice8:


ncopa:
It does not break ABI?

ncopa:
> It does not break ABI?

apparently not.

PR #10786, opened 2019-08-31
testing/singularity: upgrade to 3.4.1

OTLabs:
- https://github.com/sylabs/singularity/releases/tag/v3.4.1 3.4.1
- Enable all architectures for builders
- Add cryptsetup to makedepends

### DIFF
--- a/testing/singularity/APKBUILD
+++ b/testing/singularity/APKBUILD
@@ -1,11 +1,11 @@
 # Contributor: Oleg Titov <oleg.titov@gmail.com>
 # Maintainer: Oleg Titov <oleg.titov@gmail.com>
 pkgname=singularity
-pkgver=3.3.0
-pkgrel=3
+pkgver=3.4.1
+pkgrel=0
 pkgdesc="Application containers focused on reproducibility for scientific computing and HPC world."
 url="https://www.sylabs.io/singularity/"
-arch=
+arch="all"
 license="BSD-3-Clause BSD-3-Clause-LBNL"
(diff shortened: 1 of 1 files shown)


OTLabs:
Drone x86_64 - builds fine, x86 - fails; aarch64 - builds fine, armhf and armv7 - fail.

All builds fail with same error:

```
# github.com/sylabs/singularity/internal/pkg/util/fs/overlay
--
224 | ../internal/pkg/util/fs/overlay/overlay.go:68:26: cannot use stfs.Type (type int32) as type int64 in map index

```

Travis CI builds fine.

testing/singularity/APKBUILD
```diff
@@ -17,7 +17,7 @@ makedepends="
 	libuuid
 	util-linux-dev
 	libseccomp-dev
-	"
+	cryptsetup"
```
maxice8: please keep the double-quotes in a separate line

OTLabs: DOne. Thank you!

OTLabs:
Disabling x86, armhf, and armv7 due to build errors.

OTLabs:
Singularity was updated to v3.4.1 which fixes compilation issues on non 64-bit architectures.

Cogitri:
fwiw it fails on s390x, will disable there

Cogitri:
Nevermind, seems like it only was a temporary failure

OTLabs:
Thank you!

PR #10813, opened 2019-09-01
main/py-paramiko: upgrade to 2.6.0

J0WI:


maxice8:
py3-netmiko needs paramiko>=2.4.3 according to setup.py

PR #10930, opened 2019-09-05
 main/glib: upgrade to 2.62.0

Cogitri:
* Adopt maintainership

### DIFF
--- a/main/glib/APKBUILD
+++ b/main/glib/APKBUILD
@@ -1,7 +1,8 @@
 # Contributor: Valery Kartel <valery.kartel@gmail.com>
-# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
+# Contributor: Natanael Copa <ncopa@alpinelinux.org>
+# Maintainer: Rasmus Thomsen <oss@cogitri.dev>
 pkgname=glib
-pkgver=2.60.7
+pkgver=2.62.0
 pkgrel=0
 pkgdesc="Common C routines used by Gtk+ and other libs"
 url="https://developer.gnome.org/glib/"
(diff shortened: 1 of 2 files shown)


Cogitri:
Could someone else maybe test Qemu? I don't have VMs.

main/glib/APKBUILD
```diff
@@ -10,7 +11,7 @@ license="LGPL-2.1-or-later"
 triggers="$pkgname.trigger=/usr/share/glib-2.0/schemas:/usr/lib/gio/modules"
 depends_dev="python3 gettext-dev bzip2-dev libxml2-utils libxslt
 	docbook-xml docbook-xsl"
-makedepends="$depends_dev pcre-dev meson ninja
+makedepends="$depends_dev pcre-dev meson dbus-dev dbus
```
ncopa: `dbus-dev` here will introduce a circular dependency since dbus needs `glib-dev` to build.

Cogitri: Oh right, thanks for noticing

ncopa:
> Could someone else maybe test Qemu? I don't have VMs.

builds. :+1:

main/glib/APKBUILD
ncopa: why is it needed? it seems to build fine without dbus.

Cogitri:
Great thanks. Will have to fix this crashing gnome-shell, somehow they break something about this every minor release...

main/glib/APKBUILD
Cogitri: Ah, it's for additional tests, but we can live without that to avoid the dependency loop.

Cogitri:
Alrighty, this works with the gjs release candidate. Would it be OK to upgrade to that already, or should we wait for the actual stable release? I'd personally prefer using the gjs release candidate just so we have a more smooth rollout of GNOME since otherwise almost all of it will block on this which would be unfortunate and the gjs release should be done by September 11 (hopefully sooner).

Cogitri:
There's no change in 1.58.0 (other than the version number). I'll bump qemu for this and than this should he good to go.

Cogitri:
Mind taking another look, @ncopa?

Cogitri:
Qemu fails:

```
usr/lib/gcc/i586-alpine-linux-musl/8.3.0/../../../../i586-alpine-linux-musl/bin/ld: cannot find -lz
```

Not quite sure why since you said that it works for you though and zlib-dev is a dep (so I guess it might be on linking order?)

Cogitri:
Removed Qemu for now, we can bump it later but I don't really want to block GNOME on this :/

PR #10931, opened 2019-09-05
main/libsecret: upgrade to 0.19.0

maxice8:


Cogitri:
Are uneven numbers stable for libsecret?

maxice8:
On Fri Sep 6, 2019 at 5:35 AM Rasmus Thomsen wrote:
> Are uneven numbers stable for libsecret?
> 

No, i forgot it is a gnome project

PR #10936, opened 2019-09-06
main/linux-vanilla: PINCTRL_SINGLE is needed builtin for BeagleBoneBlack

vicencb:
Otherwise it fails to initialize the console.


ncopa:
this was fixed with 9c668293db911bf69ec23158500d96218d850b24

PR #10961, opened 2019-09-06
community/exim: security upgrade to 4.92.2

J0WI:
https://www.exim.org/static/doc/security/CVE-2019-15846.txt


maxice8:
Merged this and pushed to 3.10 as well.

PR #11062, opened 2019-09-07
main/acpi: adopt

durrendal:
Adopt maintenance of ACPI as Cag has decided to no longer maintain this package.


maxice8:
@ddevault asked before on https://gitlab.alpinelinux.org/alpine/aports/merge_requests/51.patch

durrendal:
Oh perfect, just so long as ACPI is still maintained. I'll let him take this one! Thanks for the heads up Maxice!

PR #11155, opened 2019-09-08
esting/hiawatha: fix / improve initd

itoffshore:
* adds `$pidfile` required by supervise-daemon
* adds `healthcheck_timer` so the service is monitored
* fixes stopping the daemon

	internally `hiawatha` spawns mutiple processes. To stop `hiawatha`
	it is necessary to terminate it's `supervise-daemon` process.

### DIFF
--- a/testing/hiawatha/hiawatha.initd
+++ b/testing/hiawatha/hiawatha.initd
@@ -2,12 +2,21 @@
 
 supervisor=supervise-daemon
 
-name="Webserver"
-description="Secure and Advanced Hiawatha Webserver"
+name="Hiawatha webserver"
+description="Secure and advanced webserver"
 
 command=/usr/sbin/hiawatha
-command_args_foreground="-d"
+command_args_foreground="-d 1>/dev/null"
+pidfile=/var/run/hiawatha.pid
+healthcheck_timer=60
 
 depend() {
 	after firewall
+}
+
+stop() {
+	local supervise_pid=$(pgrep -f "$supervisor $RC_SVCNAME")
+
+	einfo "Stopping $name"
+	kill -TERM $supervise_pid
 }
--- a/testing/hiawatha/APKBUILD
+++ b/testing/hiawatha/APKBUILD
@@ -3,7 +3,7 @@
 # Contributor: Stuart Cardall <developer at it-offshore dot co.uk>
 pkgname=hiawatha
 pkgver=10.9
-pkgrel=2
+pkgrel=3
 pkgdesc='Secure and advanced webserver'
 url='https://www.hiawatha-webserver.org/'
 arch="all"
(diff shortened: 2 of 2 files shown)


testing/hiawatha/hiawatha.initd
```diff
@@ -2,12 +2,21 @@
...
 
-name="Webserver"
-description="Secure and Advanced Hiawatha Webserver"
+name="Hiawatha webserver"
+description="Secure and advanced webserver"
 
 command=/usr/sbin/hiawatha
-command_args_foreground="-d"
+command_args_foreground="-d 1>/dev/null"
+pidfile=/var/run/hiawatha.pid
```
maxice8: pidfile isn't required by supervise-daemon, the whole reason to use supervise daemon is to not have to use the disgusting hack that is pidfiles.

testing/hiawatha/hiawatha.initd
```diff
@@ -2,12 +2,21 @@
...
 supervisor=supervise-daemon
 
-name="Webserver"
-description="Secure and Advanced Hiawatha Webserver"
+name="Hiawatha webserver"
+description="Secure and advanced webserver"
 
 command=/usr/sbin/hiawatha
-command_args_foreground="-d"
+command_args_foreground="-d 1>/dev/null"
```
maxice8: put it on command_args. command_args_foreground is meant for the command switch that makes the program stay on the foreground

testing/hiawatha/hiawatha.initd
```diff
@@ -2,12 +2,21 @@
...
+pidfile=/var/run/hiawatha.pid
+healthcheck_timer=60
 
 depend() {
 	after firewall
 }
+
+stop() {
```
maxice8: supervise-daemon requires the default start,stop functions. This is just replicating in a fragile form what is done reliably with supervise-daemon.

testing/hiawatha/hiawatha.initd
itoffshore: It seems to be required:

```
root@proxy [~]# service hiawatha start
 * Caching service dependencies ... [ ok ]
 * Starting Hiawatha webserver ... * supervise-daemon: --pidfile must be specified
 * failed to start Hiawatha webserver
                                                                                                                                              [ !! ]
 * ERROR: hiawatha failed to start
```
with `pidfile` set it starts

testing/hiawatha/hiawatha.initd
itoffshore: the current `initd` is broken.

Not only does it fail to start hiawatha - it does not stop it either due to the numerous processes hiawatha spawns.

maxice8:
image

image

testing/hiawatha/hiawatha.initd
itoffshore: run `netstat -lptun` - hiawatha is not stopped

maxice8: image

itoffshore: when run as a reverse proxy it does not stop

itoffshore:
Closing - the `initd` is broken in Alpine `3.9` & is ok in `3.10`

PR #11223, opened 2019-09-09
testing/py-pyparted: drop python2, clarify license

prspkt:


maxice8:
This will break ginger, work and gingerbasea and by extension kimchi

prspkt:
> This will break ginger, work and gingerbasea and by extension kimchi

I will add a hold label however, `wok` and `gingerbasea` are already broken due to `py2-pyldap` and `py2-pam` missing.

maxice8:
oh, if that is the case then fire away

PR #11234, opened 2019-09-09
testing/gnome-terminal: upgrade to 3.34.0

Cogitri:


Cogitri:
Ah, it doesn't work because it can't pull in the just built vte3 :c

Cogitri:
Needs #11743

PR #11251, opened 2019-09-09
unmaintained/letsencrypt-nosudo: move from community

maxice8:
- Requires python2
- Can be returned once https://github.com/diafygi/acme-nosudo/pull/99
  or similar is merged


maxice8:
@itoffshore

PR #11266, opened 2019-09-10
testing/clisp: Attempt to build on all platforms

durrendal:
Attempt to see if current APKBUILD builds on any platform outside of x86_64.


durrendal:
Missing ffcall for x86 and arm variants it seems.

PR #11269, opened 2019-09-10
testing/clisp: build test

durrendal:
Testing to see if clisp can build now that ffcall dependency is met.


durrendal:
Build fails with a segmentation fault on arm, aarch64 doesn't build at all, and x86 appears to still be missing ffcall.

PR #11290, opened 2019-09-10
testing/geary: depend on libgnome-keyring

maxice8:
crashes with SIGTRAP otherwise

@Cogitri


Cogitri:
Huh, it should use libsecret and it works for me without libgnome-keyring: 
Screenshot from 2019-09-11 08-29-28

PR #11323, opened 2019-09-11
main/portaudio: enable cpp bindings

Cogitri:


Cogitri:
Oh, didn't know people could do without libstdc++ to be honest. I can split it then.

Cogitri:
Splitted it into libportaudiocxx.

PR #11325, opened 2019-09-11
community/ceph: Use $pkgname instead of ceph for subpackages

iggy:
Fix up some variable use per PR feedback

### DIFF
--- a/community/ceph/APKBUILD
+++ b/community/ceph/APKBUILD
@@ -226,8 +226,8 @@ user() {
 base() {
 	pkgdesc="Base is the package that includes all the files shared amongst ceph servers"
 	depends="$_base_deps
-		ceph-user=$pkgver-r$pkgrel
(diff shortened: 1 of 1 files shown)


iggy:
The CI on this package never finishes because ceph is a huge amount of C++ code that takes forever to compile.

Cogitri:
IMO `use $pkgname instead of ceph for subpackages` would be a better commit msg

iggy:
Yeah, the commit message is wonky because I cherry-picked this from another branch that got merged already. It made more sense in that context. I'll see if I can figure out the git-fu to update the commit message.

tcely:
> I'll see if I can figure out the git-fu to update the commit message.

`git rebase -i d3739c9~` then use `reword` instead of `pick`.

iggy:
Updated the commit message and the title of the PR. Thanks for the review.

PR #11354, opened 2019-09-12
main/py-certifi: upgrade to 2019.9.11

dimon222:
For now still cross-python package, but it will be switched to `py3-` version as nested dependencies get converted from cross-python version into `py3-` only versions (py-tornado and py-requests)


dimon222:
Commited in bf52553b2b135847313e71d71597bec3bb3e943c

PR #11365, opened 2019-09-12
community/gitlab-runner: set rev&branch during build

Cogitri:
The git rev is used to determine which docker tag of gitlab-runner-helper to use

Please test @Ikke , this fixes gitlab-runner for me, but I don't want to break
Alpine's CI setup :)


Ikke:
> Please test @Ikke , this fixes gitlab-runner for me, but I don't want to break
> Alpine's CI setup :)

Right, without that information, you have to explicitly set the docker image/tag to use in the gitlab-runner config, which is what we do now anyway, so I don't think it will break anything, but I will test it when I have time.

Cogitri:
Thank you

PR #11380, opened 2019-09-13
testing/netcdf-fortran: fix nf-config

hjaekel:
workaround for bug https://github.com/Unidata/netcdf-fortran/issues/60 which leads to an unusable `nf-config`: switch from CMake to configure. As now also a static library is build, I added a -static subpackage.


maxice8:
Is there demand for the static libraries ? i'd rather they be off by default

hjaekel:
fund the option --disable-static, so I could remove the -static subpackage.

PR #11387, opened 2019-09-13
[3.10] main/openssl: Fix openssl secfix version number

mirskifa:
In 3.10, `pkgrel` is `0` not `1` as it is in the master branch.  Secfix block should reflect this.

### DIFF
--- a/main/openssl/APKBUILD
+++ b/main/openssl/APKBUILD
@@ -20,7 +20,7 @@ esac
 builddir="$srcdir/openssl-$pkgver"
 
 # secfixes:
-#   1.1.1d-r1:
+#   1.1.1d-r0:
 #     - CVE-2019-1547
 #     - CVE-2019-1549
 #     - CVE-2019-1563


mirskifa:
@ncopa It looks like you've been the person to merge the recent changes to the 3.10 and, as of today, 3.9 branches.  The secfix comment block in both of these releases is at odds with the `pkgrel` version specified (it looks like maybe the secfix block was copied from edge at the time?).  Please can we get the block to line up with the actual release builds?

mfiotgh:
@ncopa Do you need a separate PR for the recent update in the 3.9 branch (it wasn't there when I opened this PR)?

mirskifa:
@ncopa Do you need a separate PR for the 3.9 branch (1.1.1d hadn't been merged into 3.9 when I opened this PR)?

mirskifa:
Assuming separate PR needed so created #11853

ncopa:
> @ncopa Do you need a separate PR for the 3.9 branch (1.1.1d hadn't been merged into 3.9 when I opened this PR)?

technically, separate PR shouldn't be needed, but I have hard time to keep up with everything. A separate PR makes it clear what is going on and in this case it is helpful.

Thanks!

PR #11388, opened 2019-09-13
testing/xerces-c: reduce package size by creating -samples and -dev subpackages

hjaekel:


testing/xerces-c/APKBUILD
```diff
@@ -3,23 +3,21 @@
 pkgname=xerces-c
 pkgver=3.2.2
 pkgrel=0
```
maxice8: please bump pkgrel so changes take effect.

testing/xerces-c/APKBUILD
```diff
@@ -3,23 +3,21 @@
...
+pkgdesc="A validating XML parser written in a portable subset of C++"
 url="https://xerces.apache.org/index.html"
 arch="x86 x86_64"
 license="Apache-2.0"
 makedepends="
     cmake
     linux-headers
     "
-subpackages="$pkgname-doc"
+subpackages="$pkgname-samples:samples $pkgname-doc $pkgname-dev"
```
maxice8: no need for the `:samples` part, it will by default use the word after the last hyphen as the name of the function

```
$pkgname-dev -> dev()
$pkgname-libs-static -> static()
$pkgname-libs -> libs()
$pkgname-samples -> samples()
```

testing/xerces-c/APKBUILD
hjaekel: done

testing/xerces-c/APKBUILD
hjaekel: done

PR #11391, opened 2019-09-13
main/libsecret: upgrade to 0.19.1

maxice8:
@Cogitri


Cogitri:
So 0.19 is stable after all?

ncopa:
@maxice8 is this a stable release? its a gnome project.

maxice8:
It appears to be, Arch Linux uses it but i couldn't find much information in their sourcerepo.

PR #11415, opened 2019-09-15
testing/redo-c: new aport

maxice8:
https://github.com/leahneukirchen/redo-c
redo build system in portable C with zero dependencies

@Ikke


maxice8:
Now on GitLab

PR #11434, opened 2019-09-15
main/py3-vobject: drop py2

maxice8:


Ikke:
This has been merged as aa81b906, closing.

PR #11435, opened 2019-09-15
main/py3-twitter: drop py2

maxice8:


Ikke:
This has been merged as c936ade6, closing

PR #11437, opened 2019-09-15
testing/libkml: fix checkdepends and depends_dev

hjaekel:


testing/libkml/APKBUILD
```diff
@@ -13,14 +13,21 @@ makedepends="
...
 	openjdk$_jdkver-jdk
 	python3-dev
 	swig
 	uriparser-dev
 	zlib-dev
 	"
+checkdepends="
+	gtest-dev
+	"
+depends_dev="
```
maxice8: isn't it better to declare depends_dev before makedepends and then add $depends_dev to makedepends ?

hjaekel: Yes, that's much better. I changed that and force pushed.

PR #11460, opened 2019-09-16
main/vala: upgrade to 0.46.1

Cogitri:
@maxice8 did the tests fail on all arches?


Cogitri:
Tests time out on armhf, as usual.

PR #11467, opened 2019-09-16
main/gcc: add patch fixing libgomp

Cogitri:
The defect can be obversible e.g. with GNOME Photos

@ncopa could you give this a try?

### DIFF
--- a/main/gcc/APKBUILD
+++ b/main/gcc/APKBUILD
@@ -6,7 +6,7 @@ pkgver=9.2.0
 [ "$CHOST" != "$CTARGET" ] && _target="-$CTARGET_ARCH" || _target=""
 
 pkgname="$pkgname$_target"
-pkgrel=1
+pkgrel=2
 pkgdesc="The GNU Compiler Collection"
 url="http://gcc.gnu.org"
 arch="all"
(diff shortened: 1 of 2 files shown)


main/gcc/APKBUILD
```diff
@@ -171,6 +171,7 @@ source="https://gcc.gnu.org/pub/gcc/releases/gcc-${_pkgbase:-$pkgver}/gcc-${_pkg
 	320-libffi-gnulinux.patch
 
 	gcc-pure64-mips.patch
+	0016-invalid_tls_model.diff
```
ncopa: I think we need to rename it to `0016-invalid_tls_model.patch` or it will not be applied by abuild.

Cogitri: Derp, right. Thanks for noticing

Cogitri:
Tested, wfm. BTW without ADA enabled this happens:

>>> libgnat*: Running split function libgnat...
mv: can't rename '/home/builder/aports/main/gcc/pkg/gcc/usr/lib/libgna*.so': No such file or directory
>>> ERROR: libgnat*: libgnat failed
>>> ERROR: gcc*: prepare_subpackages failed
>>> ERROR: gcc: rootpkg failed
>>> gcc: Uninstalling dependencies...

PR #11473, opened 2019-09-17
testing/consul: upgrade to 1.6.1 and improve init script

mtneug:
Upgrade Consul to latest 1.6.1 and migrate init scrip to `supervise-daemon` as well as add `healthcheck` and config `reload` functions.


maxice8:
upgrade is lowercase

PR #11474, opened 2019-09-17
community/vault: upgrade to 1.2.3 and improve init script

mtneug:
Upgrade Vault to latest 1.2.3 and improve init scrip with supervise-daemon configurations and config `reload` functions.

### DIFF
--- a/community/vault/vault.initd
+++ b/community/vault/vault.initd
@@ -1,21 +1,32 @@
 #!/sbin/openrc-run
-
-supervisor=supervise-daemon
-
 name="Vault server"
 description="Vault is a tool for securely accessing secrets"
+description_reload="Reload configuration"
 
-VAULT_LOG_FILE="/var/log/${RC_SVCNAME}.log"
+extra_started_commands="reload"
 
-command=/usr/sbin/${RC_SVCNAME}
+command="/usr/sbin/${RC_SVCNAME}"
 command_args="${vault_opts}"
 command_user="${RC_SVCNAME}:${RC_SVCNAME}"
 
-start_pre() {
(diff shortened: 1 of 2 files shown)


maxice8:
upgrade is lowercase

community/vault/vault.initd
```diff
@@ -1,21 +1,33 @@
...
-command=/usr/sbin/${RC_SVCNAME}
+command="/usr/sbin/${RC_SVCNAME}"
 command_args="${vault_opts}"
 command_user="${RC_SVCNAME}:${RC_SVCNAME}"
 
-start_pre() {
-	checkpath -f -m 0644 -o "$command_user" "$VAULT_LOG_FILE"
-}
+supervisor=supervise-daemon
+pidfile="/run/${RC_SVCNAME}.pid"
```
maxice8: don't think pidfiles matter for supervise-daemon

community/vault/vault.initd
```diff
@@ -1,21 +1,33 @@
...
 }
+
+start_pre() {
+	checkpath -f -m 0644 -o "$command_user" "$output_log" "$error_log"
+}
+
+reload() {
+	start_pre \
+		&& ebegin "Reloading $RC_SVCNAME configuration" \
+		&& supervise-daemon "$RC_SVCNAME" --signal HUP --pidfile "$pidfile"
```
maxice8: use $supervisor, supervise-daemon shouldn't use pidfile

community/vault/vault.initd
mtneug: Apparently supervise-daemon will create a default PID file (`/run/supervise-vault.pid` in this case). Will update the PR.

PR #11475, opened 2019-09-17
testing/nomad: new aport

mtneug:
Add [Nomad](https://www.nomadproject.io/).


maxice8:
```
testing/nomad: new aport

$pkgdesc
$url
```

maxice8:
Is python2 really  required ? it is soon to be EOL

mtneug:
No, it's not. Python 3 also works fine. Thanks for pointing this out. I also updated the init script and removed the custom PID file.

mtneug:
I spoke too soon. On `x86_64`, where I primarily test, it is working probably because the build output is downloaded instead of generated. For creating the build output on other platforms, however, the dependency [node-gyp](https://github.com/nodejs/node-gyp) currently only support Python 2.

PR #11479, opened 2019-09-17
community/mawk: version in source download updated to version 20190203

mksully22:
Source contained in downloaded mawk archive, https://invisible-island.net/datafiles/release/mawk.tar.gz,
has been updated moving from version 20171017 to version 20190203. The pkgdate value must be updated to avoid build error:
```
/usr/bin/abuild: cd: line 16: can't cd to /home/mksully/trees/aports/community/mawk/src/mawk-1.3.4-20171017: No such file or directory
```


Cogitri:
Ugh, there are no versioned tarballs?

PR #11491, opened 2019-09-18
community/php7-pecl-ssh2: upgrade to 1.2

andypost:
Ref https://pecl.php.net/package-changelog.php?package=ssh2&release=1.2

PS download broken now, that's why WIP, see https://github.com/php/pecl-networking-ssh2/pull/36#issuecomment-532676199
PPS
- https://bugs.php.net/bug.php?id=78560
- https://bugs.php.net/bug.php?id=78567


andypost:
Fixed by using github releases, actually they are uploaded to pecl according to https://bugs.php.net/bug.php?id=78560

PR #11495, opened 2019-09-19
main/py3-flask-assets: drop py2

maxice8:


Ikke:
This seems to be missing a dependency.

maxice8:
It is now on GitLab

PR #11496, opened 2019-09-19
main/py3-flask-script: drop py2

maxice8:


Ikke:
@maxice8 This renames flask-assets to flask-script, but there is a separate [PR](11495) for flask-assets. Is that a mistake?

maxice8:
Yes, it is a mistake and is already fixed by another commit

PR #11497, opened 2019-09-19
main/py3-flask: drop py2

maxice8:


maxice8:
@ddevault

maxice8:
On Thu, 19 Sep 2019 08:13:38 -0700
Drew DeVault <notifications@github.com> wrote:

> ddevault requested changes on this pull request.
> 
> No check?
> 
> 
> 

Requires upper-repo dependencies

ddevault:
Let's move them, or move flask to community (wait, why is flask in main anyway)

maxice8:
On Thu, 19 Sep 2019 08:15:43 -0700
Drew DeVault <notifications@github.com> wrote:

> Let's move them, or move flask to community (wait, why is flask in
> main anyway)
> 

You tell me, you're the maintainer

ddevault:
I adopted it, I didn't make the call. The person who used to maintain it is no longer involved with Alpine. It doesn't look like much thought was given to putting it in main (this was back in 2014, it seems), so I don't see any reason to keep it there. I think we ought to move it.

Ikke:
> It doesn't look like much thought was given to putting it in main (this was back in 2014, it seems)

At that time, community did not exist yet, so main was the only stable repo :-)

PR #11502, opened 2019-09-19
main/py3-tappy: drop py2

maxice8:


Ikke:
Merged as bbd9d915, closing

PR #11508, opened 2019-09-19
main/py3-tox: drop py2

maxice8:


Ikke:
Merged as f47083a9b, closing.

PR #11509, opened 2019-09-19
main/py3-pytest: drop py2

maxice8:


Ikke:
Merged as c6d12c52b, closing.

PR #11519, opened 2019-09-19
main/py3-wcwidth: drop py2

maxice8:


Ikke:
Merged as 326ba5a7c, closing.

PR #11531, opened 2019-09-20
main/py3-toml: drop py2

maxice8:


Ikke:
Merged as 581053e6, closing.

PR #11543, opened 2019-09-20
testing/gdal: add py3-gdal subpackage and upgrade to 2.4.2

zarch:
Hi, 
this is my first contribution to aports, I've read the documentation on how to contribute, but please let me know if something is wrong and/or I've missed something. :-)

The commit add the subpackage: `py3-gdal` to be able to use gdal from `python3`.

I have also updated the version from `2.4.0` to `2.4.2`.

Regarding the version, there is also available the version: `3.0.1` released 2019-06-28, see the [changelog](https://github.com/OSGeo/gdal/blob/v3.0.1/gdal/NEWS), but I'm not sure about the general alpine distribution policy, it is a major release, so it might break other packages.


testing/gdal/APKBUILD
```diff
@@ -36,26 +38,43 @@ build() {
...
+	pkgdesc="$pkgname (python2 bindings)"
 
-	cd "$builddir"/swig/python
+	cd "$builddir"/swig/python2
 	python2 setup.py install --prefix=/usr --root="$subpkgdir"
 
 	chmod a+x scripts/*
 	install -d "$subpkgdir"/usr/bin
 	install -m755 scripts/*.py "$subpkgdir"/usr/bin/
 }
```
Cogitri: Just drop python2 in favour of python3, we're in the process of removing py2 anyway since it'll be EOL soon.

Cogitri:
LGTM to me otherwise, you don't have to do the major bump if you don't feel like it :)

testing/gdal/APKBUILD
tclarke: Agreed, just remove all the py2 stuff.

Cogitri:
Uhh... that doesn't look right

zarch:
> Uhh... that doesn't look right

mmh... yes I see... I did a `git pull --rebase` but this have added all the other commits on this PR.
Should I close the PR do my changes and open a new PR? Other better idea on how I can solve this?

Cogitri:
Hm, usually rebasing upon alpine master should work just fine, no clue what went wrong here

Cogitri:
Ah, you rewrote history. Put your two commits at the end of the history (to not put it in front of the already existing ones) and squash them into one please.

zarch:
@Cogitri, I did a new PR #11555 where you can find just one single commit. I close this PR.
Sorry for the noise.

PR #11550, opened 2019-09-20
testing/low-memory-monitor: new aport

maxice8:
https://gitlab.freedesktop.org/hadess/low-memory-monitor
Daemon to monitor memory usage and signal processes and the kernel


maxice8:
@ncopa

maxice8:
Now on GitLab

PR #11552, opened 2019-09-20
community/cppcheck: drop py2

maxice8:


Ikke:
Fails with error:

> g++: error: build/analyzerinfo.cpp: No such file or directory

PR #11553, opened 2019-09-20
move hyperlink,automat,pyhamcrest to main

maxice8:
Requires by py-twisted


maxice8:
Twisted was moved to community

PR #11554, opened 2019-09-20
community/emacs upgrade version to 26.3

stef:
minor version upgrade only.

### DIFF
--- a/community/emacs/APKBUILD
+++ b/community/emacs/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: stef <l0ls0fo2i@ctrlc.hu>
 # Contributor: Timo Teräs <timo.teras@iki.fi>
 pkgname=emacs
-pkgver=26.2
+pkgver=26.3
 pkgrel=2
 pkgdesc="The extensible, customizable, self-documenting real-time display editor"
 arch="all"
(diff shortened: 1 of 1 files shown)


maxice8:
Please use the correct format  for  updating packages

`<repo>/<package>: upgrade to <pkgver>`

stef:
the CI builders seem to have problems with this, however on my private builders this builds fine. we had the same problem also with the previous version.

stef:
btw i don't see that "bumping" would be incorrect: https://github.com/alpinelinux/aports/search?q=bumped&type=Commits

stef:
damn. i forgot the checksum

Cogitri:
Fails on my machine with the same error as on CI

stef:
which one? on travis-ci the problem is, that ftp is blocked by some firewall rule, on drone io some other security mechanism gets in the way. it's a problem with the ci security settings.

community/emacs/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: stef <l0ls0fo2i@ctrlc.hu>
 # Contributor: Timo Teräs <timo.teras@iki.fi>
 pkgname=emacs
-pkgver=26.2
+pkgver=26.3
 pkgrel=2
```
Cogitri: Argh, noticed it too late, but this should've been reset to 0.

stef: hrmpf! sorry! i'll try to not forget in the future.

Cogitri: No worries, I should've caught that.

PR #11555, opened 2019-09-20
testing/gdal: add py3-gdal, drop py2 and upgrade to 2.4.2

zarch:
I did a mess with my previous PR #11543 so here a clean version.


testing/gdal/APKBUILD
```diff
@@ -26,7 +26,7 @@ makedepends="
 	"
 subpackages="
 	$pkgname-dev
-	py-$pkgname:py
+	py3-$pkgname:py3
```
Cogitri: We usually do `provides` and `replaces` here for backwards compatibility, please see eg. py3-cffi for how this is done.

Cogitri:
Sorry for the long waiting time.

PR #11577, opened 2019-09-21
testing/ranger: fixed linting issue

jjsearle:


maxice8:
I'll do it when ranger gets an update

PR #11578, opened 2019-09-21
testing/xsel: fixed source + build

jjsearle:


testing/xsel/APKBUILD
```diff
@@ -2,18 +2,17 @@
...
-makedepends="libxt-dev"
+makedepends="libxt-dev autoconf automake libtool"
 subpackages="$pkgname-doc"
-source="http://www.vergenet.net/~conrad/software/$pkgname/download/$pkgname-$pkgver.tar.gz"
-builddir="$srcdir/$pkgname-$pkgver"
+source="$pkgname-$pkgver.tar.gz::https://github.com/kfish/xsel/archive/$pkgver.tar.gz"
 
 build() {
-	cd "$builddir"
+	./autogen.sh
```
maxice8: please separate the ./autogen.sh into prepare()

jjsearle: autogen.sh separated into prepare function.

testing/xsel/APKBUILD
```diff
@@ -2,18 +2,24 @@
...
-source="http://www.vergenet.net/~conrad/software/$pkgname/download/$pkgname-$pkgver.tar.gz"
-builddir="$srcdir/$pkgname-$pkgver"
+source="$pkgname-$pkgver.tar.gz::https://github.com/kfish/xsel/archive/$pkgver.tar.gz"
+
+prepare() {
+	aclocal
+	libtoolize --automake
+	autoheader
+	automake --add-missing
+	autoconf
```
Cogitri: `autoreconf -if` doesn't cut it?

nmeum: @jjsearle :question:

maxice8: can't it all be replaced with a simple call to autoreconf -fi like it is done elsewhere ?

PR #11581, opened 2019-09-21
community/icinga2: upgrade to 2.11.0

jjsearle:


community/icinga2/APKBUILD
```diff
@@ -14,26 +14,20 @@ install="$pkgname.pre-install $pkgname.post-install"
 pkgusers="icinga"
 pkggroups="icinga icingacmd"
 subpackages="$pkgname-doc $pkgname-bash-completion:bashcomp:noarch $pkgname-vim::noarch $pkgname-openrc"
-source="$pkgname-$pkgver.tar.gz::https://github.com/Icinga/$pkgname/archive/v$pkgver.tar.gz
+source="icinga2-$pkgver.tar.gz::https://github.com/Icinga/icinga2/archive/v$pkgver.tar.gz
```
Cogitri: Nitpick, but I think we usually keep the `$pkgname` in the archive name, since it makes sense to have the name of the package in the distfile since that makes mapping the distfile to the respective package easier.

jjsearle:
beaten to the punch it would seem
https://github.com/alpinelinux/aports/commit/2ea8b5069d14c893193ddbc7fb3e2acbdd45622b

PR #11586, opened 2019-09-22
testing/autojump: new aport

icyphox:
Add `autojump` package -- a faster way to navigate your filesystem.


icyphox:
It appears that the CI doesn't have the `SHELL` environment variable set, which explains why the build fails.

PR #7666, opened 2019-05-07
main/luajit: add support for s390x

J0WI:
https://github.com/alpinelinux/aports/pull/7223#issuecomment-489543894
There are currently many other packages disabled for s390x because of this. Let's see if it works with nginx first.

PR #11599, opened 2019-09-22
community/suricata: add rust support for more arches

J0WI:


community/suricata/APKBUILD
```diff
@@ -24,7 +24,7 @@ _enable_luajit=""
 _enable_rust=""
 
 case "$CARCH" in
-	armhf|armv7|x86_64)
+	aarch64|armhf|armv7|ppc64le|x86|x86_64)
```
Cogitri: I'm pretty sure luajit isn't actually functional on at least aarch64.

J0WI: AFAIK it works on all but `s390x`

Cogitri: Well, at least on my aarch64 machine it just instantly explodes

J0WI: `luajit` or `suricata`?
Btw. there's a PR to address `s390x`: #7666

Cogitri: luajit

Cogitri:
7ecfea85438f2defe607195522d8807a50239370

PR #11605, opened 2019-09-22
testing/fractal: upgrade to 4.2.1

J0WI:


Cogitri:
#11634

J0WI:
@Cogitri what about `x86`?

Cogitri:
Didn't enable it since Rust is crashing often with gtk-rs stuff on everything but x86_64 and ppc64le. Will see how the Fractal upgrade works on the builders and disable armhf, armv7 and aarch64 if they SIGSEGV again.

PR #11621, opened 2019-09-23
main/cjdns: upgrade to 20.4

maxice8:


nmeum:
@kpcyrd you maintain this package, are you ok with the changes purposed here?

PR #11623, opened 2019-09-23
testing/armagetronad:new aport

sodomon2:


testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="x86 x86_64"
+license="GPL"
+depends="ftgl libxml2 sdl_image sdl_mixer"
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc"
+options="!check"
+source="https://launchpad.net/$pkgname/0.2.8/$pkgver/+download/$pkgname-$pkgver.src.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
```
Cogitri: This isn't required as it's the default

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+pkgrel=0
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="x86 x86_64"
+license="GPL"
+depends="ftgl libxml2 sdl_image sdl_mixer"
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc"
+options="!check"
+source="https://launchpad.net/$pkgname/0.2.8/$pkgver/+download/$pkgname-$pkgver.src.tar.gz"
```
Cogitri: Please don't hardcode 0.2.8 here. Maybe `${pkgver%.*.*}` would do the trick.

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Díaz Urbaneja Diego <sodomon2@gmail.com>
+pkgname=armagetronad
+pkgver=0.2.8.3.4
+pkgrel=0
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="x86 x86_64"
+license="GPL"
+depends="ftgl libxml2 sdl_image sdl_mixer"
```
Cogitri: These are autodetected already, see https://cloud.drone.io/alpinelinux/aports/12297, line 2787

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc"
```
Cogitri: WARNING: armagetronad*: Found OpenRC directory (/etc/conf.d or /etc/init.d) but name doesn't end with -openrc

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Díaz Urbaneja Diego <sodomon2@gmail.com>
+pkgname=armagetronad
+pkgver=0.2.8.3.4
+pkgrel=0
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="x86 x86_64"
+license="GPL"
```
Cogitri: This isn't a SPDX compliant license. See https://spdx.org/licenses/ for a list of valid license expressions.

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Díaz Urbaneja Diego <sodomon2@gmail.com>
+pkgname=armagetronad
+pkgver=0.2.8.3.4
+pkgrel=0
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="x86 x86_64"
```
Cogitri: Why doesn't it work on other arches?

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+license="GPL"
+depends="ftgl libxml2 sdl_image sdl_mixer"
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc"
+options="!check"
```
Cogitri: Why are tests disabled? (Leave a comment)

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+    --libexecdir=/usr/libexec \
+    --datadir=/usr/share \
+    --sysconfdir=/usr/share \
+    --sharedstatedir=/usr/share \
+    --localstatedir=/var \
+    --libdir=/usr/share \
+    --includedir=/usr/share \
+    --docdir=/usr/share/doc/armagetronad
+  make
+}
```
Cogitri: The indention isn't correct here. Please use tools like apkbuild-lint from the atools package to avoid these mistakes.

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+license="GPL"
+depends="ftgl libxml2 sdl_image sdl_mixer"
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc"
+options="!check"
+source="https://launchpad.net/$pkgname/0.2.8/$pkgver/+download/$pkgname-$pkgver.src.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+build() {
+cd "$builddir"
```
Cogitri: No need to cd into builddir, abuild does that automatically.

testing/armagetronad/APKBUILD
sodomon2: doesn't work in other arcs, no enrialidad if they work what happens is that I don't do the packages in a chroot and I have the architecture x86_64

testing/armagetronad/APKBUILD
sodomon2: ok

testing/armagetronad/APKBUILD
sodomon2: I don't understand, what's the default value?

sodomon2: Could you tell me more clearly?

Cogitri: builddir is `$srcdr/$pkgname-$pkgver` by default already, so setting the same thing here isn't useful

testing/armagetronad/APKBUILD
Cogitri: CI can test on other arches.

testing/armagetronad/APKBUILD
Cogitri: Then please add a comment to the APKBUILD so other people working on it have am easier time knowing why the tests are disabled

testing/armagetronad/APKBUILD
sodomon2: Can you be more specific?

testing/armagetronad/APKBUILD
sodomon2: Apparently apkbuild-lint does not work, 
I've made several attempts but nothing works so I don't know how to fix the bleeding.

Cogitri:
CI on aarch64 is failing due to `config.guess` being too old. `update_config_guess` in prepare() should take care of that.

sodomon2:
> CI on aarch64 is failing due to `config.guess` being too old. `update_config_guess` in prepare() should take care of that.

You're telling me this would be enough. ?

prepare(){ 
  update_config_guess 
}

Cogitri:
```
prepare(){
    default_prepare
    update_config_guess
}
```

Would be the best. Be mindful to use tabs.

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,46 @@
+# Maintainer: Díaz Urbaneja Diego <sodomon2@gmail.com>
```
Cogitri: Please also add yourself as contributor.

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgname=armagetronad
+pkgver=0.2.8.3.4
+pkgrel=0
+pkgdesc="A Tron Clone in 3D."
+url="http://armagetronad.net/"
+arch="all"
+license="GPL-1.0-only"
+depends=""
```
Cogitri: You can remove this if it's empty

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+makedepends="sdl_image-dev sdl_mixer-dev ftgl-dev libxml2-dev"
+subpackages="$pkgname-doc $pkgname-openrc"
+options="!check" #the tests are deactivated because the make check has not been declared
+source="https://launchpad.net/$pkgname/${pkgver%.*.*}/$pkgver/+download/$pkgname-$pkgver.src.tar.gz"
```
Cogitri: Please write out $pkgname instead of using the variable name here.

sodomon2: I don't get it very well, please be more specific.

Cogitri: instead of writing `$pkgname` in source please write `armagetronad`.

sodomon2: Something like that?
source="https://launchpad.net/armagetronad/${pkgver%.*.*}/$pkgver/+download
/$pkgname-$pkgver.src.tar.gz"

sodomon2: or exchange all $pkgname from source for armagetronad?

Cogitri: Exchange all of them

testing/armagetronad/APKBUILD
```diff
@@ -0,0 +1,47 @@
...
+      --localstatedir=/var \
+      --libdir=/usr/share \
+      --includedir=/usr/share \
+      --docdir=/usr/share/doc/armagetronad
+  make
+}
+
+
+package() {
+    make DESTDIR="$pkgdir" install
```
Cogitri: Seems like there's a racecondition in the build system. Could you please change this to `make -j1 DESTDIR="$pkgdir" install` to see if that fixes it?

sodomon2: ok

Cogitri: Seems to work, thanks

Cogitri:
Pushed with a few changes:

* Indent with tabs. Please always indent with tabs instead of spaces
* Squashed all commits into one

PR #11631, opened 2019-09-23
community/openjdk[9,10,11]: Enhance and unify APKBUILDs

bratkartoffel:
Enhance and unify APKBUILDs for java builds. Add patches to enable 32 bit builds sometime later.
Travis / drone will fail on this, as the builds take too long.

### DIFF
--- a/community/openjdk10/APKBUILD
+++ b/community/openjdk10/APKBUILD
@@ -3,58 +3,56 @@
 pkgname=openjdk10
 pkgver=10.0.2_p13
 _pkgver=${pkgver/_p/+}
-pkgrel=0
+pkgrel=1
 pkgdesc="Oracle OpenJDK 10"
 url="https://hg.openjdk.java.net/jdk-updates/jdk10u"
-arch="all !x86 !armhf !armv7" # openjdk9 is not yet available on 32 bit arches
-license="GPL-2.0 with Classpath"
+arch="all !x86 !armhf !armv7" # oracle dopped support for 32 bit arches, so disable for now
+license="GPL-2.0-with-classpath-exception"
 makedepends="autoconf
-bash
-gawk
-grep
-make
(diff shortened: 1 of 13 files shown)


Cogitri:
> openjdk11: Checking sha512sums...
jdk-11.0.4+11.tar.bz2: OK
build.patch: FAILED

Cogitri:
Mind fixing the merge conflict? Then I'll merge.

bratkartoffel:
Sure, just give me a few minutes

bratkartoffel:
Done, I've also indented the source and makedepends fields and updated the licence to match spdx licence format

Cogitri:
Great, thanks. Just gonna let CI run/timeout again

Cogitri:
I guess we should solve the aarch64 build failure of OpenJFX 11 first

bratkartoffel:
> I guess we should solve the aarch64 build failure of OpenJFX 11 first
OpenJFX is not really related to this package here. I'd suggest the other way round, first enable all archs on openjdk, then taking care of openjfx

ncopa:
Do we really want to enable 32 bit builds? It does not look like Oracle supports 32 bit.
https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms

I know for sure that I will not have the capacity to provide any support for it.

bratkartoffel:
You're right, Natanael, it seems that oracle has finally dropped 32 bit support with all java releases greater than 8. There are other vendors, which still ship 32 bit builds (e.g. [adoptopenjdk](https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=hotspot)) but I think we can leave them out for now until someone wants them.
I'll change the commit and disable 32 bit again, but I still want the patches to be included here. They weren't that easy to write and may help someone later to add 32 bit builds.

PR #11636, opened 2019-09-24
non-free/yed: new aport

durrendal:
New aport for yEd

### DIFF
--- a/non-free/yed/yed.install
+++ b/non-free/yed/yed.install
@@ -1,0 +1,12 @@
+post_upgrade() {
+	post_install $1
+}
+
+post_install() {
+	echo ">>> You have to read and accept the license agreement under"
+	echo ">>> /usr/share/licenses/yed/license.html before using yEd."
+}
+
+op=$1
+shift
+$op $*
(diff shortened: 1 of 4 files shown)


testing/yed/yed.install
```diff
@@ -0,0 +1,13 @@
+post_upgrade() {
+			   post_install $1
+}
+
+post_install() {
+			   echo ">>> You have to read and accept the license agreement under"
+			   echo ">>> /usr/share/licenses/yed/license.html before using yEd."
```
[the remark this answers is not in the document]
durrendal: Right you are, I wasn't thinking when I pushed this last night. I went ahead and moved it from testing to non-free and corrected the new-line issues as well.

non-free/yed/yed.install
```diff
@@ -0,0 +1,12 @@
+post_upgrade() {
+			   post_install $1
+}
+
+post_install() {
+			   echo ">>> You have to read and accept the license agreement under"
+			   echo ">>> /usr/share/licenses/yed/license.html before using yEd."
+}
```
Cogitri: Why indent with so many tabs? :o

durrendal: I think Emacs just hates me. Maxice8 has gotten onto me about this before. Give me just a second and I can fix that.

durrendal: And it should be fixed now!

Cogitri: Seems like it's still like previously in post_upgrade :P

Cogitri:
Thank you :)

PR #11650, opened 2019-09-24
testing/logstash: upgrade to 7.3.2

jjsearle:


maxice8:
@itoffshore

jjsearle:
> I do not think`$pkgname` should be replaced with `logstash`

I have amended this

jjsearle:
> please do not remove `$pkgname` in `source=`
> 
> `$pkgname` is used wherever possible so that if a package name does change it does not require `source` to also be changed

I removed it from the source url because that is what is suggested as proper by maxice8's apkbuild-lint tool https://github.com/maxice8/atools

```
pkgname_used_in_source() {
    [ "$SKIP_PKGNAME_USED_IN_SOURCE" ] && return 0
    [ "$SKIP_AL29" ] && return 0
    scan '^source="[a-z]+://[^"]+\$\{?pkgname\}?' '$pkgname should not be used in the source url' 'AL29' 'MC'
}
```

I am not sure whether this is considered an official standard?

maxice8:
> $pkgname is used wherever possible so that if a package name does change it does not require source to also be changed

embedding the $pkgname into source= means it is harder for people creating their own package based on logstash which is more common than logstash itself renaming

maxice8:
@jjsearle pkgname_used_in_source was contributed by @Ikke

PR #11659, opened 2019-09-24
testing/aws-cli: upgrade to 1.16.244

jjsearle:


subwindow:
This broke aws-cli for us. We now get the following error whenever we use `aws-cli`:

```
Traceback (most recent call last):
  File "/usr/bin/aws", line 19, in <module>
    import awscli.clidriver
ModuleNotFoundError: No module named 'awscli'
```

PR #11672, opened 2019-09-25
testing/libdap: fix depends_dev

hjaekel:


maxice8:
Merged with revision bump

PR #11676, opened 2019-09-26
main/x265: upgrade to 3.2

jjsearle:


maxice8:
Duplicate of https://gitlab.alpinelinux.org/alpine/aports/merge_requests/141

Also doesn't do proper pkgrel bump due to soname changes. Please see above on how to do it properly

PR #11682, opened 2019-09-26
main/perl-html-scrubber: upgrade to 0.19

jjsearle:


maxice8:
Duplicate of https://gitlab.alpinelinux.org/alpine/aports/merge_requests/179 which is the successor of https://github.com/alpinelinux/aports/pull/11619

PR #11685, opened 2019-09-26
main/f2fs-tools: upgrade to 1.13.0

jjsearle:


main/f2fs-tools/APKBUILD
```diff
@@ -1,29 +1,23 @@
...
-install=""
 subpackages="$pkgname-doc $pkgname-dev $pkgname-libs"
-source="https://git.kernel.org/cgit/linux/kernel/git/jaegeuk/f2fs-tools.git/snapshot/$pkgname-$pkgver.tar.gz
+source="https://git.kernel.org/cgit/linux/kernel/git/jaegeuk/f2fs-tools.git/snapshot/f2fs-tools-$pkgver.tar.gz
 	"
-builddir="$srcdir/$pkgname-$pkgver"
 
 prepare() {
 	default_prepare
-	cd "$builddir"
```
maxice8: packages in main/ don't lose this.

https://gitlab.alpinelinux.org/Leo/atools/blob/master/apkbuild-lint#L298

jjsearle: Not sure I understand the rational behind this but I've added them back in.

maxice8: main/ is 2 year support which has versions of  Alpine that  do not support it.

jjsearle: Thanks for the explanation

maxice8:
28f45698a58920de9f35af40ae0a964c37ce9fe9

PR #11686, opened 2019-09-26
main/nginx: fix nginx tmp directory owner

urain39:
This commit/PR fix when user modified default user(and group) in `/etc/nginx/nginx.conf`, then they cannot upload large file(buffer too small) due to `/var/tmp/nginx` owner is still nginx.


urain39:
Related package: nextcloud.

PR #11693, opened 2019-09-26
testing/corehunt: upgrade to 2.7.1

jjsearle:


testing/corehunt/APKBUILD
```diff
@@ -1,15 +1,15 @@
...
+pkgver=2.7.1
 pkgrel=0
 pkgdesc="A file finder utility from the CoreApps family."
 arch="all"
 url="https://gitlab.com/cubocore"
 license="GPL-3.0-or-later"
 depends="qt5-qtsvg"
 makedepends="qt5-qtbase-dev libcprime-dev"
-source="$pkgname-$pkgver.tar.gz::https://gitlab.com/cubocore/$pkgname/-/archive/v$pkgver/$pkgname-v$pkgver.tar.gz"
+source="corehunt-$pkgver.tar.gz::https://gitlab.com/cubocore/corehunt/-/archive/v$pkgver/corehunt-v$pkgver.tar.gz"
```
Cogitri: We usually leave `$pkgname-$pkgver.tar.gz` since it makes sense to always save the archive in the package's name.

jjsearle: Ammended

PR #11694, opened 2019-09-26
testing/libfixposix: Correct Packaging

durrendal:
Add a symlink of libfixposix.3.2.0 -> libfixposix.so

### DIFF
--- a/testing/libfixposix/APKBUILD
+++ b/testing/libfixposix/APKBUILD
@@ -2,7 +2,7 @@
 # Contributor: Will Sinatra <wpsinatra@gmail.com>
 pkgname=libfixposix
 pkgver=0.4.3
-pkgrel=0
+pkgrel=1
 pkgdesc="Thin wrapper over POSIX syscalls"
 url="https://github.com/sionescu/libfixposix"
 arch="all"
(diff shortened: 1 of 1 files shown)


testing/libfixposix/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Contributor: Will Sinatra <wpsinatra@gmail.com>
 pkgname=libfixposix
 pkgver=0.4.3
-pkgrel=0
+pkgrel=1.0
```
jjsearle: Why is the pkgrel `1.0` instead of `1`? I don't think this is valid.

durrendal: I honestly couldn't remember if it was 1.0 or 1, but that's an easy fix at least.

durrendal: Regardless, it's set to 1 instead of 1.0 now.

Cogitri:
Honestly not quite sure what you're trying to do here, why would you want a symlink like that in your `/`? It provides a symlink in `/usr/lib` as it should AFAICS: https://pkgs.alpinelinux.org/contents?branch=edge&name=libfixposix&arch=x86_64&repo=testing

durrendal:
The purpose was actually to create a symlink in /usr/lib/, certain programs look for libfixposix.so specifically, rather than the .3 or .3.2.0 version.

durrendal:
Verified I missed something, .so is provided by libfixposix-dev and I just wasn't well enough aware.

PR #11698, opened 2019-09-26
[3.10] main/python3: upgrade to 3.7.4 and enable optimizations

kr0k0:


subwindow:
What are the chances of this getting pushed to 3.10.x? and/or 490cecc7f6f26962a3f7d229219444efda2a40a5

There's an asyncio fix in 3.7.4 that we really need.

PR #11702, opened 2019-09-26
main/duplicity: upgrade to 0.8.04 and move to python3

kr0k0:


kr0k0:
In this step we should also remove the python2 packages "py2-monotonic", "py-boto" and "py2-fasteners".

maxice8:
rename the py- packages in one commit to py3- and then do the changes to make them py3- in another commit.

maxice8:
Example: https://gitlab.alpinelinux.org/alpine/aports/merge_requests/169

kr0k0:
Thanks maxice8!
Is done.

PR #11704, opened 2019-09-27
main/zfs: upgrade to 0.8.2

Cogitri:


main/zfs-vanilla/APKBUILD
```diff
@@ -3,8 +3,8 @@
 
 # when changing _ver we *must* bump _rel
 _name=zfs
-_ver=0.8.1
-_rel=0
+_ver=0.8.2
+_rel=1
```
maxice8: shouldn't it be 0 ?

Cogitri: How would users be upgraded to the new version then?

HRio:
0.8.2 fixes lots of bugs:
https://github.com/zfsonlinux/zfs/releases/tag/zfs-0.8.2
would be nice to see this in v3.10 as well...

Cogitri:
Sure, once we have had this in edge for a few days, okay?

Cogitri:
@ncopa mind taking a look?

Cogitri:
Can some more people test this? I have spl's `z_rd_int` and `z_wr_iss` take up lots of CPU on my machine since the update.

ncopa:
fixed with e92eab6d0422ed7960bb7800f04f76952c43728a

HRio:
> Sure, once we have had this in edge for a few days, okay?
@Cogitri; I have loaded pinned @edge packages on a v3.10 system, to test on my side as well

Cogitri:
Ah, thank you. Actually kind of forgot about that. Thanks for reminding me!

PR #11724, opened 2019-09-28
[3.10] main/mariadb: upgrade to 10.3.18

kr0k0:


ncopa:
Fails with:
```
-- Performing Test have_CXX__Wdeprecated_declarations - Success
CMake Error at cmake/plugin.cmake:284 (MESSAGE):
  Plugin AUTH_PAM cannot be built
```

PR #11727, opened 2019-09-28
 testing/proj: upgrade to 6.2.0 

hjaekel:


Cogitri:
postgis and mapnik don't need to be rebuilt?

hjaekel:
You are right, postgis needs to be rebuilt. mapnik is broken and deactivated.

PR #11728, opened 2019-09-29
testing/bird: Fix bird init script.

epoelke:
The bird init script had a bug in it that prevented restart form working
properly.  Gentoo's bird init script seems much more mature with proper
pid file handling, updated openrc environment variables, and support for
reload via openrc.


Cogitri:
@maxice8

maxice8:
LGTM but didn't test.

PR #11731, opened 2019-09-30
testing/next: new aport

durrendal:
New aport for Atlas Engineering's NEXT browser

### DIFF
--- a/testing/next/APKBUILD
+++ b/testing/next/APKBUILD
@@ -1,0 +1,26 @@
+# Maintainer: Will Sinatra <wpsinatra@gmail.com>
+# Contributor: Will Sinatra <wpsinatra@gmail.com>
+# Contributor: Benjamin Buccianti <bebuccianti@gmail.com>
+pkgname=next
+pkgver=1.3.2
+pkgrel=0
+pkgdesc="Atlas Engineer Next Browser"
+url="https://next.atlas.engineer"
+arch="x86_64" #SBCL currently only built for x86_64, to be expanded once compatibility is met.
+license="BSD-3-Clause"
+source=next.zip::"https://github.com/atlas-engineer/next/archive/$pkgver.zip"
+makedepends="git sbcl webkit2gtk-dev"
+depends="webkit2gtk glib-networking gsettings-desktop-schemas libfixposix libfixposix-dev gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly libffi libffi-dev"
+options="!check !strip" #No testsuite exists; Stripping the package causes the
+# package to build, but the next browser is unfunctional, dropping to an SBCL repl
+# instead of properly running.
+
+
+build() {
+	make all
+}      
+
+package() {
+	make PREFIX=/usr DESTDIR="$pkgdir" install
+}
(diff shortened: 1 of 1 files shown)


Ambrevar:
Thanks!
Maybe add a comment why the package should not be stripped.

durrendal:
A solid suggestion, I went ahead and put a brief comment.

testing/next/APKBUILD
```diff
@@ -0,0 +1,24 @@
...
+pkgver=1.3.2
+pkgrel=0
+pkgdesc="Atlas Engineer Next Browser"
+url="https://next.atlas.engineer"
+arch="x86_64"
+license="BSD-3-Clause"
+source="https://github.com/atlas-engineer/next/archive/$pkgver.zip"
+makedepends="git sbcl webkit2gtk-dev"
+depends="webkit2gtk glib-networking gsettings-desktop-schemas libfixposix libfixposix-dev gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly libffi libffi-dev"
+options="!check !strip" #No testsuite exists; Stripping the package causes the package to build, but the next browser is unfunctional, dropping to an SBCL repl instead of properly running.
```
Cogitri: Please break the line after 80 chars.

testing/next/APKBUILD
```diff
@@ -0,0 +1,24 @@
...
+# Contributor: Will Sinatra <wpsinatra@gmail.com>
+# Contributor: Benjamin Buccianti <bebuccianti@gmail.com>
+pkgname=next
+pkgver=1.3.2
+pkgrel=0
+pkgdesc="Atlas Engineer Next Browser"
+url="https://next.atlas.engineer"
+arch="x86_64"
+license="BSD-3-Clause"
+source="https://github.com/atlas-engineer/next/archive/$pkgver.zip"
```
Cogitri: I think you should rename the tarball here

testing/next/APKBUILD
```diff
@@ -0,0 +1,24 @@
+# Maintainer: Will Sinatra <wpsinatra@gmail.com>
+# Contributor: Will Sinatra <wpsinatra@gmail.com>
+# Contributor: Benjamin Buccianti <bebuccianti@gmail.com>
+pkgname=next
+pkgver=1.3.2
+pkgrel=0
+pkgdesc="Atlas Engineer Next Browser"
+url="https://next.atlas.engineer"
+arch="x86_64"
```
Cogitri: Why is it only available on x86_64?

durrendal: Because at this time SBCL is only built for x86_64. I haven't been able to get it to built on the arm platforms due to issues with CLISP. The long term plan is to fix that, then update any existent lisp package with an SBCL dependency to other platforms.

testing/next/APKBUILD
durrendal: This I can do, you'll see a fix inbound for the line break as well.

testing/next/APKBUILD
Cogitri: Please add a comment briefly explaining that to the APKBUILD then

durrendal: All three should now be handled.

PR #11732, opened 2019-09-30
community/engrampa: fix g_ptr_array_copy conflicting types err by pkg upgrade

mksully22:
Build of pkg version 1.22.1 fails with error:
```
glib-utils.h:69:21: error: conflicting types for 'g_ptr_array_copy'
```
Upstream commit https://github.com/mate-desktop/engrampa/commit/66076cf781963cd1b781ba8e42d32efcf8fa9180 fixes the issue but since this commit is integrated into now available 1.23.1 version of source this PR updates to new version.

### DIFF
--- a/community/engrampa/APKBUILD
+++ b/community/engrampa/APKBUILD
@@ -2,7 +2,7 @@
 # Maintainer: Alan Lacerda <alacerda@alpinelinux.org>
 pkgname=engrampa
 pkgver=1.22.1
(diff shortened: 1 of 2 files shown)


ncopa:
i think `1.23` is an unstable release so we may want backport the patch instead.

mksully22:
Sounds good. Backported patches to 1.22.1 instead.

PR #11738, opened 2019-10-01
testing/firefox: upgrade to 69.0.1, enable build on aarch64

colona:
Not too familiar with rust, Void's version of fix-third_party.patch
might be cleaner or more portable:
https://github.com/void-linux/void-packages/blob/3a3d0ef7811ac0c200dd4caba108fa46519dc239/srcpkgs/firefox/patches/initialize_sched_param.patch

Also this change updates the checksum of rt_linux.rs used by cargo for
validation, whereas they disable checksumming entirely:
https://github.com/void-linux/void-packages/blob/3a3d0ef7811ac0c200dd4caba108fa46519dc239/srcpkgs/firefox/template#L42


Cogitri:
wfm on x86_64

testing/firefox/APKBUILD
```diff
@@ -96,11 +96,19 @@ build() {
 	# set rpath so linker finds the libs
 	export LDFLAGS="$LDFLAGS -Wl,-rpath,${_mozappdir}"
 
+	case "$CARCH" in
+		x86_64)
+			arch_config="--disable-elf-hack"
+			;;
+		*)
+			arch_config=""
```
Cogitri: Just don't set arch_config

testing/firefox/APKBUILD
```diff
@@ -96,11 +96,19 @@ build() {
 	# set rpath so linker finds the libs
 	export LDFLAGS="$LDFLAGS -Wl,-rpath,${_mozappdir}"
 
+	case "$CARCH" in
+		x86_64)
+			arch_config="--disable-elf-hack"
```
Cogitri: Why do this?

testing/firefox/APKBUILD
```diff
@@ -130,7 +138,9 @@ build() {
 		--with-system-png \
 		--with-system-zlib \
 		--with-clang-path=/usr/bin/clang \
-		--with-libclang-path=/usr/lib
+		--with-libclang-path=/usr/lib \
+		\
+		$arch_config
```
Cogitri: Why the newline?

testing/firefox/fix-rust-target.patch
```diff
@@ -1,10 +1,12 @@
...
-@@ -280,9 +280,11 @@
+@@ -280,9 +280,13 @@
              die("Don't know how to translate {} for rustc".format(
                  host_or_target.alias))
  
-+        fixed_alias = rustc_target.alias.replace("-unknown-", "-alpine-").replace("-gnu", "-musl")
++        fixed_alias = rustc_target.alias.replace("-gnu", "-musl")
++        if host_or_target.cpu != 'aarch64':
++            fixed_alias = fixed_alias.replace("-unknown-", "-alpine-")
 +
```
Cogitri: Maybe setting `RUST_TARGET` appropriately would work too? Something like:

`RUST_TARGET=${CTARGET/alpine/unknown}`

colona: That looks promising indeed, I think we'll be able to get rid of that patch altogether. I'll wait for the build to finish and post second revision.

Any idea of something to test to detect if we have rust for alpine-linux or unknown-linux in the build environment? That would probably be better than hardcoding it in the APKBUILD depending on the arch.

testing/firefox/APKBUILD
colona: I don't know what that flag does but it doesn't exist for aarch64 (it was just on the configure args list before, when only x86_64 was enabled), do you see a better way to do this?

Cogitri: No, not really, but it'd be worth adding a comment to explain this here then :)

testing/firefox/fix-rust-target.patch
Cogitri: Right now it's unknown-linux on everything but x86_64. Please just hardcode it for now :)

Cogitri: You can find the triplets in ./community/rust/APKBUILD

colona: Updated version seems to work on aarch64 without `fix-rust-target.patch`, let me know if you can test it on x86_64.

Cogitri: Just tested, sadly it doesn't work, sorry about that. I guess we'll have to keep using the patch.

(Also, how is Mozilla of all people not following the standard of RUST_TARGET...)

Cogitri: With that reversed it works for me on x86_64

colona: Alright thanks for testing, I reverted to patching `rust.configure`. I'll try to debug this and make `RUST_CONFIGURE` work when I have a x86_64 setup.

PR #11743, opened 2019-10-01
main/vte3: upgrade to 0.58.0

MartijnBraam:
Upgrades to the latest vte3 version, which switches the build system to meson.

The python dependency also doesn't seem to be needed anymore.

This is a requirement for gnome-terminal 3.34

### DIFF
--- a/main/vte3/APKBUILD
+++ b/main/vte3/APKBUILD
@@ -1,43 +1,45 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=vte3
-pkgver=0.56.3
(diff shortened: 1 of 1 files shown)


Cogitri:
#11234 :/

MartijnBraam:
Oh wtf, I didn't find that one when searching for vte3 in the github PR list...

Cogitri:
Yeah, GitHub search isn't that good :c

Cogitri:
But we can go with this and I'll just change my PR to only bump gnome-terminal.

Cogitri:
fwiw 0.58.1 has been released

PR #11749, opened 2019-10-01
community/apk-file: move from testing

drebrez:


Cogitri:
Mind force pushing to restart CI?

Cogitri:
Also, maintainer ping @drebrez

drebrez:
@Cogitri I've rebased on master and force pushed but the build still fails and I can't understand why

kaey:
CI fails because go build downloads all dependecies, puts them in $GOPATH/mod and then builds. Go unsets `w` flag on downloaded modules to prevent their corruption (from text editors or running tests).
Easy fix there would be to use vendor dir
`export GOFLAGS="-mod=vendor"` or `make GOFLAGS=-mod=vendor`

Cogitri:
Seems like it's still failing with the same error.

drebrez:
@Cogitri Found another commit (https://github.com/alpinelinux/aports/commit/bda9fed96df73fbcb57f586aa89d859541653e43) fixing the same error and applied the same changes, now build works fine.

PR #11758, opened 2019-10-02
main/raspberrypi: leave *.a files

kytart:
*.a files were deleted from the package but some *.pc files reference them, hence they have to be included

Original issue: https://gitlab.alpinelinux.org/alpine/aports/issues/9400


kytart:
@fabled

Cogitri:
I suppose you should add a $pkgname-static subpackage then.

kytart:
@Cogitri good point, I'll do that. Thanks

kytart:
@Cogitri now it should be ok

Cogitri:
lgtm, but I don't have the bits for main/

PR #11763, opened 2019-10-03
community/cachefilesd: switch from DNOTIFY to INOTIFY

tomalok:
This changes `cachefilesd` to use INOTIFY instead of the deprecated DNOTIFY to trigger cleanup of its cache graveyard directory.  Resolves https://gitlab.alpinelinux.org/alpine/aports/issues/9953.

Additionally fixes an issue which would skip cleaning up graveyard files with names greater than 2 characters in length and start with `..` (i.e. `..foo`).

### DIFF
--- a/community/cachefilesd/APKBUILD
+++ b/community/cachefilesd/APKBUILD
@@ -1,8 +1,9 @@
+# Contributor: Jake Buchholz <tomalok@gmail.com>
 # Contributor: Francesco Colista <fcolista@alpinelinux.org>
 # Maintainer: Francesco Colista <fcolista@alpinelinux.org>
 pkgname=cachefilesd
 pkgver=0.10.10
-pkgrel=0
+pkgrel=1
 pkgdesc="Userspace daemon acting as a backend for FS-Cache"
 url="http://people.redhat.com/~dhowells/fscache/"
 arch="all"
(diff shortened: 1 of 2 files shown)


tomalok:
It should be noted that the current `cachefilesd` package currently does not work because Alpine kernels (except for `ppc64le` and `s390x`)have DNOTIFY disabled.

Cogitri:
Is this upstream? I'd rather not have such significant patch downstream.

tomalok:
I agree.

I've reached out to the most recent maintainer (as of 3-ish years ago), David Howells <dhowells@redhat.com>, for guidance as to how/where to submit patch and/or git pull requests to what appears to be the main repo... https://git.kernel.org/pub/scm/linux/kernel/git/dhowells/cachefilesd.git

I have not, as of yet, received a reply.

Another alternative might be to create a separate fork, say `cachefilesd-inotify` and package that separately, until a time if/when a patch is accepted?

Cogitri:
Maybe try on https://www.redhat.com/mailman/listinfo/linux-cachefs I guess?

tomalok:
good call, I'll follow up there.  in the meantime I'll get my forked version "out there" in gitlab or github.

tomalok:
patched `cachefilesd` fork: https://gitlab.com/tomalok/cachefilesd-inotify

I've also sent a message to the linux-cachefs mailing list.

Cogitri:
Nice, should we wait on this then, or make a separate package?

tomalok:
Withdrawing in favor of #11810 -- new `cachefilesd-inotify` package until such time as the INOTIFY patch is applied to `cachefilesd`'s mainline codebase.

PR #11775, opened 2019-10-04
testing/gradio: new aport

drebrez:


Cogitri:
Thanks! Should this be added to the gnome-apps subpackage too?

drebrez:
> Thanks! Should this be added to the gnome-apps subpackage too?

Don't know, if you want I can create a PR to add it. Let me know.

Cogitri:
I guess we should. I'll make a PR for it since I have a few more packages to be added to it :)

PR #11777, opened 2019-10-04
[main][grub] Fix file /boot/grub/locale/C.gmo not found

graywolf:
Due to alpine using LANG=C.UTF-8 /etc/grub.d/00_header gets confused and
tries to load C.gmo during boot:

error: file /boot/grub/locale/C.gmo not found

which fails because it does not exist. This commit fixes this (harmless)
warning by correctly setting LANG variable.

Fixes https://gitlab.alpinelinux.org/alpine/aports/issues/10816

### DIFF
--- a/main/grub/APKBUILD
+++ b/main/grub/APKBUILD
@@ -180,6 +180,7 @@ package() {
 
 	mkdir -p "$pkgdir"/etc/default/
 	cat >"$pkgdir"/etc/default/grub <<-EOF
+	LANG=C
 	GRUB_DISTRIBUTOR="Alpine"
 	GRUB_TIMEOUT=2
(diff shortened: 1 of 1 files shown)


graywolf:
1. Rebased on master
2. No feedback received
3. CIs pass
4. ping @ncopa

ncopa:
Thanks for the patch. I think I would prefer to patch grub in this case:
```sh
if [ "x${LANG%.UTF-8}" != "xC" ] &&  [ "x${LANG}" != "x" ]; then
```

ncopa:
this was fixed with 29f0a4dcb6162bdabcda3a888230dd2d1d737090. Thanks!

PR #11786, opened 2019-10-05
community/gnome-music: add gstreamer dependency

MartijnBraam:
Without the volume and the limiter plugin from gstreamer gnome-music
will crash on launch.


Cogitri:
Nitpick, but maybe say that you're adding `gst-plugins-good` to dependencies, not gstreamer.

MartijnBraam:
Renamed the commit

PR #11792, opened 2019-10-05
community/john enabled jumbo version of jtr

stef:
core john is not very useful, the jumbo version is much more versatile in todays world.


stef:
`abuild -r` works for me and runs the check() correctly. why does it fail on the ci builders?

PR #11793, opened 2019-10-05
community/runc: upgrade to 1.0.0_rc9

tomalok:
Primarily fixes CVS-209-16884.  See https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc9 for more information.

### DIFF
--- a/community/runc/APKBUILD
+++ b/community/runc/APKBUILD
@@ -5,13 +5,13 @@ pkgname=runc
 pkgdesc="CLI tool for spawning and running containers according to the OCI specification"
 url="https://www.opencontainers.org"
 
-_commit=3e425f80a8c931f88e6d94a8c831b9d5aa481657
-pkgver=1.0.0_rc8
-pkgrel=2
+_commit=d736ef14f0288d6993a1845745d6756cfc9ddd5a
+pkgver=1.0.0_rc9
+pkgrel=0
 
-#_ver=v${pkgver/_rc/-rc}
+_ver=v${pkgver/_rc/-rc}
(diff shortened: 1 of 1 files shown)


community/runc/APKBUILD
```diff
@@ -5,13 +5,13 @@ pkgname=runc
...
-pkgrel=2
+_commit=d736ef14f0288d6993a1845745d6756cfc9ddd5a
+pkgver=1.0.0_rc9
+pkgrel=0
 
-#_ver=v${pkgver/_rc/-rc}
+_ver=v${pkgver/_rc/-rc}
 # if we're building against an explicit commit beyond pkgver, use this instead:
-_ver=${_commit}
+#_ver=${_commit}
```
andypost: why that changed?

tomalok: Prior to 1.0.0_rc9, we were building from a commit *after* 1.0.0_rc8's, and needed to pull the commit tarball.

With 1.0.0_rc9, we're back to pulling the version tarball, and so we've toggled back to that.

community/runc/APKBUILD
```diff
@@ -22,7 +22,7 @@ builddir="$srcdir/src/github.com/opencontainers/runc"
 options="!check"
 
 # secfixes:
-#   1.0.0_rc8-r2:
+#   1.0.0_rc9:
```
andypost: not sure it needs update because previous version has this already fixed, but as I understand it needs another rebuild

tomalok: related to the `go-1.13.1` rebuild you mentioned above?

andypost: I guessed)

PR #11799, opened 2019-10-06
community/eigen do install pkgconf eigen3.pc with eigen-dev

stef:
some packages try to find eigen using pkgconf for those this change is important.


Cogitri:
Can't we use upstream's CMake build system instead of doing that ourselves?

stef:
sure we could, we could even enable tests for "check", i have my own version of eigen: https://github.com/aports-ugly/aports/blob/master/ugly/eigen/APKBUILD

but i thought this sed approach is more lightweight. we can go either way.

Cogitri:
I don't see the point of going "lightweight" in the build process at the expense of maintenance cost and having a fragile build :)

PR #11804, opened 2019-10-06
community/eigen install eigen3.pc, use cmake

stef:
this PR replaces https://github.com/alpinelinux/aports/pull/11799 which installs eigen3.pc. however this PR uses the cmake build system, which also enables us to enable `check()`.

### DIFF
--- a/community/eigen/APKBUILD
+++ b/community/eigen/APKBUILD
@@ -1,36 +1,55 @@
+# Contributor: stef <l0ls0fo2i@ctrlc.hu>
 # Contributor: Bradley J Chambers <brad.chambers@gmail.com>
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
 pkgname=eigen
 pkgver=3.3.7
-pkgrel=0
+pkgrel=1
 pkgdesc="Eigen is a C++ template library for linear algebra"
 url="http://eigen.tuxfamily.org/index.php?title=Main_Page"
 arch="noarch"
-options="!check"  # Headers-only: no tests are possible.
 license="MPL-2.0"
 depends=""
-makedepends=""
+options="!check"  # checks take a long time, and tend to fail randomly
+makedepends="cmake openblas-dev suitesparse-dev mpfr-dev fftw-dev boost-dev gmp-dev glu-dev freeglut-dev"
 install=""
 subpackages="$pkgname-dev $pkgname-doc"
-source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz"
+source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz
+	fftw-test-fix.patch"
 builddir="$srcdir/$pkgname-$pkgname"
 
 build() {
-        return 0
+	mkdir -p build
+	cd build
+	# needs -std so c11 tests are being built
+	CXXFLAGS="-std=c++11" \
+	cmake -DCMAKE_INSTALL_PREFIX=/usr "$builddir"
+	make
 }
 
 prepare() {
-        mv "$builddir"-* "$builddir"  # directory name contains hash
-        default_prepare
+	mv "$builddir"-* "$builddir"  # directory name contains hash
+	default_prepare
+}
+
+check() {
+	cd "$builddir/build"
+	make check
+}
+
+dev() {
+	default_dev
+	mkdir -p ${subpkgdir}/usr/share/cmake/Modules
+	mv $pkgdir/usr/share/eigen3/cmake/*.cmake ${subpkgdir}/usr/share/cmake/Modules
 }
 
 package() {
-        mkdir -p "$pkgdir"/usr/include/eigen3
-        cp -r "$builddir"/Eigen "$pkgdir"/usr/include/eigen3
-        cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
-        cp "$builddir"/signature_of_eigen3_matrix_library "$pkgdir"/usr/include/eigen3
-        mkdir -p "$pkgdir"/usr/share/doc/$pkgname
-        cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	cd "$builddir/build"
+	make DESTDIR="$pkgdir" install
+	mkdir -p "$pkgdir"/usr/share/doc/$pkgname
+	cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	mkdir -p $pkgdir/usr/lib/pkgconfig
+	mv $pkgdir/usr/share/pkgconfig/eigen3.pc $pkgdir/usr/lib/pkgconfig
 }
 
-sha512sums="34cf600914cce719d61511577ef9cd26fbdcb7a6fad1d0ab8396f98b887fac6a5577d3967e84a8f56225cc50de38f3b91f34f447d14312028383e32b34ea1972  eigen-3.3.7.tar.gz"
+sha512sums="34cf600914cce719d61511577ef9cd26fbdcb7a6fad1d0ab8396f98b887fac6a5577d3967e84a8f56225cc50de38f3b91f34f447d14312028383e32b34ea1972  eigen-3.3.7.tar.gz
+c1f9d7d8c9025b4b2b3a679f80519e494f206d7bee232cd10dd45df454b1ce6697858547bc0d956a5818481d344948704db8dbb39d04ac69ff7de99961492384  fftw-test-fix.patch"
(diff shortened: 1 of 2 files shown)


stef:
@Cogitri the build is now taking much longer since it also executes the `check` which is quite long.

Cogitri:
Seems like tests don't build. pass `-DBUILD_TESTING=FALSE` to CMake to disable building them.

stef:
yeah the tests take to long and the CI builders kill them after some time.

Cogitri:
Huh? It fails to build them on any platform on Drone CI.

stef:
that is super strange indeed, it says:
> make: *** No rule to make target 'check'.  Stop.
> >>> ERROR: eigen: check failed

however if i do `abuild check` locally it actually can start the `make check` from the `check()` rule.

Cogitri:
Odd. Maybe you can run CTest directly?

stef:
i'll try an abuild -r

stef:
indeed, abuild -r does fail.

stef:
yeah, so after passing `-DBUILD_TESTING=FALSE` to cmake the makefile lacks the `check` target. so that explains it. so i removed the `-DBUILD_TESTING=FALSE` and also the `check()` function.

stef:
now it succeeds

stef:
i disabled checks in options since it seems that tests randomly fail, also the tests take too long to build and thus are killed on travis.

Cogitri:
Don't worry about timeouts on Travis, it won't timeout on the builders where it actually counts. The random failures are of course bad though.

community/eigen/APKBUILD
```diff
@@ -1,36 +1,49 @@
 # Contributor: Bradley J Chambers <brad.chambers@gmail.com>
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+# Contributor: stef <l0ls0fo2i@ctrlc.hu>
 pkgname=eigen
 pkgver=3.3.7
-pkgrel=0
+pkgrel=2
```
Cogitri: Why bump pkgrel twice?

community/eigen/APKBUILD
```diff
@@ -1,36 +1,49 @@
 # Contributor: Bradley J Chambers <brad.chambers@gmail.com>
 # Maintainer: Bradley J Chambers <brad.chambers@gmail.com>
+# Contributor: stef <l0ls0fo2i@ctrlc.hu>
```
Cogitri: Add yourself to the top, not the bottom of the contributor graph.

community/eigen/APKBUILD
```diff
@@ -1,36 +1,49 @@
...
-        return 0
+  cd "$builddir"
+  mkdir -p build
+  cd build
+  CXXFLAGS="-std=c++11" \
+  cmake -DCMAKE_INSTALL_PREFIX=/usr "$builddir"
+  make
 }
 
+
```
Cogitri: Why the random newline here?

community/eigen/APKBUILD
```diff
@@ -1,36 +1,49 @@
...
+source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz
+	fftw-test-fix.patch"
 builddir="$srcdir/$pkgname-$pkgname"
 
 build() {
-        return 0
+  cd "$builddir"
+  mkdir -p build
+  cd build
+  CXXFLAGS="-std=c++11" \
```
Cogitri: Why add this?

community/eigen/APKBUILD
```diff
@@ -1,36 +1,49 @@
...
 install=""
 subpackages="$pkgname-dev $pkgname-doc"
-source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz"
+source="$pkgname-$pkgver.tar.gz::http://bitbucket.org/eigen/$pkgname/get/$pkgver.tar.gz
+	fftw-test-fix.patch"
 builddir="$srcdir/$pkgname-$pkgname"
 
 build() {
-        return 0
+  cd "$builddir"
```
Cogitri: No need to do this, abuild does that for you

community/eigen/fftw-test-fix.patch
```diff
@@ -0,0 +1,11 @@
+diff -urw src/eigen-eigen/unsupported/test/FFTW.cpp eigen-eigen/unsupported/test/FFTW.cpp
+--- src/eigen-eigen/unsupported/test/FFTW.cpp	2018-12-11 17:57:55.000000000 +0000
++++ eigen-eigen/unsupported/test/FFTW.cpp	2019-10-06 22:32:27.239613362 +0000
+@@ -8,6 +8,7 @@
+ // with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ 
+ #include "main.h"
++#include <stdint.h>
```
Cogitri: Is upstream aware of this?

community/eigen/APKBUILD
stef: otherwise cmake disables:
```
-- C++11:             OFF
```

stef: i mean c++11 tests of course.

community/eigen/fftw-test-fix.patch
stef: yes i sent them the patch via mail, since their bugtracker has registrations disabled and pull requests they only accept on bitbucket.

community/eigen/APKBUILD
stef: fixed.

community/eigen/APKBUILD
stef: good to know. fixed.

community/eigen/APKBUILD
stef: no reason. removed.

community/eigen/APKBUILD
stef: historical remnant. removed.

stef:
i just force pushed a version where all review comments are addressed.

community/eigen/APKBUILD
Cogitri: Alright, please add a comment then.

community/eigen/APKBUILD
```diff
@@ -1,36 +1,46 @@
...
-        cp -r "$builddir"/Eigen "$pkgdir"/usr/include/eigen3
-        cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
-        cp "$builddir"/signature_of_eigen3_matrix_library "$pkgdir"/usr/include/eigen3
-        mkdir -p "$pkgdir"/usr/share/doc/$pkgname
-        cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	cd "$builddir/build"
+	make DESTDIR="$pkgdir" install
+	mkdir -p "$pkgdir"/usr/share/doc/$pkgname
+	cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	install -Dm0644 eigen3.pc "$pkgdir/usr/lib/pkgconfig/eigen3.pc"
```
Cogitri: Huh, CMake doesn't automatically install this?

stef: uh, yeah, it does. however it gets bundled in the main pkg, not in -dev. i changed it so it gets into -dev. however the main package is now empty. but i dunno how to disable the creation of an empty apk file. would be nice to suppress it.

community/eigen/APKBUILD
stef: done.

community/eigen/APKBUILD
```diff
@@ -1,36 +1,51 @@
...
-        cp -r "$builddir"/unsupported "$pkgdir"/usr/include/eigen3
-        cp "$builddir"/signature_of_eigen3_matrix_library "$pkgdir"/usr/include/eigen3
-        mkdir -p "$pkgdir"/usr/share/doc/$pkgname
-        cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	cd "$builddir/build"
+	make DESTDIR="$pkgdir" install
+	mkdir -p "$pkgdir"/usr/share/doc/$pkgname
+	cp -r "$builddir"/doc/ "$pkgdir"/usr/share/doc/$pkgname
+	mkdir -p ${pkgdir}-dev/usr/share/cmake/Modules
+	mv $pkgdir/usr/share/eigen3/cmake/*.cmake ${pkgdir}-dev/usr/share/cmake/Modules
```
Cogitri: Doesn't the -dev package do this automatically? If not, this is the wrong approach: You'd do a `dev() {}` function, which runs `default_dev` and then installs these files into `"$subpkgdir"`, not `$pkgdir-dev`.

community/eigen/APKBUILD
Cogitri: It's fine as-is.

stef: really? i would expect cmake and pkgconf things go into -dev, makes little sense to have them in the main pkg.

community/eigen/APKBUILD
stef: aaah. thanks! will do that.

community/eigen/APKBUILD
Cogitri: Ah sorry if I wasn't clear, I mean it's fine `as-is` == `the way you're doing it right now`. It'll be clearer what you're trying to do once you've moved things into the `dev()` function :)

community/eigen/APKBUILD
stef: done. seems to work as expected (locally).

community/eigen/APKBUILD
stef: i've moved it to a 'dev()' function and force pushed an update.

PR #11805, opened 2019-10-06
[WIP] community/nextcloud-client: upgrade to 2.6.0

Cogitri:
@PureTryOut mind giving it a shot?


PureTryOut:
You should add `graphviz` to `$makedepends` as Doxygen complains about it missing.

Seems to work fine. Version info in the application itself says it's "2.6.0git", not sure if we can somehow change that to just "2.6.0?

Cogitri:
Thanks for taking a look, will look into fixing that later today.

PR #11808, opened 2019-10-06
community/libreoffice: upgrade to 6.3.2

J0WI:


Cogitri:
Ah, didn't see this. Could you please do some style fixes (move `depends_dev` to makedepends since there's no -dev subpackage) and also do what `apkbuild-lint` wants?

Cogitri:
Also, should we maybe add an `install_if` to libreoffice-gnome which automatically installs it if gtk3 is installed?

Cogitri:
Thank you

J0WI:
> Also, should we maybe add an `install_if` to `libreoffice-gnome` which automatically installs it if `gtk3` is installed?

Does Alpine Linux have a guideline for this? Maybe people just want to use the cli, even if `gtk3` is present on their system.

Cogitri:
Possibly, it just annoyed me a bit that I thought we didn't have the gtk3 version enabled at first only to find out it's just not installed after a few hours.

J0WI:
btw `libreoffice` depends on `libreoffice-common` which depends on `gtk+3.0`. So `libreoffice-gnome` would always be installed. Maybe `*-desktop` would be a better condition?

Cogitri:
A -desktop doesn't really make sense since LO also offers a Qt5 and KDE5 interface so the desktop subpackage would be kind of confusing IMHO.

Cogitri:
Maybe it'd make sense to not split the gnome (gtk3) since it's pretty small (compared to LO's size) and GTK3 is the default interface

Cogitri:
Merged as c1bf371130005e3e21c7d9aa815aa7a33645bb03, e8aab477f3c8c716ddd3dde2e36c9154b7bd6e86, thank you

PR #11809, opened 2019-10-06
community/john enabled jumbo version of jtr

stef:
the packaged version of john is only the core version although it downloads the jumbo version. this change enables building the full jumbo patch.

### DIFF
--- a/community/john/APKBUILD
+++ b/community/john/APKBUILD
@@ -4,31 +4,23 @@
 pkgname=john
 pkgver=1.9.0
 _pkgrel=jumbo-1
-pkgrel=1
+pkgrel=2
 pkgdesc="John the Ripper password cracker"
 url="https://www.openwall.com/john"
 arch="all !s390x !x86 !ppc64le"
 license="GPL-2.0-or-later"
 depends="nss"
-options="!strip"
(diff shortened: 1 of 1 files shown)


stef:
the built john executable looks for its configuration files in /usr/share/john, which is not available during build/check. shall i disable the check()?

stef:
i disabled checks in options. since john expects all configfiles to be available in the install destination which is not the case during build/check.

community/john/APKBUILD
```diff
@@ -4,31 +4,23 @@
...
-	case "$CARCH" in
-		aarch64) _arch=arm64le;;
-		arm*) _arch=arm32le;;
-		ppc64le) _arch=ppc64;;
-		x86) _arch=x86-any;;
-		x86_64) _arch=x86-64;
-	esac
-
-	make linux-$_arch
+   ./configure --prefix=/usr
```
ncopa: seem to be whitespace damage here.

community/john/APKBUILD
```diff
@@ -37,11 +29,18 @@ check() {
 }
 
 package() {
+	cd "$builddir/src"
+   make DESTDIR="$pkgdir" install
```
ncopa: more white space damage, including trailing whitespace

PR #11841, opened 2019-10-08
community/kafkacat: move from testing

jchipmunk:
kafkacat has been in testing for some time and it is quite stable


Cogitri:
I think we usually do `<newrepo>/<package>: move from <oldrepo>`

jchipmunk:
> I think we usually do `<newrepo>/<package>: move from <oldrepo>`

I updated commit message

PR #11839, opened 2019-10-07
main/ceph: upgrade to 14.2.4

J0WI:

PR #11852, opened 2019-10-09
main/ceph: Move ceph from community

iggy:
Move ceph from community to main

This move should allow us to enable ceph support in Qemu


iggy:
Build timed out because c++ (and massive project)

iggy:
@fcolista @TBK @sanmai-NL  moving `oath-toolkit` `leveldb` `xmlstarlet` `xmlsec` to main so I can move `ceph` to main so we can turn on `ceph` support in `qemu`

tcely:
Why does qemu require packages in main?

ncopa:
> Why does qemu require packages in main?

Because we support main for 2 years but community for 6 months (latest release). It wouldn't make sense to support a package if we also don't support its dependencies.

tcely:
See also #11839 and #11661

PR #11866, opened 2019-10-09
main/vte3: upgrade to 0.58.1

MartijnBraam:
@Cogitri Actually updates to 0.58.1 instead of only updating the commit message...


Cogitri:
Thank you :)

Cogitri:
There's 0.58.2 now

PR #11868, opened 2019-10-09
testing/telegraf: upgrade to 1.12.3

kaey:


kaey:
Test failure looks like a flake, restarting build should fix it

Cogitri:
Please just force push to restart the build :)

kaey:
Ok this is not a flake, but nothing changed in telegraf code since 1.12.2 in `http_response` package.
This probably caused by changed builder config, either firewall doesn't allow access to `127.0.0.11` or loopback is not configured as `127.0.0.1/8`.

kaey:
Test failure is not reproduced on my rpi3

Cogitri:
Ah, I guess that might be on port restrictions in Docker, could you disable that test please?

PR #11879, opened 2019-10-11
testing/wayfire: new aport

Danct12:
Wayfire is a Compiz-like Wayland compositor.

Signed-off-by: Danct12 <danct12@disroot.org>

### DIFF
--- a/testing/wayfire/APKBUILD
+++ b/testing/wayfire/APKBUILD
@@ -1,0 +1,27 @@
+# Contributor: Danct12 <danct12@disroot.org>
+# Maintainer: Danct12 <danct12@disroot.org>
+pkgname=wayfire
+pkgver=0.3
+pkgrel=0
+pkgdesc="Compiz-clone Wayland compositor"
+url="https://wayfire.org"
+arch="all"
+license="MIT"
+install="$pkgname.post-install"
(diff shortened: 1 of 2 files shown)


testing/wayfire/wayfire.post-install
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+echo
```
Cogitri: Wouldn't just a "\n" instead of echo here do the trick?

testing/wayfire/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+pkgdesc="Compiz-clone Wayland compositor"
+url="https://wayfire.org"
+arch="all"
+license="MIT"
+install="$pkgname.post-install"
+makedepends="wlroots-dev glm-dev wf-config-dev meson ninja cmake cairo-dev libexecinfo-dev"
+options="!check" # no testsuite
+subpackages="$pkgname-dev"
+source="https://github.com/WayfireWM/$pkgname/releases/download/$pkgver.0/$pkgname-$pkgver.tar.xz"
+builddir="$srcdir/$pkgname-$pkgver"
```
Cogitri: No need for that, that's the default already.

PR #11887, opened 2019-10-12
testing/fs-uae: upgrade to 3.0.0

tanertas:


testing/fs-uae/APKBUILD
```diff
@@ -14,8 +14,6 @@ source="https://fs-uae.net/stable/${pkgver}/${pkgname}-${pkgver}.tar.gz"
 options="!check"
 
 build() {
-	export CC=clang
-	export CXX=clang++
```
Cogitri: I guess we won't need `clang-dev` as makedepend then?

tanertas: Indeed. Fixed.

testing/fs-uae-launcher/APKBUILD
```diff
@@ -1,20 +1,18 @@
...
 pkgname=fs-uae-launcher
-pkgver=2.8.4
+pkgver=3.0.0
 pkgrel=0
 pkgdesc="FS-UAE emulator launcher."
 url="https://fs-uae.net/launcher"
 arch="noarch !armhf !armv7 !aarch64 !s390x"  # fs-uae not avail
 license="GPL-2.0-only"
-depends="fs-uae~$pkgver py3-qt5"
+depends="py3-qt5 py3-sip-pyqt5"
```
Cogitri: I don't know the package, but does the launcher really not need the actual emulator?

testing/fs-uae/APKBUILD
Cogitri: Thanks

testing/fs-uae-launcher/APKBUILD
tanertas: It causes other problems during package build. This is better in my opinion. If you think otherwise, feel free to add it for yourself.

Cogitri: define "other problems" please.

tanertas: It couldn't detect proper version of the fs-uae thus unable to build the package. This shouldn't be happen since fs-uae is not a build dependency.

Is this definition enough? Or should I file a full report about the problem sir?

Cogitri: Just depend on fs-uae=$pkgver and it should work.

tanertas: Ok. Done.

testing/fs-uae-launcher/APKBUILD
```diff
@@ -1,20 +1,18 @@
...
 pkgname=fs-uae-launcher
-pkgver=2.8.4
+pkgver=3.0.0
 pkgrel=0
 pkgdesc="FS-UAE emulator launcher."
 url="https://fs-uae.net/launcher"
 arch="noarch !armhf !armv7 !aarch64 !s390x"  # fs-uae not avail
 license="GPL-2.0-only"
-depends="fs-uae~$pkgver py3-qt5"
+depends="fs-uae=$pkgver py3-qt5 py3-sip-pyqt5"
```
Cogitri: Err...sorry, should be `~$pkgver` not `=$pkgver`, my bad. Seems to work for me then.

tanertas: I did it with updated commit message.

PR #11888, opened 2019-10-12
testing/mame: upgrade to 0.214

tanertas:
* Use bundled asio due to incompability with system asio (mamedev/mame#5721)
* Temporary disable on aarch64 due to buld system problem (mamedev/mame#5741)
* Remove obselete patches


Cogitri:
Why not use the system asio?

Cogitri:
Also seems to fail on CI on aarch64

tanertas:
> Why not use the system asio?

https://github.com/mamedev/mame/issues/5721

> Also seems to fail on CI on aarch64

https://github.com/mamedev/mame/issues/5741

Cogitri:
Please add those issues to the commit message for further reference then.

tanertas:
> Please add those issues to the commit message for further reference then.

No I won't. I don't think that it is necessary at all. If you feel otherwise, feel free to do it for yourself.

tanertas:
@prspkt @maxice8 @andypost can you take a look at this? This package is already 7 version behind.

andypost:
@tanertas as I see it still fails on aarch64

tanertas:
> @tanertas as I see it still fails on aarch64

Ok. PR revised. I temporary disabled on aarch64.

PR #11891, opened 2019-10-14
testing/h1-cli: Upgrade to 1.8.0

ad-m:


ad-m:
We have held the package without updates at Alpine for some time to get information from users about any problems. No problems were reported, so a full range of Alpine Linux users can enjoy the new release.

PR #11892, opened 2019-10-14
main/sudo: security upgrade to 1.8.28

J0WI:


J0WI:
All tests pass but the target fails.

J0WI:
The relevant fix is [here](https://www.sudo.ws/repos/sudo/rev/83db8dba09e7) (+[tests](https://www.sudo.ws/repos/sudo/rev/db06a8336c09)), but the patch does not apply cleanly.

ncopa:
this was fixed with 87cda3c142a7ab418e7a25d79f226a7470ec3867, which also fixes the test suite

PR #11897, opened 2019-10-14
main/linux-vanilla: enable CONFIG_IIO

Cogitri:
It's required for most orientation sensors and for iio-sensor-proxy (as the name
suggests)

I guess we should also enable this for other arches, @ncopa?


Cogitri:
I think this is required for iio-sensor-proxy (or at least it makes it work for me) @z3ntu

main/linux-vanilla/APKBUILD
```diff
@@ -7,7 +7,7 @@ case $pkgver in
 	*.*.*)	_kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=0
+pkgrel=1
```
Cogitri: This would of course necessitate a bump of other modules, but it'd be best if we could just apply the configuration differences with the next kernel bump

z3ntu:
@Cogitri I've tested iio-sensor-proxy with a custom kernel on an aarch64 board which presumably had CONFIG_IIO enabled.

Cogitri:
Ah yes, that'd make sense.

ncopa:
I guess we want this for x86, aarch64 and arm* too?

Cogitri:
Yes, but I'm not sure how to configure the kernel for those 😅

Cogitri:
Enabled it on aarch64 and x86 too. arm* already has it enabled AFAICS

Cogitri:
Mind taking a look, @ncopa?

PR #11904, opened 2019-10-15
main/aria2: fix aria2-daemon crashes

urain39:


main/aria2/aria2.initd
```diff
@@ -28,5 +28,8 @@ depend() {
 start_pre() {
 	local _logfile=$(sed -En 's|^log=([^#]+)|\1|p' "$cfgfile")
 
+	su - "$command_user" -s /bin/sh -c "
+		touch ~/aria2.session
+	"
```
Cogitri: IMHO `touch` and then `chown` would be better.

urain39:
Sounds good, so i decided to move this changes to daemon function due to `chmod` in a init.d script may be user doesn't want.

Cogitri:
### Sorry?

PR #11907, opened 2019-10-15
main/gcr: upgrade to 3.34.0

Cogitri:
- enable tests
- adopt maintainership


Cogitri:
Mind taking a look, @ncopa?

PR #11909, opened 2019-10-16
testing/libtss2: move from community

whooo:


asac:
hi @whooo  we discussed that libtss2  is a dupe of tpm2-tss package so we should rather remove it (if it really got landed).

asac:
lets please close this PR... rather make a clean PR against tpm2-tss package that does not move/rename the package itself .... then we can discuss if this is better in testing or main etc... but given that @Cogitri moved tpm2-tss to community for a reason i believe for now thats a great place until we move this to main eventually.

Cogitri:
Yes, we need tpm2-tss for fwupd which is in community.

PR #11911, opened 2019-10-16
main/wget: add iri support

SuperSandro2000:
I need support for remote encoding in wget and I thought this is useful in upstream.

Currently this command ``wget --remote-encoding=utf-8 http://example.com`` returns:
````
This version does not have support for IRIs
````
After this patch it works as intended.

### DIFF
--- a/main/wget/APKBUILD
+++ b/main/wget/APKBUILD
@@ -1,15 +1,16 @@
+# Contributor: Sandro Jäckel <sandro.jaeckel@posteo.de>
 # Contributor: Sergei Lukin <sergej.lukin@gmail.com>
 # Contributor: Carlo Landmeter <clandmeter@gmail.com>
 # Maintainer: Carlo Landmeter <clandmeter@gmail.com>
 pkgname=wget
 pkgver=1.20.3
-pkgrel=0
+pkgrel=1
 pkgdesc="A network utility to retrieve files from the Web"
 url="https://www.gnu.org/software/wget/wget.html"
 arch="all"
 license="GPL-3.0-or-later"
 depends=""
-makedepends="openssl-dev perl"
+makedepends="libidn2-dev openssl-dev perl"
 checkdepends="perl-http-daemon"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by SuperSandro2000** [comment]
Sadly I missed the libc6-compat dependency. Maybe this should be split into an wget-full package or is this still acceptable?

**Comment by Cogitri** [comment]
Why does this depend on libc6-compat? It seems to work without it for me even. I personally don't mind adding the libidn2 dep to wget since it's not that heavy and we use curl by default anyway.

**Comment by SuperSandro2000** [comment]
Without it I got the following error:
````
wget
Error relocating /usr/bin/wget: secure_getenv: symbol not
````
Should have documented it but I forgot.

File: main/wget/APKBUILD [hunk 1]
```diff
@@ -1,15 +1,16 @@
...
 pkgver=1.20.3
-pkgrel=0
+pkgrel=1
 pkgdesc="A network utility to retrieve files from the Web"
 url="https://www.gnu.org/software/wget/wget.html"
 arch="all"
 license="GPL-3.0-or-later"
-depends=""
-makedepends="openssl-dev perl"
+depends="libc6-compat"
```
andypost: I build it with `libintl` dependency instead of that and it works for me

SuperSandro2000: Building it with ``libintl`` does not work.
````
configure: WARNING: *** libpsl was not found. Fallback to builtin cookie checking.
configure: WARNING: *** LIBIDN2 was not found. You will not be able to use IDN2008 support
````
At runtime it does work but libintl is way bigger than libc6-compat

maxice8:
secure_getenv is a symbol now in musl 1.1.24

SuperSandro2000:
> secure_getenv is a symbol now in musl 1.1.24

Do I need to add a version constraint to ``depends``?

maxice8:
No

SuperSandro2000:
Removed libc-compat

PR #11912, opened 2019-10-16
testing/rclone: upgrade to 1.49.5

SuperSandro2000:


SuperSandro2000:
When I tested this locally I always got this:
````
$ abuild build
# github.com/rclone/rclone
loadinternal: cannot find runtime/cgo
````
Not sure why. Anyone any idea?

Cogitri:
Hm, I think that warning is normal for us

SuperSandro2000:
Maybe someone should take a look at this if it is fatal or can be ignored in abuild.

Cogitri:
I don't really know Go, maybe you could ask over on IRC?

PR #11915, opened 2019-10-16
testing/libexosip2: new aport

dyfet:
This includes libexosip2 and libosip2 as a dependency and also a new aport:

    testing/libexosip2: new aport
    
    http://savannah.nongnu.org/projects/exosip
    Extended osip2 library

    testing/libosip2: new aport
    
    https://www.gnu.org/software/osip
    oSIP is an implementation of SIP

### DIFF
--- a/testing/libosip2/APKBUILD
+++ b/testing/libosip2/APKBUILD
@@ -7,16 +7,11 @@ pkgdesc="oSIP is an implementation of SIP"
 url="https://www.gnu.org/software/osip/"
 arch="all"
 license="LGPL-2.1-or-later"
-makedepends="
-	autoconf
-	automake
-	libtool
-"
+makedepends="autoconf automake libtool"
 options="libtool"
 subpackages="$pkgname-dev $pkgname-doc"
-source="
-	http://ftp.gnu.org/gnu/osip/$pkgname-$pkgver.tar.gz
-	"
+source="http://ftp.gnu.org/gnu/osip/libosip2-$pkgver.tar.gz"
+
 prepare() {
 	default_prepare
 	./autogen.sh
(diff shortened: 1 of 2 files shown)


testing/libosip2/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+license="LGPL-2.1-or-later"
+makedepends="
+	autoconf
+	automake
+	libtool
+"
+options="libtool"
+subpackages="$pkgname-dev $pkgname-doc"
+source="
+	http://ftp.gnu.org/gnu/osip/$pkgname-$pkgver.tar.gz
```
Cogitri: Please don't use $pkgname in source URLs

testing/libosip2/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+	"
```
Cogitri: super nitty, but please leave a newline here.

dyfet: I took some time to look at other packages.  I used to have a patch and didn't need that anymore, so I actually should have consolidated the source line back into one line.  I also now see other packages do indeed typically use name-$pkgvar.tar.gz in source url's.  Especially to do my first packages, it did help me to compare more carefully with existing packages.

Cogitri: I can recommend you to use `apkbuild-lint` from the `atools ` package which will spot these things for you :)

dyfet: Actually apkbuild-lint returned no warnings...so it did not check for that.   I originally had a bunch of cd $builddir which it did warn me about, and a few other minor things, though.

Cogitri:
Thanks. For the future: Please just amend these changes to the respective commit :)

PR #11916, opened 2019-10-17
testing/thunderbird: update to 60.9.0

jbenden:
This patch upgrades the version of Thunderbird to 60.9.0.

Local testing against Alpine edge.

### DIFF
--- a/testing/thunderbird/rust-cssparser.patch
+++ b/testing/thunderbird/rust-cssparser.patch
@@ -1,0 +1,92 @@
+This is a fix for rust version 1.38.0, which causes the warning below to become an error.
+
+backport of:
+
+From 3c98d22c5de3b696bf1fde2b6c90069812312aa6 Mon Sep 17 00:00:00 2001
+From: Simon Sapin <simon.sapin@exyr.org>
+Date: Tue, 23 Apr 2019 13:47:25 +0200
+Subject: [PATCH] Fix a future-compat warning
+
+```
+warning[E0506]: cannot assign to `self.input.cached_token` because it is borrowed
+   --> src/parser.rs:591:17
+    |
+566 |     pub fn next_including_whitespace_and_comments(&mut self) -> Result<&Token<'i>, BasicParseError<'i>> {
+    |                                                   - let's call the lifetime of this reference `'1`
+...
+579 |             Some(ref cached_token)
+    |                  ---------------- borrow of `self.input.cached_token` occurs here
+...
+591 |                 self.input.cached_token = Some(CachedToken {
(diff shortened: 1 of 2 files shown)


jbenden:
The build will always exceed the limits on Travis, unless `ccache` is enabled for it; however, then cache size may become an issue.

testing/thunderbird/rust-cssparser.patch
```diff
@@ -0,0 +1,90 @@
...
+...
+591 |                 self.input.cached_token = Some(CachedToken {
+    |                 ^^^^^^^^^^^^^^^^^^^^^^^ assignment to borrowed `self.input.cached_token` occurs here
+...
+603 |         Ok(token)
+    |         --------- returning this value requires that `self.input.cached_token.0` is borrowed for `'1`
+    |
+    = warning: this error has been downgraded to a warning for backwards compatibility with previous releases
+    = warning: this represents potential undefined behavior in your code and this warning will become a hard error in the future
+```
```
Cogitri: Why do we need to backport a fix for a warning?

Cogitri:
Don't worry about Travis failing.

testing/thunderbird/rust-cssparser.patch
jbenden: Because it's an error condition with rust version 1.38.0. It simply wouldn't build without applying the patch.

Cogitri:
It'd be worth mentioning that then

testing/thunderbird/rust-cssparser.patch
jbenden: FWIW: upstream already knows and fixed it, but this patch is from them to fix the temporary issue at hand.

Cogitri:
Yes, but please put a comment in the patch so people know what it's for.

jbenden:
I'd rather comment in the `APKBUILD`, if required, because the patch isn't mine and would cause a mismatch if someone tried verifying it against the original one.

Also, I feel the patch does share enough information in the title and body, if you're in the realm of being the unlucky one whom stumbles on the specific error.

Thoughts, comments, and suggestions are welcomed,
-Joe

Cogitri:
> I'd rather comment in the APKBUILD, if required, because the patch isn't mine and would cause a mismatch if someone tried verifying it against the original one.

Just put it in the patch header, that's standard practice

> Also, I feel the patch does share enough information in the title and body, if you're in the realm of being the unlucky one whom stumbles on the specific error.

The patch says it's fixing a warning, so I'd intuitively just drop the thing since it supposedly is'nt required for us then, wasting a good amount of time until I notice that it is required after all.

jbenden:
Ok, I've placed a note at the top of the patch file.

Cogitri:

Cogitri:
That was meant to say a thank you :)

PR #11922, opened 2019-10-17
community/qt5-qtwebengine fixed UB in ProcessMetrics::GetMallocUsage

stef:
The .patch file introduced an #ifdef condition, but neglected to also return an int, which this function required and thus the compiler replaced the whole function with `ud2` (on x86_64) which triggers an SIGILL and thus made the qtwebengine renderer always crash. This function seems to be optional, since for fuchsia it is ok to return 0 and also there is variants for windows which return 0. The problem is that this function relies on mallinfo which musl does not implement and seems more buggy than useful (see also https://www.openwall.com/lists/musl/2018/01/17/2)

The bug gets triggered in `src/3rdparty/chromium/content/renderer/render_thread_impl.cc:1828`, the problematic `ProcessMetrics::GetMallocUsage` can be found in `src/3rdparty/chromium/base/process/process_metrics_posix.cc:95`

The fixed package has been tested against a small testprogram and against qutebrowser-1.8.1 which both work as expected.


ncopa:
Does this solve https://gitlab.alpinelinux.org/alpine/aports/issues/10405?

Would be nice to have a `fixes #10405` in commit message if so.

stef:
i added that fixes message, but in retrospect i dunno, i guess it does fix it, but this is speculation nothing more.

ncopa:
Lets assume it fixes it. We can always reopen if it doesn't.

PR #11923, opened 2019-10-17
testing/libyang: update to 1.0-r4

cfra:
This is an update to the latest released version of libyang.

See: <https://github.com/CESNET/libyang/releases/tag/v1.0-r4>


Cogitri:
Please make the commit msg to `update to 1.0_p4` since that's the pkgver you've set. It's a bit confusing otherwise since we use `-r$num` as `pkgrel` in the APKBUILD.

cfra:
Yea, I didn't come up with the upstream versioning scheme. :wink:

Should be fixed now.

Cogitri:
Thank you

PR #11924, opened 2019-10-17
main/gtk+3.0: add patch fixing crashing applications w/ stylus on Wayland

Cogitri:


Cogitri:
Will wait for the next gtk version.

PR #11925, opened 2019-10-17
community/wxgtk fix bad include path in pkgconf to the correct location

stef:
during install the include dir gets moved from /usr/include/wx-3.0 to /usr/include, however the pkgconf does not get updated and points to the wrong location, this PR fixes this.


stef:
btw whoever bumped this to 3.1.2 please reconsider this, since according to the website itself:
> Please notice that while 3.1.1 is officially a “development” version because it is not fully compatible with the “stable” 3.0.x

ncopa:
> btw whoever bumped this to 3.1.2 please reconsider this, since according to the website itself:
> 
> > Please notice that while 3.1.1 is officially a “development” version because it is not fully compatible with the “stable” 3.0.x

Yeah, this was a mistake. I need to revert it. Sorry.

stef:
thanks a lot!

PR #11931, opened 2019-10-19
testing/ldc: upgrade to 1.18.0

Geod24:
LDC is upgraded to [v1.18.0](https://github.com/ldc-developers/ldc/releases/tag/v1.18.0), released yesterday. DUB, the D package manager, is added, as well as `rdmd`, which is a handy and simple build tool.

LDC relies on `ld.gold` by default, so it was added as a dependency.

CC @JohanEngelen @ZombineDev


Geod24:
Ah, the CI builds them out of order (or more precisely: in alphabetical order), so `dtools` and `dub` will have to be in another PR.

testing/ldc/APKBUILD
```diff
@@ -10,73 +10,85 @@ url="https://github.com/ldc-developers/ldc"
...
 subpackages="$pkgname-runtime $pkgname-doc $pkgname-static"
 provides="dlang-compiler"
-source="https://github.com/ldc-developers/ldc/releases/download/v$pkgver/ldc2-$pkgver-alpine-linux-x86_64.tar.xz
+source="https://github.com/ldc-developers/ldc/releases/download/v$pkgver/ldc-$pkgver-src.tar.gz
 		ldc2.package.conf"
 builddir="$srcdir/"
 
 build() {
-	rootdir="$srcdir/ldc2-$pkgver-alpine-linux-x86_64/"
+	rootdir="$srcdir/ldc-$pkgver-src/"
```
Cogitri: Can't you just set "$builddir" to this now?

testing/ldc/APKBUILD
```diff
@@ -10,73 +10,85 @@ url="https://github.com/ldc-developers/ldc"
...
-	libroot="$builddir/ldc-runtime-build/lib/"
+	rootdir="$srcdir/ldc-$pkgver-src/"
 
-	echo 'import std.stdio; void main () { try throw new Exception(null); catch (Exception) writeln("Hello World!"); }' > hello_world.d
-	if [ "$($rootdir/bin/ldmd2 -conf='' -I$rootdir/import/ldc -I$rootdir/import/ -L-L$libroot -defaultlib=phobos2-ldc,druntime-ldc -L-lexecinfo -L-rpath=$libroot -run hello_world.d)" != "Hello World!" ]; then
-		return 1
-	fi
+    cd "$rootdir"
+    make test
 	:
```
Cogitri: I don't think this is required anymore

testing/ldc/APKBUILD
```diff
@@ -10,73 +10,85 @@ url="https://github.com/ldc-developers/ldc"
...
 
-	# Install the static runtime libraries
-	for libn in "libdruntime-ldc" "libdruntime-ldc-debug" "libphobos2-ldc" "libphobos2-ldc-debug";
+	# Install Druntime import files
+	install -D "$rootdir/runtime/druntime/src/object.d" "$importdir/object.d"
+	for importpkg in "core" "etc" "gc" "ldc" "rt";
 	do
-		install -D "$builddir/ldc-runtime-build/lib/$libn.a" "$pkgdir/usr/lib/$libn.a"
+		cp -R "$rootdir/runtime/druntime/src/$importpkg/" "$importdir/"
 	done
```
Cogitri: There's no handy `make install` which does this for us?

Cogitri:
```
Unable to find executable: python

99% tests passed, 20 tests failed out of 1651

Total Test time (real) = 3607.78 sec

The following tests FAILED:
	299 - std.datetime.systime (Failed)
	300 - std.datetime.timezone (Failed)
	397 - std.stdio (Failed)
	699 - std.datetime.systime-debug (Failed)
	700 - std.datetime.timezone-debug (Failed)
	797 - std.stdio-debug (Failed)
	1099 - std.datetime.systime-shared (Failed)
	1100 - std.datetime.timezone-shared (Failed)
	1197 - std.stdio-shared (Failed)
	1499 - std.datetime.systime-debug-shared (Failed)
	1500 - std.datetime.timezone-debug-shared (Failed)
	1597 - std.stdio-debug-shared (Failed)
	1620 - druntime-test-coverage (Failed)
	1626 - druntime-test-exceptions (Failed)
	1636 - druntime-test-profile (Failed)
	1638 - druntime-test-shared (Failed)
	1640 - druntime-test-stdcpp (Failed)
	1648 - dmd-testsuite-debug (Failed)
	1650 - dmd-testsuite (Failed)
	1651 - lit-tests (Not Run)
Errors while running CTest
make: *** [Makefile:119: test] Error 8
```

Geod24:
Indeed some unittests are failing. I need to go through them and see if they are legit failure (hence need more work upstream) or down to my package.

Geod24:
The `std.datetime` failure was a missing dependency, so now reduced to 12 failures:

```console
The following tests FAILED:
--
4913 | 397 - std.stdio (Failed)
4914 | 797 - std.stdio-debug (Failed)
4915 | 1197 - std.stdio-shared (Failed)
4916 | 1597 - std.stdio-debug-shared (Failed)
4917 | 1620 - druntime-test-coverage (Failed)
4918 | 1626 - druntime-test-exceptions (Failed)
4919 | 1636 - druntime-test-profile (Failed)
4920 | 1638 - druntime-test-shared (Failed)
4921 | 1640 - druntime-test-stdcpp (Failed)
4922 | 1648 - dmd-testsuite-debug (Failed)
4923 | 1650 - dmd-testsuite (Failed)
4924 | 1651 - lit-tests (Failed)
4925 | Errors while running CTest
```

The failing `std.stdio` is [this](https://github.com/ldc-developers/phobos/blob/b60fda9da5744385ad0dfbb42cec4555a81fc078/std/stdio.d#L1251).

The other tests all have the same error message:
```console
1612: /home/effortman/aports/ldc/src/ldc-1.18.0-src/bin/ldmd2   -w -I../../src -I../../import -Isrc -defaultlib= -debuglib= -dip1000 -L-lpthread -L-lm -L/home/effortman/aports/ldc/src/ldc-1.18.0-src/lib/libdruntime-ldc.a -O -release -of/home/effortman/aports/ldc/src/ldc-1.18.0-src/runtime/druntime-test-aa/test_aa src/test_aa.d
1612: /usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld.gold: fatal error: /home/effortman/aports/ldc/src/ldc-1.18.0-src/runtime/druntime-test-aa/test_aa: No error information
```

So all failing tests seems to be related to exceptions.

Cogitri:
Hm, I guess it'd be best to ask upstream about that?

Geod24:
> The failing `std.stdio` is [this](https://github.com/ldc-developers/phobos/blob/b60fda9da5744385ad0dfbb42cec4555a81fc078/std/stdio.d#L1251).

So I figured this one out. The test ultimately calls `fseeko` with `whence = 3`. 
Now what is 3 ? On Glibc the whence values are:
```
/* The possibilities for the third argument to `fseek'.
--
These values should not be changed.  */
#define SEEK_SET        0        /* Seek from beginning of file.  */
#define SEEK_CUR        1        /* Seek from current position.  */
#define SEEK_END        2        /* Seek from end of file.  */
#ifdef __USE_GNU
# define SEEK_DATA        3        /* Seek to next data.  */
# define SEEK_HOLE        4        /* Seek to next hole.  */
#endif
```

So 3 is a Linux-only value, introduced in 3.1. On the other hand, on Musl, the definitions are missing:
```
#undef SEEK_SET
#undef SEEK_CUR
#undef SEEK_END
#define SEEK_SET 0
#define SEEK_CUR 1
#define SEEK_END 2
```

That value was introduced with the assumption that `3` was always invalid. I'll fix it, and will add a patch for LDC. See the  [man](http://man7.org/linux/man-pages/man2/lseek.2.html), for more informations.

testing/ldc/APKBUILD
Geod24: You mean the column (`:`) right ? Actually I was wondering what it was for... Does it have a name ? Google hasn't been helpful.

Cogitri: Not quite sure what it's name, but you use it when you have an empty function, since this would be a syntax error:

```
check() {

}
```

whereas

```
check() {
    : 
}
```

isn't. You could put a `true` in there too, I suppose.

Cogitri:
Nice, good job at figuring that out! :)

Cogitri:
Please request a review from me/ping me once you've figured the tests out :)

Cogitri:
BTW, are you interested in maintaining more of the D ecosystem? I'm thinking about packaging gtkd and tilix in a bit but I don't really know about D :)

If you want to you can just disable the few failing tests for now

Geod24:
> If you want to you can just disable the few failing tests for now

I actually found out that `make test` is not the way upstream test things. So I update it to [something closer](https://github.com/ldc-developers/ldc/blob/a0787d956d6345b25ed18ba21535893d7f5d9a66/.travis.yml#L113-L126).

One of the problem I see is that LDC decided to go with `ld.gold` by default and this fails on Alpine. However, switching back to BFD is not an issue in this case. I haven't identified the reason for the failure yet, but it might be due to [musl design](https://wiki.musl-libc.org/design-concepts.html).
Anyway I'll try to get the package working soon.

> BTW, are you interested in maintaining more of the D ecosystem? I'm thinking about packaging gtkd and tilix in a bit but I don't really know about D :)

I don't mind maintaining what I use. I have 2 more packages coming, which I can be maintainer for, one including some tools (rdmd, ddemangle), and one using `dub`, the D package manager. I'd be happy to lend you a hand with tilix / gtkd packaging, but you probably want to have `dub` available first, but as I don't currently use those two I won't be able to keep up with the updates.
Note: D usually does not package libraries the way C/C++ does. The approach we use is closer to Go/Rust.

Cogitri:
> One of the problem I see is that LDC decided to go with ld.gold by default and this fails on Alpine. However, switching back to BFD is not an issue in this case. I haven't identified the reason for the failure yet, but it might be due to musl design.

Ah yes, I think ld.gold didn't work for Void Linux musl either.

> Note: D usually does not package libraries the way C/C++ does. The approach we use is closer to Go/Rust.

Ah, but I think it's better for distros to package it like C/C++ does it and all distros which I know which package D do it like that too.

Geod24:
@Cogitri : Green (at last) on Drone. I mistakenly pushed some experiment on gcc earlier hence the `main` label (the bot adds it but does not remove it).

PR #11934, opened 2019-10-20
main/ghostscript: upgrade to 9.50

J0WI:


J0WI:
Packages depending on `ghostscript` should be rebuild against the new version.

Cogitri:
Why? The soname hasn't changed.

J0WI:
https://www.ghostscript.com/doc/9.50/News.htm
https://abi-laboratory.pro/index.php?view=objects_report&l=ghostscript&v1=9.27&v2=9.50

There are some breaking changes. Especially `imagemagick` makes use of the changed `-dSAFER` option.

Cogitri:
Ugh and they didn't increase the soname, that's a bit annoying...

Anyway, mind bumping the deps in this PR then too?

J0WI:
I'm a bit worried that this will lead in merge conflicts, but if you're going to merge it right now I can do this.

Cogitri:
Well, I don't have access to main, but I'll review it tomorrow and hopefully someone with the magic bits can take a look soon then.

J0WI:
Done

Cogitri:
FWIW one of imagemagick6's tests fails:

`FAIL: tests/validate-composite.tap 1`

PR #11942, opened 2019-10-21
main/nodejs: upgrade to 12.13.0

timbru31:
Node.js v12.13.0 ('Erbium') is the new LTS branch: https://nodejs.org/en/blog/release/v12.13.0/

This PR uses the build steps from the nodejs-current package as this already is updated to build v12 versions of Node.js

### DIFF
--- a/main/nodejs/APKBUILD
+++ b/main/nodejs/APKBUILD
@@ -44,7 +44,7 @@
 pkgname=nodejs
 # Note: Update only to even-numbered versions (e.g. 6.y.z, 8.y.z)!
 # Odd-numbered versions are supported only for 9 months by upstream.
-pkgver=10.16.3
+pkgver=12.13.0
 pkgrel=0
 pkgdesc="JavaScript runtime built on V8 engine - LTS version"
 url="https://nodejs.org/"
@@ -52,10 +52,9 @@ arch="all !mips64 !mips64el"
 license="MIT"
 depends="ca-certificates"
 depends_dev="libuv"
-# gold is needed for mksnapshot
-makedepends="$depends_dev python2 openssl-dev zlib-dev libuv-dev linux-headers
-	paxmark binutils-gold http-parser-dev ca-certificates c-ares-dev"
-subpackages="$pkgname-dev $pkgname-doc npm::noarch"
+makedepends="linux-headers python2 paxmark
+	zlib-dev libuv-dev openssl-dev c-ares-dev nghttp2-dev"
+subpackages="$pkgname-dev $pkgname-doc"
 provides="nodejs-lts=$pkgver"  # for backward compatibility
 replaces="nodejs-current nodejs-lts"  # nodejs-lts for backward compatibility
 source="https://nodejs.org/dist/v$pkgver/node-v$pkgver.tar.gz
@@ -68,7 +67,7 @@ prepare() {
 	default_prepare
 
 	# Remove bundled dependencies that we're not using.
-	rm -rf deps/http_parser deps/openssl deps/uv deps/zlib
+	rm -rf deps/cares deps/openssl deps/uv deps/zlib
 }
 
 build() {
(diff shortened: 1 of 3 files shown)


timbru31:
```
Package nghttp2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `nghttp2.pc'
to the PKG_CONFIG_PATH environment variable
Package 'nghttp2', required by 'virtual:world', not found
INFO: Using floating patch "tools/icu/patches/64/source/common/putil.cpp" from "tools/icu"
gyp: /home/buildozer/aports/main/nodejs/src/node-v12.13.0/deps/http_parser/http_parser.gyp not found (cwd: /home/buildozer/aports/main/nodejs/src/node-v12.13.0) while loading dependencies of /home/buildozer/aports/main/nodejs/src/node-v12.13.0/node.gyp while trying to load /home/buildozer/aports/main/nodejs/src/node-v12.13.0/node.gyp
INFO: Using floating patch "tools/icu/patches/64/source/i18n/dtptngen.cpp" from "tools/icu"
Error running GYP
```

I'll take a look

timbru31:
Drone succeeded, TravisCI timed out (as always)

main/nodejs/APKBUILD
```diff
@@ -44,18 +44,17 @@
...
 arch="all !mips64 !mips64el"
 license="MIT"
 depends="ca-certificates"
 depends_dev="libuv"
-# gold is needed for mksnapshot
-makedepends="$depends_dev python2 openssl-dev zlib-dev libuv-dev linux-headers
-	paxmark binutils-gold http-parser-dev ca-certificates c-ares-dev"
-subpackages="$pkgname-dev $pkgname-doc npm::noarch"
+makedepends="linux-headers python2 paxmark
+	zlib-dev libuv-dev openssl-dev c-ares-dev nghttp2-dev"
```
Cogitri: Just to make sure, we can't use py3?

main/nodejs/APKBUILD
```diff
@@ -83,11 +82,12 @@ build() {
 		--shared-zlib \
 		--shared-libuv \
 		--shared-openssl \
-		--shared-http-parser \
+		--experimental-http-parser \
```
Cogitri: experimental doesn't sound nice, we can't use an un-experimental version? :/

timbru31:
> Not really involved with nodejs stuff, but can we just bump major versions like that without breaking the stuff that needs it?

Thanks a lot for your question. I'm not 100% sure what the Alpine polciy is, but we simply upgraded the version to the latest LTS in the past:
- https://github.com/alpinelinux/aports/commit/617771fdbcdadc12574efaf287732d1ee9c9418c#diff-37650f4e428b216e6aa5e4ab2c932010 (6 -> 8)
- https://github.com/alpinelinux/aports/pull/5558 (8 -> 10)

Personally, I haven't encountered any issues in the past years of using Node.js when we upgraded the versions. :)

main/nodejs/APKBUILD
timbru31: I've just copied the current `APKBUILD` of https://github.com/alpinelinux/aports/blob/master/community/nodejs-current/APKBUILD since this is/was already upgraded for Node.js v10. AFAIK python2 is still the recommended way to build Node.js: https://github.com/nodejs/node/blob/master/BUILDING.md#note-about-python-2-and-python-3

main/nodejs/APKBUILD
timbru31: Good catch, the new ("experimental") parser lhttp should be the default with Node.js v12: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#semver-major-commits - I'll update the PR

timbru31: Done

main/nodejs/APKBUILD
Cogitri: Well, if py2 is recommended but we can use py3 it'd be best if we used py3 since we want to get py2 out of 3.11 if possible.

timbru31: I've just tried some ways (`alias` and `ln`), but I it seems python2 is still required:
- https://cloud.drone.io/alpinelinux/aports/13105/2/1 (line 124)
- https://cloud.drone.io/alpinelinux/aports/13099/2/1 (line 122)

Cogitri: Ah, that's unfortunate, but thanks for trying

PR #11949, opened 2019-10-22
main/qemu: fix undefined SYNC_MAP on ppc64le

mksully22:
MAP_SYNC is explicitly undefined ppc64le in musl provided headers. This fix defines a MAP_SYNC fallback on ppc64le in a minimally invasive way to support changes introduced in qemu 4.1.0.


mksully22:
ncopa has already fixed in https://git.alpinelinux.org/aports/commit/?id=b10efa20

PR #11960, opened 2019-10-23
testing/vpnc: modernize, build with TLS support, use patched fork

J0WI:
Maybe this can now be moved to community?


nmeum:
I just ran into issues with vpnc as well and came up with a different solution before looking at the PR queue (see #11961).

J0WI:
I'm not too familiar with both of them, but I'd keep them both as other distros do

nmeum:
I would definitely keep openconnect. However, if openconnect is the only use-case for vpnc I would personally rather rely on the vpnc version [advocated by the openconnect developers](https://www.infradead.org/openconnect/vpnc-script.html) instead of relying on some GitHub fork.

Maybe @fcolista (openconnect maintainer) has some thoughts on this, but after all both PRs solve the bug, outlined in #11961, the currently packaged vpnc version has.

nmeum:
```
15:56:24 (fcolista) I don't use openconnect, so I would go with 11960 (which is what the other distro does), if you agree.
```

merged and closed #11961. thanks!

PR #11961, opened 2019-10-23
testing/vpnc: remove aport

nmeum:
upstream is dead, last release is from 2008. Additionally, the script
doesn't seem to work properly. When using it with testing/openconnect
with the vpnc script I get the following error

	Error: either "to" is duplicate, or "uid" is a garbage.

This is due to the fact that /etc/vpnc/vpnc-script calls `ip route add`
with `uid 0` which is incorrect. While this could be fixed, removing the
aport makes more sense to me as upstream is dead. Since nobody seems to
have reported this issue I am assuming that this aport isn't used
anyhow.

The openconnect aport, which previously depended on vpnc, downloads
a custom fork of the vpnc script provided by openconnect (from a git
repository) upstream. Using this script fixes the error mentioned above.

**Alternative solution**: Package a fork. For example:

* http://git.infradead.org/users/dwmw2/vpnc-scripts.git
* https://github.com/streambinder/vpnc
* …


testing/openconnect/APKBUILD
```diff
@@ -2,17 +2,19 @@
...
 arch="all"
 license="LGPL-2.1"
-depends="vpnc iproute2"
+depends="iproute2"
 makedepends="autoconf automake intltool openssl-dev libxml2-dev 
 	krb5-dev lz4-dev libproxy-dev linux-headers stoken-dev
 	pcsc-lite-dev oath-toolkit-dev python-dev"
 subpackages="$pkgname-doc $pkgname-dev $pkgname-openrc"
 source="ftp://ftp.infradead.org/pub/$pkgname/$pkgname-$pkgver.tar.gz
+	$pkgname-vpnc::http://git.infradead.org/users/dwmw2/vpnc-scripts.git/blob_plain/${_vpnc}:/vpnc-script
```
nmeum: file name should be changed to `$pkgname-$pkgver-vpnc`.

nmeum:
Superseded by #11960

PR #11963, opened 2019-10-23
testing/openjfx12: New aport

bratkartoffel:
This PR add OpenJFX to alpine, the latest JavaFX release developed by the community.
Note: Builds on Drone and Travis fail because the build takes extremely long.

### DIFF
--- a/testing/openjfx12/APKBUILD
+++ b/testing/openjfx12/APKBUILD
@@ -1,0 +1,241 @@
+# Contributor: Simon Frankenberger <simon-alpine@fraho.eu>
+# Maintainer: Simon Frankenberger <simon-alpine@fraho.eu>
+pkgname=openjfx12
+pkgver=12.0.2_p1
+_pkgver=${pkgver/_p/+}
+pkgrel=0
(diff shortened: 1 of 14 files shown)


testing/openjfx12/APKBUILD
```diff
@@ -0,0 +1,241 @@
...
+		"$_fromroot"/jmods/javafx.media.jmod \
+		"$_toroot/jmods"
+	mv \
+		"$_fromroot/legal/javafx.media" \
+		"$_toroot/legal/"
+}
+
+_libs() {
+	pkgdesc="$pkgdesc (libs)"
+	depends=""
```
Cogitri: Nitty, but you can remove these empty variables

bratkartoffel: The "root" package has a depends, so i wanted to make sure this subpackage does not inherit the root one. Can I remove this empty assignment?

Cogitri: Hm, I'm like 99% sure that the subpkg won't inherit the deps, but let me check

PR #11967, opened 2019-10-24
main/libssh2: Update release version for CVE patch

mirskifa:
https://github.com/alpinelinux/aports/commit/6c763143a08a56997ee6f88f9329cfc17d6b56b5 introduced a patch for `CVE-2019-17498` but did not bump the package release version.

This makes it very difficult to track whether a fix has been applied.  In addition the secfixes block indicates that -r1 was the intention:
```
# security fixes:
#   1.9.0-r1:
#     - CVE-2019-17498
```

This fix needs merging into all the same branches that https://github.com/alpinelinux/aports/commit/6c763143a08a56997ee6f88f9329cfc17d6b56b5 was pulled into.

### DIFF
--- a/main/libssh2/APKBUILD
+++ b/main/libssh2/APKBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=libssh2
 pkgver=1.9.0
-pkgrel=0
+pkgrel=1
 pkgdesc="library for accessing ssh1/ssh2 protocol servers"
 url="https://libssh2.org/"
 arch="all"


Cogitri:
> This makes it very difficult to track whether a fix has been applied.

It hasn't been applied yet since not increasing the pkgrel means the builders haven't rebuilt it. Thanks for noticing!

mirskifa:
I wasn't able to tell if it wouldn't build or it would just silently replace the existing build.  Good to know which behaviour to expect.

mirskifa:
In terms of things like this being noticed. Would a PR for the build pipeline to highlight changes where the package version+release don't match the `security fixes` block be useful/welcome?  We have automation that highlights when the security fixes block shows fixes are available but are ahead of an installed version so have a particular interest in this info being valid, but in this case it would have caught that the fix block references a version that hasn't been released yet.

Cogitri:
That warning certainly sounds interesting, but it'd be best if it was integrated into our new fancy Gitlab CI pipeline: https://gitlab.alpinelinux.org/alpine/aports

nmeum:
> Would a PR for the build pipeline to highlight changes where the package version+release don't match the `security fixes` block be useful/welcome?

IMHO there is definitely some room to improve tooling regarding the secfixes comments. For instance, I noticed that (occasionally) comments are wrongly indented and don't constitute valid YAML, which is also something that could be detected by an abuild sanitycheck.

Either way, probably makes sense to move this discussion to the alpine-devel mailing list.

mirskifa:
Are you able to give a rough idea of the timeline for picking this commit into the branches that https://github.com/alpinelinux/aports/commit/6c763143a08a56997ee6f88f9329cfc17d6b56b5 went to (3.7 through 3.10 I believe - I'm particularly interested in the timeline for the 3.9 and 3.10 versions containing the CVE fix?)

nmeum:
Just cherry picked the commit, thanks for reminding me.

PR #11971, opened 2019-10-24
main/nss: upgrade to 3.47

J0WI:
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/NSS_3.47_release_notes
https://abi-laboratory.pro/index.php?view=timeline&l=nss


ncopa:
fixed with ca77e4a20b550ac7a55d06e614052a9b85339e4e

PR #11982, opened 2019-10-25
community/nodejs-current: upgrade to 13.0.1

andypost:
Ref https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.0.1

LTS upgrade is https://github.com/alpinelinux/aports/pull/11942


andypost:
Experimental should be removed according to https://github.com/nodejs/node/commit/2cb8f247512d611d3dfe81c2441e8028bfb0e8f0

PR #11984, opened 2019-10-25
community/mpfr3: move from main

J0WI:
This reverts commit 4edefa6fef74b64167cf818f486a337a6cb903d1
`mpfr3` is required by `gcc6` and the whole `openjdk` buildchain.
cc @ncopa


J0WI:
@ncopa would be great if you could have a look at this because it's currently broken.

PR #11994, opened 2019-10-26
testing/vpnc: upgrade to fa0689c

J0WI:


Cogitri:
What's the status of vpnc now? Didn't we want to just package it with openconnect?

J0WI:
https://github.com/alpinelinux/aports/pull/11960#issuecomment-546369294

testing/vpnc/APKBUILD
```diff
@@ -2,16 +2,15 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=vpnc
 pkgver=0.5.3
-_commit=101208be5b74039ea70b9e007ce0d6c9fbe44d82
-pkgrel=5
+_commit=fa0689cc9a67d02a614c7a2406dfe2e794825a31
```
Cogitri: I think it'd be better if the commit was in the pkgver, can we do that?

J0WI: What should the version look like?

Cogitri: The best would be `0_git<date of commit` and is what we use for other packages, bur I'm not sure if people would be upgraded to that to be honest

PR #11995, opened 2019-10-27
testing/mozjs68: new aport

J0WI:


testing/mozjs68/APKBUILD
```diff
@@ -0,0 +1,114 @@
...
+		armv7)
+			export RUST_TARGET="armv7-unknown-linux-musleabihf"
+			;;
+		armhf)
+			export RUST_TARGET="arm-unknown-linux-musleabihf"
+			;;
+		ppc64le)
+			export RUST_TARGET="powerpc64le-unknown-linux-musl"
+			;;
+	esac
```
Cogitri: I think you're using an older version of the fix-rust-target.patch where you don't have to export RUST_TARGET

J0WI: Done. But I still messing with paths

Cogitri:
Please request a review when you're ready

maxice8:
Any progress ? 

@Cogitri would be nice to get this on 3.11 ?

Cogitri:
As of now it's not necessary, but polkit might require it down the line.

J0WI:
It's important because mozjs60 is EOL

Cogitri:
Yes, but if applications don't switch to mozjs68's API we can't do anything about that

PR #11997, opened 2019-10-27
community/firefox-esr: enable on armhf, ppc64le and x86

J0WI:


Cogitri:
Could you maybe file this and thunderbird as MR on https://gitlab.alpinelinux.org/alpine/aports ? There we don't have CI restrictions (as long as you set a higher timeout for your project)

Cogitri:
Oh, you did, oops :sweat_smile:

J0WI:
They are not more reliable. Even the existing architectures do not pass there.

Cogitri:
Huh, odd

J0WI:
Could you try to build them on real hardware and post the logs if they fail?

Cogitri:
Ah, I only have a aarch64 on hand, let me check if I can get the other arches too...

Cogitri:
On armv7: http://arvanta.net/ff-esr.log.gz tested by mps

PR #11999, opened 2019-10-27
testing/py3-catkin-pkg: renamed from py3-catkin-pkg-modules

russkel:
Looks like the project has renamed in pypi https://pypi.org/project/catkin-pkg/ to reflect the name of the github repo and ROS wiki. Changed to suit here as well.


russkel:
@Cogitri even if no one has actually used this? I'm 99.99% sure no one has installed this package.

Cogitri:
Well, better to be safe than sorry I suppose

PR #12002, opened 2019-10-27
testing/py3-pydocstyle: new aport

russkel:


testing/py3-pydocstyle/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+source="$_pkgname-$pkgver.tar.gz::https://github.com/PyCQA/$_pkgname/archive/$pkgver.tar.gz"
+builddir="$srcdir/$_pkgname-$pkgver"
+
+build() {
+	python3 setup.py build
+}
+
+check() {
+	# integration tests call pip which doesn't exist
+    PYTHONPATH="$builddir/build/lib" py.test --deselect src/tests/test_integration.py -vv src/tests
```
Cogitri: fwiw you've accidentally used spaces instead of tabs here.

testing/py3-pydocstyle/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+pkgname="py3-pydocstyle"
+_pkgname="pydocstyle"
+pkgver="4.0.1"
+pkgrel=0
+pkgdesc="Static analysis tool for checking compliance with Python docstring conventions"
+url="https://pydocstyle.org/"
+arch="noarch"
+license="MIT"
+depends="python3 py3-snowballstemmer"
+makedepends="python3 py3-pytest py3-mock py3-six"
```
Cogitri: depends will be installed during build too, so adding python3 to both makedepends and depends does nothing (just having it in depends is sufficient)

Cogitri: This should most likely have py3-setuptools in its makedepends though

testing/py3-pydocstyle/APKBUILD
russkel: Oops. need to fix my editor to show this.

russkel:
Thanks @Cogitri. Fixed as requested.

PR #11591, opened 2019-09-22
community/ocaml: upgrade to 4.09.0

jjsearle:

PR #12008, opened 2019-10-28
community/ocaml: update pkgrel for binutils update

XVilka:
The package requires rebuilding to match the upgrade of binutils

Fixing this bug: https://gitlab.alpinelinux.org/alpine/aports/issues/10888:
```
  so:libbfd-2.32.so (missing):
    required by: ocaml-4.08.1-r0[so:libbfd-2.32.so]
```

### DIFF
--- a/community/ocaml/APKBUILD
+++ b/community/ocaml/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Jakub Jirutka <jakub@jirutka.cz>
 pkgname=ocaml
 pkgver=4.08.1
-pkgrel=0
+pkgrel=1
 pkgdesc="Main implementation of the Caml programming language"
 url="http://ocaml.org/"
 arch="aarch64 ppc64le x86_64"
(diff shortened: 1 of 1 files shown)


Cogitri:
It already correctly finds libbfd as dep, you just need to rebuild ocaml (and you rebuilding the package is what actually fixed it, not adding the binutils-dev dependency)

XVilka:
It uses binutils-dev as a built time dependency, not a runtime, but seems it requires it at runtime stage too. So my change should be fine. Also, why upgrade of binutils didn't trigger rebuild of ocaml package? I think it's because of the missing runtime dependency.

Cogitri:
No, it's simply because someone forgot to bump the pkgrel :)

XVilka:
Ok, what should be done then to fix the existing 4.08.1 release?

russkel:
No point bumping when there is a new version available. Something like this is preferable: https://github.com/alpinelinux/aports/pull/11591

PR #12009, opened 2019-10-28
community/wireguard-go: new aport

kr0k0:
I know, we should use the kernel module on linux systems.
But I think it's useful, when you only have a hosted container somewhere.

### DIFF
--- a/community/wireguard-go/APKBUILD
+++ b/community/wireguard-go/APKBUILD
@@ -1,0 +1,26 @@
+# Contributor: Stefan Reiff <kroko87@hotmail.com>
+# Maintainer: Stefan Reiff <kroko87@hotmail.com>
+pkgname=wireguard-go
+pkgver=0.0.20191012
+pkgrel=0
+pkgdesc="Next generation secure network tunnel: userspace implementation in go"
+arch="all"
+url="https://www.wireguard.com"
+license="GPL-2.0-only"
+makedepends="go"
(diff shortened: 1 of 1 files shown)


community/wireguard-go/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+# Maintainer: Stefan Reiff <kroko87@hotmail.com>
+pkgname=wireguard-go
+pkgver=0.0.20191012
+pkgrel=0
+pkgdesc="Next generation secure network tunnel: userspace implementation in go"
+arch="all"
+url="https://www.wireguard.com"
+# SPDX identifier headers tells us 'GPL-2.0' but considering it
+# is a kernel project i think it is safe to assume it is GPL-2.0-only just
+# like the kernel.
```
Cogitri: You could just grep the source for "at your option" since that's part of GPL-2.0-or-later

kr0k0: Couldn't find such string, is "custom" ok for this package?

Cogitri: If there's no such string then it's GPL-2.0-only :)

kr0k0: Ah ok ^^. Thanks!

maxice8:
Merged with a modified commit, please take a look at it to see what information is needed in the body of the commit when adding a new package to aports.

PR #12006, opened 2019-10-28
main/gcc: Add support for the D programming language

Geod24:
```
GDC, the GCC-based D compiler, is part of GCC since GCC9.
Upstream support for Musl was added progressively (D relies heavily on libc),
and some key components were missing as of GCC 9.2.0 release (2019-08-12).
A stream of patches, from which the massive patch included in this PR derives,
was submitted to the gcc-patches mailing list on the 2019-08-21.
This patch could be removed after/if GCC 9.3.0 is released.
```

This is still very much a WIP, as dealing with GCC is much more complicated than the LDC bootstrapping.
However, should this work, it would allow D to bootstrap on all platform Alpine support, which is kind of cool.

PR #12015, opened 2019-10-29
testing/ldc: Install ldc-static alongside compiler

Geod24:
```
The previous dependency means that 'ldc' would not work out of the box.
One had to install ldc and ldc-static, or ldc and ldc-runtime and compile with flags.
This makes sure that ldc-static is installed so basic compilation works out of the box.
People wanting to compile with shared libraries will still need
to install the 'runtime' package explicitly.
```

CC @Cogitri 

EDIT: Hum, `abuild` ignore `ldc-static` in makedepends, probably because it is a subpackage. One way to solve it would be to merge `ldc-static` with `ldc`. The `static` package was originally here because `abuild` outputs a warning when a static library is present without a `-static` package, but I don't think that makes much sense for compilers (and, in fact, GCC outputs plenty of those warnings).

Another way is to just wait on #12006, given it's a much better bootstrapping solution long term (and the package currently works, just not out of the box).

### DIFF
--- a/testing/ldc/APKBUILD
+++ b/testing/ldc/APKBUILD
@@ -11,7 +11,7 @@ url="https://github.com/ldc-developers/ldc"
 arch="x86_64"
 license="BSD-3-Clause AND BSL-1.0 AND (Artistic-1.0 OR GPL-2.0-or-later) AND NCSA AND MIT"
 depends="libexecinfo tzdata"
-makedepends="bash chrpath cmake curl-dev diffutils dlang-compiler gdb grep llvm9-dev llvm9-static libedit-dev libexecinfo-static py-pip zlib-dev"
+makedepends="bash chrpath cmake curl-dev diffutils dlang-compiler ldc-static gdb grep llvm9-dev llvm9-static libedit-dev libexecinfo-static py-pip zlib-dev"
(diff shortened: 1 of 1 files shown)


Cogitri:
> EDIT: Hum, abuild ignore ldc-static in makedepends, probably because it is a subpackage.

Yup, abuild removes that dep to break the circular dep.

One way to solve it would be to merge ldc-static with ldc.

You can play abuild by adding a `provides` like rust does with rust-bootstrap

The static package was originally here because abuild outputs a warning when a static library is present without a -static package, but I don't think that makes much sense for compilers (and, in fact, GCC outputs plenty of those warnings).

But just removing the -static subpkg most likely is the best way to do this.

Geod24:
> But just removing the -static subpkg most likely is the best way to do this.

The problem is that LDC does not do the bootstrapping, it merely relies on the previous package being present. So currently it needs `ldc-static` to build. I could remove the `ldc-static` package but that would need 2 steps:
- depend on the `-static` and remove it
- remove the `-static` dependency

That's why I'm thinking to just update the package once #12006 is in. This way, I can also extend the arch support.

Geod24:
I'll just rework this once #12006 is in.

PR #12016, opened 2019-10-29
testing/perl-module-install-authorrequires: new aport

timlegge:


maxice8:
Merged with modified commits, please check to see what needs to be changed for next time

PR #12022, opened 2019-10-30
testing/mariadb: fixed plugin directory

madstk1:
When using MariaDB without services - in this case, in a Docker container - `mariadb-install-db` will fail, as it tries to load `/usr/lib*/plugin` and `/usr/lib*/mysql/plugin`. None of which exist, in a MariaDB installation. The correct path would be `/usr/lib*/mariadb/plugin`.
Failure to load plugins, leads to inability to load PAM and inherently fails the installation.

### Steps to reproduce bug
1. `docker run --rm -i -t alpine:edge /bin/ash`
1. `apk add mariadb`
1. `mariadb-install-db --basedir=/usr --datadir=/var/lib/mysql --user=mysql`

#### Expected output:
```
Installing MariaDB/MySQL system tables in '/var/lib/mysql' ...
OK
[...]
```

#### Actual output:
```
chown: /auth_pam_tool_dir: No such file or directory
Cannot change ownership of the '/auth_pam_tool_dir' directory
 to the 'mysql' user. Check that you have the necessary permissions and try again.
```

### Fix
Main patch: `./testing/mariadb/fix-plugin-dir.patch`


ncopa:
Looks like this should be reported upstream to mariadb?

madstk1:
> Looks like this should be reported upstream to mariadb?

The MariaDB repository already includes this change, but it seems to be omitted in the Alpine version.

Appropriate line can be seen [here.](https://github.com/MariaDB/server/blob/10.5/scripts/mysql_install_db.sh#L361)

Cheers,
Natamo

maxice8:
f12f2b778ee936808ce7577efd837e62d877c67a

PR #12025, opened 2019-10-30
main/linux-firmware: upgrade to 20191022

HRio:


HRio:
@ncopa contains new amd microcode. This should be backported as well. Its a but manual work to update the checksums as its needs a snapshot. So please check if I got the checksum right.

Ref:
https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/?id=2b016afc348ba4b5fb2016ffcb2822f4a293da0c

HRio:
@ncopa also it would seem the snapshot function does not care for the version in pkgver but use the current date, so not sure how to update this package to a tagged release from git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git

Should we always package latest master instead of tagged releases for this upstream?

ncopa:
> Should we always package latest master instead of tagged releases for this upstream?

I think we should use tagged release.

PR #12046, opened 2019-11-04
testing/dub: new aport

Geod24:
Add dub v1.18.0, the D package manager.

### DIFF
--- a/testing/dub/APKBUILD
+++ b/testing/dub/APKBUILD
@@ -1,0 +1,26 @@
+# Contributor: Mathias LANG <pro.mathias.lang@gmail.com>
+# Maintainer: Mathias LANG <pro.mathias.lang@gmail.com>
+pkgname=dub
+pkgver=1.18.0
+pkgrel=0
+pkgdesc="Package and build management system for D"
+url="http://code.dlang.org/"
+arch="x86_64"
+license="MIT"
+depends="libcurl"
+makedepends="ldc ldc-static bash curl-dev"
+source="$pkgname-$pkgver.zip::https://github.com/dlang/dub/archive/v$pkgver.zip"
+
+build() {
+	DMD=ldmd2 GITVER="v$pkgver" ./build.sh
+}
+
+check() {
+	bin/dub test
+}
+
+package() {
+	install -s -D "$builddir/bin/dub" "$pkgdir/usr/bin/dub"
+}
+
+sha512sums="454479867cce89d6a555505bb6b51b11192c18ae622239c01b0b9914d0c007cc6aa9b68274e80baa2eea068ae4c41c6838eef516c7ed57a947759cb37452dfa8  dub-1.18.0.zip"


Geod24:
CC @Cogitri

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+arch="x86_64"
+license="MIT"
+depends="libcurl"
+makedepends="ldc ldc-static bash curl-dev"
+subpackages="$pkgname-doc"
+source="https://github.com/dlang/dub/archive/v${pkgver}.zip"
+builddir="$srcdir/dub-$pkgver/"
+
+build() {
+	DMD=ldmd2 GITVER="v${pkgver}" ./build.sh
```
maxice8: unnecessary usage of brackets ${ }

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+pkgver=1.18.0
+pkgrel=0
+pkgdesc="Package and build management system for D"
+url="http://code.dlang.org/"
+arch="x86_64"
+license="MIT"
+depends="libcurl"
+makedepends="ldc ldc-static bash curl-dev"
+subpackages="$pkgname-doc"
+source="https://github.com/dlang/dub/archive/v${pkgver}.zip"
```
maxice8: unnecessary usage of brackets ${ }

maxice8: if possible use the .tar.gz, it is **highly** preferred over .zip files.

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+builddir="$srcdir/dub-$pkgver/"
+
+build() {
+	DMD=ldmd2 GITVER="v${pkgver}" ./build.sh
+}
+
+check () {
```
maxice8: remove the space between `check` and `()`

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+    bin/dub test
```
maxice8: indent with tabs, not spaces

testing/dub/APKBUILD
maxice8: prefix the URL with `$pkgname-$pkgver.tar.gz::`

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+pkgrel=0
+pkgdesc="Package and build management system for D"
+url="http://code.dlang.org/"
+arch="x86_64"
+license="MIT"
+depends="libcurl"
+makedepends="ldc ldc-static bash curl-dev"
+subpackages="$pkgname-doc"
+source="https://github.com/dlang/dub/archive/v${pkgver}.zip"
+builddir="$srcdir/dub-$pkgver/"
```
maxice8: This can be removed as it is the default value

testing/dub/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+build() {
+	DMD=ldmd2 GITVER="v${pkgver}" ./build.sh
+}
+
+check () {
+    bin/dub test
+}
+
+package() {
+	install -D "$builddir/LICENSE" "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
```
maxice8: It doesn't necessarily need to be installed but it is your call

Geod24:
Addressed all of @maxice8 review's except for the `zip` one. When I switched from `zip` to `tar.gz`, I got an unpacking error:
```console
>>> dub: Unpacking /var/cache/distfiles/dub-1.18.0.tar.gz...
tar: dub-1.18.0/test/issue1336-registry/packages/gitcompatibledubpackage/1.0.2.zip: Cannot utime: No such file or directory
tar: Exiting with failure status due to previous errors
>>> ERROR: dub: unpack failed
```

The file in question is a symbolic link. Is there any issue related to symlink with Alpine's `tar` ?

testing/dub/APKBUILD
```diff
@@ -0,0 +1,26 @@
+# Contributor: Mathias LANG <pro.mathias.lang@gmail.com>
+# Maintainer: Mathias LANG <pro.mathias.lang@gmail.com>
+pkgname=dub
+pkgver=1.18.0
+pkgrel=0
+pkgdesc="Package and build management system for D"
+url="http://code.dlang.org/"
+arch="x86_64"
+license="MIT"
+depends="libcurl"
```
maxice8: is this really required ? from my testing it doesn't link to  libcurl automatically so no binary uses it

Geod24: Yes. Phobos (the D standard library) does not link to `curl`, instead it loads it [at runtime](https://github.com/dlang/phobos/blob/e70d86b75d82f756ceec3b1398b77d1c1e9248bd/std/net/curl.d#L4165-L4195)

wilzbach: It's possible to build `dub` without curl though (if desired) by relying on Vibe.d (see the `net` config). However, this requires one to have a dub build in the first place, so this could only be done after bootstrapping.

testing/dub/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+makedepends="ldc ldc-static bash curl-dev"
+source="$pkgname-$pkgver.zip::https://github.com/dlang/dub/archive/v$pkgver.zip"
```
maxice8: There is a .tar.gz tarball autogenerated by GitHub like the .zip, isn't it usable ?

PR #12047, opened 2019-11-04
main/cython: upgrade to 0.29.14

OTLabs:
Ref https://github.com/cython/cython/blob/master/CHANGES.rst#02914-2019-11-01


maxice8:
Duplicate of https://gitlab.alpinelinux.org/alpine/aports/merge_requests/1048

PR #12056, opened 2019-11-07
testing/mailutils: security upgrade to 3.8

J0WI:
No CVE available: https://lists.gnu.org/archive/html/bug-mailutils/2019-11/msg00000.html


maxice8:
Merged

PR #12062, opened 2019-11-08
testing/moosefs: upgrade to 3.0.107

uu:
also changed source url to the main site.


testing/moosefs/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Contributor: Michael Pirogov <vbnet.ru@gmail.com>
 # Maintainer: Michael Pirogov <vbnet.ru@gmail.com>
 pkgname="moosefs"
-pkgver="3.0.105"
+pkgver="3.0.107"
 pkgrel=1
```
maxice8: pkgrel needs to be reset to 0 when pkgver is changed.

uu: thanks, done ;)

PR #12065, opened 2019-11-08
main/jq: fix jq rc status

mor1:
Prompted by https://github.com/alpinelinux/alpine-secdb/pull/2

Signed-off-by: Richard Mortier <mort@cantab.net>

### DIFF
--- a/main/jq/APKBUILD
+++ b/main/jq/APKBUILD
@@ -16,7 +16,7 @@ source="https://github.com/stedolan/jq/archive/$pkgname-$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$pkgname-$pkgver"
 
 # secfixes:
-#   1.6_rc1-r0:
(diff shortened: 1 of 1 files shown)


ncopa:
I believe this change is incorrect.

According https://github.com/stedolan/jq/issues/1136#issuecomment-471258094, the commit fixing the CVE-2016-4074 issue is https://github.com/stedolan/jq/commit/83e2cf607f3599d208b6b3129092fa7deb2e5292 which is reachable from the `jq-1.6rc1` tag, which means that `alpine:3.9` users with `jq-1.6_rc1-r0` installed are not affected by this CVE.

Or am I missing something here?

ncopa:
Seems like the real issue is a [bug in Clair scanner](https://github.com/alpinelinux/alpine-secdb/pull/2#issuecomment-553323633).

I don´t think we should feed the database with incorrect information.

PR #12086, opened 2019-11-11
main/libssh: upgrade to 0.9.2

J0WI:


maxice8:
Merged with a few fixes, please check the commit. It was missing -DCMAKE_INSTALL_LIBDIR=lib and was installing to lib64 instead which broke everything

J0WI:
dbe22637aab590140357c626d8f9467593ea432e

PR #12110, opened 2019-11-14
WiP testing/py3-blis: enable x86 and aarch64

OTLabs:


maxice8:
I merged a cleaned up version that enabled only on 'x86', because 'aarch64' segafults.

The commit credit goes to you, my version is just a cleaned up version of yours.

PR #12122, opened 2019-11-17
main/opensmtpd: workaround for broken mail.lmtp

fgma:
In the current release of opensmtpd delivery via lmtp is broken on musl based sytems because of a non-standard implementation of fflush() in musl. This patch adds a workaround until either musl or opensmtpd is fixed properly.

For details on the issue see https://github.com/OpenSMTPD/OpenSMTPD/issues/994


fgma:
The official fix for the problem was merged upstream in https://github.com/OpenSMTPD/OpenSMTPD/commit/0c92211d87a21ee0f5e8ca1a823e43738c666e88. This bugfix is still working for the current alpine version until it is replaced with a new upstream release.

maxice8:
Merged with commits squashed into one

PR #12136, opened 2019-11-20
main/nodejs: upgrade to 12.13.1

timbru31:
This updates Node.js to v12.13.1 (see https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.13.1).

~I've also switched to Python3 - given the fact that the changelog states that the building process is improved. Finger crossed 🤞~
Nope. `./configure: exec: line 4: python: not found`


timbru31:
Travis timed out, as always

PR #12137, opened 2019-11-20
community/nodejs-current: switch to python3 for building

timbru31:


timbru31:
Failure due to https://github.com/nodejs/node-gyp/pull/1947 - I will watch this and re-open if Node.js is updated with the updated node-gyp version.

andypost:
Maybe just use a patch https://github.com/nodejs/node-gyp/commit/c0282daa48cf1803c4283998304c563390866c8d

timbru31:
We could try that, too, if wanted. I'll update the PR later.

PR #12138, opened 2019-11-20
community/nodejs-current: switch to python3 for building

timbru31:


community/nodejs-current/APKBUILD
```diff
@@ -21,21 +21,22 @@
...
-makedepends="linux-headers python2 paxmark
+makedepends="linux-headers python3 paxmark
 	zlib-dev libuv-dev openssl-dev c-ares-dev nghttp2-dev"
 subpackages="$pkgname-dev $pkgname-doc"
 provides="nodejs"
 replaces="nodejs nodejs-lts"  # nodejs-lts for backward compatibility
 source="https://nodejs.org/dist/v$pkgver/node-v$pkgver.tar.gz
 	dont-run-gyp-files-for-bundled-deps.patch
 	link-with-libatomic-on-mips32.patch
+	fix-gyp-python3-incompatibility.patch
```
andypost: There's more then it in https://github.com/nodejs/node-gyp/pull/1972

timbru31:
The question we should discuss is if we want to include so many patches and differ from the upstream release or just wait a bit longer until the upstream release is actually python3 compatible.

andypost:
Better to wait, sametime gyp related update https://github.com/alpinelinux/aports/pull/12144

PR #12139, opened 2019-11-20
main/mesa: fix PIPE_ARCH_LITTLE_ENDIAN define for ppc64le

mksully22:
with update to mesa 19.2.5 building on ppc64le fails with error:
```
../src/gallium/auxiliary/gallivm/lp_bld_misc.cpp:708:28: error: #if with no expression
  708 | #if PIPE_ARCH_LITTLE_ENDIAN
```
This patch will fix the issue


maxice8:
Fixed on 32f331505e6e2ada549f3b62a584a7d185f3039d

PR #12144, opened 2019-11-20
testing/gyp: upgrade to 20190716

OTLabs:
- Add cleanup_srcdir()

### DIFF
--- a/testing/gyp/APKBUILD
+++ b/testing/gyp/APKBUILD
@@ -1,8 +1,8 @@
 # Contributor: Leo <thinkabit.ukim@gmail.com>
 # Maintainer: Leo <thinkabit.ukim@gmail.com>
 pkgname=gyp
-pkgver=20190707
-_commit=aca1e2c3d346d704adfa60944e6b4dd06f4728be
+pkgver=20190716
+_commit=fcd686f1880fa52a1ee78d3e98af1b88cb334528
 pkgrel=0
 pkgdesc="Generate Your Projects Meta-Build system"
 options="!check" # Tests currently fail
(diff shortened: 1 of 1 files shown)


OTLabs:
@maxice8 could you please review proposed upgrade and changes? Thank you!

maxice8:
Can you check that stuff like telegram-desktop still work ? I also did a gyp to python3 conversion once but telegram-desktop failed to build

OTLabs:
I will check the possibility to switch to python3 on telegram-desktop, thank you for a tip. Meanwhile I have refactored the PR to update gyp to latest version and added cleanup_srcdir() and removed switch to python3.

JFYI, I was able to run all tests except this one: test/generator-output/gyptest-actions.py

testing/gyp/APKBUILD
```diff
@@ -21,4 +21,9 @@ package() {
 	python2 setup.py install --root="$pkgdir"
 }
 
-sha512sums="bfe8484290e7216888c001638a5b6e42fc254ecf13320c9398a1c2a71e968f8338aebe2fe6edc64de3fe75203ad704bd8976caef21affe21f97430b838fefc52  aca1e2c3d346d704adfa60944e6b4dd06f4728be.tar.gz"
+cleanup_srcdir() {
+	[ -d src ] && chmod -R +w src
```
maxice8: Any reason why this is necessary ? as far as i am aware the only ones that need it are go packages that use modules because thay are made read-only.

OTLabs: For unknown reason I got same problem as you mention for go packages during several unsuccessful builds of this package. I have not dig to much in this issue, just solve it this way.

PR #12150, opened 2019-11-21
testing/monero: upgrade to 0.15.0.1

andypost:
Ref https://github.com/monero-project/monero/releases


Cogitri:
Fails on CI, seems like some submodule isn't in the release tarball.

andypost:
Yes, https://github.com/monero-project/monero/tree/v0.15.0.1/external the randomX vendored but not included in source tarball

PR #9813, opened 2019-08-01
community/openjdk7: security upgrade to 7.231.2.6.19

J0WI:

PR #12153, opened 2019-11-22
[3.10] community/openjdk7: security upgrade to 7.232.2.6.20

J0WI:
#9813


Cogitri:
I suppose we want the other PR in first (and CI fails here too)

PR #12159, opened 2019-11-24
community/dnscrypt-proxy: upgrade to 2.0.33-r1

ianbashford:
Add latest config file changes


maxice8:
This is not a version upgrade. So you shouldn't use `<repository>/<name>: upgrade to <version>`

maxice8:
Also when doing version upgrades never put the pkgrel= value, which is always 0

maxice8:
Have it be `repository/package: <what you're doing in the present tense>`

maxice8:
Add latest config file changes

can be the title but with some minor fixes.

`sync config file with latest changes`

Is the best title i can come up with right now.

maxice8:
Merged with the recommended fixes

PR #12165, opened 2019-11-27
testing/dlib: new aport

agboom:
http://dlib.net
A toolkit for making real world machine learning and data analysis applications in C++

This is my first Alpine package, so any feedback is appreciated. The `APKBUILD` is loosely based on the [AUR PKGBUILD](https://aur.archlinux.org/packages/dlib), but I did leave out CUDA, because I couldn't find it in the Alpine repos. If there is such a package, please let me know.

### DIFF
--- a/testing/dlib/APKBUILD
+++ b/testing/dlib/APKBUILD
@@ -1,0 +1,31 @@
+# Contributor: Davis E. King <davis@dlib.net>
+# Maintainer: Adriaan Groenenboom <agboom@pm.me>
+pkgname=dlib
+pkgver=19.18
+pkgrel=0
+pkgdesc="A toolkit for making real world machine learning and data analysis applications in C++"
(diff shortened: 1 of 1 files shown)


testing/dlib/APKBUILD
```diff
@@ -0,0 +1,31 @@
+# Contributor: Davis E. King <davis@dlib.net>
+# Maintainer: Adriaan Groenenboom <agboom@pm.me>
+pkgname=dlib
+pkgver=19.18
+pkgrel=0
+pkgdesc="A toolkit for making real world machine learning and data analysis applications in C++"
+url="http://dlib.net"
+arch="all"
+license="BSL-1.0"
+depends="jpeg libpng openblas"
```
maxice8: those shouldn't be necessary.

PR #12170, opened 2019-11-27
main/busybox: remove timestamp

J0WI:
before:
```
$ busybox
BusyBox v1.31.1 (2019-10-29 11:54:15 UTC) multi-call binary.
```
after:
```
$ busybox 
BusyBox v1.31.1 () multi-call binary.
```
This makes it a bit more reproducible

### DIFF
--- a/main/busybox/APKBUILD
+++ b/main/busybox/APKBUILD
@@ -3,7 +3,7 @@
 # Maintainer: Natanael Copa <ncopa@alpinelinux.org>
 pkgname=busybox
 pkgver=1.31.1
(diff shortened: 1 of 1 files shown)


ncopa:
can we set the build to `$SOURCE_DATE_EPOCH` instead? then will we get the git commit date.

J0WI:
I think this would require a patch, I haven't found a flag to customize it.

ncopa:
merged with 27b3358e513ce45eb29c3b2aa2e4c6ef96ebaf77. thanks!

PR #12172, opened 2019-11-28
main/cmake: upgrade to 3.16.0

andypost:
Ref https://blog.kitware.com/cmake-3-16-0-available-for-download/


ncopa:
I want wait with this til after v3.11 is released. thanks!

andypost:
Fixed with eeb7758e

PR #12181, opened 2019-12-05
community/gitea: upgrade to 1.10.1

6543:
as title


6543:
@clandmeter I'll added me as Contributor, so if someon has questions he can write me too

6543:
@maxice8 because of securety and letsencrypt fixes this would be nice if it get merged ...

PR #12182, opened 2019-12-06
testing/sblim-wbemcli: new aport

goricz:
https://sourceforge.net/projects/sblim/files/sblim-wbemcli
WBEM Command Line Interface for CIMOM access

### DIFF
--- a/testing/sblim-wbemcli/APKBUILD
+++ b/testing/sblim-wbemcli/APKBUILD
@@ -6,13 +6,10 @@ pkgrel=0
 pkgdesc="WBEM Command Line Interface for CIMOM access"
 url="https://sourceforge.net/projects/sblim/files/sblim-wbemcli"
 arch="all"
-license="CPL-1.0,EPL-1.0"
-depends="libcurl"
+license="CPL-1.0 AND EPL-1.0"
 makedepends="curl-dev"
 subpackages="$pkgname-doc"
 source="https://downloads.sourceforge.net/project/sblim/$pkgname/$pkgver/$pkgname-$pkgver.tar.bz2"
-options="!check"
-# builddir="$srcdir/$pkgname-$pkgver"
 
 build() {
 	./configure \
--- a/testing/sblim-wbemcli/APKBUILD
+++ b/testing/sblim-wbemcli/APKBUILD
@@ -10,6 +10,12 @@ license="CPL-1.0 AND EPL-1.0"
 makedepends="curl-dev"
 subpackages="$pkgname-doc"
 source="https://downloads.sourceforge.net/project/sblim/$pkgname/$pkgver/$pkgname-$pkgver.tar.bz2"
+options="!check" # there is no testsuite
+
+prepare() {
+	default_prepare
+	update_config_sub
+}
 
 build() {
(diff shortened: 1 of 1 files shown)


testing/sblim-wbemcli/APKBUILD
```diff
@@ -0,0 +1,27 @@
+# Contributor: Jirka Dutka <jirka@dutka.net>
+# Maintainer: Jirka Dutka <jirka@dutka.net>
+pkgname=sblim-wbemcli
+pkgver=1.6.3
+pkgrel=0
+pkgdesc="WBEM Command Line Interface for CIMOM access"
+url="https://sourceforge.net/projects/sblim/files/sblim-wbemcli"
+arch="all"
+license="CPL-1.0,EPL-1.0"
```
maxice8: the license is separated using a SPDX identifier.

AND -> When the project is under both licenses
OR -> If the user can pick under which license to use

testing/sblim-wbemcli/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+depends="libcurl"
```
maxice8: I believe this is found automatically via our so: providers/depends system

testing/sblim-wbemcli/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+makedepends="curl-dev"
+subpackages="$pkgname-doc"
+source="https://downloads.sourceforge.net/project/sblim/$pkgname/$pkgver/$pkgname-$pkgver.tar.bz2"
+options="!check"
+# builddir="$srcdir/$pkgname-$pkgver"
```
maxice8: this can be removed

testing/sblim-wbemcli/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+pkgrel=0
+pkgdesc="WBEM Command Line Interface for CIMOM access"
+url="https://sourceforge.net/projects/sblim/files/sblim-wbemcli"
+arch="all"
+license="CPL-1.0,EPL-1.0"
+depends="libcurl"
+makedepends="curl-dev"
+subpackages="$pkgname-doc"
+source="https://downloads.sourceforge.net/project/sblim/$pkgname/$pkgver/$pkgname-$pkgver.tar.bz2"
+options="!check"
```
maxice8: Please add a comment on why this is not enabled

Cogitri:
Seems like this doesn't build on aarch64:

`configure: error: cannot guess build type; you must specify one`

This can be fixed with `update_config_sub`, see other packages for how to do this.

goricz:
> Seems like this doesn't build on aarch64:
> 
> `configure: error: cannot guess build type; you must specify one`
> This can be fixed with `update_config_sub`, see other packages for how to do this.

Hi, 

tried to add prepare section with update_config_sub, no luck, still failing for aarch64. 
Can I just exclude this architecture?

Cogitri:
Seems to work now, good job :)

Since my laptop is in repair (*sigh*, again) @maxice8 can you please take another look & merge?

PR #12188, opened 2019-12-28
testing/podman: new aports

kohnish:


maxice8:
It is preferable that gitlab.alpinelinux.org is used.

kohnish:
Will do from the next patch. Thanks for merging.

